PR #1, opened 2026-04-15
Improve Preview Rendering Reliability with Compose Test Rules and Roborazzi

ithinkihaveacat:
This PR updates the standalone preview rendering engine to improve stability, resolve blank image generation issues, and ensure the toolchain operates reliably across different developer environments. 

Below is a detailed breakdown of the architectural choices and specific build logic implementations included in this change.

---

#### Why are we eliminating custom lifecycle management?
The previous system attempted to manually instantiate a `ComponentActivity`, manually drive its lifecycle, and manually pump execution frames to force layouts to converge. This custom orchestration was brittle. Because it wasn't perfectly handling the cascade of parent-to-child measure constraints, child nodes in the Compose tree often failed to receive correct dimensions in time, leading to truncated or missing layouts. Removing this in favor of the platform-provided `createComposeRule()` delegates that complex lifecycle and measurement orchestration back to the core AndroidX libraries, drastically improving reliability.

#### Why JUnit / `ui-test-junit4`?
Jetpack Compose's testing architecture is designed around the `ComposeTestRule`, which relies on the JUnit4 test runner infrastructure. We introduced this dependency because it provides the only standardized mechanism to control the `BroadcastFrameClock` and automatically synchronize the rendering passes with the main thread message queue. Without this rule, we would have to manually replicate massive amounts of internal framework logic just to figure out when a Composable has settled and is ready to be drawn.

#### Why Roborazzi?
The legacy pipeline relied on a highly experimental and internal Robolectric feature via reflection (`HardwareRenderingScreenshot`). In our setup, that approach frequently yielded completely empty (blank) images because it captured the view before Compose finished its internal multi-pass measurements. 
Roborazzi bridges this gap reliably. It hooks directly into the Compose testing tree to guarantee that the UI is completely idle and fully measured before pulling any pixels. Additionally, it handles complex rendering features (like elevation and shadows) that pure Robolectric often fails to synchronize correctly in headless environments.

---

### ⚙️ Classpath & Build Resolution Strategies (Potential Points of Discussion)

While implementing the above, I had to introduce a few specialized build-script strategies in the Gradle plugin to ensure smooth execution:

*   **Pulling the entire AGP Test Task Classpath:** The plugin now resolves execution classpaths by pulling directly from `agpTestTask.classpath`. 
    *   *Justification:* This guarantees that the isolated rendering engine automatically inherits any specific testing tools or custom layout engines that the host application is already using, preventing "class not found" errors during headless runs.
*   **Fallback to Sibling Repository Paths:** If the renderer isn't found in the root project, the plugin falls back to searching in a hardcoded relative path (`../../compose-ai-tools/renderer-android`).
    *   *Justification:* While this assumes a specific directory structure on the developer's machine, it is currently necessary to bridge the app workspace and the tools repository locally since the tools are not yet published as standard dependencies.
*   **Resilient `android.jar` Resolution:** Added logic to search through the SDK `platforms/` directory if the exact requested `android-$compileSdk` folder isn't found.
    *   *Justification:* Different development environments (and different versions of the Android SDK manager) lay out platform files slightly differently (e.g., `android-35` vs `android-35.0`). This prevents build failures due to minor variations in local SDK structures.

### DIFF
--- a/renderer-android/src/main/kotlin/ee/schimke/composeai/renderer/RobolectricRenderTest.kt
+++ b/renderer-android/src/main/kotlin/ee/schimke/composeai/renderer/RobolectricRenderTest.kt
@@ -23,10 +23,15 @@ import androidx.compose.ui.platform.AndroidUiDispatcher
 import androidx.compose.ui.platform.LocalInspectionMode
 import androidx.compose.ui.platform.WindowRecomposerFactory
 import androidx.compose.ui.platform.WindowRecomposerPolicy
+import androidx.compose.ui.test.junit4.createComposeRule
+import com.github.takahirom.roborazzi.captureRoboImage
+import java.io.File
+import java.io.FileOutputStream
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import kotlinx.serialization.json.Json
+import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.ParameterizedRobolectricTestRunner
@@ -35,8 +40,6 @@ import org.robolectric.Shadows
 import org.robolectric.annotation.Config
 import org.robolectric.annotation.GraphicsMode
 import org.robolectric.shadows.ShadowLooper
-import java.io.File
-import java.io.FileOutputStream
 import java.util.concurrent.TimeUnit
 
 /**
@@ -101,7 +104,7 @@ object PreviewManifestLoader {
 // SDK 34 is the highest API where Robolectric 4.16.1 still shadows
 // `ShadowNativeImageReaderSurfaceImage.nativeCreatePlanes`. Above that, the
 // HardwareRenderingScreenshot path returns an Image with `planes[0] == null`.
-@Config(sdk = [34])
+@Config(sdk = [35], qualifiers = "w227dp-h227dp-small-notlong-round-watch-xhdpi-keyshidden-nonav")
 @GraphicsMode(GraphicsMode.Mode.NATIVE)
 abstract class RobolectricRenderTestBase(private val preview: RenderPreviewEntry) {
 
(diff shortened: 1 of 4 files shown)


renderer-android/src/main/kotlin/ee/schimke/composeai/renderer/RobolectricRenderTest.kt
```diff
@@ -101,7 +104,7 @@ object PreviewManifestLoader {
 // SDK 34 is the highest API where Robolectric 4.16.1 still shadows
 // `ShadowNativeImageReaderSurfaceImage.nativeCreatePlanes`. Above that, the
 // HardwareRenderingScreenshot path returns an Image with `planes[0] == null`.
-@Config(sdk = [34])
+@Config(sdk = [35], qualifiers = "w227dp-h227dp-small-notlong-round-watch-xhdpi-keyshidden-nonav")
```
yschimke, line 107: I don't think this is a good default. Can we set the ones that aren't Wear specific?

renderer-android/src/main/kotlin/ee/schimke/composeai/renderer/RobolectricRenderTest.kt
```diff
@@ -115,72 +118,26 @@ abstract class RobolectricRenderTestBase(private val preview: RenderPreviewEntry
...
+        composeTestRule.setContent {
+            println("Entering setContent composition")
             CompositionLocalProvider(LocalInspectionMode provides true) {
                 val clazz = Class.forName(preview.className)
                 val composableMethod = clazz.getDeclaredComposableMethod(preview.functionName)
                 val bgColor = when {
                     preview.params.backgroundColor != 0L -> Color(preview.params.backgroundColor.toInt())
                     preview.params.showBackground -> Color.White
-                    else -> Color.Transparent
+                    else -> Color.Black
```
yschimke, line 134: Is this Wear specific?

yschimke:
folded into https://github.com/yschimke/compose-ai-tools/pull/5

PR #14, opened 2026-04-16
Fix extreme slowness and OOM in Android preview rendering

ithinkihaveacat:
### Problem
Running `./gradlew :sample-android:renderAllPreviews` was taking over 12 minutes to complete in restricted or headless environments (like virtual machines) and frequently failing with `OutOfMemoryError`.

### Cause
1. **Infinite Animation:** The `LoadingPreview` used a `CircularProgressIndicator()` without arguments, which creates an infinite spinning animation. Since the renderer waits for the pixel checksum to stabilize for 2 consecutive frames to ensure the UI has settled, an infinite animation prevents stabilization. This forced the renderer to run for the maximum allowed limit (20 frames) for this preview, doing heavy layout and JNI graphics work on every frame.
2. **Memory Limit:** The test worker process defaulted to 512MB of heap, which was insufficient for handling the load, leading to frequent garbage collection pauses and OOM crashes.

### Solution
1. **Make Preview Static:** Changed `CircularProgressIndicator()` to use a static progress lambda `progress = { 0.5f }` in `LoadingPreview`. This allows the renderer to stabilize immediately on frame 1 or 2.
2. **Increase Memory:** Set `jvmArgs("-Xmx2048m")` in the `testOptions` of `sample-android/build.gradle.kts` to ensure the test worker process has enough memory.

### Results
* **Render time dropped from ~12.5 minutes to 15 seconds!** (approx. 50x speedup).
* Resolved the `OutOfMemoryError` crashes.

### DIFF
--- a/sample-android/src/main/kotlin/com/example/sampleandroid/Previews.kt
+++ b/sample-android/src/main/kotlin/com/example/sampleandroid/Previews.kt
@@ -72,7 +72,7 @@ fun GreetingPreview() {
 fun LoadingPreview() {
     MaterialTheme {
         Column(horizontalAlignment = Alignment.CenterHorizontally) {
-            CircularProgressIndicator()
+            CircularProgressIndicator(progress = { 0.5f })
             Text("Loading...")
         }
     }
--- a/sample-android/build.gradle.kts
+++ b/sample-android/build.gradle.kts
@@ -24,6 +24,12 @@ android {
     buildFeatures {
         compose = true
     }
+
+    testOptions {
+        unitTests.all {
+            it.jvmArgs("-Xmx2048m")
+        }
+    }
 }
 
 dependencies {


sample-android/src/main/kotlin/com/example/sampleandroid/Previews.kt
```diff
@@ -72,7 +72,7 @@ fun GreetingPreview() {
 fun LoadingPreview() {
     MaterialTheme {
         Column(horizontalAlignment = Alignment.CenterHorizontally) {
-            CircularProgressIndicator()
+            CircularProgressIndicator(progress = { 0.5f })
```
yschimke, line 75: We should add a TODO. And should we add a tight timeout (can't use junit timesout with robolectric)? as follow up. Will land.

PR #16, opened 2026-04-16
Support External Dependency Consumption

ithinkihaveacat:
## Summary

This PR introduces changes to support consuming `compose-ai-tools` as an
external library dependency, rather than having to embed the source code
directly in the consumer project.

## Motivation

This change is necessary to support consuming `compose-ai-tools` as an external library dependency.

The alternative approach of copying the library's source code directly into the consumer repository is hard to maintain, leads to code duplication, and makes it difficult to share improvements across different projects. By supporting external consumption, we enable a much cleaner architecture where the library can be evolved independently and consumed via standard dependency management.

This enables better integration with sample apps like [WearWidgetKotlin](https://github.com/ithinkihaveacat/wear-os-samples/tree/wear-widgets/WearWidgetKotlin) and any other app that wants to use this library for rendering previews.

### Gradle Plugin (`gradle-plugin`)

- Refactored `AndroidPreviewSupport.kt` to support dynamic classpath resolution.
- It now detects if it is running as a local project or an external dependency.
- It looks for test classes in the consumer project's build directory (both
  Kotlin and Java output dirs).

### Android Renderer (`renderer-android`)

- Added `maven-publish` configuration to enable publishing to Maven repositories (like `mavenLocal()`).
- Added stub classes for `org.conscrypt.OpenSSLProvider` and `org.conscrypt.OkHostnameVerifier` in `src/main/java`.

## Conscrypt Stubs Rationale & Safety

We encountered an environment issue where Robolectric tests in the consumer
project failed to load native libraries for Conscrypt (specifically looking for
`libstdc++.so.6` which was not compatible or missing in the specific test
environment).

To provide a smooth experience out-of-the-box, we added minimal stub
implementations for the required Conscrypt classes.

- **Safety**: These stubs are only intended to satisfy the class loader during
  Robolectric initialization in environments where Conscrypt fails to load. If a
  consumer project requires real Conscrypt functionality for their tests, they
  may need to ensure proper native library setup and override these stubs. For
  preview rendering purposes, these stubs are sufficient and safe.

## Gradle Plugin Testing Notes

We have not added automated tests for the Gradle plugin changes in this PR. Here is a short explanation of why, and what would be required to test it:

### Why tests were not added
Testing the specific logic for detecting and resolving external dependencies requires that the library artifact (`renderer-android`) be available in a repository (like `mavenLocal()`) *before* the test runs. This creates a cyclic dependency in the build process (we need to publish to test, but tests run during the build) which is not easily supported by the current `GradleRunner` setup in `functionalTest`.

### What would be required to test
To test this conventionally, we would need to:
1. Use `GradleRunner` to create a temporary test project.
2. Ensure `renderer-android` is published to `mavenLocal()` before running the test.
3. Configure the test project to resolve `renderer-android` from `mavenLocal()`.
4. Verify that the plugin correctly locates the classes and sets up the classpath.

### Safety of the change without tests
This change is considered safe because:
- It falls back to the existing local project resolution logic if the external dependency is not found.
- It only affects the test runtime classpath setup for the rendering task, not the main application build or runtime.
- We manually verified that it works correctly in the `WearWidgetKotlin` sample app, which consumed the published artifact.

## Verification & Environment

These changes were thoroughly tested and verified:
- **OS**: Linux
- **Robolectric**: 4.16.1
- **Kotlin**: 2.2.21
- **Java Versions**: Verified with both **Java 11** and **Java 21** in the consumer application.

We added a unit test (`ConscryptStubTest`) in `renderer-android` to verify that the stub classes can be loaded and instantiated correctly.

The rendering task was able to generate 55 valid screenshots (3 broken previews were filtered out on the app side to allow the task to succeed).

## Manual Verification with WearWidgetKotlin

To verify these changes manually using the `WearWidgetKotlin` codebase, follow these steps:

### Prerequisites
1.  Check out the `wear-widgets` branch in the `wear-os-samples` repository (specifically at [WearWidgetKotlin](https://github.com/ithinkihaveacat/wear-os-samples/tree/wear-widgets/WearWidgetKotlin)).
2.  Ensure you have this PR's branch (`feature/option-2-support`) checked out in `compose-ai-tools` (specifically commit `a092386d94d4bc01e9436ab542e5097ff6eb7a2a` or later).

### Steps to Reproduce

1.  **Publish the library and plugin from `compose-ai-tools`**:
    In the `compose-ai-tools` directory, run the following command to publish both:
    ```bash
    ./gradlew publishToMavenLocal && ./gradlew -p gradle-plugin publishToMavenLocal
    ```

2.  **Run previews in `WearWidgetKotlin`**:
    In the `WearWidgetKotlin` directory, run:
    ```bash
    ./gradlew renderPreviews
    ```
    This will execute the rendering task using the published library.

3.  **Generate individual previews**:
    You can also use the `widget-screenshot` script to generate and inspect individual previews. For example:
    ```bash
    ./widget-screenshot WidgetButtonSamplePreview
    ```
    This will generate the screenshot for `WidgetButtonSamplePreview` and place it in `app/screenshots/`.

### What to Expect
- The build should complete successfully (BUILD SUCCESSFUL).
- All executed tests should pass.
- The generated screenshots should be available in `app/build/compose-previews/renders/` and should not be blank or corrupted.

### DIFF
--- a/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
+++ b/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
@@ -82,28 +82,39 @@ internal object AndroidPreviewSupport {
         val hasAndroidRenderer = rendererProjectDir.resolve("build.gradle.kts").exists()
                 || rendererProjectDir.resolve("build.gradle").exists()
 
-        if (hasAndroidRenderer) {
-            val rendererClassDirs = project.files(
+        val rendererClassDirs = if (hasAndroidRenderer) {
+            project.files(
                 rendererProjectDir.resolve("build/intermediates/built_in_kotlinc/$variantName/compile${capVariant}Kotlin/classes"),
                 rendererProjectDir.resolve("build/tmp/kotlin-classes/$variantName"),
             )
+        } else {
+            project.files(
+                project.layout.buildDirectory.dir("intermediates/built_in_kotlinc/${variantName}UnitTest/compile${capVariant}UnitTestKotlin/classes"),
+                project.layout.buildDirectory.dir("intermediates/javac/${variantName}UnitTest/compile${capVariant}UnitTestJavaWithJavac/classes")
+            )
+        }
 
-            // Renderer's transitive runtime dependencies come through a dedicated
-            // resolvable configuration in *this* project. Attributes are copied
-            // from the sample's unit-test runtime classpath so Gradle picks the
-            // right Android variant without us declaring them by hand.
-            val rendererConfig = project.configurations.maybeCreate("composePreviewAndroidRenderer$capVariant").apply {
-                isCanBeResolved = true
-                isCanBeConsumed = false
-                if (testConfig != null) {
-                    copyAttributes(attributes, testConfig.attributes)
-                }
+        // Renderer's transitive runtime dependencies come through a dedicated
+        // resolvable configuration in *this* project. Attributes are copied
+        // from the sample's unit-test runtime classpath so Gradle picks the
+        // right Android variant without us declaring them by hand.
+        val rendererConfig = project.configurations.maybeCreate("composePreviewAndroidRenderer$capVariant").apply {
+            isCanBeResolved = true
+            isCanBeConsumed = false
+            if (testConfig != null) {
+                copyAttributes(attributes, testConfig.attributes)
             }
+        }
+
+        if (hasAndroidRenderer) {
             try {
                 project.dependencies.add(rendererConfig.name, project.dependencies.project(mapOf("path" to ":renderer-android")))
             } catch (e: org.gradle.api.UnknownProjectException) {
                 project.logger.debug("compose-ai-tools: :renderer-android project not found, skipping", e)
             }
+        } else {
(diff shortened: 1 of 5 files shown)


gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
```diff
@@ -243,13 +256,17 @@ internal object AndroidPreviewSupport {
                 // Roborazzi defaults to "compare" mode (which doesn't write pixels
                 // unless the expected baseline exists). Force "record" so every run
                 // writes fresh PNGs.
-                systemProperty("roborazzi.test.record", "true")
+                jvmArgs("-Droborazzi.test.record=true")
```
ithinkihaveacat, line 259: Rationale for this change:

> We changed `systemProperty("roborazzi.test.record", "true")` to `jvmArgs("-Droborazzi.test.record=true")` because Roborazzi reads this system property very early during its initialization (likely within a static initializer).
>
> When configured via `systemProperty` in the Gradle `Test` task, the property is sometimes set too late in the test JVM lifecycle for Roborazzi to detect it, causing it to fall back to its default "compare" mode (which fails if no baseline image exists). Passing it directly via `jvmArgs` ensures it is available at the very start of the JVM process, guaranteeing that Roborazzi correctly enters "record" mode and writes out the fresh screenshots every time.
>
> This is a common pattern when dealing with tools that inspect environment or system properties during class loading or static initialization rather than during test execution.

PR #63, opened 2026-04-18
feat(a11y): annotated screenshot overlay + interactive VSCode legend

yschimke:
## Summary

Phase 2 of the ATF accessibility work (follows #58). With `accessibilityChecks.enabled = true` each preview now gets a sidecar `<id>.a11y.png` showing the findings as numbered badges on the screenshot + a legend panel, and the VSCode extension renders an interactive legend below the clean PNG where hovering a row highlights the matching bounds on the image.

Three modes are now explicit:

| Mode | Gradle | `-P` override |
|---|---|---|
| Off (default) | `accessibilityChecks { enabled = false }` | — |
| ATF only | `enabled = true; annotateScreenshots = false` | `-PcomposePreview.accessibilityChecks.enabled=true -PcomposePreview.accessibilityChecks.annotateScreenshots=false` |
| ATF + overlay | `enabled = true` | `-PcomposePreview.accessibilityChecks.enabled=true` |

## How it works

**Overlay renderer** (new `AccessibilityOverlay.kt`). Native Canvas — coloured outlines + numbered badges on the screenshot, legend panel listing rule + level + message + element descriptor. Adaptive layout: portrait mobile → side-by-side (520px legend right of screenshot); Wear / square / landscape → stacked (legend below, full width). Verified on sample-android (840×470 side-by-side) and sample-wear (520×432 stacked).

**Config-cache-friendly plumbing.** `verifyAccessibility` is always registered with an `onlyIf(Provider<Boolean>)` gate; renderer-side ATF system properties go through a `CommandLineArgumentProvider` with `@Input` providers so they resolve at task execution time. Net result: toggling `-PcomposePreview.accessibilityChecks.enabled` reuses the configuration cache (~0s reconfigure vs ~5-8s previously) while still re-running renders when the flag flips (task up-to-date keys correctly differ). Same pattern is the template for any future VSCode-toggleable option.

**Collapsed render path.** After #62 landed the paused-`mainClock` default path which already uses `createAndroidComposeRule`, the separate a11y path was pure duplication. The ATF call is now inline in `renderDefault`: `LocalInspectionMode` flips to `false` when a11y is on (so the semantics tree actually gets populated), and after capture we pull `(onRoot.fetchSemanticsNode().root as ViewRootForTest).view` for ATF — exactly what `roborazzi-accessibility-check` walks internally.

**VSCode interactive legend.** Each card with findings gets an `.a11y-legend` below the image and an `.a11y-overlay` layer inside the image container. Overlay boxes are positioned as % of the image's natural dimensions so they stay correct across layout changes without a resize handler. Hovering a legend row adds `.a11y-active` to the matching overlay box; leaving clears. Colour-coded by level (red/orange/blue).

## Test plan

- [x] `./gradlew check` — green
- [x] `./gradlew :sample-android:renderAllPreviews` — default-off behaviour unchanged
- [x] `./gradlew :sample-android:renderAllPreviews -PcomposePreview.accessibilityChecks.enabled=true` — writes `accessibility.json` + `<id>.a11y.png` for `BadButtonPreview`; `SpeakableTextPresentCheck` error surfaces
- [x] `./gradlew :sample-wear:renderAllPreviews -PcomposePreview.accessibilityChecks.enabled=true` — Wear-stacked overlay renders correctly; tiles also produce `TextSizeCheck` findings
- [x] Config cache reused across three consecutive toggles (off → on → off): `Configuration cache entry reused` on runs 2 and 3; task up-to-date tracking re-runs `renderPreviews` when the flag flips
- [x] VSCode extension compiles; legend + overlay rendering smoke-tested via generated sample outputs
- [ ] Eyeball VSCode extension in a running instance (reviewer hand-test — the webview HTML is compiled output; no automated webview test covers hover styling)

Sample outputs (committed on the branch, not checked in):

- sample-android mobile overlay (side-by-side, 840×470): BadButton with red badge + SpeakableTextPresentCheck row
- sample-wear round overlay (stacked, 520×432): same check on the Wear-shaped preview

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/Commands.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/Commands.kt
@@ -60,6 +60,7 @@ data class AccessibilityFinding(
 data class AccessibilityEntry(
     val previewId: String,
     val findings: List<AccessibilityFinding>,
+    val annotatedPath: String? = null,
 )
 
 @Serializable
@@ -85,6 +86,12 @@ data class PreviewResult(
      * disabled for this module. Empty list means checks ran and found nothing.
      */
     val a11yFindings: List<AccessibilityFinding>? = null,
+    /**
+     * Absolute path to an annotated screenshot showing each finding as a
+     * numbered badge + legend. `null` when there were no findings or
+     * accessibility checks are disabled.
+     */
+    val a11yAnnotatedPath: String? = null,
 )
 
 @Serializable
@@ -161,10 +168,16 @@ abstract class Command(protected val args: List<String>) {
      * findings disappear from CLI output; we never report stale reports from
      * a prior opt-in run.
      */
+    /**
+     * Loads each module's accessibility report and returns a lookup from
+     * `"<module>/<previewId>"` to (findings, annotatedPathAbsolute). The
+     * annotated path is resolved against the report file's parent so the
+     * value we emit is an absolute filesystem path agents can open directly.
+     */
     protected fun readAllA11yReports(
         manifests: List<Pair<String, PreviewManifest>>,
-    ): Map<String, List<AccessibilityFinding>> {
-        val out = mutableMapOf<String, List<AccessibilityFinding>>()
+    ): Map<String, Pair<List<AccessibilityFinding>, String?>> {
+        val out = mutableMapOf<String, Pair<List<AccessibilityFinding>, String?>>()
         for ((module, manifest) in manifests) {
             val pointer = manifest.accessibilityReport ?: continue
             val reportFile = projectDir.resolve("$module/build/compose-previews/$pointer")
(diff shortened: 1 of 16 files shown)


yschimke:
Pushed a fix for the Integration CI red on WearTilesKotlin (pre-existing on main after #62):

**Root cause.** The paused-`mainClock` default path wrapped `setContent` in `Box(Modifier.fillMaxSize().background(bg)) { … }`. compose-compiler 2.2.21 emits `ComposeUiNode.setCompositeKeyHash` getter calls for every layout node in our bytecode; that accessor only exists on compose-ui 1.9+ (compose-bom ≥ 2025.01.00). Consumers on older BOMs (WearTilesKotlin pins 2024.04.00) hit `NoSuchMethodError` the instant `renderPreviews` starts. #60's compileOnly fix covers the transitive-dep axis; this fix covers the direct-compose-API axis.

**Fix.**

- Paint the preview background natively on `rule.activity.window.decorView.setBackgroundColor(bg.toArgb())` before `setContent`. Drop the outer `Box(...)` wrapper entirely.
- Same treatment for `ComposePreviewStrategy` — drop its inner `Box(Modifier.fillMaxSize().background(bgColor))`. fillMaxSize becomes the consumer composable's responsibility (every real `@Preview` already fills its bounds).

Net effect: the renderer's bytecode stops referencing any post-1.8 compose layout-node APIs. The preview body itself runs through the consumer's own compose-compiler, so its bytecode naturally targets the consumer's runtime.

**Doctor check added.** `compose-preview doctor` now scans `gradle/libs.versions.toml` + every `build.gradle*` under the workspace for `compose-bom` / inline `platform("androidx.compose:compose-bom:…")` declarations and warns if any are below 2025.01.00. Runs without Gradle — same shape as the other pre-flight checks.

Verified locally: `./gradlew :sample-android:renderAllPreviews` + `:sample-wear:renderAllPreviews -PcomposePreview.accessibilityChecks.enabled=true` both green; `compose-preview doctor` flags an intentionally-old BOM with a clear message.

Issue #142, opened 2026-04-21
`compose-ai-tools` Classpath and Test Discovery Failures


ithinkihaveacat:
**Applies to version**: `0.7.5-SNAPSHOT` (commit
`09732a3885c118607e4680942e66defe831b826b` of `compose-ai-tools`).

## Problem Statement

The `compose-ai-tools` library's preview rendering infrastructure fails to work
out-of-the-box with certain Android project configurations, specifically
regarding test class discovery and Android platform class visibility in
Robolectric environments.

## Problem Explanation

When running the custom `renderPreviews` task provided by the `compose-ai-tools`
Gradle plugin, the following issue was observed:

1. **ClassNotFoundException for Android Platform Classes**: The `renderPreviews`
   task is registered as a plain Gradle `Test` task, rather than an Android unit
   test task. Consequently, it does not automatically inherit the Android SDK's
   `android.jar` on its classpath. When JUnit attempts to introspect the test
   class (e.g., `RobolectricRenderTest`), it encounters references to Android
   platform classes (like `android.app.Application`) in method signatures or
   annotations and fails with `ClassNotFoundException` because these types are
   not present in the standard Java boot classpath used by the test worker.

## What Was Tried to Fix It

- **Observations**:
  - The test class was successfully compiled and present in the build directory.
  - The plugin attempted to add `consumerTestClasses` to the classpath, but it
    hardcoded the path to a specific directory structure that might not match
    all AGP versions or project layouts.
  - Adding the classes directory directly to `renderPreviews.classpath` in an
    `afterEvaluate` block did not resolve the `ClassNotFoundException`.
- **Speculation**:
  - It is possible that the `Test` task's `classpath` property behaves
    differently regarding directory entries vs. JARs for class discovery by
    JUnit's worker process.
  - Attempting to use `bootstrapClasspath` to force `android.jar` onto the boot
    classloader failed because the underlying option (`-Xbootclasspath`) is no
    longer supported in modern Java versions (Java 9+).

## Workaround Applied to a Project

To bypass the classpath limitations of the custom `renderPreviews` task while
maintaining the ability to generate previews, the following workaround can be
applied to a project:

1. **Leverage `testDebugUnitTest`**: Instead of using the custom
   `renderPreviews` task, use the standard Android unit test task provided by
   AGP, which correctly handles visibility of Android platform classes.
2. **Direct Classpath Injection**: In the module's `build.gradle.kts`,
[...]

PR #163, opened 2026-04-23
Add Diagnostic Warning for `ANDROID_JAR`

ithinkihaveacat:
## Title

feat: add diagnostic warning for `ANDROID_JAR` environment variable

## Description

This PR adds a diagnostic warning to the `DiscoverPreviewsTask` to alert users
when the `ANDROID_JAR` environment variable is set while running on Java 9 or
higher.

### Background

During investigation of issue #142 and subsequent testing, it was discovered
that having `ANDROID_JAR` set in the environment can cause Gradle's plain `Test`
tasks (like `renderPreviews`) to automatically append the Android SDK jars to
the `-Xbootclasspath` JVM argument on some platforms (specifically observed on
Linux with specific JDKs). This is problematic because `-Xbootclasspath` is a
legacy option that was removed in Java 9 and later. While not all JVM 21
implementations are affected (for example, the JetBrains Runtime on macOS did
not exhibit this failure), it causes the JVM to refuse to start on at least one
specific environment (observed on Linux with a specific OpenJDK 21 build),
resulting in a fatal error instead of executing the tests:
`Could not create the Java Virtual Machine.`
`-Xbootclasspath is no longer a supported option.` This makes the custom task
unusable in such environments unless the variable is unset.

Additionally, standard `println` is used because `logger.lifecycle` output was
found to be suppressed by Gradle's rich console on macOS for this task action.

### Changes

- Added a check in `DiscoverPreviewsTask.discover()` to inspect the
  `ANDROID_JAR` environment variable.
- If set and running on Java 9+, it logs a warning suggesting to unset it.
- Uses standard `println` for this warning to ensure visibility on all
  platforms.

### Verification Results

- Verified that unsetting `ANDROID_JAR` allows the tests to run successfully on
  the affected Linux environment.
- Verified that the warning appears visibly on both Linux and macOS when the
  variable is set.


ithinkihaveacat:
Closing, not confident this is a good approach.

PR #178, opened 2026-04-24
fix(plugin): widen dep-jar filter + fail fast when @Preview not on classpath (#162)

### LINKED ISSUE #162: ./gradlew discoverPreviews doesn't discover any Previews
Here's the report from the gradle task
```
> Task :app:discoverPreviews
Discovered 0 preview(s) in module 'app':
```

This is the resulting previews.json:
```json
{
    "module": "app",
    "variant": "uatDebug",
    "previews": [
    ],
    "accessibilityReport": null
}
```

The plugin is registered in the top-level `build.gradle`:
```groovy
plugins {
    ...
    alias libs.plugins.compose.ai.preview apply false
}
```

Module-level `build.gradle`:
```groovy
plugins {
    ...
    alias libs.plugins.compose.ai.preview
}

composePreview {
    variant.set('uatDebug')
    sdkVersion.set(35)     // Robolectric SDK version (default: 35)
    enabled.set(true)      // set false to skip task registration
}
```

From `libs.versions.toml`:
```
composeAiPreview = "0.7.7"
...
compose-ai-preview = { id = "ee.schimke.composeai.preview", version.ref = "composeAiPreview" }
```

Running `compose-preview doctor` says everything but the GH token is fine. I'll gladly update with whatever else might be needed since I realize this isn't much to go on.

yschimke:
## Summary

- The `dependencyJars` filter in `DiscoverPreviewsTask` matched on `file.name`, but AGP 9.x + KGP 2.3 resolves AAR deps to `<cache>/transforms/<hash>/transformed/<library>/jars/classes.jar` — `file.name` is just `classes.jar`, so **every AAR-extracted jar was dropped**. That disables multi-preview annotation resolution (it calls `scanResult.getClassInfo(ann.name)`). Root cause of #162.
- Switched the filter to match on `file.absolutePath` so the library dir qualifies. Still precise: only paths containing `preview|tooling|compose|annotation`.
- Added an unconditional fail in `discoverPreviews` when the scan found classes but **no known @Preview FQN is reachable** on the ClassGraph classpath. Catches the #162 shape without requiring `failOnEmpty=true`, and the existing diagnostics block now reports `reachable @Preview annotation classes on ClassGraph classpath` so the lifecycle log disambiguates "jar missing" from "annotation FQN doesn't match".

## Test plan

- [x] `./gradlew :gradle-plugin:test`
- [x] `./gradlew :gradle-plugin:functionalTest --rerun-tasks`
- [x] `./gradlew :sample-wear:discoverPreviews --rerun-tasks` — 17 previews discovered (unchanged from main)
- [ ] Manual verification against the #162 repro once reporter runs the fix branch

Closes #162

### DIFF
--- a/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
+++ b/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
@@ -35,6 +35,25 @@ internal object AndroidPreviewSupport {
         "tiles-tooling",
     )
 
+    /**
+     * `(group, name)` of every artifact whose presence in a module's
+     * declared deps marks it as a "valid preview module" — the plugin
+     * registers its tasks and runs discovery only when at least one
+     * matches. Convention-plugin-everywhere setups (e.g. applying
+     * `composePreview` to every Android module) stay silent no-ops on
+     * utility modules without any preview surface.
+     *
+     * Group+name match only (no version): cheap, IP-safe, doesn't
+     * trigger dependency resolution.
+     */
+    private val previewArtifactSignals = setOf(
+        "androidx.compose.ui" to "ui-tooling-preview",
+        "androidx.compose.ui" to "ui-tooling-preview-android",
+        "androidx.wear.tiles" to "tiles-tooling-preview",
+        // CMP-only; AGP consumers never declare it but the helper is shared.
+        "org.jetbrains.compose.components" to "components-ui-tooling-preview",
+    )
+
     fun configure(project: Project, extension: PreviewExtension) {
         val androidComponents = project.extensions.getByType(AndroidComponentsExtension::class.java)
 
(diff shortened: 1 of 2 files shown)


yschimke:
Pushed a follow-up: 05adf8f — gates Android task registration on "module declares a known preview artifact" (group+name match: `androidx.compose.ui:ui-tooling-preview`, `androidx.wear.tiles:tiles-tooling-preview`, `org.jetbrains.compose.components:components-ui-tooling-preview`). Convention-plugin-everywhere setups (e.g. Androidify) now silently skip on utility modules with no Compose surface — task registration just doesn't happen, and the unconditional fail-on-missing-Preview inside `discoverPreviews` is now safe because it only runs on valid preview modules.

Extra wins: utility modules stop getting `ui-test-manifest` / `ui-test-junit4` auto-injected into their test classpath.

Scoped to the AGP path for now — CMP (`AndroidPreviewSupport.configure` only) can pick up the same gate as a follow-up if needed.

PR #249, opened 2026-04-26
docs(daemon): lock IPC protocol contract (P0.4)

yschimke:
## Summary

- Adds [docs/daemon/PROTOCOL.md](docs/daemon/PROTOCOL.md) as the authoritative wire-format spec for the JSON-RPC channel between the VS Code extension and the per-module preview daemon.
- Seeds [docs/daemon/protocol-fixtures/](docs/daemon/protocol-fixtures/) as the shared golden-message corpus that both Kotlin and TypeScript test suites will consume in Phase 1.
- Marks P0.4 done in [docs/daemon/TODO.md](docs/daemon/TODO.md); links the protocol doc from the daemon docs index.

This is **Phase 0 task P0.4** from the daemon work breakdown. Pure docs — no code risk. Unblocks parallel work on Stream B (daemon Kotlin protocol types, B1.2) and Stream C (VS Code TypeScript types, C1.1).

## What's locked in v1

- Transport: stdio with LSP-style \`Content-Length\` framing, UTF-8 only.
- Base: JSON-RPC 2.0 with five daemon-specific error codes in the reserved -32000..-32099 range (\`NotInitialized\`, \`ClasspathDirty\`, \`SandboxRecycling\`, \`UnknownPreview\`, \`RenderFailed\`).
- Lifecycle: \`initialize\` (request) → \`initialized\` (notification) → traffic → \`shutdown\` (request) → \`exit\` (notification).
- Client → daemon: \`setVisible\`, \`setFocus\`, \`fileChanged\` notifications; \`renderNow\` request.
- Daemon → client: \`discoveryUpdated\`, \`renderStarted\` / \`renderFinished\` / \`renderFailed\`, \`classpathDirty\`, \`sandboxRecycle\`, \`daemonWarming\` / \`daemonReady\`, \`log\`.
- Versioning: \`protocolVersion: 1\`. Additive changes don't bump; breaking changes do and require coordinated daemon + extension release.

## Test plan

- [ ] Render the docs locally / on GitHub to verify all internal links resolve (PROTOCOL.md ↔ DESIGN.md ↔ TODO.md ↔ protocol-fixtures/README.md).
- [ ] Sanity-check schema field names against the corresponding existing types: \`PreviewInfo\` shape vs. \`gradle-plugin/.../PreviewData.kt\`, render strategy vs. \`AndroidPreviewSupport.kt\`.
[...]

PR #250, opened 2026-04-26
refactor(plugin): hoist renderPreviews classpath/JVM-args helpers (P0.3 + P0.2 wrap-up)

yschimke:
> Stacked on top of #249 (P0.4). Review-and-merge order is #249 → this PR.

## Summary

- Extracts the renderPreviews test classpath, static JVM open flags, and static system-property set out of the inline \`project.tasks.register(\"renderPreviews\", Test::class.java) { ... }\` block in [AndroidPreviewSupport.kt](gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt) into a new package-private helper [AndroidPreviewClasspath.kt](gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewClasspath.kt). Each helper returns data; nothing calls Test DSL methods.
- Wraps up **P0.2** while in the area: \`PreviewInfo.sourceFile\` was already shipped before the daemon work breakdown was written, so this PR adds the missing functional-test guard (\`DiscoveryFunctionalTest\` asserts non-blank for at least one preview) and marks P0.2 done in [TODO.md](docs/daemon/TODO.md).

## Why

The upcoming preview daemon (see [docs/daemon/DESIGN.md](docs/daemon/DESIGN.md)) needs to construct the same classpath / JVM args / sysprops to launch its own JVM. Sharing the construction logic now means we don't duplicate the inline DSL block when wiring \`DaemonBootstrapTask\` (Phase 1 / Stream A / A1.1).

## What stays inline

The Test task lambda still:

- Composes the final classpath (appends AGP unit-test classes / classpath; only resolvable late via \`findByName(\"test\${Cap}UnitTest\")\`).
- Inherits AGP's \`javaLauncher\`.
- Registers the dynamic \`ArgumentProvider\`s (a11y, tier) which need lazy \`Provider<>\` evaluation at task-execution time to avoid config-cache invalidation on \`-P\` toggles.
- Copies AGP's test-task JVM args.

## Verification

- \`./gradlew :samples:android:renderAllPreviews\` — **pass**, no PNG diffs vs. main (\`git status\` clean across \`samples/android/\`).
- \`./gradlew :gradle-plugin:test\` — **pass**.
- \`./gradlew :gradle-plugin:functionalTest\` — **pass**, including the new \`sourceFile\` non-null assertion.

File ordering inside the classpath \`FileCollection\`, JVM-arg order, and the system-property set match the previous inline construction byte-for-byte.

## Test plan

- [ ] CI green on this branch.
- [ ] Render a preview locally and confirm output matches main (only needed if a reviewer doubts the byte-identical claim — \`renderAllPreviews\` already covers it).

### DIFF
--- a/docs/daemon/TODO.md
+++ b/docs/daemon/TODO.md
@@ -29,14 +29,16 @@ Build `:samples:android-daemon-bench` (skeleton sample module) with a `benchPrev
 - **Depends on:** none
 - **DoD:** CSV checked into `docs/daemon/baseline-latency.csv` for the reference dev machine. Numbers referenced from `DESIGN.md` § 13 are validated or corrected.
 
-### P0.2 — Add `sourceFile` to `PreviewInfo` [Stream A] [shared seam]
+### P0.2 — Add `sourceFile` to `PreviewInfo` [Stream A] [shared seam] ✅
 
 Extend `PreviewInfo` (in `gradle-plugin/.../PreviewData.kt`) with `sourceFile: String?` populated from `ClassInfo.sourceFile` (ClassGraph exposes the bytecode `SourceFile` attribute). Wire through `DiscoverPreviewsTask`. Update `previews.json` schema.
+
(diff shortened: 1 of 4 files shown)


yschimke:
Closing per author request — keeping branch agent/preview-daemon-sourcefile for continued chain work.

PR #395, opened 2026-05-01
feat(daemon): interactive/start|stop|input RPC + frame dedup

yschimke:
Daemon-side counterpart to #394's panel work. Implements the `interactive/*` RPC surface
the VS Code extension already drives, plus a generally-useful frame deduplication pass on
every `renderFinished`.

## Summary

- **`interactive/start` (request, client → daemon)** — `{ previewId } → { frameStreamId }`. Stashes a single `(previewId, frameStreamId)` as the active interactive target and wipes the cached frame hash for that preview so the bootstrap frame paints. Blank previewIds reject with `InvalidParams (-32602)`.
- **`interactive/stop` (notification)** — `{ frameStreamId }`. Idempotent on stale or unknown ids.
- **`interactive/input` (notification)** — `{ frameStreamId, kind, pixelX?, pixelY?, keyCode? }`. Drops silently when the streamId doesn't match the active target. Otherwise enqueues a render of the target preview through the same `submitRenderAsync` path `renderNow` uses; the result demuxes through the existing render-watcher thread back into `renderFinished`.
- **Frame deduplication** — every `renderFinished` SHA-256s the rendered PNG bytes; on match against the prior hash for the same `previewId`, sets `renderFinished.unchanged: true` and skips the history archive. Tracked always (not just under interactive mode): a save that doesn't move pixels is just as redundant as an input that doesn't change state.
- **Click forwarding wired in `extension.ts`** — `recordInteractiveClick` from the panel now feeds `client.interactiveInput(...)`; falls back to legacy `setFocus + renderNow` when the daemon doesn't speak the new RPC (`MethodNotFound`). The `unchanged: true` field short-circuits `DaemonScheduler.handleRenderFinished` so the panel never re-paints identical bytes.

## Wire compatibility

Additive per [PROTOCOL.md § 7](docs/daemon/PROTOCOL.md#7-versioning) — no `protocolVersion` bump.
- Old daemons reject `interactive/start` with `MethodNotFound`; the panel logs and falls back.
- Old clients drop `unknown notifications` and ignore the unknown `unchanged` field on `renderFinished`.

## What today's `interactive/input` does — and what it doesn't

v1 routes the click coords as far as the daemon and triggers a fresh render of the target preview, but the renderer body still composes with `LocalInspectionMode = true`, so the input does **not** reach the active composition. v2 flips that local during interactive frames and dispatches the pixel coords through `ImageComposeScene`'s pointer-input pipeline. This lets the wire shape settle before any renderer surgery — see `docs/daemon/INTERACTIVE.md` § 8 "Open questions".

## Design + spec

Updated `docs/daemon/INTERACTIVE.md`:
- Status bumped to v1.
- New § 5 "Frame deduplication" documents the SHA-256 / `unchanged: true` pass and the wipe-on-start trick.
- § 8 documents the v1 dispatch flow and the `LocalInspectionMode` v2 follow-up.
- § 9 marks "click-into-composition" out of scope for v1.

## Tests

- New `daemon/core/.../InteractiveRpcIntegrationTest.kt` — drives a `JsonRpcServer` over piped streams against a deterministic bytes-aware fake host:
  - **`click_input_triggers_renderFinished_then_dedups_identical_frame`** — full lifecycle: start → input → first frame paints → second input on identical bytes → `unchanged: true` → swap bytes → third input → fresh frame → stop → stale input is dropped.
  - **`start_with_blank_previewId_rejects_with_invalid_params`** — rejects blank ids.
  - **`start_overwrites_prior_target_so_input_against_stale_streamId_is_dropped`** — second start invalidates the prior streamId; inputs against it are dropped.
- New `vscode-extension/.../daemonScheduler.test.ts` test asserts `unchanged: true` short-circuits the read+post hop without producing a failure path.

All 295 existing TS tests still pass; new suite adds 4 tests across the two ecosystems.

Pre-existing `LocalFsHistorySourcePruneTest.dedup_by_hash_…` was already failing on `main` before this PR (date-encoded fixture; unrelated).

## Follow-up — multi-target support (left for the next agent)

The daemon currently enforces single-target via:

```kotlin
private val interactiveTarget = AtomicReference<InteractiveTarget?>(null)
```

The wire shape doesn't preclude multiple concurrent streams (each `interactive/start` returns a fresh monotonic `streamId`), the renderer can already serve concurrent renders for different previews, and `lastFrameHashes` is already keyed per-previewId — so the dedup story carries through unchanged. Lifting this is roughly:

1. **`JsonRpcServer.kt`** — swap the `AtomicReference` for `ConcurrentHashMap<String /*streamId*/, InteractiveTarget>`. `handleInteractiveStart` does `put` instead of `set`. `handleInteractiveStop` does `remove(streamId)`. `handleInteractiveInput` looks up by `streamId` instead of comparing against a single slot.
2. **Tests** — replace the "second start overwrites prior" test with a "two concurrent streams route inputs independently; stop on one doesn't affect the other" test.
3. **`docs/daemon/INTERACTIVE.md`** — drop the "single-target by design" / "Multiple `start` calls overwrite the prior target" lines from §§ 7, 8, 9. Note that the UI deliberately keeps a single `.live` chip for now; multi-target is a daemon-protocol capability for hypothetical programmatic clients (e.g. side-by-side comparison views, CI agents driving multiple previews over one stdio pair).
4. **No changes to `extension.ts` / `previewPanel.ts`** — the panel's `interactivePreviewId` is intentionally kept as a single value; the click-forwarding map is already keyed per-preview so it works either way.

Should be ~100 lines across the three layers, no protocol bump required. The reason it's split out is purely scope discipline — the v1 wire shape is the load-bearing part, and the multi-target lift is an implementation refinement that doesn't touch the contract.

## Test plan

- [x] `LANG=C.UTF-8 LC_ALL=C.UTF-8 ./gradlew :daemon:core:test --tests "ee.schimke.composeai.daemon.InteractiveRpcIntegrationTest"` — 3/3 pass
- [x] `LANG=C.UTF-8 LC_ALL=C.UTF-8 ./gradlew :daemon:core:test --tests "ee.schimke.composeai.daemon.JsonRpcServerIntegrationTest"` — pre-existing suite still green
- [x] `cd vscode-extension && npm test` — 296/296 pass (including the new dedup short-circuit test)
- [x] `LANG=C.UTF-8 LC_ALL=C.UTF-8 ./gradlew :daemon:core:ktfmtFormatMain :daemon:core:ktfmtFormatTest` — clean
- [ ] Manual smoke test in a running extension host (LIVE toggle → click → daemon roundtrip) — not exercised here

---
_Generated by [Claude Code](https://claude.ai/code/session_017NDGZk17VhvctNNoSrVZsK)_

### DIFF
--- a/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/JsonRpcServer.kt
+++ b/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/JsonRpcServer.kt
@@ -234,23 +234,31 @@ class JsonRpcServer(
   // ----------------------------------------------------------------------
   // Interactive (live-stream) mode state — see docs/daemon/INTERACTIVE.md.
   //
-  // [interactiveTarget] is the single (previewId, frameStreamId) tuple the daemon is
-  // serving as the user's interactive target. Single-target by design: a second
-  // `interactive/start` overwrites this slot, and any input notification carrying a
-  // stale frameStreamId is silently dropped.
+  // [interactiveTargets] is the set of currently-active interactive streams, keyed by
+  // [InteractiveTarget.frameStreamId]. The wire shape is multi-target by design: each
+  // `interactive/start` registers a fresh slot and returns a unique stream id; concurrent
+  // streams targeting different (or even the same) preview ids coexist. Inputs route by
+  // stream id, so a stale id from a `stop`'d or never-started stream is dropped without
+  // touching live targets.
+  //
+  // The current panel UI is single-target — only one card carries `.live` at a time —
+  // but the protocol does not require that. Lifting it on the wire keeps the daemon
+  // useful for hypothetical programmatic clients (side-by-side comparison views, CI
+  // agents driving multiple previews over one stdio pair) without any contract change.
   //
   // [lastFrameHashes] is a per-previewId cache of the SHA-256 of the most recently
   // notified PNG bytes. Populated on every `renderFinished` that the watcher actually
   // emits; consulted to set `unchanged: true` (and skip history) when the next render
   // produces byte-identical output. Always tracked, not just when interactive mode is
-  // on — the dedup is generally useful (a save that doesn't move pixels shouldn't
-  // burn IPC + a panel repaint).
+  // on — the dedup is generally useful (a save that doesn't move pixels shouldn't burn
+  // IPC + a panel repaint). Per-preview key, not per-stream: two streams targeting the
+  // same preview share dedup state, which is what we want — the bytes are the same
+  // bytes regardless of which stream's input triggered the render.
   // ----------------------------------------------------------------------
 
   private data class InteractiveTarget(val previewId: String, val frameStreamId: String)
 
-  private val interactiveTarget =
-    java.util.concurrent.atomic.AtomicReference<InteractiveTarget?>(null)
+  private val interactiveTargets = ConcurrentHashMap<String, InteractiveTarget>()
 
   private val lastFrameHashes = ConcurrentHashMap<String, String>()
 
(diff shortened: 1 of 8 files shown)


yschimke:
Superseded by #400 (which landed). Closing.

---

PR #396, opened 2026-05-01
feat(daemon): support multiple concurrent interactive streams

yschimke:
**Stacked on #395.** Targets that PR's branch so the diff here shows only the multi-target lift; rebase to `main` once #395 merges.

## What changes

Lifts the daemon-side single-target enforcement that #395 documented as a "for the next agent" follow-up. The wire shape didn't preclude multi-target — each `interactive/start` already returned a fresh monotonic streamId — so no protocol bump.

- **State**: `AtomicReference<InteractiveTarget?>` → `ConcurrentHashMap<String /*streamId*/, InteractiveTarget>`.
- **`interactive/start`**: `put(streamId, …)` instead of `set(...)`. Each start registers a fresh slot; concurrent streams targeting different (or even the same) preview ids coexist. Dedup hash wipe stays per-preview by design — two streams on the same preview share dedup state because the bytes are the same bytes regardless of which stream's input triggered the render.
- **`interactive/stop`**: `remove(streamId)`. Idempotent on stale or unknown ids — `remove` is a no-op when the key isn't present, which is exactly the contract we want.
- **`interactive/input`**: `interactiveTargets[streamId] ?: return`. Stale or unknown stream ids drop silently without touching live targets.

## Why the panel UI stays single-target

The webview's `interactivePreviewId` stays a single value and only one card carries `.live`. The argument for surfacing multi-target to a human is weak — the user can only meaningfully focus on one live preview at a time, and a multi-`.live` view would mostly burn renders without changing what the user is looking at. Programmatic clients are the load-bearing case for multi-target on the wire today (side-by-side comparison views, CI agents driving multiple previews over one stdio pair).

`INTERACTIVE.md § 9` documents this: wire supports multi-target, UI doesn't surface it.

## Test changes

Replaces `start_overwrites_prior_target_so_input_against_stale_streamId_is_dropped` (which asserted the old single-target invariant) with `multiple_concurrent_streams_route_inputs_independently`:

1. Start on preview-A → streamA; start on preview-B → streamB. Assert distinct streamIds.
2. Input on streamA → renderFinished for preview-A.
3. Input on streamB → renderFinished for preview-B (proves both streams alive).
4. Stop streamA. Stale input on streamA → no renderFinished within the wait window.
5. Input on streamB → still renders preview-B (untouched by A's stop).

The other two tests (`click_input_triggers_renderFinished_then_dedups_identical_frame`, `start_with_blank_previewId_rejects_with_invalid_params`) are unchanged — single-stream behaviour is identical to #395.

## Test plan

- [x] `LANG=C.UTF-8 LC_ALL=C.UTF-8 ./gradlew :daemon:core:test --tests "ee.schimke.composeai.daemon.InteractiveRpcIntegrationTest"` — 3/3 pass
- [x] `LANG=C.UTF-8 LC_ALL=C.UTF-8 ./gradlew :daemon:core:test --tests "ee.schimke.composeai.daemon.JsonRpcServerIntegrationTest"` — pre-existing suite still green
- [x] `LANG=C.UTF-8 LC_ALL=C.UTF-8 ./gradlew :daemon:core:ktfmtFormatMain :daemon:core:ktfmtFormatTest` — clean
- [ ] Manual smoke test with two simultaneous programmatic streams against the daemon — not exercised here (no current consumer)

---
_Generated by [Claude Code](https://claude.ai/code/session_017NDGZk17VhvctNNoSrVZsK)_

### DIFF
--- a/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/JsonRpcServer.kt
+++ b/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/JsonRpcServer.kt
@@ -234,23 +234,31 @@ class JsonRpcServer(
   // ----------------------------------------------------------------------
   // Interactive (live-stream) mode state — see docs/daemon/INTERACTIVE.md.
   //
-  // [interactiveTarget] is the single (previewId, frameStreamId) tuple the daemon is
-  // serving as the user's interactive target. Single-target by design: a second
-  // `interactive/start` overwrites this slot, and any input notification carrying a
-  // stale frameStreamId is silently dropped.
+  // [interactiveTargets] is the set of currently-active interactive streams, keyed by
+  // [InteractiveTarget.frameStreamId]. The wire shape is multi-target by design: each
+  // `interactive/start` registers a fresh slot and returns a unique stream id; concurrent
+  // streams targeting different (or even the same) preview ids coexist. Inputs route by
+  // stream id, so a stale id from a `stop`'d or never-started stream is dropped without
+  // touching live targets.
+  //
+  // The current panel UI is single-target — only one card carries `.live` at a time —
+  // but the protocol does not require that. Lifting it on the wire keeps the daemon
+  // useful for hypothetical programmatic clients (side-by-side comparison views, CI
+  // agents driving multiple previews over one stdio pair) without any contract change.
   //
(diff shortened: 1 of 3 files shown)


yschimke:
Superseded by #400 (which landed) — the multi-target lift is included in the squashed commit there. Closing.

---

PR #417, opened 2026-05-01
feat(daemon): held compose scene for interactive sessions (v2 PR-2)

yschimke:
## Summary

PR-2 of the Interactive v2 rollout — desktop-side held scene so `interactive/input` actually mutates the composition. Implements [INTERACTIVE.md § 9](docs/daemon/INTERACTIVE.md#9-v2--click-dispatch-into-composition).

- `RenderEngine.render` is split into `setUp` / `renderOnce` / `tearDown`; the existing one-shot `render` becomes a thin wrapper. Held scenes are the new `HeldScene` data class.
- New `DesktopInteractiveSession` implements the `:daemon:core` `InteractiveSession` interface — `dispatch(input)` routes pixel coords through `ImageComposeScene.sendPointerEvent` (with `LocalInspectionMode = false` so `pointerInput` modifiers fire), `render()` re-encodes a PNG, `close()` tears down.
- `DesktopHost.acquireInteractiveSession` returns a per-stream proxy over a ref-counted shared `DesktopInteractiveSession`; two `interactive/start` calls on the same `previewId` share one composition state per § 9.7.
- Android keeps the default `UnsupportedOperationException` — `JsonRpcServer.handleInteractiveStart` translates that to v1 fallback.

## Notable deviations from the spec

1. **Inter-event render between Press and Release.** `Modifier.clickable`'s gesture detector cancels events that arrive in the same dispatch tick — `dispatch(CLICK)` does `sendPointerEvent(Press) → scene.render() → sendPointerEvent(Release)`. The post-input `render()` then settles the post-Release frame normally. Wire shape unchanged.
2. **`PointerButton.Primary` on every Press/Release.** § 9.4's sketch elides the `button` parameter; `Modifier.clickable` only fires for primary clicks, so the session sets it explicitly.
3. **Bootstrap render in `acquireInteractiveSession`.** Compose's hit-testing has no measured tree until the first render — bootstrap pixels aren't shipped, only the layout pass matters.
4. **`resolveInteractiveSpec` hook on `DesktopHost`.** § 9.8's interface only carries `previewId` + `classLoader`; production needs a way to map previewId → `RenderSpec`. Added a `protected open fun resolveInteractiveSpec(previewId): RenderSpec?`. Default returns `null` (→ `UnsupportedOperationException` → v1 fallback). `PreviewManifestRouter` overrides it; tests subclass to pin a fixture.
5. **Per-call context-classloader install.** The original one-shot installed once around the whole render. The split moves install responsibility outward — `dispatch` and `render` install on whatever thread the JsonRpcServer worker uses.
6. **Headless AWT.** Key dispatch needs an AWT `Component` as event source — switched from `Label` (which throws `HeadlessException`) to a lazy `Canvas()` wrapped in try/catch. Click events don't touch this path.
7. **Pixel tolerance 32 (vs `RenderEngineTest`'s 8)** in the new integration test — `ImageComposeScene` renders against transparent and unpremultiply drift is several LSB. Still trivially distinguishes red ↔ green.

## Test plan

- [x] `:daemon:desktop:test` — 18 tests green, including 2 new:
  - `DesktopInteractiveSessionTest` — unit-level dispatch/render/close coverage.
  - `InteractiveDesktopRpcIntegrationTest` — full JSON-RPC round trip against a stateful `ClickToGreenSquare` fixture (red on first frame, green after one click).
- [x] `:daemon:core:test` — all green, including the existing `InteractiveSessionPlumbingTest` and `InteractiveRpcIntegrationTest`.
- [ ] Manual: drive an interactive session via the panel (post-merge) once a v2-supporting host advertises capability.

## Follow-ups (not in this PR)

- Wire a `interactive: Boolean` capability bit on `InitializeResult.capabilities` so clients can render an "unsupported host" hint cleanly. Small follow-up PR.
- Widen `PreviewInfoDto` in `:daemon:core` with `widthDp` / `heightDp` / `density` from the gradle plugin's discovery output so interactive scenes match `@Preview(widthDp=…)` exactly. Today the `resolveInteractiveSpec` hook gives subclasses a place to thread this; the wire/discovery widening is a separate PR.
- `LocalInspectionMode` opt-in story (currently hardcoded `false` for v2 sessions). Some previews assume `true`; design call pending.
- Robolectric `acquireInteractiveSession` (Android v3) — INTERACTIVE.md § 9.10 already calls this out as its own design pass.
- Pin `LANG=C.UTF-8 LC_ALL=C.UTF-8` in `gradle.properties` — the Kotlin compile daemon defaults to ASCII for `sun.jnu.encoding` and chokes on em-dashes in `@Test` names (pre-existing repo-wide; only affects fresh checkouts that don't inherit the user's locale).
- `RenderEngine.setUp` could opt the scene into an opaque background when `showBackground=true` to avoid the unpremultiply drift that forced tolerance 32 on the new tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WYvTgEGDJxSu9gAGUTF9ub)_

### DIFF
--- a/daemon/desktop/src/main/kotlin/ee/schimke/composeai/daemon/DesktopHost.kt
+++ b/daemon/desktop/src/main/kotlin/ee/schimke/composeai/daemon/DesktopHost.kt
@@ -78,6 +78,24 @@ open class DesktopHost(
   override val userClassloaderHolder: UserClassLoaderHolder? = null,
 ) : RenderHost {
 
+  /**
+   * v2 — interactive session sharing per `previewId`, ref-counted by the number of subscribed
+   * streams (INTERACTIVE.md § 9.7 — "shared session per previewId, ref-counted by streamId").
+   * Two `interactive/start` calls for the same preview share one held [HeldScene] and one
+   * underlying [DesktopInteractiveSession]; the session lives until the last subscribed stream
+   * stops. Inputs from any stream drive the same composition state.
+   *
+   * Keyed by previewId. Each acquire returns a fresh per-stream proxy [InteractiveSession] whose
+   * [InteractiveSession.close] decrements the refcount; the inner shared session is torn down
+   * once the count hits zero.
+   */
+  private val sharedInteractiveSessions = ConcurrentHashMap<String, SharedInteractiveSession>()
+
+  private data class SharedInteractiveSession(
+    val inner: DesktopInteractiveSession,
+    val refCount: java.util.concurrent.atomic.AtomicInteger,
+  )
+
   private val requests: LinkedBlockingQueue<RenderRequest> = LinkedBlockingQueue()
   private val results: ConcurrentHashMap<Long, LinkedBlockingQueue<Any>> = ConcurrentHashMap()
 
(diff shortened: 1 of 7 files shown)


yschimke:
Closing as a duplicate of #408 — that PR landed a parallel implementation of INTERACTIVE.md § 9 PR-2 against the same files (`DesktopHost.kt`, `DesktopInteractiveSession.kt`, `RenderEngine.kt` split, `DesktopInteractiveSessionTest.kt`, `RedFixturePreviews.kt`). Rebasing this PR would produce a duplicate after resolving 5 conflicts, which isn't worth the cost.

One bit on this branch isn't on main and is worth lifting separately: `InteractiveDesktopRpcIntegrationTest.kt` — a full RPC-level round-trip that asserts a `click_reaches_composition_red_becomes_green`. The existing `DesktopInteractiveSessionTest` is unit-level (no JsonRpcServer) and `JsonRpcDesktopIntegrationTest` is non-interactive. Will salvage as a tiny follow-up PR.

The capability-bit follow-up (review #5: `InitializeResult.capabilities.interactive`) survives the parallel landings and will land as its own PR off main.

---

PR #418, opened 2026-05-01
feat(vscode): auto-subscribe a11y/atf for diagnostics in daemon mode

yschimke:
## Summary

Splits the D2 a11y kinds along the design doc's "ambient vs. opt-in" axis so `PreviewA11yDiagnostics` works in daemon mode without requiring the user to enter focus mode and toggle the overlay on.

- `a11y/atf` is the cheap-always-on kind. Auto-subscribe per visible preview (driven by `setVisible`) so findings flow into the `PreviewRegistry` ambient — the diagnostic squigglies in the Problems panel read straight off it.
- `a11y/hierarchy` stays toggle-gated via the focus-mode "Show accessibility overlay" button. It's the heavier visual layer; no point firing it for cards outside the spotlight.

`PreviewA11yDiagnostics` itself doesn't change — it already reads `entry.preview.a11yFindings` from the registry and listens to `registry.onDidChange`. The migration is making sure the daemon path feeds the registry ambient, matching what the Gradle sidecar reader already does at manifest-load time. Both paths now converge on the same registry; diagnostics is source-agnostic.

## Test plan

- [x] `vscode-extension/` `npm test` — 317 tests pass; scheduler tests updated to assert the split (setVisible auto-subscribes atf only; hierarchy absent until the toggle drives it).
- [x] TypeScript compile clean (`npm run compile`).
- [ ] Manual smoke (consumer side): start daemon-mode panel without focus, verify ATF-derived squigglies appear on `@Preview` annotations whose composables fail an ATF check; enter focus + click eye toggle, verify hierarchy overlay also paints.

Spec: [`docs/daemon/DATA-PRODUCTS.md`](https://github.com/yschimke/compose-ai-tools/blob/main/docs/daemon/DATA-PRODUCTS.md) § "Catalogue" — `a11y/atf` is flagged as low-cost / candidate for global attach; this PR realises that.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NVoxnr38t4n1LEatFc1yGe)_

### DIFF
--- a/vscode-extension/src/daemon/daemonScheduler.ts
+++ b/vscode-extension/src/daemon/daemonScheduler.ts
@@ -64,11 +64,20 @@ export interface SchedulerEvents {
 const HEAVY_TIER_DEFAULT: RenderTier = 'fast';
 
 /**
- * D2 — kinds the focus-mode "Show a11y overlay" button toggles. Pinned to the a11y producer
- * so the local finding + hierarchy overlays light up together; a future panel that wants to
- * subscribe to other kinds (`compose/recomposition`, `layout/tree`) will export its own list.
+ * D2 — kinds the focus-mode "Show accessibility overlay" toggle adds on top of the ambient
+ * subscriptions. `a11y/atf` is ambient (see [AMBIENT_A11Y_KINDS]) so the diagnostic squigglies
+ * stay live without a user gesture; the toggle layers `a11y/hierarchy` for the visual overlay.
  */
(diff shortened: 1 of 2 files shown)


yschimke:
Abandoning — defaulting a11y on per visible preview walks back from the design doc's "Default = nothing attached" stance, and the cost is real (every render runs ATF and flips `LocalInspectionMode = false`, which interrupts animation parking). Opt-in via the focus-mode toggle is the right baseline; users who want ambient diagnostics in daemon mode can opt in via the eye-icon button. Gradle mode keeps the sidecar reader so diagnostics stay ambient there as today.

---

PR #595, opened 2026-05-02
Move scroll screenshots to data products

yschimke:
## Summary
- map @ScrollingPreview LONG/GIF modes to annotation-sourced data product specs instead of primary captures
- render scroll data products under build/compose-previews/data while keeping TOP/END as normal captures
- include data-product cost in renderer sharding and VS Code heavy-output stale detection

## Tests
- ./gradlew :gradle-plugin:compileKotlin :renderer-android:compileDebugKotlin :gradle-plugin:test --tests ee.schimke.composeai.plugin.PreviewDataTest :renderer-android:testDebugUnitTest --tests ee.schimke.composeai.renderer.PreviewManifestLoaderShardTest

Note: npm run compile in vscode-extension is blocked in this workspace because node_modules is missing @types/mocha and @types/node.

### DIFF
--- a/docs/daemon/DATA-PRODUCTS.md
+++ b/docs/daemon/DATA-PRODUCTS.md
@@ -558,6 +558,24 @@ for rectangular/default devices, or:
 for round Wear-style devices. It does not require a render pass and is
 available on both Android and desktop daemons.
 
+`render/scroll/long` and `render/scroll/gif` are annotation-sourced render
+products produced from `@ScrollingPreview(modes = [LONG, GIF])`. The typed
+annotation remains the authoring API; discovery maps those modes to data-product
+requests instead of adding long/GIF artefacts to the primary capture list. Gradle
+renders write path-backed product files under `build/compose-previews/data/`:
+
+```json
+{
+  "kind": "render/scroll/long",
+  "schemaVersion": 1,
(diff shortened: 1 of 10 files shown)


yschimke:
Status update:

- Pushed the current branch state: `feature/scroll-captures-data-products` at `042c5bba`.
- Added the missing `data/scroll/core` module next to `data/a11y/core`.
- Moved the reusable scroll product implementation out of `renderer-android` into `data-scroll-core`:
  - scroll drivers
  - LONG screenshot slice stitching / Wear pill clipping
  - GIF scroll script + encoder
  - core scroll tests for GIF script and stitcher
- Renderer now depends on `:data-scroll-core` and only adapts renderer-specific preview metadata / composition setup into the scroll product producer calls.

Verification so far:
- PASS: `./gradlew :data-scroll-core:compileDebugKotlin :renderer-android:compileDebugKotlin`
- Earlier targeted tests for manifest/sharding passed before this module split.
- Known existing failing regression remains: `LongScrollGhostOverlayTest > Wear anchor path cuts at last list item regardless of pill jitter`. I left that test under `renderer-android` rather than moving it into `data-scroll-core` so the new product module is not immediately blocked by the known failure.

Current caveat:
- This is now structurally aligned with the requested data-product module layout, but I have not yet rerun the full targeted test matrix after the split beyond the compile check above.

Issue #675, opened 2026-05-03
State Restoration Checks


yschimke:
Check that resume builds the same UI, after some edit.

yschimke:
I looked at the state restoration path and put up a proposed harness design in #681.

Conclusion: `StateRestorationTester` is the right model for a backend-level Android test, but it is not enough by itself for this daemon issue. The end-to-end check should also exercise the held interactive session, `fileChanged(source)`, forced session teardown, user-classloader swap, and restore into the new composition.

The proposed S11 flow is: mutate a `rememberSaveable` counter in an interactive preview, snapshot primitive saveable state, edit/recompile a visible version label, send `fileChanged(source)`, assert the old held session closes, restart the preview, and verify the resumed render shows both the old saved state and the new bytecode.

yschimke:
Updated conclusion after review: do not land docs or agent guidance for unimplemented state-restoration commands.

The implementation direction should be:
- use the existing script/recording mode as the user-facing surface for lifecycle/state events;
- add real script events such as save state, restore state, Activity recreate/resume, reload preview, and probe semantics/text;
- emit structured evidence from an actual run before adding this to the agent audit guidance;
- only add a convenience check once the primitive script events work.

I closed #681 because it was documenting proposed commands rather than implemented behavior.

PR #681, opened 2026-05-03
Design state restoration extension primitives

yschimke:
## Summary

- Rework the issue #675 state-restoration design around the newer composable preview-extension pipeline instead of a harness-only special case.
- Make `state-restoration` primitive-first for agents: `probe`, `save`, `restore`, `lifecycle`, and `reload` produce `test/stateRestoration` evidence.
- Treat `state-restoration.verify` as an optional composed policy over those primitives, not the primary/only interface.
- Document the explicit source-edit flow: mutate -> save checkpoint -> source edit/recompile -> `fileChanged(source)` -> reload -> restore -> probe.
- Add lifecycle helper direction for script/recording mode and update the consumer-facing state restoration audit guidance with a sample restorable counter composable, MCP call sequence, and tried command-surface note.

## Validation

- `./gradlew :cli:installDist`
- `cli/build/install/compose-preview/bin/compose-preview --help`
- `cli/build/install/compose-preview/bin/compose-preview extensions --help` (confirms extension commands are MCP-first today)
- `git diff --check`

Refs #675

### DIFF
--- a/docs/daemon/DATA-PRODUCTS.md
+++ b/docs/daemon/DATA-PRODUCTS.md
@@ -766,6 +766,75 @@ renders write path-backed product files under `build/compose-previews/data/`:
 showing the scroll from top to bottom. Non-scrollable previews no-op through the
 same fallback behaviour as the renderer's scroll capture path.
 
+`test/stateRestoration` is a proposed Android-first evidence product owned by a
+future `state-restoration` preview extension. It should be produced by
+agent-invoked lifecycle/state primitives first, with higher-level checks layered
+on top later. The extension uses composable hooks to install or observe
+state-restoration probes, then asks the Android host to run operations such as:
+
+- local `rememberSaveable` save/restore, equivalent to AndroidX Compose
+  `StateRestorationTester`;
+- mocked `LifecycleOwner` plus `SavedStateRegistryOwner` read/write through
+  `SavedStateRegistryController`;
+- true Activity background/resume or `ActivityScenario.recreate()` when the
+  host can launch the preview in an Activity-backed composition.
+
+Proposed primitive commands:
+
+```text
+compose-preview extensions run state-restoration.probe --id <preview-id> --json
+compose-preview extensions run state-restoration.save --id <preview-id> --json
+compose-preview extensions run state-restoration.reload --id <preview-id> --json
+compose-preview extensions run state-restoration.restore --id <preview-id> --checkpoint <id> --json
+compose-preview extensions run state-restoration.lifecycle --id <preview-id> --event recreateActivity --json
+```
+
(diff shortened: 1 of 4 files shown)


yschimke:
Updated the design to add that as the first backend-level rung: test the composable with mocked `LifecycleOwner` and `SavedStateRegistryOwner`, backed by `LifecycleRegistry` and `SavedStateRegistryController`; save into a `Bundle`; recreate the owner from that bundle; and assert the composable reads the saved value. That gives us a cheap read/write proof before the `StateRestorationTester` recreation check and the full daemon/classloader harness scenario.

yschimke:
Reworked this for the current composable preview-extension model.

Key changes:
- Treat state restoration as a future `state-restoration` preview extension with command `state-restoration.verify` and data product `test/stateRestoration`, rather than a harness-only one-off.
- Split the evidence ladder into local `StateRestorationTester`, mocked lifecycle/saved-state owners, real Activity resume/recreate, and the daemon source-edit/classloader rebuild scenario.
- Added direction that script/recording mode should grow lifecycle helpers (`saveState`, `resume`, `recreateActivity`, etc.) because pointer/key events alone cannot verify restoration.
- Added a consumer-facing state restoration/lifecycle audit recipe for agents in the compose-preview-review skill docs.

yschimke:
Updated again to make the extension primitive-first for agents.

The design now leads with reusable operations: `state-restoration.probe`, `save`, `restore`, `lifecycle`, and `reload`, each emitting `test/stateRestoration` evidence. `state-restoration.verify` is now explicitly optional: a later convenience policy composed over those primitives, not a separate hidden check path.

The source-edit workflow is documented as an explicit checkpoint sequence: mutate -> save -> source edit/recompile -> `fileChanged(source)` -> reload -> restore -> probe. The agent audit guidance now prefers citing those primitive operation results over a single pass/fail wrapper.

yschimke:
Added the concrete audit example requested:

- sample `RestorableCounterPreview` with both `rememberSaveable` and plain `remember` counters;
- MCP primitive sequence using `probe`, `record_preview` for mutation, `save`, `lifecycle(recreateActivity)`, `restore`, and final `probe`;
- expected evidence: `saved` returns to 1 while `volatile` resets to 0;
- note from local trial that `compose-preview extensions --help` is not a cold-shell CLI command today and extension command routing is MCP-first.

Also updated PR validation to include the CLI install/help commands I ran.

yschimke:
Closing this rather than keeping a PR that documents unimplemented `state-restoration` commands/events. The useful design conclusion belongs on #675 until there is working script-mode support or an implementation PR.

Implementation direction to carry forward: use the existing script/recording mode for lifecycle/state events rather than separate commands, and only update agent audit guidance once those events can actually be run.

PR #1304, opened 2026-05-21
feat(daemon-desktop): touch-event visualization extension + multi-pointer pinch dispatch

yschimke:
`TouchOverlayExtension` is an `AroundComposableHook` registered as the
`data/touch-overlay` data extension. When `overrides.touchOverlay = true`
(or implicitly for `live = true` recording sessions, where it's the only
ground truth the agent has for where their dispatched touches landed),
the extension wraps the held composition with a `Modifier.pointerInput`
observer on the Initial pass and a `Canvas` overlay above the content.
Active pointers get a translucent cyan ring + crosshair; down/up events
emit short-lived expanding pulses on Compose's frame clock. Same shape
as Android's "Show touches" developer-mode toggle.

The around-composable form is portable — it works on any backend that
runs Compose, observes every event the composition sees regardless of
whether dispatch came from the recording session, an `interactive/input`
notification, or future direct UI input, and fades on the same virtual
frame clock as the rest of the preview. No PNG post-processing, no
per-backend special cases in the renderer or recording session.

`DesktopRecordingSession` now dispatches pointer events via the
multi-pointer `sendPointerEvent(pointers: List<ComposeScenePointer>)`
overload, keyed by a new `pointerId` field on `RecordingScriptEvent` /
`InteractiveInputParams`. This is what makes pinch-to-zoom (and any
other two-finger gesture) actually reach `Modifier.transformable`'s
zoom callback — without it, two simultaneous `pointerDown`s arrive as
two independent single-pointer gestures that the detector ignores.

Integration test scripts a 5-step two-finger outward pinch against a
`PinchableSquare` fixture, asserts the captured frames carry the cyan
overlay ring pixels (visualization fired) and that the blue square's
post-pinch coverage exceeds its initial coverage (zoom callback fired,
multi-pointer dispatch worked). The test also stitches the PNG frames
into `build/touch-overlay-artifacts/touch-overlay-pinch.gif` so the PR
comment can embed the result inline (GitHub web doesn't autoplay APNG).

`PinchToZoomPreview` lands in the cmp sample so a human can run
`./gradlew :samples:cmp:renderAllPreviews` and see the same fixture
the test exercises.


yschimke:
Preview capture — 3-finger swirl on `MultiTouchDrawingPreview` with `overrides.touchOverlay = true`. The three coloured trails (amber / pink / green) are what the composition saw; the cyan rings are what coords the daemon dispatched. They line up, so the multi-pointer dispatch path is wired end-to-end.

multi-touch drawing GIF

Generated by `TouchOverlayDrawingRecordingTest.three_finger_swirl_paints_three_distinct_trails_with_overlay` and committed to `docs/screenshots/touch-overlay-drawing.gif` so this comment renders inline on the web and in mobile.

---

Issue #1203, opened 2026-05-17
Interactive + recording sessions silently drop key / rotary events on both backends


yschimke:
## Summary

`InteractiveSession.dispatch` and `RecordingScriptHandlerRegistry` accept `KEY_DOWN` / `KEY_UP` / `ROTARY_SCROLL` on the wire but drop them on the floor on both backends. Forward-looking clients can send them without rejection (good), but the inputs never reach Compose (silent failure — bad). This issue tracks landing real dispatch on both sides.

---

**Resolved.** All four backend-x-mode combinations now dispatch key + rotary events through to Compose; the silent-drop / `desktopUnsupported` / `androidUnsupported` branches for these inputs are gone.

- Desktop interactive: `daemon/desktop/src/main/kotlin/ee/schimke/composeai/daemon/DesktopInteractiveSession.kt:137-159` — `ROTARY_SCROLL` routes via `sendPointerEvent(Scroll)`; `KEY_DOWN` / `KEY_UP` route via `sendKeyEvent` with `androidKeycodeToComposeKey` translation (KDoc explicitly references #1203).
- Android interactive: `daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidInteractiveSession.kt:210-229` — `KEY_DOWN` / `KEY_UP` / `ROTARY_SCROLL` all flow through the dispatch path with `KeyboardController` notification.
- Desktop recording: `daemon/desktop/src/main/kotlin/ee/schimke/composeai/daemon/DesktopRecordingSession.kt:303-304` — `KEY_DOWN_EVENT` / `KEY_UP_EVENT` registered via `keyHandler(KeyEventType.KeyDown/KeyUp)`.
- Android recording: `daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidRecordingSession.kt:274-283` — `ROTARY_SCROLL_EVENT` / `KEY_DOWN_EVENT` / `KEY_UP_EVENT` routed via `interactiveDispatchHandler`.

Harness scenarios from the acceptance section also exist:

- `daemon/harness/src/test/kotlin/ee/schimke/composeai/daemon/harness/SInteractiveKeyDispatchRealModeTest.kt`
- `daemon/harness/src/test/kotlin/ee/schimke/composeai/daemon/harness/SRecordingKeyDispatchRealModeTest.kt`

[...]

PR #1308, opened 2026-05-21
feat(vscode): per-card touch-overlay + keyboard-band toggle buttons

yschimke:
`PreviewOverrides.touchOverlay` and `PreviewOverrides.keyboard` have
been on the wire since the data-extension PRs, but the VS Code
extension had no UI to flip either. The recording / streaming /
interactive start calls all hand-rolled a `{ previewId }` payload with
no overrides field at all, so end-users had to drop into the daemon
CLI or MCP to enable touch visualization.

Adds two per-card icon buttons next to the existing #1203 controls
toggle:

* `.card-touch-overlay-toggle-btn` (codicon-target) — sets
  `overrides.touchOverlay = true` on the next `requestStreamStart`.
  Daemon installs the `TouchOverlayExtension` `AroundComposable` for
  the session; streamed frames carry cyan rings at every pressed
  pointer plus down/up pulses.
* `.card-keyboard-band-toggle-btn` (codicon-symbol-keyword) — sets
  `overrides.keyboard = { visible: true }`. Daemon's always-active
  `data/keyboard` connector forces its Gboard band visible regardless
  of what the app's `LocalSoftwareKeyboardController` / focus state
  would naturally do.

Both buttons are sticky per preview id. Flipping either toggle while
the card is live restarts the stream so the daemon session rebuilds
with the new override set (overrides are applied at session-acquire
time, can't be patched mid-session).

Plumbing:

* `daemonProtocol.ts`: TS `PreviewOverrides` was stale (no
  `touchOverlay`, no `keyboard`). Now mirrors the Kotlin shape; adds
  `KeyboardOverride` interface and `StreamStartParams.overrides`.
* `liveCommand.ts`: `LiveCommand`'s `requestStreamStart` carries
  optional `overrides`; collapses to the no-overrides shape when every
  field is undefined so the wire payload stays byte-identical for the
  common case.
* `types.ts` / `bundleViewerPanel.ts`: `WebviewToExtension`'s
  `requestStreamStart` shape gets `overrides?`.
* `liveCardControls.ts`: two new `ensure*ToggleButton` /
  `remove*ToggleButton` helper pairs — idempotent across re-stamps,
  click handlers bound exactly once.
* `liveState.ts`: per-card state Sets, `overridesForPreview()` accessor
  that builds the payload, restart-on-toggle helper that posts
  stop+start when flipping a live card.
* `extension.ts` + `bundleViewerPanel.ts`: `handleRequestStreamStart`
  reads `msg.overrides`, threads it to `client.streamStart`.

Tests: stamp / aria-pressed / idempotent re-stamp / click / removal
coverage for each new button; pinning the `liveToggleCommand` wire
shape — overrides flow on start, never on stop, collapse to
no-overrides when every field is undefined. 1135 webview tests pass.

Gating: the new buttons appear on every card when the focused
module's daemon reports interactive support; the underlying overrides
are no-ops on backends that don't ship the matching extension. The
existing #1203 controls button keeps its narrower
`requiresInteractive=true` gate.


yschimke:
Visual preview of the new buttons in the panel:

panel preview — per-card toggle buttons

Left card is the default state, all three toggles off. Right card is live with touch overlay + keyboard band both pressed (cyan ring icons turn blue, matching the existing #1203 controls toggle's pressed state). Rendered against the real `media/preview.css` + `vscode-theme.css` + codicons via headless Chromium so the visual treatment matches what users actually see in the panel — the row of three top-left toggles + live-stop top-right matches `liveCardControls.ts` + the new CSS positioning.

---

PR #1562, opened 2026-05-28
Would you be open to support private previews? Private previews PoC

joakimtall:
## Rationale

Disclosure: This is written by a human, but AI was used for the minimal code example.

We have all private preview functions in our big codebase, and reluctant to change this. I created this proof of concept (with the help of AI), and it worked very well for our repo. This is not meant to be merged as is, just as a proof that its possible, and check if you are interested. Without private preview support we are forced to either fork, or rename thousands of previews -- both doable, but not desired :)

If you want to support it -- I imagine you would rather do the code yourself. But let me know if you want help

Very cool and welcome project btw, previously I relied on a (bad) Paparazzi hack to render previews

## Result

Three boxes were flipped to private fun in this PR's diff — discovery and render pick them up unchanged

Render log [link](https://github.com/joakimtall/compose-ai-tools/blob/private-preview-render-proof/docs/private-preview-render-proof/composePreviewRenderAll.log)

```
Task :samples:android:composePreviewDiscover
Discovered 79 preview(s) in module 'android':
  ...
  com.example.sampleandroid.PreviewsKt.RedBoxPreview  [Red Box]
  com.example.sampleandroid.PreviewsKt.BlueBoxPreview  [Blue Box]
  com.example.sampleandroid.PreviewsKt.GreenBoxPreview  [Green Box]
  ...
Task :samples:android:composePreviewRender
Task :samples:android:composePreviewRenderAll
BUILD SUCCESSFUL in 27s
230 actionable tasks: 230 executed
```

Json output [link](https://github.com/joakimtall/compose-ai-tools/blob/private-preview-render-proof/docs/private-preview-render-proof/previews.json)

```
{
  "id": "com.example.sampleandroid.PreviewsKt.RedBoxPreview_Red Box",
  "functionName": "RedBoxPreview",
  "className": "com.example.sampleandroid.PreviewsKt",
  "sourceFile": "src/main/kotlin/com/example/sampleandroid/Previews.kt",
  "params": { "name": "Red Box", "kind": "COMPOSE", ... },
  "captures": [ { "renderOutput": "renders/RedBoxPreview_Red_Box.png", "cost": 1.0 } ]
}
```

| `private fun RedBoxPreview` | `private fun BlueBoxPreview` | `private fun GreenBoxPreview` |
|:---:|:---:|:---:|
| Red | Blue | Green |

### DIFF
--- a/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewDiscovery.kt
+++ b/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewDiscovery.kt
@@ -210,14 +210,6 @@ object PreviewDiscovery {
       ClassGraph()
         .enableMethodInfo()
         .enableAnnotationInfo()
-        // Surface JVM-private methods so the loop below can actively warn
-        // when a `private fun` is annotated with @Preview (the renderer's
-        // `getDeclaredComposableMethod` fails on private composables, so
-        // we drop them — see the `method.isPrivate` branch). Without this
-        // flag ClassGraph's default visibility filter would drop them
-        // silently and the user would never learn why their preview
-        // disappeared. `internal fun` compiles to JVM `public` (with the
-        // `name$module` mangle) and passes the default filter regardless.
         .ignoreMethodVisibility()
         .overrideClasspath(classpath.map { it.absolutePath })
         .ignoreParentClassLoaders()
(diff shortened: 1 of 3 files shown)


yschimke:
Thanks for this — really nicely scoped PoC, and the render log + the three PNGs are exactly the proof I needed.

Yes, I'm happy to support private previews. Some history so you know where this sits: we actually had discovery working for `private fun` previews back in #188, but the render-time invocation was never verified end-to-end. They were reaching the daemon and failing, so in #967 we took the cheap way out and dropped them at discovery with that "make it `internal` or `public`" warning rather than fixing the renderer.

Your PR shows that was the wrong call. The diagnosis in #967 (`getDeclaredComposableMethod` can't resolve private methods → `NoSuchMethodException`) is off — resolution goes through `declaredMethods`, which includes private methods, so it resolves fine. The real failure is the `IllegalAccessException` at `invoke()` time, and your one-liner (`asMethod().isAccessible = true`) is exactly the right fix. We already do the same thing for receivers a few lines down in `resolvePreviewReceiver`, so it's very much in keeping.

I'll take it from here and land a proper version — no need for you to do more. For my own notes, what I'll change on top of your diff:

- Invert the discovery tests that currently assert the skip (`DiscoveryFunctionalTest`'s two "skips private" cases and `PreviewDiscoveryFailureWarningsTest`) to assert discovered + rendered, finally checking the box #188 left open.
- Wrap the `setAccessible` in a `runCatching` for the SecurityManager / strong-encapsulation case, consistent with `resolvePreviewReceiver`.
- Keep a public + private mix in the sample rather than flipping all three, so both paths stay covered.

Appreciate you taking the time to PoC it instead of just filing an issue — made the decision trivial. And glad the project's been useful; rendering previews outside Studio is exactly the itch it's meant to scratch.

---

Issue #292, opened 2026-04-28
confetti: preview workflow failure


yschimke:
https://github.com/joreilly/Confetti/actions/runs/25041706734/job/73346372201?pr=1687

Run python3 "$ACTION_PATH/../lib/compare-previews.py" copy-changed _previews.json \
Traceback (most recent call last):
  File "/home/runner/work/_actions/yschimke/compose-ai-tools/v0.8.10/.github/actions/preview-comment/../lib/compare-previews.py", line 1016, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/runner/work/_actions/yschimke/compose-ai-tools/v0.8.10/.github/actions/preview-comment/../lib/compare-previews.py", line 1012, in main
    return handlers[args.command](args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/_actions/yschimke/compose-ai-tools/v0.8.10/.github/actions/preview-comment/../lib/compare-previews.py", line 515, in cmd_copy_changed
    current = load_cli_output(cli_json)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/_actions/yschimke/compose-ai-tools/v0.8.10/.github/actions/preview-comment/../lib/compare-previews.py", line 195, in load_cli_output
    raw = json.loads(cli_json_path.read_text())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

yschimke:
Quick summary of what I found from the log:

v0.8.12 of the CLI is installed correctly (sha verified).
compose-preview show --json ran for ~7 minutes and exited 0 (next step ran).
[...]

PR #1617, opened 2026-05-29
fix: discover preview modules without realizing the whole task graph

yschimke:
## Summary

`compose-preview` preview discovery fetched the Tooling-API `GradleProject` model and read `project.tasks` to detect the `composePreviewDiscover` task. Building `GradleProject` **eagerly realizes every task provider in every module**, so configuration-time side effects in unrelated modules fire during mere discovery. Concretely, a module applying `org.graalvm.buildtools.native` with a `nativeCompile` task that pins a Java toolchain gets that toolchain provisioned every time you list previews:

- **Offline / restricted networks** (CI sandboxes, cloud agent environments): provisioning fails, the model query dies, and `show --json` emits nothing → the downstream `_previews.json is empty` symptom (upstream cause of #292).
- **Online**: a large, pointless JDK/GraalVM download just to *list* previews.

## What changed

- New `DiscoverPreviewModulesAction` (`:gradle-preview-driver`) walks the lightweight `GradleBuild` model — whose `BasicGradleProject`s carry only `path` + `projectDirectory`, no tasks — and asks each project for the existing `ComposePreviewModel`. The model builder detects the plugin via a single targeted `findByName("composePreviewDiscover")`, which realizes only that one named task rather than the whole graph. Unrelated modules' eager tasks are never realized.
- Per-project `findModel` calls are isolated, so a configuration failure in a module that contributes **no** previews is skipped, not fatal — discovery returns the modules that did resolve (issue's "isolate failures per module" point).
- `findPreviewModules()` now runs the build action instead of the `GradleProject` query.
- `ComposePreviewModel` moves from `:cli` into `:gradle-preview-driver` so the driver's build action can reference it directly; `:cli` keeps seeing it through its transitive `api` dependency, so existing `ee.schimke.composeai.plugin.tooling` imports resolve unchanged.

No plugin-side change required — this reuses the model the plugin already registers (and that `doctor` already consumes), so it works against already-published plugin versions.

## Test plan

- [x] `DiscoverPreviewModulesActionTest` (unit) — fake `BuildController`; asserts only plugin-applied modules are returned, the root project is never emitted, a throwing `findModel` is skipped not fatal, and the `GradleProject` model is **never** queried.
- [x] `DiscoverPreviewModulesIntegrationTest` (real Tooling-API connection) — synthetic build whose `:native` module registers a task that **throws during configuration**. Asserts `DiscoverPreviewModulesAction` completes while the old `GradleProject` query fails on that poison task. Ran (not skipped): passes.
- [x] `:gradle-preview-driver:test`, `:cli:compileTestKotlin`, ktfmt checks on touched modules.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QLJ2gZ1XnWaA75KqYApkW2)_

### DIFF
--- a/api/gradle-preview-driver/build.gradle.kts
+++ b/api/gradle-preview-driver/build.gradle.kts
@@ -41,6 +41,15 @@ dependencies {
   testImplementation(kotlin("test"))
 }
 
+tasks.withType<Test>().configureEach {
+  // The unpacked Gradle distribution this build runs from. `DiscoverPreviewModulesIntegrationTest`
+  // points a real Tooling-API connection at it via `useInstallation(...)` so it reuses the
+  // already-present distribution instead of downloading one (the test `assumeTrue`s out when this
+  // is absent — e.g. a bare IDE run). Read at configuration time so the configuration cache
+  // captures it.
+  systemProperty("composeai.test.gradleHome", gradle.gradleHomeDir?.absolutePath ?: "")
+}
+
 composeAiMavenPublishing {
   coordinates(
     artifactId = "gradle-preview-driver",
--- a/api/gradle-preview-driver/src/main/kotlin/ee/schimke/composeai/cli/DiscoverPreviewModulesAction.kt
+++ b/api/gradle-preview-driver/src/main/kotlin/ee/schimke/composeai/cli/DiscoverPreviewModulesAction.kt
@@ -1,0 +1,60 @@
+package ee.schimke.composeai.cli
+
+import ee.schimke.composeai.plugin.tooling.ComposePreviewModel
+import java.io.Serializable
+import org.gradle.tooling.BuildAction
+import org.gradle.tooling.BuildController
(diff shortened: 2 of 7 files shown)


yschimke:
Re: the pre-0.11.13 task-based fallback suggestion — we're intentionally not adding it.

The only Tooling-API way to detect a task by name is the `GradleProject` model, which is exactly the call that realizes the whole task graph and triggers #1620. So any task-name fallback re-admits the footgun this PR removes; there's no lightweight task-name probe.

`ComposePreviewModel` (and its plugin-side builder) has shipped since 0.11.13, and the CLI's default auto-inject always supplies a current plugin, so the only affected case is a project that manually pins a pre-0.11.13 plugin *and* disables auto-inject — which we've decided not to keep supporting. The `discoverModules()` KDoc has been updated to document the model requirement (3bb34ef).

---

PR #1621, opened 2026-05-30
fix(daemon): pass classpath via @argfile to avoid spawn E2BIG

yschimke:
## Summary

The VS Code e2e tests (`test:e2e` and `test:e2e-external`) were failing during daemon warm-up with:

```
[daemon] spawn failed for :androidApp: spawn E2BIG
```

`spawnDaemon` joined the entire module classpath into a single `-cp` argv entry. On Linux a single argv string is capped at `MAX_ARG_STRLEN` (128 KB), so for a large Android/Compose classpath `child_process.spawn` rejects with `E2BIG` before the JVM ever starts — aborting the warm path and failing the e2e suite (the `:samples:cmp` interactive failures were downstream of the daemon never coming up).

## Fix

Hand the classpath to `java` through a temporary `@argfile` instead of an inline `-cp`. The launcher reads the file from disk, so argv stays tiny regardless of classpath size. `@argfile` support has been in the JDK since Java 9, which the daemon already requires.

- Classpath entries are joined and written as `-cp "<entries>"` into a per-spawn temp dir.
- Paths are wrapped in quotes and backslashes rewritten to `/` so the JDK argfile tokenizer doesn't choke on spaces or mangle Windows paths (the tokenizer treats `\` as an escape char; `java` accepts `/` separators on every platform).
- The temp dir is removed once the daemon process exits (clean shutdown, crash, or the SIGTERM from a failed `initialize`).

## Test plan

- New unit tests for `formatClasspathArgfileContent`: single quoted `-cp` option, a 5000-entry classpath that exceeds 128 KB staying in one token, and Windows backslash rewriting.
- `npm test` — 1455 passing.
- `npx eslint` clean on both changed files.

---
_Generated by [Claude Code](https://claude.ai/code/session_012wMRzBNZjSm9m1ac3vZBJi)_

### DIFF
--- a/vscode-extension/src/daemon/daemonProcess.ts
+++ b/vscode-extension/src/daemon/daemonProcess.ts
@@ -1,5 +1,6 @@
 import { ChildProcess, spawn } from "child_process";
 import * as fs from "fs";
+import * as os from "os";
 import * as path from "path";
 import {
     DaemonClient,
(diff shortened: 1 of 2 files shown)


yschimke:
Good catch on `--disable-@files` — fixed in 70a4c14. The classpath `@argfile` reference now goes first, ahead of `descriptor.jvmArgs`, so it's always expanded before any user JVM arg could disable `@file` expansion. The `-cp` precedence is unaffected in practice since the descriptor keeps the classpath separate from `jvmArgs`. All 1405 unit tests still pass.

---

PR #1701, opened 2026-06-02
feat(preview-data-api): Kotlin SpatialScene mirror of the wire contract

yschimke:
Producer-side counterpart to the SpatialScene contract (#1697).

Adds `ee.schimke.composeai.xr.SpatialScene` — a `@Serializable` Kotlin mirror of the wire format — so the offline renderer (`:renderer-xr`, when built) and other tooling can emit/read the same shape the VS Code 3D viewer consumes. It lives in `:preview-data-api` alongside the other published wire DTOs (kotlinx-serialization is already set up there), which is the idiomatic home for a wire type — no new module needed yet.

## Cross-language contract lock
`SpatialSceneTest` deserializes the **single committed fixture** (`vscode-extension/preview-harness/fixtures/spatial-scene/scene.json`) rather than a copy, so the Kotlin and TypeScript shapes can't drift — change one side and the test fails until the other follows. It also round-trips through JSON and checks the `version`/`units` defaults.

Property names are the JSON keys and match `spatialScene.ts` exactly; conventions (dp, right-handed axes, quaternion rotation) match the spec.

## Test plan
- [x] `:preview-data-api:test` — 3 passed (fixture deserializes to the contract shape; JSON round-trip; default version/units)
- [x] `:preview-data-api:ktfmtCheck`
- [x] Branched fresh from `main` after #1697/#1700 merged
- Doc: `SPATIAL_SCENE_CONTRACT.md` producer note now points at the real home

Next on the producer side: the `:renderer-xr` render mode that recovers poses (per `SubspaceLayoutPoseTest`) + renders panel textures and writes a `SpatialScene` using this type.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WUMD4xmJ9LUtL5dBsqPU2S)_

### DIFF
--- a/api/preview-data-api/src/main/kotlin/ee/schimke/composeai/xr/SpatialScene.kt
+++ b/api/preview-data-api/src/main/kotlin/ee/schimke/composeai/xr/SpatialScene.kt
@@ -1,0 +1,90 @@
+package ee.schimke.composeai.xr
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Kotlin mirror of the `SpatialScene` wire contract — the format the offline renderer
+ * (`:renderer-xr`, the producer) emits and the VS Code webview's 3D spatial-layout viewer (the
+ * consumer) reads.
+ *
+ * The **authoritative** definition is the TypeScript source
+ * (`vscode-extension/src/webview/shared/spatialScene.ts`) plus the prose spec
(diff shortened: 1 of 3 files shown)


yschimke:
Superseded by #1703 — this branch was stacked beneath it, so its squash-merge carried the `SpatialScene` DTO, `SpatialSceneTest`, and the contract-doc note onto `main`. Closing as redundant.

---

PR #1862, opened 2026-06-14
feat(daemon): history/diff mode=SEMANTICS — pixel-free semantics regression diff (#1785)

### LINKED ISSUE #1785: Semantics-tree text diff as a cheap, deterministic regression signal (aria-snapshot analogue)
## Context

Part of the agent-loop-ergonomics series (Playwright-informed). Playwright's recommended regression mechanism is **`toMatchAriaSnapshot`** — a deterministic *text* diff of the accessibility tree — with pixel diffing secondary. Text diffs are token-cheap, immune to anti-aliasing / font-hinting noise, and *semantic* ("Button 'Submit' lost its label"), which is exactly what an LLM reviewer wants to assert on.

Our review loop diffs **pixels** today: `history/diff` is metadata-only in 1.0, `history/diff/regions` is a pixel bbox, and the CLI `--changed-only` path costs ~3k tokens per changed preview (two PNG reads) per [`docs/TOKEN_USAGE.md`](../blob/main/docs/TOKEN_USAGE.md). We already produce the raw material — `compose/semantics` (testTag/role/mergeMode/bounds) and `a11y/hierarchy` — but nothing diffs them.

This is **complementary to**, not a duplicate of, the pixel-mode `history/diff` planned under #929 / the H5 roadmap item: it's a cheaper, structural gate that catches semantic regressions a pixel diff can't explain.

## Scope

1. **A structural differ** over `compose/semantics` (and/or `a11y/hierarchy`) producing a typed delta: nodes added/removed, and per-node field changes (text, contentDescription, role, states, testTag). Stable node identity keys the match (pairs with the `ref` work in the semantic-targeting issue).
2. **Two entry points:**
   - **Two history entries** — extend `history/diff` with a `mode: SEMANTICS` (alongside the existing METADATA / reserved PIXEL), returning the typed delta.
   - **Base vs head** — a CLI surface for the `compose-preview-review` skill, e.g. `compose-preview diff --semantics --filter … --json`, that renders both sides and diffs the trees.
3. **Human + JSON rendering.** JSON for agents (versioned `schema`); a compact human summary ("3 changed: Button 'Submit' label removed; Text 'Hello'→'Helloo'; Icon added") for terminals and PR comments.
4. **Noise control.** Ignore purely positional bounds churn by default (that's the pixel diff's job); focus on semantic fields. Make the ignored-field set configurable.

## Acceptance criteria

- A `compose/semantics` text diff is available both via `history/diff mode=SEMANTICS` and a CLI command, emitting a versioned JSON delta + a human summary.
- Deterministic: identical trees → empty delta; reordering without semantic change is reported distinctly from add/remove.
- Token cost per changed preview is materially below the two-PNG pixel path (document the measured delta in `TOKEN_USAGE.md`).
- Wired into the `compose-preview-review` skill story (doc + example PR-review sentence).
- Tests: differ unit tests over fixture trees (add/remove/field-change/reorder); a CLI golden test.

## Out of scope

- Pixel-mode `history/diff` (H5 / #929) — this lands alongside it, not instead of it.
- VS Code panel visualisation of the semantic delta (follow-up; relates to #1015 agent monitor mode).

## Seams

`compose/semantics` payload + schema (`:data-layoutinspector-core`), `history/diff` handler (`daemon/core`), a new `ExtensionReportRenderer`-style CLI command (`cli/.../*Command.kt`), `HistoryDiffMode` enum.

yschimke:
## Summary

Closes #1785.

The structural semantics-tree differ, the `compose-preview diff-semantics` CLI, and the MCP `diff_semantics` tool all landed in #1803. This PR adds the **one remaining acceptance-criteria item**: the second entry point — `history/diff mode=SEMANTICS`.

`history/diff` now accepts `mode=semantics`, diffing two history entries' captured `compose/semantics` trees and returning a typed `SemanticsDelta` (`compose-semantics-diff/v1` — added / removed / changed nodes, matched by each node's stable `ref`). It reuses the same `SemanticsDiff` as the CLI/MCP, so all three surfaces agree, and reads no PNGs.

To make this work, the semantics tree is snapshotted into each history entry at record time:

- **Storage.** A new optional `HistoryEntry.semantics` (`JsonElement`). It's heavy, so it lives **only** in the per-entry sidecar — stripped from the lean `index.jsonl` (exactly like `previewMetadata`) and never echoed on the `history/list` / `history/read` / `historyAdded` / metadata-mode `history/diff` wire surfaces. `history/diff mode=semantics` is the single reader that loads it back off disk. No change to the `HistorySource.write` signature.
- **Capture.** `recordHistoryForRender` fetches `compose/semantics` inline from the data-product registry. It's opportunistic — a missing artefact records `semantics = null` and never forces a re-render.
- **Errors.** A diff where either entry has no captured snapshot returns `ERR_HISTORY_SEMANTICS_NOT_CAPTURED` (-32013), distinct from "entry not found".

Also: `SemanticsDelta.schema` is now `@EncodeDefault`, so the versioned schema rides every wire surface — including the daemon's `encodeDefaults=false` encoder (previously an empty-or-default delta would serialize without its `schema`).

`history/diff` remains gated experimental (`-Dcomposeai.experimental.historyDiff=true`), so this is additive and off by default in production.

This is data/protocol plumbing with no visual surface, so text-only evidence is appropriate (per the repo's visual-evidence rule).

## Test plan

- [x] `:daemon:core:test --tests "*HistoryDiffTest*"` — new SEMANTICS-mode tests: field-change on the same ref, identical-trees → empty delta, missing-snapshot → -32013 error.
- [x] Full `:daemon:core:test` green except a pre-existing, timing-flaky `InteractiveCoalescingTest` (confirmed failing identically on clean `main` with these changes stashed; unrelated to this PR).
- [x] `:data-layoutinspector-core:test`, `:cli:test --tests "*SemanticsDiffCommand*"`, `:mcp:test --tests "*DaemonMcpServerTest*"`, `:daemon:core:test --tests "*HistoryEntryRoundTrip*"` — green (covers the `@EncodeDefault` change and the new `HistoryEntry` field round-trip).
- [x] `ktfmtCheckMain` / `ktfmtCheckTest` on touched modules; downstream `:mcp` / `:daemon:desktop` / `:daemon:harness` compile against the new daemon-core ABI.
- [x] Docs updated: `docs/daemon/HISTORY.md`, `docs/daemon/PROTOCOL.md` (mode, result field, sidecar field, error code).

---
_Generated by [Claude Code](https://claude.ai/code/session_01XDk94Aru4kWn27Ct21NU67)_

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -125,6 +125,7 @@ tasks.register("functionalTestWithAndroid") {
 //       api :data-displayfilter-core
 //       api :daemon:core
 //         api :renderer-xr-client (daemon-core fronts the native XR render server — RENDERER_SERVICE)
+//         api :data-layoutinspector-core (semantics models + differ for `history/diff mode=semantics`, #1785)
 //     implementation :lottie-preview-runtime (Compottie-backed kind=LOTTIE render path)
 //   plus :common-io (the Okio file-IO foundation those modules read/write through).
 val bundleRenderFunctionalTestPublishTargets =
@@ -138,6 +139,7 @@ val bundleRenderFunctionalTestPublishTargets =
     ":data-displayfilter-core",
     ":daemon:core",
     ":renderer-xr-client",
+    ":data-layoutinspector-core",
     ":lottie-preview-runtime",
     ":common-io",
   )
--- a/daemon/core/build.gradle.kts
+++ b/daemon/core/build.gradle.kts
@@ -22,6 +22,12 @@ plugins {
 dependencies {
   api(project(":data-render-core"))
 
+  // Semantics-tree models + structural differ (issue #1785). `api`, not `implementation`: the
+  // published `HistoryDiffResult.semanticsDelta` field is a `SemanticsDelta`, so the type is part
+  // of this module's compile ABI. Pure-JVM core module (no Compose/Android) — safe on the
+  // renderer-agnostic daemon classpath.
+  api(project(":data-layoutinspector-core"))
+
   // Okio-based file IO (`SystemFileSystem`) for data-product reads, history sidecars, bundle IR
   // replay, and forensics dumps. `implementation` — Okio stays an internal detail; daemon/core's
   // public surface keeps its java.io.File signatures.
--- a/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/JsonRpcServer.kt
+++ b/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/JsonRpcServer.kt
@@ -1206,12 +1206,14 @@ class JsonRpcServer(
     // history write failure never blocks the renderFinished wire-format. The render's notification
     // has already been sent above; history is observation, not state.
     //
-    // Skip history when the frame is byte-identical to the prior one for this preview — there's
-    // nothing new to archive, and a duplicate sidecar would inflate `history/list` results
-    // without changing what the user sees on disk.
-    if (!isUnchanged) {
-      recordHistoryForRender(previewId = previewId, result = result, finished = finished)
-    }
+    // Dedup is the history source's job, not this gate's: a byte-identical frame may still carry a
+    // changed `compose/semantics` tree (a dropped contentDescription leaves pixels untouched), and
+    // that's exactly what `history/diff mode=semantics` exists to catch (issue #1785). So we record
+    // regardless of the pixel-only `isUnchanged` flag and let `LocalFsHistorySource` skip only when
+    // BOTH the bytes and the semantics tree are unchanged — `recordHistoryForRender` returns null
+    // on
+    // that skip, so a truly-redundant frame still emits no `historyAdded` and adds no sidecar.
+    recordHistoryForRender(previewId = previewId, result = result, finished = finished)
     inFlightRenders.remove(result.id)
     previewIdsWithOverridesInFlight.remove(previewId)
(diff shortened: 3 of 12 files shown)


yschimke:
Good catch on the semantics-only dedup gap (P1) — fixed in 5d4ecd6.

History dedup was pixel-only on two gates (`emitRenderFinished` skipped any byte-identical frame; `LocalFsHistorySource` tier-1 skipped any same-`pngHash` write), so a non-visual semantics change never recorded a second entry. Now the FS source owns dedup and is semantics-aware: it skips only when **both** the `pngHash` and the semantics tree are unchanged. The tree comparison goes through `SemanticsDiff` (not a raw byte hash), so the volatile per-render `nodeId` and positional bounds don't defeat dedup — a true duplicate still skips, but a dropped `contentDescription`/`role`/`testTag` lands. The PNG is still pointer-deduped, so a semantics-only change costs one sidecar + index line, no duplicate bytes. Added FS-level tests for all three cases (semantics-only-change → recorded, identical → skipped, nodeId-churn-only → skipped).

The same push also fixes the Bundle Render / Android Bundle Daemon E2E failures: daemon-core's new `api` dep on `data-layoutinspector-core` extends `renderer-desktop`'s published transitive closure, so it now needs publishing to mavenLocal — added to `bundleRenderFunctionalTestPublishTargets`.

---

Issue #1969, opened 2026-06-20
feat(recording): smoke "crawl" — auto-tap every clickable semantics node (Robo-style)


yschimke:
Follow-up to PR #1963 / `docs/daemon/RECORDING-ASSERTIONS.md`. Robo / Monkey auto-explore an app by tapping things. A preview is a single held scene, not an app graph, so "crawl" reduces to a cheap smoke check: tap every clickable node and confirm nothing throws.

## Reuse existing machinery (no new dispatch)

- Enumerate clickable nodes from the existing `recording.probe` semantics snapshot — `ComposeSemanticsNode.clickable` / `RecordingProbeNode.clickable` already carry the flag.
- Dispatch a plain `input.click` at each (reuse the existing input handler + `SemanticsInputTarget` resolution — by `ref`/`testTag`).
- Snapshot frames via the normal recording loop; surface any node whose tap throws / latches a tick-thread failure as an error.

This is generated entirely from existing event kinds — it could even be a script the CLI synthesises rather than a new protocol event.

## Shape

`compose-preview record --crawl --out crawl.gif` (or a `--emit-script` mode that writes the generated tap-every-node script for the user to edit). Fails (exit 2) if any tap crashes the composition; otherwise produces a recording that visits every clickable affordance.

## Non-goals

Not a real crawler — no navigation graph, no state-space exploration. A "does anything crash when tapped" smoke pass over one scene. Deeper exploration stays out of scope.

## Acceptance

- Against a preview with N clickable nodes, the crawl taps all N and reports; a node that throws on tap fails the run.

PR #1977, opened 2026-06-20
feat(recording): wire assert.visible / assert.notVisible on the Android backend

### LINKED ISSUE #1964: feat(recording): wire assert.visible / assert.notVisible on the Android (Robolectric) backend
Follow-up to the recording-assertions work (PR #1963, which shipped desktop-only). See `docs/daemon/RECORDING-ASSERTIONS.md`.

## Goal

Make `assert.visible` / `assert.notVisible` work against the Android (Robolectric) recording backend, so `compose-preview record` can gate Android previews the same way it gates desktop.

## Reuse existing features (don't invent new syntax)

Everything on the wire already exists — this is backend wiring only:

- The event kinds (`assert.visible` / `assert.notVisible`), `RecordingScriptEventStatus.FAILED`, the `failedEvidence(...)` builder, and `RecordingScriptDataExtensions.assertionDescriptor` already exist from PR #1963.
- The verdict logic (`evaluateVisibilityAssertion`) is pure and backend-agnostic — reuse it as-is.
- Mirror `DesktopRecordingSession.assertVisibilityHandler` in `AndroidRecordingSession.buildScriptHandlers()`.

## The actual work

The blocker is target resolution: `AndroidRecordingSession`'s probe snapshot (`captureProbeSemantics`) doesn't carry the refs `SemanticsTargets.resolve` needs. **Reuse the Android target-resolution path that `input.click` already uses** on the held compose rule (the Android side of #1784) to resolve a `SemanticsInputTarget` → node match count, then feed that into the shared `evaluateVisibilityAssertion`.

Then add `RecordingScriptDataExtensions.assertionDescriptor` to `RobolectricHost.recordingScriptEventDescriptors()` so MCP advertises (and accepts) `assert.*` for Android daemons.

## Acceptance

- `AndroidRecordingSessionTest` mirrors the desktop integration test: APPLIED for a present `testTag`, FAILED (+ candidates) for an absent one, and the inverse for `assert.notVisible`.
- Removes the CLI's "unsupported by this backend" exit-2 path for Android (it currently fails `assert.*` there because the backend doesn't advertise support — by design until this lands).

yschimke:
Closes #1964. Second of the #1964–#1969 follow-ups; brings the recording-assertion visibility checks to the **Android (Robolectric)** backend so `compose-preview record` can gate Android previews the same way it already gates desktop.

## Reuses existing machinery (no new bridge command)

The desktop trick — call `SemanticsTargets.resolve` in-handler — doesn't apply on Android because of the host/sandbox classloader boundary. But instead of a heavyweight new `InteractiveCommand`, this reuses the **already-bridged `captureProbeSemantics()`** (the exact path the `recording.probe` event uses):

- Resolve the target against the probe snapshot by `testTag` / `role`+`text`, count matches, feed the shared verdict logic. **No new sandbox command, no sandbox-loop changes.**
- The pure verdict logic (`AssertionVerdict` / `evaluateVisibilityAssertion` / `evaluateTextEqualsAssertion` / `resolvedNodeText`) **moves from `:daemon:desktop` into `:daemon:core`** (now `public`) so desktop and Android share one implementation. The pure test moves with it.
- `RobolectricHost` advertises a visibility-only `assertionVisibilityDescriptor` (`assert.visible` / `assert.notVisible`).

## Honest limitations (documented)

- **`ref` targets** fail with a clear message on Android — the probe snapshot carries no refs. Use `testTag` or `role`+`text`.
- **`assert.textEquals` stays desktop-only** — the flat probe snapshot can't do the merged-descendant text fallback the desktop unmerged tree gives — so `record_preview` rejects it on Android up front (the descriptor doesn't advertise it) rather than silently no-op'ing.

## Test plan

- [x] Pure `RecordingAssertionsTest` now in `:daemon:core` — **ran green.**
- [x] `:daemon:core:compileKotlin` + `:daemon:desktop:compileKotlin` — confirms the `internal`→`public` cross-module move (desktop still resolves the shared functions). **Green.**
- [x] New `AndroidRecordingSessionTest.scriptedVisibilityAssertionsResolveAgainstProbeSnapshot` — drives `assert.visible`/`notVisible` against a fake `InteractiveSession` with canned probe nodes (present tag → APPLIED, absent → FAILED, inverse for not-visible, `ref` target → FAILED with a clear message). **`:daemon:android compileDebugKotlin + testDebugUnitTest` ran green locally** (Robolectric not needed — the fake supplies the probe snapshot).
- [x] `ktfmtFormat` applied across touched modules.

## Visual evidence

No rendered surface changes — protocol/host/handler plumbing. Behaviour is `FAILED`/`APPLIED` evidence + the CLI's existing non-zero exit, verified by the tests above.

---
_Generated by [Claude Code](https://claude.ai/code/session_015XGE7oXG27kAaotihUDCag)_


yschimke:
Good catch on the `role`+`text` false-pass — fixed in `c06f77b`, but with a narrower approach than the two suggested, for a reason worth recording:

- **"Use the full semantics resolver"** isn't available on Android here — the recording handler runs host-side, across the classloader boundary from the held rule; the whole point of resolving against `captureProbeSemantics()` was to avoid a new sandbox round-trip.
- **"Include merged descendant text in the probe node"** would help `Button { Text(...) }`, but the probe snapshot is a *flat* list, so I'd also have to re-introduce parent/child structure to know which text belongs to which role — a bigger, sandbox-side change.

So this PR restricts the Android backend to **`testTag` targets** (reliably one node) and fails `ref` and `role`+`text` with a clear "set a testTag" message — eliminating the false-pass rather than risking it. The test now includes a `role`+`text` `assert.notVisible` that must FAIL (not silently pass). Enriching the probe snapshot to support `role`+`text` is filed as a follow-up rather than rushed here.

---

PR #1980, opened 2026-06-20
feat(recording): record --fail-on a11y gates a recording on ATF findings

### LINKED ISSUE #1966: feat(recording): gate a scripted recording on accessibility findings (record --fail-on a11y)
Follow-up to PR #1963 / `docs/daemon/RECORDING-ASSERTIONS.md`. Espresso/ATF fail a test on accessibility violations; we can do the same for a scripted recording, and it pairs naturally with the TalkBack spec (sibling of #1955).

## Reuse the existing a11y product

The daemon already produces the `a11y/atf` data product — reachable via `RenderSession.fetchData(kind = "a11y/atf")`, already aggregated by `DaemonA11yFetcher`, already surfaced by `compose-preview a11y --fail-on errors|warnings`. **Reuse all of it.** No new findings engine.

## Two shapes (prefer the flag)

1. **`compose-preview record --fail-on a11y[=errors|warnings]`** (preferred) — after `recordingStop`, run the existing `DaemonA11yFetcher` against the recorded preview and gate the exit using the **same `--fail-on` semantics `compose-preview a11y` already has**. Pure CLI + existing fetch; no protocol change.
2. `assert.a11y` script event — fetch `a11y/atf` at the event's `tMs` and emit `FAILED` evidence (reusing `failedEvidence`) when findings exceed the threshold. More granular (assert mid-script), but needs a new event kind.

Start with (1); add (2) only if per-step a11y assertions are wanted.

## Acceptance

- A recording of a preview with a known ATF violation exits non-zero under `--fail-on a11y`, and exits 0 when clean — mirroring `compose-preview a11y`.

yschimke:
Closes #1966. Third of the #1964–#1969 follow-ups. Adds the Espresso/ATF "fail the test on accessibility violations" model to `compose-preview record`.

```
compose-preview record --module :app --preview MyScreen --script flow.json --out flow.gif --fail-on a11y
# records flow.gif, then exits 2 if the preview has ATF accessibility errors
```

`--fail-on a11y[=errors|warnings]` fetches the preview's `a11y/atf` findings through the **already-open** `RenderSession` after recording and exits non-zero (code 2) when the threshold is tripped — combined with the existing assertion gate.

## Reuses existing machinery

- The daemon's `a11y/atf` data product — the same one `compose-preview a11y` drives — via `session.fetchData(previewId, "a11y/atf")` + `enableExtensions(["a11y"])`. No new producer.
- The existing `AccessibilityFinding` wire model for parsing.
- The verdict logic (threshold parse + severity count) is a **pure, unit-tested** function (`RecordA11yGate.kt`).

## Backend honesty (per the discussion on the issue)

ATF findings are **Android-produced** — the framework runs against an Android `View` hierarchy. The desktop a11y path is overlay-only and emits no findings, so on a desktop preview the findings list is empty and the gate is a no-op (with a clear note on stderr). The gate is meaningful for **Android** recordings. This flag gates the preview's a11y; per-step *post-interaction* a11y assertions are a future `assert.a11y` event (noted in the roadmap).

## Test plan

- [x] `:cli:test --tests "*RecordA11yGateTest"` — threshold parsing (errors/warnings, singular/plural, junk), severity counting, errors-vs-warnings thresholds, empty-findings (desktop) never trips, case-insensitive levels. **Ran green.**
- [x] `:cli:compileKotlin` + `ktfmtFormat` clean.
- [ ] End-to-end against an Android preview with a real ATF violation — relies on Android CI (the fetch path is the same one `compose-preview a11y` already exercises).

## Visual evidence

No rendered surface changes — CLI plumbing reusing the existing a11y data product. Behaviour is a non-zero exit on findings, verified by the unit tests above.

---
_Generated by [Claude Code](https://claude.ai/code/session_015XGE7oXG27kAaotihUDCag)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/Main.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/Main.kt
@@ -133,7 +133,9 @@ private fun printUsage() {
                        --out extension unless --format is given. gif/apng are always available
                        (pure-JVM); mp4/webm need ffmpeg on PATH. Scripts may include Maestro-style
                        assert.visible / assert.notVisible events (with a target); a failed assertion
-                       still writes the recording but exits non-zero (code 2).
+                       still writes the recording but exits non-zero (code 2). --fail-on
+                       a11y[=errors|warnings] additionally gates on the preview's ATF accessibility
+                       findings (Android backend; desktop has no ATF data).
       a11y             Render previews with the a11y data extension on and
                        print ATF findings (thin wrapper over `--with-extension a11y`)
       diff-semantics   Diff two compose/semantics trees (base vs head) and report what
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/RecordA11yGate.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/RecordA11yGate.kt
@@ -32,6 +32,22 @@ internal enum class A11yThreshold {
   }
 }
 
+/**
+ * What happened when the record command tried to gate on a11y. Lets the caller fail *closed*: a
+ * producer error ([EvaluationFailed]) must fail the command, while a backend that legitimately has
+ * no ATF data ([NotApplicable], e.g. desktop) must not.
+ */
+internal sealed interface A11yGateOutcome {
+  /** The gate ran; [result] carries the verdict. */
+  data class Evaluated(val result: A11yGateResult) : A11yGateOutcome
+
+  /** The backend produces no ATF findings here (desktop overlay-only / kind not advertised). */
+  data class NotApplicable(val reason: String) : A11yGateOutcome
+
+  /** The a11y producer errored, so the requested check could not run — the caller fails closed. */
+  data class EvaluationFailed(val reason: String) : A11yGateOutcome
+}
+
 /** Outcome of the a11y gate: per-severity counts plus whether the [threshold] was tripped. */
 internal data class A11yGateResult(
   val errorCount: Int,
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/RecordA11yGate.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/RecordA11yGate.kt
@@ -1,0 +1,64 @@
+package ee.schimke.composeai.cli
+
+/**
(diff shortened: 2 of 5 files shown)


yschimke:
Thanks — addressed in `d042715` (also fixes the CI break: `failOn` is on `ReportCommand`, not the base `Command`; a stale local incremental compile masked it, CI's clean build caught it).

- **P1 — fail closed:** ✅ Now distinguishes `DataProductException.code`. `NOT_AVAILABLE` / `UNKNOWN` (desktop overlay-only, or kind not advertised) → `NotApplicable`, don't trip. `FETCH_FAILED` / `BUDGET_EXCEEDED` (producer errored) → `EvaluationFailed`, the command **exits non-zero** rather than passing on an unevaluated check. New `A11yGateOutcome` sealed type carries the three cases.

- **P1 — stale a11y file / P2 — overrides:** ✅ partially, with a documented limit. The gate now forces a fresh `renderNow(previewId, overrides = recordingOverrides)` before the fetch, so findings aren't served from a stale `a11y-atf.json` and reflect the recorded `--overrides` config. **Honest caveat (now in `docs/daemon/RECORDING-ASSERTIONS.md`):** this gates the *preview's* a11y via `data/fetch` — it is not per-step a11y of a specific post-interaction frame. If the a11y producer re-renders independently of the pre-render it can still fall back to the discovery spec; the fully-correct design (a11y evaluated against the held recording scene at a script point) is the future `assert.a11y` event, which I'd file as a follow-up.

If you'd rather not ship the partial `data/fetch` gate and instead go straight to `assert.a11y` (evaluated against the recorded scene), say so and I'll pivot #1966 — the reviewer's two render-coupling points are exactly why that's the more correct shape.

---

yschimke:
Closing in favor of the correct design. Per discussion, the `--fail-on a11y` `data/fetch` approach is render-coupled (the two points Codex raised: stale/independent re-render + override mismatch). Rather than ship a partial gate with that caveat, #1966 will be re-done as an **`assert.a11y`** script event evaluated against the **held recording scene** at a script point — which gates the actual recorded/post-interaction state, Android-backed (ATF runs against the View hierarchy).

New PR to follow. Thanks @chatgpt-codex-connector for catching the coupling — it's exactly why the script-event shape is the right one.

---

PR #2036, opened 2026-06-21
fix(fake-emulator): move the a11y subscription with the launched preview

yschimke:
Follow-up fix to #2034. Auto-merge landed #2034 from its original head before the Codex-review fix was pushed, so `main` has the buggy version — this PR carries just that fix.

## What
`RenderSessionFrameSource.applyA11y` short-circuited on a boolean (`talkBack == a11ySubscribed`). With TalkBack enabled, launching a *different* preview never moved the a11y data-product subscription: daemon subscriptions are keyed by `(previewId, kind)`, so it stayed attached to the previous preview, and toggling TalkBack off afterward unsubscribed the wrong preview and leaked the old one.

Now we track the **subscribed preview id**: launching a new preview unsubscribes the old preview and subscribes the new one; `close()` drops it too.

## Test plan
`:fake-emulator` compiles and is ktfmt-clean. The behavioural change is in the daemon-backed bridge (`RenderSessionFrameSource`), which has no standalone unit test (it needs a live `RenderSession`); the logic is small and guarded with `runCatching`.

```
./gradlew :fake-emulator:compileKotlin :fake-emulator:ktfmtCheck
```

### DIFF
--- a/fake-emulator/app/src/main/kotlin/ee/schimke/composeai/fakeemulator/app/RenderSessionFrameSource.kt
+++ b/fake-emulator/app/src/main/kotlin/ee/schimke/composeai/fakeemulator/app/RenderSessionFrameSource.kt
@@ -45,7 +45,8 @@ class RenderSessionFrameSource(
   private val lock = Any()
   private var currentPreviewId: String? = null
   private var currentStreamId: String? = null
-  private var a11ySubscribed = false
+  /** The preview id the a11y data product is currently subscribed on, or `null` if none. */
+  private var a11yPreviewId: String? = null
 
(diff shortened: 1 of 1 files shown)


yschimke:
Good catch — fixed in 69028e5. The daemon reader thread delivers both `streamFrame` notifications and the RPC responses, so holding the frame lock across `streamStart`/`subscribeData` could stall it. Now `currentStreamId` is `@Volatile` and read lock-free in `onStreamFrame`, and the launch/settings restarts + a11y subscription changes are serialised under a separate `opLock` the reader never touches — so the RPCs run safely without blocking the reader (this covers `restartStream` too, not just `applyA11y`).

---

PR #2096, opened 2026-06-29
docs(serve): spike report + design for in-browser CMP via Kotlin/Wasm

yschimke:
## Why

The last tier of the public-server render model is **CMP rendered in the browser via Kotlin/Wasm**. It's also the largest and most uncertain piece, so — per a spike-first approach — this records the feasibility findings and a concrete build/packaging/viewer design **before** committing a heavy Compose/Wasm build, so that build can be executed and *measured* as a focused follow-up.

## What it captures

- **Confirmed feasible:** the toolchain (Kotlin `2.3.21` + Compose Multiplatform `1.10.3`) ships the `wasmJs` target with Compose and the `org.jetbrains.compose` plugin is already on the classpath; the serve viewer already has the `data-mode="live"` seam (with a `cp-canvas` beside `cp-img`); and the bundle reserves a `web/` carriage dir. So the risk is **not** the toolchain.
- **The core design call:** a Wasm app can only render composables compiled into it. Recommends a **fixed `wasmJs` catalog app** (built from modules we own, served for `--catalogs`) first, then optional **per-bundle Wasm** behind `--with-wasm`.
- **Packaging:** `web/wasm/` + a v9 `webRender` manifest descriptor (additive, like v5–v8).
- **Viewer integration:** mount the Wasm app in a sandboxed iframe at the live seam (Wasm runs in the browser sandbox → safe even for unverified bundles), falling back RemoteCompose → server frames → baked PNG.
- **Open questions for the measured build:** artifact size / cold-load, headless render confirmation (Chromium), build cost — with explicit **go/no-go** criteria. No-go fallback (server frames + baked PNG) is already shipped, so nothing regresses if Wasm is deferred.

## Test plan

Docs-only. The build it designs is the follow-up that produces the measured evidence (artifact size + a headless render screenshot).

---
_Generated by [Claude Code](https://claude.ai/code/session_01K8nJEfEYGM3ijyoAXDNLGH)_

### DIFF
--- a/docs/wasm-cmp-spike.md
+++ b/docs/wasm-cmp-spike.md
@@ -38,23 +38,37 @@ cold-load, and (c) confirming a headless in-browser render. Those need a measure
 A Wasm app can only render composables **compiled into it** — it can't reflectively render an arbitrary
 uploaded consumer composable the way the server's classpath player can. Two models:
 
-1. **Fixed catalog app (recommended first).** Build one `wasmJs` app from a module **we control** (the
-   `design-catalog-*` catalogs) that registers its composables by preview id. Serve it for our
-   *published* catalogs (`--catalogs`). No per-bundle build; one artifact per design system. This
-   directly delivers "CMP renders in the browser" for the catalogs the public server already hosts.
+1. **Fixed catalog app (recommended first).** Build one `wasmJs` app from a CMP module **we control**
+   that registers its composables by preview id, and serve it for our *published* catalogs
+   (`--catalogs`). No per-bundle build; one artifact per design system. This directly delivers "CMP
+   renders in the browser" for the catalogs the public server already hosts.
+
+   **Important constraint (the existing catalogs aren't wasm-compilable as-is).** The current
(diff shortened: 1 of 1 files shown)


yschimke:
Addressing the two Codex review notes:

- **P1 (agent identity) — false positive.** The commit is authored *and* committed as `Yuri Schimke <yuri@schimke.ee>` (`git log -1 --pretty='%an <%ae> | %cn <%ce>'` → `Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>`), and the **`Reject agent attribution` check passed**. No `codex@openai.com` / agent email is present, so there's nothing to re-author.
- **P2 (Android-only catalog sources) — fixed in 8b72a0d.** Correct catch: `design-catalog-m3` / `wear-m3` are Android application modules using `@Preview` tooling / `Configuration` / `androidx.wear.compose.*`, which `wasmJs` can't compile. The doc now states that model 1 requires re-authoring the M3 component set into a `commonMain` CMP surface (no `@Preview`), that the Wear catalog stays server-only (no wasm target for `wear.compose`), and that the M3 port is part of the follow-up — not free reuse.

---

PR #2099, opened 2026-06-29
feat(serve): carry the CMP Wasm app on the trusted catalog branch + cache it

yschimke:
## Summary

Answers "can the Wasm be served from the target branches, and cached?" — **yes to both**, server half here. The in-browser CMP tier can now be delivered from the `design-artifacts/<system>` branch (the same trust anchor as the catalog) instead of requiring a local `--wasm-dir`, and the heavy payload is cached.

- **Branch-sourced Wasm.** When a fetched `catalog.json` declares a `webRender` (`kind: "compose-wasm"`, a `path`, and a `files` list), `ServeCatalogStore` fetches the app from the same branch into `<dir>/web/wasm/` — each file path-contained, size-capped, and traversal-rejected exactly like the catalog images — and registers it so the server serves it at `/wasm/<system>/`, **trusted by the same branch origin**. A deployed `--public --catalogs` server then needs **no local wasm build**; `--wasm-dir` still overrides per-system for local iteration.
- **Caching.** The `/wasm/` route now sends `Cache-Control: public, max-age=…` + a size+mtime `ETag` and answers `If-None-Match` with `304`, so the ~8 MB-gzipped skiko + app wasm is cached and revalidated cheaply instead of re-downloaded on every viewer load.

Why not jsDelivr in this PR: the *dev* wasm is 21 MB raw (at/over jsDelivr's per-file limit), so the server route is the robust delivery; jsDelivr/Pages over the branch is a documented option for the smaller `wasm-opt` build, covered in the follow-up.

## Test plan

- `./gradlew :cli:test` — new `ServeCatalogStore` coverage: a `webRender` catalog fetches + registers the `web/wasm/` dir; a `../../escape.html` entry in the `files` list is rejected (stays out of the dir); no `webRender` registers no wasm dir. Existing catalog/auth/fixture tests still pass.
- Content-type + viewer-mount coverage from the prior PRs still applies; the route change is additive (cache validators).

## Follow-up (PR-2)

`design-artifacts.yml` + the catalog generator build `wasmCatalogDist` and publish `web/wasm/` + the `webRender` descriptor into the branch (git content-addressing makes an unchanged re-push ~free), the gallery links it, and jsDelivr serves the `wasm-opt` build. Text-only PR (server/data-plumbing; no rendered-surface change).

---
_Generated by [Claude Code](https://claude.ai/code/session_01K8nJEfEYGM3ijyoAXDNLGH)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt
@@ -270,7 +270,9 @@ class ServeCommand(args: List<String>) : Command(args) {
       registry.register(id, host = bundleHost, pinned = true)
     }
     // Serve our published design systems from their trusted `design-artifacts/<system>` branches.
-    if (catalogs.isNotEmpty()) registerCatalogs(registry)
+    // A catalog that carries a `web/wasm/` app yields a system→dir entry so the in-browser tier
+    // rides the same trusted branch (no local --wasm-dir build needed).
+    val catalogWasm = if (catalogs.isNotEmpty()) registerCatalogs(registry) else emptyMap()
     // Runtime ingestion (--accept-bundles): clients POST a bundle (or a ?url= to one) and it's
     // registered as a pinned session. Unpacked under a temp dir for this server's lifetime.
     val bundleStore =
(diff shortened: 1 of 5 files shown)


yschimke:
Thanks — both addressed:

- **Require all declared files before registering** (P2, `fetchWasmApp`): fixed in e473d5c. It's now all-or-nothing — register only if every declared file is fetched and an `index.html` is present; on any miss (404/timeout, traversal/escaping entry, or a list over the cap) it deletes the partial dir, warns, and leaves the in-browser tier off so the viewer never advertises a "Run in browser (Wasm)" toggle whose iframe would 404. Tests cover the complete / missing-required-file / traversal cases.

- **No `webRender` writer yet**: correct, and intentional. This PR is deliberately the *consumer* half — the producer side (`design-artifacts.yml` + `generate-design-catalog.mjs` writing `web/wasm/` + the `webRender` descriptor into the branch) lands as the follow-up PR-2, called out in the PR description. Until then the branch-backed map stays empty and `--wasm-dir` supplies the app, exactly as you noted; nothing regresses. The fail-closed change above also means a half-published branch can't light up a broken tier before PR-2 lands.

---

PR #2232, opened 2026-07-05
fix(design-artifacts): embed the measured Roboto face in published figma-svg exports

yschimke:
## Summary

The published `compose/figma-svg` stickers rendered each text label with
`font-family="sans-serif"` and **no embedded face**, so a browser substituted
whatever generic sans it had. The layered SVG left-anchors each label at the
glyph origin the daemon measured against **Roboto**, so any wider substitute
runs the label past the box we measured — the centering skew reported on
`button-elevated.svg` and other labelled components.

This sets `composeai.figma.embedFonts=true` on the `design-artifacts.yml` render
step (via `JAVA_TOOL_OPTIONS`, inherited by the `bundle pack` daemon
subprocess). The export now emits an `@font-face` carrying the real Roboto
WOFF2 and `font-family="Roboto"`, so the SVG renders the **exact measured face**
in any viewer instead of depending on font substitution. The daemon fetches the
WOFF2 from the same Google-Fonts source it already uses for the PNG renders, and
degrades to `sans-serif` if unreachable — so it never fails a render, only
improves it.

## Why the label drifted (measured)

`button-elevated.svg` emits the label at `x=105`, left-anchored, sized to the
Roboto advance of **140.0px**. Rendered advance of the same label at 36.75px/500:

| Face the viewer resolves `sans-serif` to | advance | Δ vs Roboto | right-edge drift |
|---|--:|--:|--:|
| **Roboto** (measured — what we now embed) | 140.0px | — | centred |
| Liberation Sans / Arial-metrics | 143.0px | +3.0px | ~1.5px |
| **DejaVu Sans** (common Linux/browser default) | 160.7px | +20.7px | **~10px** |

So the drift is entirely viewer-dependent; embedding pins every viewer to the
measured 140px Roboto.

## Visual evidence

The rendered before/after comparison is committed at
**`docs/images/figma-svg-embed-fonts-button-elevated.png`** and renders inline
in this PR's **Files changed** tab. (Inline image/URL markdown in the PR
body gets neutralised by the exfiltration guard, so the image is carried as a
committed file rather than embedded here.)

It overlays both labels at the same emitted origin (`x=105`, left-anchored):
solid purple = embedded Roboto (this PR), sitting on the pill centre; red
outline = a wide fallback (DejaVu Sans) as a viewer without Roboto saw it —
overshooting the centre by ~10px.

Verified end-to-end on a real `bundle pack --module :samples:design-catalog-m3`
render with the flag set: every sticker's `compose-figma.svg` now contains
`@font-face` + `font-family="Roboto"` (was `font-family="sans-serif"`, no
`@font-face`).

## Test plan

- [x] `compose-preview bundle pack --module :samples:design-catalog-m3 --with-semantics` with `JAVA_TOOL_OPTIONS=-Dcomposeai.figma.embedFonts=true` → 62/62 previews render; on-disk `compose-figma.svg` sidecars carry `@font-face` + `font-family="Roboto"`.
- [x] Rasterised `button-elevated.svg` before/after in Chromium; label sits on the pill centre with the embedded face.
- [ ] After merge: re-dispatch `design-artifacts.yml` (ref `main`) so `design-artifacts/{compose-m3,wear-m3,remote-m3}` republish with the embedded face, and confirm the run succeeds.

## Follow-up (out of scope)

Serif / monospace text specimens still capture a `null` font family and fall
back to the default face (already `sans-serif` before this change; `Roboto`
after) — the figma-svg text layer doesn't yet capture non-default families.
Tracked as a separate fidelity gap, not a regression here.


yschimke:
### Visual evidence — `button-elevated.svg`

Both labels share the emitted origin (`x=105`, left-anchored). **Solid purple** = embedded Roboto (this PR) — sits on the pill centre (dashed line). **Red outline** = a wide fallback (DejaVu Sans) as a viewer without Roboto saw it — overshoots the centre by ~10px, which is the reported drift.

button-elevated: wide fallback overshoots the centre; embedded Roboto sits on it

---

PR #2305, opened 2026-07-08
fix(tiles): render tile images registered via ProtoLayoutScope

yschimke:
## Summary

Wear tiles that supply images through the **modern protolayout scope API** — `materialScopeWithResources` + `avatarImage` / `basicImage` — render **blank** in the preview pipeline, while text content renders fine. This is why the wear-os-samples Social tile shows empty contact avatars.

**Root cause:** those APIs register `ImageResource`s into the `TileRequest`'s `ProtoLayoutScope` during `onTileRequest`, *not* through `onTileResourceRequest`. `TilePreviewData`'s default `onTileResourceRequest` returns an empty bundle, and `TileRenderer` only inflates the `Resources` we hand it — so `TilePreviewComposable` dropped every scope-registered image.

**Fix:** after `onTileRequest`, harvest `tileRequest.scope.collectResources()` and merge it on top of the `onTileResourceRequest` map before inflating (and before the IR-sidecar capture, so exported bundles carry the images too). The classic explicit-`onTileResourceRequest` mapping path is byte-for-byte unchanged when the scope is empty.

Adds `ScopeImageTilePreview` to `samples/wear` as regression coverage for the scope path, alongside the inline/drawable previews from #2301.

## Before (the bug)

wear-os-samples Social tile — avatar contacts blank while the text chip renders; and the all-avatar variant fully blank:

 

## After / verification

I could **not** render locally: this project is on AGP 9.2.1 (needs Gradle 9.x) and the Gradle 9.6.1 distribution is policy-blocked in my environment. What I did verify:

- **`kotlinc` type-check** of both changed files against the real `protolayout` 1.4.0 / `tiles` 1.6.0 jars — clean, 0 errors/warnings.
- Confirmed the renderer resource path and the `ProtoLayoutScope` collect/merge semantics by decompiling `protolayout-renderer` / `tiles-renderer`.

The **Compose Preview** CI workflow renders `ScopeImageTilePreview` on this PR — that is the "after" proof (the avatar fills the tile instead of rendering blank). Once this merges, the wear-tiles integration avatars above fill in too.

## Test plan

- [ ] CI **Compose Preview** renders `ScopeImageTilePreview` with a visible avatar (not blank)
- [ ] `InlineImageTilePreview` and `DrawableImageTilePreview` still render unchanged
- [ ] No change to tiles that use the explicit `onTileResourceRequest` mapping

_Generated by [Claude Code]_

---
_Generated by [Claude Code](https://claude.ai/code/session_01NXcdivu4Qiqb6wEoG6M4YH)_

### DIFF
--- a/renderers/android/src/main/kotlin/ee/schimke/composeai/renderer/TilePreviewRenderer.kt
+++ b/renderers/android/src/main/kotlin/ee/schimke/composeai/renderer/TilePreviewRenderer.kt
@@ -13,6 +13,7 @@ import androidx.compose.ui.viewinterop.AndroidView
 import androidx.wear.protolayout.DeviceParametersBuilders
 import androidx.wear.protolayout.DeviceParametersBuilders.DeviceParameters
 import androidx.wear.protolayout.LayoutElementBuilders
+import androidx.wear.protolayout.ProtoLayoutScope
 import androidx.wear.protolayout.ResourceBuilders
 import androidx.wear.protolayout.proto.LayoutElementProto
 import androidx.wear.protolayout.proto.ResourceProto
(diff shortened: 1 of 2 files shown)


yschimke:
Merge conflicts

PR #2385, opened 2026-07-10
fix(figma-svg): raster a container fill whose painter isn't a plain ColorPainter

yschimke:
## Summary

A Wear `SwitchButton` pill and the lazy-list `Card`s in the scaffold templates fill their container via `Modifier.paint(painter)` where the painter is a **component-private `Painter`** — `SwitchButtonKt$…$colorPainter$1$1` (an animated colour painter) or `androidx.wear.compose.material3.lazy.BackgroundPainter` — **not** a plain `ColorPainter`. Its captured string is just a class name, so the token resolver (which only understands `ColorPainter(color=Color(…))`) can't read a flat colour, `backgroundColor` stays null, and **the whole fill silently vanished from the export** — the switch pill disappeared and every scaffold card went missing.

Rather than teach the resolver each component's private painter type (reflection into opaque classes we'd have to study ahead of time), this falls back to the frame: **in hybrid mode, a paint-filled node whose painter we can't vectorise captures its painted region as an `<image>`** — exactly how an opaque `Image`/`Icon` is already handled. The one painter we *can* turn into a flat vector fill is a `ColorPainter`, so **"the painter string isn't `ColorPainter(…)`" is the general signal to raster** — keyed only off the captured painter string and the `paint` modifier's own drawn bounds, with zero per-component knowledge. A resolvable `ColorPainter` (the wear buttons/cards fixed earlier) still stays a clean vector fill; a vector-only export (no frame to crop) emits no dangling `<image>`.

## Visual evidence

Reference render vs figma-svg **before** (fill dropped) vs **after** (raster fallback), from the committed wear capture trees at density 2.0:

wear paint-raster fallback before/after``

- **SwitchButtonOn**: the purple pill + toggle were gone (only "Wifi" survived) → fully restored.
- **TimeTextScaffold**: the dark list cards were missing → present again. (The `10:10` curved TimeText at the very top is still absent — a separate follow-up.)

## Offline fidelity (content-bbox SSIM vs the reference renders)

Regenerated the figma-svg for every wear preview from the committed `layout-inspector.json` capture trees (the export is a pure JVM function of those + the frame) and scored content-bbox SSIM. Biggest movers:

| component | before | after |
|---|---:|---:|
| SwitchButtonOff | 11.8 | **99.7** |
| SwitchButtonOn | 16.7 | **99.8** |
| BlankListLayout (×3 sizes) | 17–24 | **70–71** |
| EdgeButton* / ScalingList* (×3) | 20–28 | **67–71** |
| TimeText / EdgeButton scaffolds (×3) | 20–26 | **65–70** |

Switches clear 98%; the scaffold templates jump +40–55pt but plateau ~70% — their remaining gap (curved TimeText, and text penalised by the offline scorer not embedding fonts) is a separate follow-up.

## Test plan

- `FigmaLayeredSvgTest` (core, pure-JVM) — **new**:
  - `aPaintFillWhosePainterIsNotAColorPainterRastersFromTheFrameInHybridMode` — a non-`ColorPainter` paint fill rasters over its drawn region and drops the subtree;
  - `aPlainColorPainterFillStaysVectorAndDoesNotRaster` — a solid `ColorPainter` stays a vector rect even in hybrid mode;
  - `anUnvectorizablePaintFillStaysVectorOnlyOutsideHybridMode` — vector-only export emits no dangling `<image>`.
- `DesktopLayoutInspectorTest` (daemon:desktop) — **new** `a_standard_background_fill_stays_vector_and_never_hits_the_paint_raster_fallback`: a real `material3.Button` (fills via `BackgroundElement`, not `paint`) produces no raster target even in hybrid mode — pins the compose-m3 catalog against the fallback ever cropping a standard button/card.
- `./gradlew :data-layoutinspector-core:test :daemon:desktop:test --tests DesktopLayoutInspectorTest :data-layoutinspector-connector:test --tests ComposeFigmaSvgRasterTest` — green.

---
_Generated by [Claude Code](https://claude.ai/code/session_014nC2f2D5MWRPnJeqnrrq5Y)_


yschimke:
Good catch — fixed in 5e4fe5f. A `ColorPainter` carrying a `colorFilter` now falls to the raster path too (the resolver leaves it unresolved because the tint can't flatten, and its painter string starts with `ColorPainter(` so the prefix check alone would have skipped it). An unfiltered `ColorPainter` that didn't resolve is a fully-transparent fill and is still left un-rastered. Added `aColorPainterFillWithAColorFilterRastersBecauseTheTintCannotVectorise` + `aTransparentColorPainterFillWithoutAFilterDoesNotRaster` to pin both.

---

PR #2395, opened 2026-07-11
feat(figma-svg): export Wear curved TimeText as an SVG textPath

yschimke:
## Summary

A Wear `CurvedLayout`/`TimeText` draws its clock along an arc via a `CurvedTextChild` that **no `LayoutNode` represents** — so a normal layout-tree walk never sees it, and the figma-svg export dropped it entirely. Every round `TimeText`/`EdgeButton` scaffold was missing its "10:10".

Now the layout-inspector capture reflects the curved-child tree out of the `CurvedLayout` node's measure policy / draw modifiers (following `ContainerChild` children + single-child wrappers to each `CurvedTextChild`) and records, per run, the string + its **baseline arc** (`centre` + `radius` + `startAngle`/`sweep` from `CurvedLayoutInfo`, decoded to root-pixel space) + font size / weight / colour. The export renders each as an editable SVG `<path>` baseline arc + `<textPath>`, so the clock comes back as **real curved text** — editable in Figma, not dropped and not baked to a raster — reading across the top of the round scaffolds exactly as the render draws it.

All reflection is best-effort and typed-guarded: any failure yields no curved text and the node is unchanged, so non-Wear captures are unaffected. `layout/inspector` schema → v4 (additive `curvedTexts`).

Verified end-to-end against a local Android SDK: the real Robolectric render + bundle emits the `<textPath>` clock on all 9 TimeText/EdgeButton scaffolds, and the extracted geometry (e.g. Small Round: centre (192,192), start 4.465 rad, sweep 0.494 rad, r≈160) places "10:10" exactly where the render draws it.

## Visual evidence

Reference render vs figma-svg **before** (clock dropped) vs **after** (editable `<textPath>` clock):

wear curved TimeText before/after``

The "10:10" is restored, curved along the top arc at the captured weight, matching the render.

## Notes / follow-ups

- This is primarily a **correctness + editability** win (the clock is now present as real curved text). Its SSIM contribution is small — "10:10" is ~2% of the frame — so it doesn't by itself cross the round scaffolds over 98%; the baseline radius has a ~few-px vertical offset left to tune.
- The capture path is deliberately isolated behind a "component name contains `Curved`" gate + all-reflective extraction, so it adds nothing to the hot path for ordinary previews.

## Test plan

- `FigmaLayeredSvgTest` (core, pure-JVM) — **new** `curvedTextIsEmittedAsATextPathOnItsBaselineArc`: a synthetic `LayoutInspectorCurvedText` emits a baseline-arc `<path>` + `<textPath>` carrying the string, captured weight, and alpha-stripped fill.
- `:data-layoutinspector-core:test` green; `:data-layoutinspector-connector` + `:daemon:android:compileDebugKotlin` compile (verified against the Android SDK).
- Real `:samples:design-catalog-wear-m3` render + `bundle pack`: `<textPath>` present on the 9 TimeText scaffolds, "10:10" placed on the top arc (see evidence).

---
_Generated by [Claude Code](https://claude.ai/code/session_014nC2f2D5MWRPnJeqnrrq5Y)_


yschimke:
Good catch — fixed in 5d7e592. Curved-text arc ids now come from a document-wide sequence (`curve-c0`, `curve-c1`, …) threaded through the render walk, so two same-named `CurvedLayout`s no longer collide and every `<textPath href>` resolves to its own path. Added `twoCurvedLayoutsWithTheSameNameGetDistinctPathIds` to pin it.

---

PR #2397, opened 2026-07-11
fix(mcp): resolve daemon descriptors for projectDir-remapped modules; bump stale descriptor schema check

shovel-kun:
## Problem

On projects whose Gradle module paths don't mirror the on-disk layout — e.g. a flat `:featureTasks` remapped to `shared/features/tasks` via `project(":featureTasks").projectDir = ...` in `settings.gradle.kts` — the MCP warm-daemon path is unusable:

- **Every `render_preview` fails** with `Missing daemon launch descriptor for :featureTasks …`, even though `mcp install` / `composePreviewDaemonStart` wrote the descriptor correctly.
- **`mcp doctor` reports every module** as `schemaVersion=2, expected 1 (plugin upgraded — re-run mcp install)`, and re-running install never clears it.

## Root causes

1. **`DaemonSupervisor.readingFromDisk`** resolves the descriptor by converting the Gradle path to a directory (`gradlePathToFile`: `:featureTasks` → `<root>/featureTasks`). For a remapped `projectDir` the real descriptor lives at `shared/features/tasks/build/compose-previews/daemon-launch.json`, so the lookup always misses. `mcp install` writes to the right place (it has `module.projectDir`), but the supervisor guesses from the path string.
2. **`EXPECTED_DESCRIPTOR_SCHEMA_VERSION` is pinned to `1`** in `McpCommand.kt`, while every writer (`DAEMON_DESCRIPTOR_SCHEMA_VERSION`, `DAEMON_LAUNCH_SCHEMA_VERSION`) stamps `2`. The constant's own comment says *"keep in sync — bump together"*; it just wasn't bumped, so the doctor check is a false positive at every release.

## Fix

1. Keep the layout fast-path, but when it misses, index descriptors by the `modulePath` recorded *inside* each `daemon-launch.json` (a pruned, cached filesystem scan — skips `.git` / `.gradle` / `node_modules` / `src` and non-`compose-previews` `build/` subtrees). Robust to any `projectDir` remapping. Added a regression test reproducing the remapped layout.
2. Bump `EXPECTED_DESCRIPTOR_SCHEMA_VERSION` to `2`.

## Testing

- New hermetic test `readingFromDisk resolves descriptor for a projectDir-remapped module` passes.
- Against a real remapped project (Compose Multiplatform, 11 modules): `mcp doctor` → all modules `ok`; a standalone `mcp serve` renders a remapped-module preview end-to-end (previously "Missing descriptor").

These are two independent fixes — happy to split into separate PRs if preferred.


yschimke (approved):
Thanks for this — both fixes are real and correctly targeted, and I verified them against the tree:

- **Schema bump 1→2** is right: every writer stamps `2` (`DAEMON_DESCRIPTOR_SCHEMA_VERSION` in `DaemonClasspathDescriptor.kt`, `DAEMON_LAUNCH_SCHEMA_VERSION` in `ServeBundleDaemon.kt`) while `EXPECTED_DESCRIPTOR_SCHEMA_VERSION` was still `1`, so `mcp doctor` was a false-positive at every release. The comment even says "keep in sync — bump together."
- **Remapped `projectDir` resolution** is the right general approach: `gradlePathToFile` can't reconstruct a `projectDir` remap, so indexing descriptors by the `modulePath` recorded *inside* each `daemon-launch.json` is exactly the robust signal. Fast-path-first keeps normal layouts free, and the walk pruning keeps the fallback cheap.

The Codex **P2 (re-scan after a cached negative miss)** is valid — I reached the same conclusion independently. If the server is already running and the descriptor is generated afterward (exactly what the error message tells the user to do), the cached index keeps returning the miss until restart. That said, this PR is a **strict improvement** — remapped modules resolve zero of the time on `main` today, and the stale-cache case only bites a descriptor written *after* the first miss and is recoverable by a restart — so I'm landing it as-is and we'll **follow up** with the negative-cache invalidation (rebuild the index on a miss for an absent module) in a separate change. No need for you to do anything here.

Approving and merging. Thanks again for the clear write-up and the regression test.

---

PR #2409, opened 2026-07-11
fix(cloud): seed Gradle distribution from repo.gradle.org proxy before warm

yschimke:
## Summary

The SessionStart Gradle warm (`./gradlew help`) and the first real `./gradlew` died in locked-down cloud sandboxes (Claude Code on the web's "custom" network mode) with:

```
Downloading https://services.gradle.org/distributions/gradle-9.6.1-bin.zip
Attempt 1/1 failed. Reason: Server returned HTTP response code: 403
java.io.IOException: Server returned HTTP response code: 403
```

**Root cause:** the wrapper's `distributionUrl` (`services.gradle.org`) 307-redirects to a `github.com` release asset (`gradle/gradle-distributions/releases/...`), and the egress policy blocks `github.com` (403). So the distribution could never be fetched, and no Gradle task worked.

**Fix:** new `scripts/seed-gradle-dist.sh` seeds the wrapper distribution into `$GRADLE_USER_HOME/wrapper/dists` from `repo.gradle.org`'s github-downloads-proxy — which serves the identical, checksum-verified bytes and is an allowlisted Gradle host. Gradle then unpacks it offline on the first invocation (no repo files are written; the wrapper properties are read-only). The SessionStart hook runs it before the JDK setup and warm.

## Details

- **Single- and multi-repo checkouts:** discovers every `gradle-wrapper.properties` across the side-by-side clones (workspace root = parent of this repo), dedups by `distributionUrl`, and seeds each distinct Gradle version. A single-repo layout falls back to seeding just this repo.
- **Idempotent + best-effort:** no-ops when the `.zip` or `.zip.ok` marker is already cached — a warm box, a prior run, or an environment where `coo.ee/env`'s java module already seeded it. Any failure (blocked mirror, checksum mismatch, no wrapper) warns and leaves the download to Gradle without aborting the session.
- **Checksum-verified:** prefers the wrapper's pinned `distributionSha256Sum`, else the mirror's published `.sha256`; a mismatch refuses to seed.
- Only rewrites the stock `services.gradle.org` distribution URL; a custom/self-hosted `distributionUrl` is left untouched.

## Test plan

Exercised in a live cloud sandbox (github.com blocked, 403):

- `services.gradle.org/distributions/gradle-9.6.1-bin.zip` → 403 (reproduced the failure); `repo.gradle.org/gradle/github-downloads-proxy/.../v9.6.1/gradle-9.6.1-bin.zip` → 200.
- Clean cache → `seed-gradle-dist.sh` seeds from the proxy, checksum verified against the published `.sha256`, zip lands in the exact wrapper cache dir (`gradle-9.6.1-bin/4ticwg1pgcbps2hj28r8so764/`).
- `./gradlew --version` then runs fully offline from the seeded distribution.
- Second run is a clean no-op (`already present; nothing to seed`) — confirms idempotency after coo.ee/env or a prior run.
- The `base36(md5(distributionUrl))` cache-dir hash matches Gradle's own `PathAssembler#getHash` output.

## Notes

A related, separate blocker remains out of scope here: `scripts/setup-cloud-jdk.sh` provisions JDK 17 from `github.com/adoptium/...` release assets, which the same policy blocks (403). With no JDK 17 on the box, tasks that start the daemon still fail toolchain resolution even once the distribution is seeded. Happy to follow up with a non-GitHub JDK source if wanted.

_Generated by [Claude Code]_

---
_Generated by [Claude Code](https://claude.ai/code/session_012iPYAXfPHfVuLj5vEeyRYm)_

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -31,6 +31,20 @@ if [ "${CLAUDE_CODE_REMOTE:-}" != "true" ]; then
   exit 0
 fi
 
+# Seed the Gradle wrapper distribution(s) from the allowlisted repo.gradle.org
+# github-downloads-proxy BEFORE any `./gradlew` runs. The wrapper's
+# distributionUrl (services.gradle.org) 307-redirects to a github.com release
+# asset, which the "custom" cloud egress policy blocks (403) — so the warm below
+# would otherwise die with `Server returned HTTP response code: 403` before the
+# build starts. The proxy serves the identical, checksum-verified bytes and is an
+# allowlisted Gradle host. Covers single- and multi-repo checkouts (every wrapper
+# across the side-by-side clones). Best-effort: a failure leaves the download to
+# Gradle and never aborts the session.
+if [ -x scripts/seed-gradle-dist.sh ]; then
+  scripts/seed-gradle-dist.sh >&2 || \
+    echo "[session-start] seed-gradle-dist.sh failed; Gradle will fetch the distribution itself" >&2
+fi
+
 # Ensure a JDK 17 toolchain exists before warming Gradle. The build pins
 # `toolchainVersion=17` (gradle/gradle-daemon-jvm.properties), but cloud
 # containers commonly ship only JDK 21, and on the allowlist network policy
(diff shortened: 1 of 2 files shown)


yschimke:
Closing in favor of centralizing this in the env setup (`env.coo.ee`). This environment provisions with `curl -fsSL https://env.coo.ee/compose,android-emulator,java,android | bash`, and coo-ee-env's `java` module already seeds the Gradle wrapper distribution from the `repo.gradle.org` github-downloads-proxy across all checkouts — so a seed script here duplicates that. The one net-new improvement from the Codex review (honoring `zipStoreBase`/`zipStorePath`) is being applied to coo-ee-env's seeder instead, which is the single source of truth.

---

PR #2496, opened 2026-07-15
feat(serve): fold component states into one grid card + add a viewer state switcher

yschimke:
## What & why

The preview server (`compose-preview serve`, hosted at preview.coo.ee) showed each baked component **state** as its own grid card — `checkbox-checked` beside `checkbox-unchecked`, `button-filled` beside its pressed/focused/disabled renders — and offered **no way to flip a component between its states** in the viewer. This change:

1. **Folds non-default states out of the landing grid**, so a component shows **one** (default) card instead of a card per state.
2. **Adds a state switcher to the viewer** — a `<nav class="cp-states">` of plain links to the component's other states *in the same theme*, the current one marked `aria-current="page"`. No daemon, no JS state machine, so a scripting-off client still navigates.

The catalog + generator were already correct (states are `variant`s folded onto one component's `images[]`, tagged with `state`/`theme`); the gap was entirely **serve-side**.

### Mechanics
- `ServeCatalogStore` writes a `previews/variants.json` manifest from each catalog `Image`'s `state`/`theme` (atomic, before the staging swap).
- `ServeBundleHost` reads it back (best-effort) and tags each `ServePreview` with `state`/`theme`.
- `ServeWeb` filters `isNonDefaultState` out of the grid and renders the switcher (`stateSwitcherHtml`).
- Plain (stateless) bundles carry no manifest → unchanged everywhere.

## Visual evidence

**Landing grid** — non-default states fold into the default card (4 cards → 2):

| Before | After |
| --- | --- |
| grid before | grid after |

**Viewer** — the new `Default` / `Unchecked` state switcher (absent before):

| Before | After |
| --- | --- |
| viewer before | viewer after |

(Component renders in the shots are the real baked checkbox/radio PNGs; the `/render` lane has no backend in a static capture.)

## Tests
- **`ServeWebTest`** (new): the grid folds a non-default state into one card; the viewer renders a same-theme switcher with the current state active; a single-state component renders no switcher; a plain stateless catalog is unchanged.
- **`ServeCatalogStoreTest`**: a state-bearing catalog writes `variants.json` (null keys omitted) that round-trips onto host previews.
- **`ServeBundleHostTest`**: previews are tagged from the manifest; a plain bundle keeps null `state`/`theme`.
- **`ServeWebFixtureTest`**: goldens regenerated; two new fixtures (`serve-landing-states`, `serve-viewer-states`) added so the CI visual-diff bot captures the grid fold + switcher on every future PR.

Run: `./gradlew :cli:test --tests "*.serve.*"` — green.

## Notes
- Design write-up kept at `docs/design/SERVE_STATE_TOGGLE_SPEC.md`.
- compose-m3 `RadioButton` needs an `unselected` variant to actually toggle — that's a separate catalog PR (#follow-up on `agent/radiobutton-unselected`).

_Generated by [Claude Code]_

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y3BJh2uPPLS2ErfxJmdqgL)_


yschimke:
Good catch on the state-switcher scoping (P2) — fixed in `db83cd5`. The switcher now groups siblings by a **state-invariant key** (the id with only its state segment removed) plus theme, so it toggles only the state axis and holds every other axis (theme, `content`/`size`/props) fixed. A `content=icon+label` render therefore keeps its props segment and no longer cross-links to the label-only button, and a variant with no state sibling of its own shows no switcher. Added a regression test (`the state switcher stays within the current variant axis, not just the slug`).

---

PR #2508, opened 2026-07-16
fix(serve): keep the live render lane up for knob overrides and Wear

yschimke:
## Summary

Two failure modes on the public preview server (`preview.coo.ee`) collapsed the live, daemon-backed render lane to the baked-PNG snapshot:

### 1. `?knob.…=` on `/render/{id}.png` → `500 timed out after 10s` (compose-m3)

Reproduced against production (`/version` = `0.16.50`):
- `…/render/button-filled__ideal__default__dark.png` → **200**, PNG ✓
- `…?knob.label=Hello` → **500**, body `timed out after 10s waiting for render`

A knob override is a valid string override that flips `/render` off the instant baked-PNG lane onto a **live desktop re-render**. That render is governed by two budgets in `ServeRenderHost`: a cold budget (`renderTimeoutSeconds`, lifted to 900s by the deploy image) and a warm-frame budget (`frameRenderTimeoutSeconds`, **10s**, not lifted). Because the deploy image runs `warmInBackground`, the background prewarm flips the host-wide `warmedUp` flag, so the user's *first real override render* is charged the tight 10s frame cap and times out.

**Fix:** `ServeRenderHost` now gives the **first override-bearing render** the cold budget too (a prewarm/no-override render no longer starves it); steady-state override frames stay capped so a wedged render can't pin the slot. The deploy image also sets `composeai.serve.frameRenderTimeoutSeconds` alongside the cold budget it already lifts (this also unstarves warm Wear/Robolectric frames, which the 10s default is too tight for).

### 2. wear-m3 live streams → `input requires a live stream`

The `/wear-m3/…` live daemon logs showed `RobolectricHost SandboxRunner[3] failed: Unable to load Robolectric native runtime library`. The Android daemon boots a **5-sandbox pool** (`warmSpare`), and under memory pressure on the 8 GB box one sandbox intermittently fails to load the native runtime. `RobolectricHost.start()` was **all-or-nothing** — any single slot's boot failure aborted the whole host, so the live lane fell back to the snapshot session, whose `Input` handler emits exactly `input requires a live stream`. (Logs prove the native runtime loads fine normally on that box — an earlier full pool booted and rendered — so the failure is transient, not a hard incompatibility.)

**Fix:** `RobolectricHost.start()` now **retries a transient per-slot boot failure** (2 retries; a fresh worker + reset latch each time) before giving up, so one flaky sandbox no longer takes down the whole live lane; an unrecoverable slot still fails exactly as before. The deploy image pins the pooled Android daemons to `sandboxCount=3` (interactive needs ≥2) to relieve the memory pressure that triggers the failure.

## Changes
- `cli/…/serve/ServeRenderHost.kt` — first override render gets the cold budget.
- `daemon/…/RobolectricHost.kt` — bounded per-slot boot retry in `start()`.
- `deploy/image/Dockerfile` — `frameRenderTimeoutSeconds=120` + `sandboxCount=3` in the runtime env.

## Test plan
- `:cli:test` — `ServeRenderHostTest` (22) + `ServeRenderHostStreamTest` (7): **green**.
- `:daemon:android:compileDebugKotlin` + `:cli:compileKotlin`: **clean**.
- Reproduced the production 500 and root-caused it to the warm-frame budget above.
- The `:daemon:android` Robolectric pool tests can't run in the agent container (a nix-vs-glibc `libstdc++` mismatch blocks Robolectric's native libs — it blocks unmodified code too); the new retry path was confirmed to execute and surface the real underlying cause. CI exercises those tests where `libstdc++` is present.

## Rollout note
This is non-visual (it restores rendering; pixels are unchanged), so text-only evidence. The deploy-image env changes reach `preview.coo.ee` on the next image roll; the `ServeRenderHost`/`RobolectricHost` changes ship with the next CLI release (Watchtower).

---
_Generated by [Claude Code](https://claude.ai/code/session_01SLoRFBUZCqf4rMWJbSvDFZ)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeRenderHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeRenderHost.kt
@@ -280,6 +280,16 @@ internal constructor(
   // [frameRenderTimeoutSeconds] so a single wedged render can't hold the only render slot.
   private val warmedUp = AtomicBoolean(false)
 
+  // An override-bearing render (a `?knob.…=` edit, device/locale/theme override, …) forces a real
+  // recomposition and is much slower than a plain re-emit — its first occurrence is effectively
+  // cold even when [warmedUp] is already set by a background prewarm (a throwaway default render
+  // that flips [warmedUp] without ever exercising the override path). Without a separate gate the
+  // very first override render on `preview.coo.ee` — where `warmInBackground` prewarms — is charged
+  // the tight [frameRenderTimeoutSeconds] cap and times out (the public `?knob.…` 500). Give the
+  // first override render the generous [renderTimeoutSeconds] budget too; subsequent override frames
+  // are capped like any other so a wedged one can't pin the slot.
+  private val overridesWarmedUp = AtomicBoolean(false)
+
   // Fans one upstream daemon stream out to all watchers of the same preview/overrides/codec/fps, so
   // many browsers cost one held session instead of one each. Built on [startStream]; shared because
   // there's one host per server.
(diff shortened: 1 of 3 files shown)


yschimke:
Pushed `53f8f59` addressing the review + CI:

- **Codex P2 (`RobolectricHost` boot-retry race)** — good catch. The retry is now gated on a **recorded** `workerBootErrors` entry (set only after the worker's JUnit run returns, i.e. it has demonstrably exited and never claimed its slot). A bare `awaitSandboxReady` timeout — where the original worker may still be inside Robolectric bootstrap — stays a **clean startup failure** as before, so there's no path where a replacement worker races the original to `registerSandbox` / the slot latch.
- **`ktfmt` / `Build CLI` failures** — same root: `ServeRenderHost.kt` wasn't ktfmt-formatted. Auto-formatted; `:cli:ktfmtCheck` + `:daemon:android:ktfmtCheck` are clean locally now.
- **`Plugin Functional Tests`** — unrelated flaky `OutOfMemoryError: Java heap space` in `:gradle-plugin`'s `BundleFunctionalTest` (ClassGraph scan in `BundlePreviewTask.closureWalk`); this PR doesn't touch the plugin. Expecting it to clear on this fresh run — will re-check.

Local verification: `:cli:test` (`ServeRenderHostTest` 22 + `ServeRenderHostStreamTest` 7) green, `:daemon:android:compileDebugKotlin` + `:cli:compileKotlin` clean.

---

PR #2570, opened 2026-07-18
feat(bundle): add --knob key=value to bake previews under a theme override

yschimke:
## Summary

Adds `bundle pack --knob key=value` (repeatable) so a batch bake can render every preview under a **theme override** instead of always stock. This is the seam the theme-variant work needs: `bundle pack --module :samples:design-catalog-m3 --knob theme.colors=scheme:…` re-skins the whole compose-m3 catalog under an app's palette — the same `PreviewOverride` seed the **live serve daemon** injects (`PreviewOverridesOverrideExtension`), now mirrored on the **batch** path.

**Threading** (mirrors the existing `displayFilterFilters` pipeline):
- **CLI** — `PackSubcommand` parses repeatable `--knob key=value` (split on the first `=`, so a value keeps its own `=`/`,`/`;` — a serialized `scheme:` blob is safe), Base64-packs the set, emits `-PcomposePreview.knobs=<b64>`.
- **Gradle** — `AndroidPreviewSupport.resolvePreviewOverrideKnobs` reads that property into a new `@Input @Optional previewOverrideKnobs` on `RenderPreviewsTask` (folded into the cache key, so a re-theme **re-bakes**; unset stays cache-valid and byte-identical), forwarded as the `composeai.previewoverride.knobs` system property.
- **Renderer** — `DesktopRendererMain` decodes it once and calls `PreviewOverrideController.set(seed)` before `scene.setContent`, so `CatalogSticker`'s `previewOverride*` lookups resolve the override (the default `ControllerPreviewOverrideHost` routes them). No `--knob` ⇒ empty seed ⇒ the stock render is untouched.

Theme knobs (`theme.colors` / `theme.font` / `theme.shapes` / `theme.typography`) are all string-valued, so each entry is a `StringValue` under the knob key. **No behaviour change without `--knob`.** Foundation for the per-mode bake (via `catalogThemeModes`, #2569) and the global theme selector.

## Visual evidence — Material 3 components re-skinned by a `theme.colors` override

**Stock** (`M3` palette) vs an **app `theme.colors`** (a meshcore-teal `scheme:…`):

| Stock `M3` | App `theme.colors` (teal) |
| --- | --- |
| stock | themed |

_(Rendered via `ImageComposeScene` — the same engine `DesktopRendererMain` bakes with — to show the re-skin a `--knob theme.colors=…` produces once the seed reaches the composition. The `scheme:` → `ColorScheme` decode is covered by the shared module's `CatalogColorSchemeTest`; that the seed reaches the render is asserted by `PreviewOverrideKnobSeedTest`.)_

## Test plan

- `:renderer-desktop:test --tests '*PreviewOverrideKnobSeedTest*'` — green (3 tests): the Base64 `key=value` blob decodes (splitting on the first `=`); blank/malformed → empty seed (a stock render); and **a seeded knob flips the rendered pixel** the override host drives (same `ImageComposeScene` + skiko path as the committed `UiModeSystemThemeTest`).
- `:cli`, `:gradle-plugin`, `:renderer-desktop` compile; `ktfmtCheck` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KBurh9kb3NJ2KsmYRFhU62)_


yschimke:
Closing in favor of a **bundle + overrides** approach.

This PR re-themes by rebuilding compose-m3 **from source** (`bundle pack --knob` → the Gradle `composePreviewRender` task), which only works where compose-m3's source is present (compose-ai-tools' own CI). The actual consumer — **meshcore-mobile reusing compose-m3's catalog under its theme** — has compose-m3's *published bundle* (`design-artifacts/compose-m3`), not its source, so a source rebuild can't serve that case.

Pivoting to render the **published bundle's** previews to PNG with theme overrides via the daemon's existing `PreviewOverrides` support (the same `PreviewOverrideController.set(namedOverrides)` seed the live serve path already applies) — no source rebuild required. Superseded by that work.

---

PR #2578, opened 2026-07-19
feat(theme): capture the node shape leg of the M3 triad from the render

yschimke:
## Summary

Shape attribution landed in #2574 — `ThemeConsumerAttribution` maps a node's
resolved `Shape` to the theme's shape scale — but nothing populated
`NodeThemeFacts.shape` from the render, so the leg was dark end to end. This adds
the **capture**:

- `ThemeConsumerCapture` reads each node's resolved clip/container `Shape` off
  its `LayoutNode` modifiers (`getModifierInfo()`): the inspector `shape`
  element first, then a reflected `shape` field, skipping the default
  `RectangleShape` — the same foundation-free path `ModifierTokenResolver` walks.
- A shape-bearing node is now emitted even when it carries no text (a
  `Surface`/`Card` is a non-text node, so its shape is the reason to appear);
  colour/typography still ride the text node's `TextLayoutResult` as before.

This closes the M3 triad on the capture side: a component preview now attributes
colour, typography, **and** corner shape per node, so a `DeviceBodyPreview`-style
design-parity token diff stops flagging `radius`/`shape` corner tokens as
"missing from candidate".

## Test plan

- `./gradlew :data-theme-connector:test` — `ThemeConsumerCaptureShapeTest`
  extracts the shape from `background(color, shape)` and `clip(shape)` off a real
  modifier chain, and returns `null` for a plain `RectangleShape` background and
  a shapeless `padding` modifier.
- `./gradlew :daemon:android:testDebugUnitTest --tests '*ThemeConsumerCaptureRenderTest*'`
  — renders a `Surface` clipped to `MaterialTheme.shapes.medium` under
  Robolectric, then asserts `extractFacts` captures that shape and
  `ThemeConsumerAttribution` maps it to the `medium` role. Green (1 test, 0
  failures).

Non-visual change (data-product plumbing + tests), so no rendered before/after
is applicable — the Robolectric render test is the behavioural evidence.

_Generated by [Claude Code](https://claude.ai/code/session_01PFaMPQ3D2PmcxAqsCZHD1e)_

---
_Generated by [Claude Code](https://claude.ai/code/session_01PFaMPQ3D2PmcxAqsCZHD1e)_

### DIFF
--- a/daemon/android/src/test/kotlin/ee/schimke/composeai/daemon/ThemeConsumerCaptureRenderTest.kt
+++ b/daemon/android/src/test/kotlin/ee/schimke/composeai/daemon/ThemeConsumerCaptureRenderTest.kt
@@ -1,0 +1,68 @@
+package ee.schimke.composeai.daemon
+
+import androidx.activity.ComponentActivity
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
+import androidx.compose.material3.Text
+import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onRoot
+import ee.schimke.composeai.data.theme.ResolvedThemeTokens
+import ee.schimke.composeai.data.theme.ThemeConsumerAttribution
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+/**
+ * End-to-end render check on the Robolectric (Android) backend: a `Surface` clipped to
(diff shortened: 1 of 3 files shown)


yschimke:
Good catch on the `BackgroundPainter` leg — the capture's `shapeStringOf` was mirroring only two of `ModifierTokenResolver.shapeOf`'s three fallbacks. Fixed in 496280b: it now also reflects `Modifier.paint(BackgroundPainter).shape`, so Wear `TransformingLazyColumn` scaling cards attribute their corner shape instead of coming back shapeless. Added unit coverage for the painter path (and the negative case where a non-`BackgroundPainter` behind a paint modifier still yields `null`).

---

PR #2653, opened 2026-07-22
feat(figma-svg): ship a figma-svg per preview variant, not one per component

yschimke:
## Problem

The bundle already carries `previews/<id>.figma.svg` for **every** rendered preview —
`DaemonSemanticsFetcher.Outcome.Ok.figmaSvgById` is keyed by preview id with no theme
gating, and `injectFigmaSvgIntoBundle` writes one entry per id.

But `figmaSvgByFunction` collapses that map to one vector per component *function*,
preferring the light variant:

> Prefer the light variant so a single deterministic sticker is emitted per component.

So every dark / locale / size vector is rendered, carried across the wire, and then
dropped at the last step. A consumer wanting an editable dark vector has to fall back
to the raster PNG.

## Change

Ship them, mirroring the raster set 1:1:

    images/<slug>/ideal__default__dark__compact.png     (today)
    figma/<slug>/ideal__default__dark__compact.svg      (new)

- **`figmaSvgById(bundle)`** — the uncollapsed carry (`previewId → svg`).
  `figmaSvgByFunction` is untouched and now documented as the narrowing of it.
- **`figmaVariantSvgPath(imagePath)`** — `images/<slug>/<variant>.png` →
  `figma/<slug>/<variant>.svg`, rejecting anything that isn't a plain `images/…png`
  and any `..` / absolute / backslash / empty segment, so a hostile manifest can't
  write outside `figma/`.
- **The emit is driven from the written manifest's `images[]`**, not from a naming
  scheme of its own. Each image already carries the `previewId` that keys the bundle's
  vectors *and* the `path` the export engine chose, so the mapping inherits
  `imagePath`'s collision-safe naming and a vector cannot drift away from the PNG it
  depicts.
- **Gaps are reported.** An image with a `previewId` but no carried vector increments
  `figmaVariantGapCount` and warns. A silently missing vector is exactly the failure
  this emit exists to prevent.
- **Hybrid crops key on the variant basename**, not the component slug — two variants
  of one component each carry their own `<node>.png`, which a per-slug dir would
  overwrite. `rewriteRasterHrefs`' second arg is now a generic `key`; behaviour is
  unchanged for existing callers.

**Purely additive.** `figma/<slug>.svg` and its per-slug crop dir are written exactly
as before — `index.html`, `compare.html`, design-parity's `FIGMA_IMPORT.md` and
meshcore-mobile's seeding runbook all reference that path today. The back-compat file
lives *beside* the new `figma/<slug>/` directory, so they can't collide.

## Worth knowing

`previewId` is **not** universal on manifest images: `buildCatalog` doesn't emit it —
`bridgeLivePreviewIds` stamps it on, and that only runs for a carried live bundle or a
`--source-module`. The workflow always passes `--source-module`, so every published
catalog is covered, but a manual run without it emits zero per-variant vectors *and*
zero gaps. Documented in the block comment rather than papered over with a fallback
derivation.

Images `bridgeLivePreviewIds` deliberately leaves unbridged (a state with no desktop
source, or a function the Android-only supplement overrode) are skipped rather than
counted as gaps — there's no daemon-rendered vector for them to be missing.

Consumers aren't updated: `index.html` / `compare.html` / `README.md` still only link
`figma/<slug>.svg`, so the per-variant set ships on the branch but is unlinked. That
matches the additive scope; surfacing it to a browsing designer is the natural
follow-up.

## Test plan

- [x] `node --test "scripts/design-artifacts/*.test.mjs"` — **88 pass, 0 fail**
      (15 in `figma-svg-catalog.test.mjs`, 6 of them new: `figmaSvgById` keeps every
      variant / drops non-SVG payloads, `figmaVariantSvgPath` mapping + rejection
      table, variant-basename raster keying, and a manifest-driven emit test that
      asserts both the written set and the gap count)
- [x] every pre-existing assertion kept unchanged
- [x] `node --check scripts/design-artifacts/generate-design-catalog.mjs`
- [ ] a real `design-artifacts.yml` run — the generator needs the private
      `@design-parity/catalog-export` and a real render bundle, so it can't be
      smoke-run locally

Text-only evidence: this is export-driver plumbing, not a renderable surface — no
`@Preview` output, webview or overlay changes.

Unblocks the theme axis of yschimke/meshcore-mobile#261.

### DIFF
--- a/scripts/design-artifacts/figma-svg-catalog.test.mjs
+++ b/scripts/design-artifacts/figma-svg-catalog.test.mjs
@@ -4,7 +4,12 @@
  * table reports the figma-svg count. The driver's `figmaSvgByFunction` reader + emit loop are
  * exercised end-to-end by the bundle round-trip; here we pin the two presentation surfaces.
  *
- * Run with `node --test scripts/design-artifacts/`.
+ * Also covers the per-variant emit (`figma/<slug>/<variant>.svg`, one per rendered preview,
+ * mirroring `images/<slug>/<variant>.png`): its bundle readers (`figmaSvgById` / `figmaSvgByIds`,
+ * including the `--extra-renders` fold), its manifest-driven path mapping (`figmaVariantSvgPath`)
+ * and the variant-keyed hybrid raster dir.
+ *
+ * Run with `node --test "scripts/design-artifacts/*.test.mjs"`.
  */
 import assert from "node:assert/strict";
 import { test } from "node:test";
@@ -15,6 +20,9 @@ import {
   figmaRastersForId,
   figmaSvgByFunction,
   figmaSvgByFunctions,
+  figmaSvgById,
+  figmaSvgByIds,
+  figmaVariantSvgPath,
   rewriteRasterHrefs,
 } from "./figma-svg-emit.mjs";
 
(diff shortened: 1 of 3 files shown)


yschimke:
Rebuilt on current `main` — the branch was originally cut from a stale local `main`,
about twenty commits behind, which is where the conflicts came from.

Re-applying rather than resolving the conflict turned out to matter, because
[#2654](https://github.com/yschimke/compose-ai-tools/pull/2654) and its predecessors
changed the shape underneath:

**`main` now folds multiple bundles.** `figmaSvgByFunctions([bundle, extraBundle])`
exists so an `--extra-renders`-only sticker still gets its vector, and the driver
merges raster crops across both bundles. The original per-variant loop read only the
primary bundle, so it would have emitted **zero** per-variant vectors for every
`--extra-renders`-only component — silently, no error, just missing files. That's the
common case for screens rendered from a second CMP-desktop module, and it's exactly
the first consumer's setup (meshcore-mobile's `Chat/*`, `Settings/*` and
`Device/Cached` all come from its `:meshcore-components` extra render), so the feature
would have shipped broken for the thing it was written for.

Now fixed via `figmaSvgByIds(bundles)`, mirroring `figmaSvgByFunctions`, with crop
merging in the same shape as the back-compat loop. Covered by a test using an
extra-renders-only preview id whose vector comes only from the second bundle.

Also added a guard the first pass lacked: `figmaVariantSvgPath` rejects a flat
`images/<slug>.png`, which would otherwise map onto `figma/<slug>.svg` and clobber the
back-compat per-component vector this PR promises to leave alone.

Tests: **174 pass, 0 fail** (`node --test "scripts/design-artifacts/*.test.mjs"`).

---

### Pre-existing bug found while checking #2654 — not fixed here

#2654 added the `theme` variant axis, so a screen whose light/dark renders are two
separate `@Preview` functions folds into one component with `…__light` / `…__dark`
stickers. But `bridge-live-preview-ids.mjs` can't key on theme:

- `previewForState` only registers a variant when `v.state || v.props`, so a
  **theme-only** variant is never registered, and `variantKey` has no theme component.
- A folded theme variant keeps `state: "default"`, so its image resolves back to the
  component's default preview function. `daemonIdFor.get(`${fn} dark`)` misses —
  the split-function case is precisely when `FooScreen` has no `_Dark` multipreview
  id — and it falls through to `unthemedIdFor.get(fn)`, the **light** preview's id.

So a theme-variant dark sticker currently carries the light preview's `previewId`.
That already mis-points `image.livePreview` and `ServeCatalogStore` aliasing; this PR
inherits it faithfully and would write the light vector at
`figma/<slug>/…__dark.svg`.

Deliberately out of scope: fixing it means adding theme to `variantKey` and the
lookup, which changes live-preview daemon aliasing for every catalog and needs its own
tests. Worth its own issue — happy to file it.

PR #2656, opened 2026-07-22
fix(design-artifacts): key the live-preview bridge on theme, not just state/props

yschimke:
## The bug

A **theme variant** (#2654) folds a screen whose light and dark renders are two separate
`@Preview` functions into one component. `bridgeLivePreviewIds` could not represent that:

- `previewForState` only registered a variant when `v.state || v.props` — a **theme-only**
  variant was never registered at all;
- `variantKey` had no theme segment — so even a registered one would have collided with
  the component's default entry.

A folded theme variant keeps `state: "default"`, so its image resolved back to the
component's **default** function. The daemon-id lookup then missed on `${fn}\0dark` — the
split-function case is precisely when `FooScreen` has no `_Dark` multipreview id — and
fell through to `unthemedIdFor.get(fn)`, which is the **light** preview's id.

Net effect: **a dark sticker carried the light preview's `previewId`.** That mis-points
`image.livePreview` and `ServeCatalogStore` aliasing today. And now that the per-variant
figma-svg emit (#2653) keys off `previewId`, it would write the **light vector** at
`figma/<slug>/…__dark.svg` — wrong content at a right-looking path, which is the kind of
failure nobody notices until a designer wonders why dark mode looks odd.

Found while checking #2654's interaction with #2653; deliberately left out of that PR
because this changes live-preview aliasing for every catalog.

## The fix

Add `theme` as an optional trailing segment of `variantKey`, and register theme-only
variants. The resolver tries the **theme-qualified key first, bare key second**:

- a **theme-folded** component registers only the qualified key → resolves to its own
  per-theme function;
- a **multipreview-themed** component (compose-m3: one `@Preview` rendered `_Light`/`_Dark`)
  has no theme in its spec → misses the first lookup → lands on the same entry it always did;
- an **un-themed state-variant** catalog (wear-m3 / remote-m3) never passes a theme at all.

So both pre-existing catalog shapes take exactly the path they took before — pinned by the
two existing tests, which pass unchanged.

The daemon-id lookup below needed no change: a per-theme function whose preview id ends in
`dark` resolves via `daemonIdFor`, and one whose id carries no theme suffix falls back to
`unthemedIdFor` — correct either way, since the function is already theme-specific.

## Test plan

- [x] `node --test "scripts/design-artifacts/*.test.mjs"` — **177 pass, 0 fail**
- [x] **Three new tests, all of which FAIL against the unfixed resolver** (verified by
      reverting the source and re-running — 3 fail, and the 2 pre-existing catalog-shape
      tests still pass, so they pin the bug rather than describing current behaviour):
  - theme-folded catalog: each theme maps to its own function, and the two ids must differ
  - theme-folded: a per-theme function whose preview id has no light/dark suffix still resolves
  - theme and state coexist: a themed variant of a non-default state keys on both
- [x] both pre-existing tests (themed compose-m3, un-themed wear-m3) unchanged and passing

Text-only evidence: this is catalog-manifest plumbing, no renderable surface changes.

## Drive-by

Two **raw NUL bytes** in the daemon-id keys, committed since #2492, made git classify the
whole file as binary — `git diff` showed `Bin 4981 -> 6671 bytes` and the change could not
be reviewed. Replaced with `\0` escapes: identical string at runtime, file is text again.
(That's why the diffstat still shows `Bin` — the *old* side is the binary one.)


yschimke:
**This is now live, not theoretical — worth prioritising.**

meshcore-mobile adopted the theme variant axis in
[yschimke/meshcore-mobile#263](https://github.com/yschimke/meshcore-mobile/pull/263)
("pair split light/dark screen renders so night mode stays baked"). Its
`catalog.spec.json` on `main` now carries **five** theme-only variants:

    { "theme": "dark", "preview": "CachedDeviceBodyDarkPreview" }
    { "theme": "dark", "preview": "ContactChatDarkPreview" }
    { "theme": "dark", "preview": "ChannelChatDarkPreview" }
    { "theme": "dark", "preview": "CommandsDarkPreview" }
    { "theme": "dark", "preview": "DeviceSettingsDarkPreview" }

Traced against the unfixed resolver, for `Chat/Contact`:

1. the dark image has `state: "default"`, so it looks up `variantKey(id, "default")`
   → **`ContactChatPreview`**, the light function (the theme-only variant was never
   registered);
2. `daemonIdFor.get("ContactChatPreview\0dark")` misses — the light preview's id is
   `…ContactChatPreview_Contact chat`, which ends in `chat`, so `themeOfPreviewId`
   returns null and it went into `unthemedIdFor`, not `daemonIdFor`;
3. fall through to `unthemedIdFor.get("ContactChatPreview")` → the **light** id.

So all five dark stickers currently take the light preview's `previewId`. With #2653
merged, the next `design-artifacts.yml` run would also write each light vector at
`figma/<slug>/…__dark.svg`.

Recommend landing this before the next design-artifacts dispatch for that repo.

PR #2765, opened 2026-07-25
ci: reject agent authorship on every branch, not just PRs

yschimke:
## Why

`main` carries three `Co-authored-by: Claude <noreply@anthropic.com>` trailers (5aacb786, 0da37499, 2e5f89b2). Nobody typed them.

They came from the commit **identity**: cloud agent sessions run with `git user.name = Claude` / `user.email = noreply@anthropic.com`, and GitHub's squash credits every distinct branch commit author as a co-author — so the trailer lands on `main` even when the commits and the PR body are spotless. (This session started that way too; the fix is in the first commit of this branch.)

The existing gates ran at commit-message time and at PR time. Neither sees a commit that was amended, rebased, cherry-picked, imported from another clone, or made before the hooks were installed — `commit-msg` does not re-run for any of those.

## What changed

A `pre-push` hook that scans every commit a push would actually add, **on any branch**, plus all detection moved behind one shared script so the hooks and CI cannot drift apart:

| layer | file | catches |
| --- | --- | --- |
| commit | `.githooks/commit-msg` | the commit as it's written |
| push | `.githooks/pre-push` **(new)** | anything `commit-msg` never saw, any branch |
| PR | `no-agent-attribution.yml` → `scan` | the branch + PR title/body (the squash message) |
| post-merge | `no-agent-attribution.yml` → `drift` | breaches that got in anyway |

All four call `.github/scripts/agent-attribution-scan.sh`, which gained `--ident` (for the commit-msg hook, where the commit object doesn't exist yet) and `--rev-args` (for pre-push's `<sha> --not --remotes=origin`). The agent name/email lists are widened to match the sibling repos (`cursor`, `devin`, `aider`, `jules`).

**Rejected:** an agent author/committer identity (agent email domain, or an agent name set literally as the identity), and an agent `Co-authored-by:` trailer at trailer position.

**Left alone:** agent session links and the "🤖 Generated with …" / `_Generated by [Claude Code]_` footers — they don't claim authorship. A GitHub App bot *identity* (`claude[bot]`) stays exempt; a bot co-author *trailer* does not, since a trailer is an authorship claim wherever it came from.

## Test plan

New `scripts/test-agent-attribution.sh` — 30 assertions in both directions (what must be rejected **and** what must be left alone), driving the real detector and both hooks end to end against a scratch repo. Wired into CI as the `self-test` job so it runs on every PR.

```
$ scripts/test-agent-attribution.sh
...
30 passed, 0 failed
```

Covers, among others: `Claude <noreply@anthropic.com>` rejected; human named "Claude Martin" accepted; `claude[bot]` identity exempt; "Generated with Claude Code" + session links accepted; prose *mentioning* a trailer accepted; pre-push catching a `--no-verify`'d commit and an amended-in agent identity; a clean new branch not tripping over pre-existing history.

The detector also fails closed — an unresolvable revision exits 2, never 0. A gate that silently scans an empty set and reports "clean" is worse than no gate.

Non-visual change (git plumbing + CI), so no render evidence.

## Follow-up

This stops new attribution. It does not clean the three trailers already on `main` — that needs a history rewrite, which is being handled separately.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N1MRxt2KnF2z4GzPGG3hqY)_


yschimke:
Codex's P1 on `.githooks/pre-push:40` is correct and is fixed in 22484e98.

Exclusions for a new branch now come from the remote's currently advertised heads (`git ls-remote --heads`) instead of this clone's cached `refs/remotes/<remote>/*`. git has already contacted the remote to build the hook's stdin, so the extra round trip is against a remote known to be reachable; if it fails, the hook fails closed rather than guessing. An advertised head not present in the clone can't serve as an exclusion, so the scan widens rather than narrows, and prints a note pointing at `git fetch`.

Regression test added, built the way the report described it: push a dirty branch, cache its ref, delete it on the remote, then push a new branch carrying that commit. The old code accepted it; the new code rejects it.

Two further gaps found reviewing the same change in the sibling repos are fixed in the same commit, since all four repos vendor this detector:

- The shared matcher had folded the trailer vocabulary into the identity vocabulary, dropping providers the pre-existing gates rejected (Tabnine, Codeium, Amazon Q, CodeWhisperer, Cody, Cursor). Trailers now match a separate, deliberately wider `AGENT_TRAILER` list. Verified against every distinct `Co-authored-by:` trailer across all four repos' histories: it flags the agent ones and none of the ~60 human ones, nor `renovate[bot]` / `github-actions[bot]` / `dependabot[bot]`, which are automation rather than agents.
- `commit-msg` exempted `Merge` / `Revert` / `fixup!` / `squash!` / `amend!` subjects by returning early, which skipped the attribution scan along with the Conventional Commits check. A merge or a revert carries an identity and a trailer like any other commit, so the exemption now applies only to the Conventional Commits rule.

Suite is at 44 assertions, run by CI as `Detector self-test`.

---

PR #2766, opened 2026-07-25
fix(gradle-plugin): order composePreviewDaemonStart after composePreviewDiscover

yschimke:
## Summary

Fixes the implicit-dependency validation failure I found while testing #2761 and flagged there. Reproduced on unmodified `main` before attributing it, so it predates that change.

`DaemonBootstrapTask.previewsManifest` is an `@InputFile` pointing at `composePreviewDiscover`'s `previews.json`, but it's wired from a **layout directory** Provider:

```kotlin
previewsManifest.fileProvider(
  previewOutputDir.flatMap { dir ->
    project.providers.provider {
      val f = dir.file("previews.json").asFile
      if (f.isFile) f else null
    }
  }
)
```

That Provider carries no build dependency — nothing connects it to the task that writes the file — so Gradle's strict validation rejects the pair the moment both tasks are in one graph:

```
Task ':composePreviewDaemonStart' uses this output of task ':composePreviewDiscover'
without declaring an explicit or implicit dependency.
```

It's only reproducible with both tasks requested together, which is why it survived: the VS Code warm path and the render path each request one.

## `mustRunAfter`, not `dependsOn`

Gradle offers both as fixes and `dependsOn` is the reflex, but it's wrong here.

The daemon must stay warmable **before anything has been discovered** — that's the entire reason `previewsManifest` is `@Optional`, and `DaemonMain`'s `manifestFile.isFile` check is the other half of the same design. `dependsOn` would satisfy the validator and quietly make every editor warm pay for a full discovery pass, deleting the fresh-module path the optionality exists to serve.

Ordering-only is the relation that actually holds: discovery first when it's in the graph, nothing changed when it isn't.

Applied to both backends — the Android registration and the CMP/desktop one, which had identical wiring.

## Test plan

Both directions verified on `:samples:android`:

| | before | after |
| --- | --- | --- |
| `composePreviewRender composePreviewDaemonStart` together | **FAILED** — implicit dependency | **SUCCESS**, discover ordered ahead |
| `composePreviewDaemonStart` alone, `previews.json` deleted | success | **success in 10s, discover does NOT run**, descriptor still written |

The second row is the one that matters for the `mustRunAfter` choice — `dependsOn` would have dragged discovery in.

Two functional tests added to `DaemonBootstrapFunctionalTest` covering both. I confirmed the combined-invocation test **fails with the fix reverted**, so it guards the regression rather than passing vacuously.

`:gradle-plugin:test` and `:gradle-plugin:ktfmtCheck` green.

Non-visual — task-graph wiring only, no renderer or output changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ULa9AYFHcJgEQmKbzJPYtx)_

### DIFF
--- a/gradle-plugin/src/functionalTest/kotlin/ee/schimke/composeai/plugin/DaemonBootstrapFunctionalTest.kt
+++ b/gradle-plugin/src/functionalTest/kotlin/ee/schimke/composeai/plugin/DaemonBootstrapFunctionalTest.kt
@@ -42,6 +42,54 @@ class DaemonBootstrapFunctionalTest {
       .isIn(listOf(TaskOutcome.UP_TO_DATE, TaskOutcome.FROM_CACHE))
   }
 
+  @Test
+  fun `discover and daemonStart in one invocation pass validation, discover first`() {
+    // `previewsManifest` is an @InputFile pointing at composePreviewDiscover's `previews.json`, but
+    // it's wired from a layout directory Provider that carries no build dependency. Gradle's strict
+    // validation rejected the pair the moment both tasks were in one graph:
+    //
+    //   Task ':composePreviewDaemonStart' uses this output of task ':composePreviewDiscover'
+    //   without declaring an explicit or implicit dependency.
+    //
+    // Only reproducible with both tasks requested together, which is why it survived until a
+    // combined invocation happened to be run by hand.
+    val projectDir = createCmpTestProject()
+
+    val result =
+      GradleRunner.create()
+        .withProjectDir(projectDir)
+        .withArguments("composePreviewDiscover", "composePreviewDaemonStart", "--stacktrace")
+        .withPluginClasspath()
+        .build()
+
+    assertThat(result.task(":composePreviewDaemonStart")?.outcome).isEqualTo(TaskOutcome.SUCCESS)
+
+    val executed = result.tasks.map { it.path }
+    assertThat(executed).containsAtLeast(":composePreviewDiscover", ":composePreviewDaemonStart")
+    assertThat(executed.indexOf(":composePreviewDiscover"))
+      .isLessThan(executed.indexOf(":composePreviewDaemonStart"))
+  }
+
+  @Test
+  fun `daemonStart alone does not drag in discovery`() {
+    // Guards the choice of `mustRunAfter` over `dependsOn`. The daemon has to be warmable before
+    // anything has been discovered — that's why `previewsManifest` is @Optional (see its kdoc, and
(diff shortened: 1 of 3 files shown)


yschimke:
The preview-diff comment above contradicts this PR's "no renderer or output changes" claim. It's a stale baseline, not this branch — and the root cause is now fixed in #2767.

**The 19 New + 19 Removed in `samples:design-catalog-remote-m3` are not from this branch.** #2765 — a pure `.githooks` / CI change — reports the *identical* list against the *identical* base render commit `7c601964`. Same 19 names in both the New and Removed lists, the "New" ones carrying `_dpi_320` suffixes. That's #2763 (`declare Wear render density (2.0)`), which changes the render key and therefore the filename.

#2763 is on `main` and is an ancestor of this branch, so it should have been on both sides. It wasn't, because **its baseline run was cancelled 48 seconds after it started**:

```
30167490283 2640f88b9 push cancelled   #2762  (17:22:16)
30167461027 16a64b33a push cancelled   #2763  (17:21:28)   <- density change
30167357457 d9f5e0f5f push cancelled   #2761  (17:18:18)
30165267442 f5b5c0f47 push success     #2759  (16:15:37)   <- 7c601964, the base above
```

`compose-preview.yml` had `cancel-in-progress: true` on a group that push-to-main shares, so each merge killed the previous merge's baseline refresh. 24 of the last 30 main runs were cancelled; the baseline sat four commits behind.

**The lottie APNG line is the same artifact**, and it finally explains the case I flagged as reasoning-not-proof on #2761: `2640f88b9` is #2762, *"settle each Lottie APNG step so the sweep stops dropping frames"* — a real pixel change, on `main`, whose baseline run was also cancelled. So the "Before" image here is pre-#2762 and the "After" is #2762's fix. Not nondeterminism, and not this branch.

Nothing to change on #2766 — the body's claim holds. Verification stands as described: the combined invocation fails without the `mustRunAfter` and passes with it, and `composePreviewDaemonStart` alone still doesn't drag in discovery.

---

yschimke:
/rerun previews

The stale-baseline root cause is fixed and merged (#2767), and main's baseline run for `22ee83076` completed **successfully** at 18:16:48 — so the published "before" state is now current with `main` instead of sitting four commits behind at `7c601964`.

Re-running previews here to confirm the diagnosis rather than just assert it: the 19 New + 19 Removed `design-catalog-remote-m3` variants and the `lottie/spin.json` change should both disappear from the sticky comment above, leaving this PR with no preview changes — which is what "task-graph wiring only" predicts.

---

PR #2820, opened 2026-07-27
build(deps): update AGP to 9.3.0

yschimke:
Split out of the grouped Renovate bump (#2805). AGP drives every module's compile and variant wiring, so a regression here should not be ambiguous with a library update landing in the same commit.

`com.android.tools.build:gradle`, `com.android.library`, `com.android.application`: `9.2.1` → `9.3.0`.

## Test plan

Locally on JDK 21:

- `./gradlew publishToMavenLocal :cli:installDist` — the integration job's build step, green.
- `./gradlew :samples:android:composePreviewDiscover :samples:cmp:composePreviewDiscover :samples:android-screenshot-test:composePreviewRenderAll` — preview discovery + full render, green.
- `./gradlew :renderer-android:testDebugUnitTest` — green, including the `WearScrollSvgGrowthTest` capsule-fixture drift guard, so the AGP bump doesn't move rendered vector output.

Non-visual: no renderer, catalog or fixture source changes, and the render suite above reproduces the committed output.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V9QFVfWmNCNEWVXAf9gdPs)_

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,7 +14,7 @@ kotlin = "2.4.0"
 # `configureKotlinCompatibility(...)`; remove the flag (and bump this to
 # `2.1.x`) when the removal lands and we're ready to drop 2.0.x consumers.
 kotlinCoreLibraries = "2.0.21"
-agp = "9.2.1"
+agp = "9.3.0"
 ktfmt = "0.26.0"
 tapmoc = "0.4.2"
 robolectric = "4.17-beta-2"


yschimke:
CI blocker, not caused by this diff.

`Apply compose-preview` fails here with:

```
> composePreviewRenderAll: render produced no output file for 1 of 15 preview(s).
One or more compose-preview pipelines exited non-zero (rc=2)
```

The identical failure landed on all three open dependency PRs within minutes of each other — this one (AGP only), #2821 (roborazzi only) and #2805 (library bumps only). Those diffs are disjoint and every other check on all three is green (CI, Integration, Samples-SDK21, CodeQL, Daemon Harness, dependency-review). What the three share is the base, `main` @ 3b72252, plus the runner environment — so the render regression is upstream of the bumps.

Two candidates under investigation:

1. a preview that stopped rendering under the XR beta01 line that landed in #2819, which would reproduce on `main` untouched;
2. the `xr-composite` provisioning 404 in the same job — it fetches `xr-composite-linux-x86_64-0.18.4-SNAPSHOT.tar.gz`, gets HTTP 404 (latest published release is v0.18.3), and skips the composite stills. With `MISSING_RENDERS=fail`, a skipped still is exactly a "produced no output file" failure.

A full-workspace render is running locally to identify the module and preview. Holding this PR rather than papering over a `main` regression inside a dependency bump; will update here with the cause.

---

yschimke:
is this safe for apps using AGP 9.2.1?

yschimke:
> is this safe for apps using AGP 9.2.1?

Yes — consumers keep their own AGP; this bump doesn't reach them.

- **The plugin takes AGP as `compileOnly`** ([`gradle-plugin/build.gradle.kts:63`](https://github.com/yschimke/compose-ai-tools/blob/agent/agp-9.3.0/gradle-plugin/build.gradle.kts#L63)): `compileOnly("com.android.tools.build:gradle:${libs.versions.agp.get()}")`. It's a compile-time-only classpath entry, so it doesn't appear in the published POM and can't pull AGP 9.3.0 onto a consumer's buildscript classpath. Whatever AGP the consumer applies is the one the plugin binds against at runtime.
- **This diff is one version string** in `gradle/libs.versions.toml` (`agp = "9.2.1"` → `"9.3.0"`), which drives this repo's own modules and samples plus the AGP the plugin compiles against. No plugin source changed, so no new-in-9.3.0 API is referenced — the risk shape for a `compileOnly` bump is "plugin code starts calling an API that only exists in the newer AGP", and there's no code change here to introduce one.
- **The floor is covered by CI, not by assumption**: the `agp8-min` job in `integration.yml` builds the in-repo fixture at `.github/ci/fixtures/agp8-min/` — Gradle 8.13 + **AGP 8.13.0** + Kotlin 2.0.21 + JDK 17 — against the plugin built from this branch, and applies it through the same init-script auto-inject path a real consumer uses. It runs on PRs whenever the change-scope gate says the diff can affect discovery/rendering, which a `libs.versions.toml` bump does. AGP 8.13 passing is a stronger statement than 9.2.1 passing, since 9.2.1 sits between it and 9.3.0.

The one thing this PR does move for consumers who build *from source* or use the samples as a reference is the toolchain those samples expect; published-artifact consumers on 9.2.1 are unaffected.

Worth noting separately: the `Apply compose-preview` failure on this PR is still the pre-existing one described above — `samples:xr-spatial` (its 15 previews are the "1 of 15" in the message), not AGP. Diagnosis continues; nothing about it is specific to this diff.

---

PR #2821, opened 2026-07-27
build(deps): update roborazzi to 1.70.0

yschimke:
Split out of the grouped Renovate bump (#2805). Roborazzi owns the capture path behind every rendered PNG, so it lands alone where a pixel change would be unambiguous.

`io.github.takahirom.roborazzi:{roborazzi,roborazzi-compose,roborazzi-annotations,roborazzi-accessibility-check}`: `1.68.0` → `1.70.0`.

What's in the two releases that touches us:

- **1.69.0** rewrote the `Bitmap` → `BufferedImage` conversion to write straight into the backing raster instead of a per-pixel `setRGB` pass. Upstream reports byte-identical recorded output, verified against their full sample baseline.
- **1.70.0** republished `roborazzi-annotations` as a Kotlin Multiplatform library — same Maven coordinate, same signatures, packaging only.

The new opt-in features (Compose Desktop preview generation, UI tree dump) are off by default and not enabled here.

## Test plan

Locally on JDK 21:

- `./gradlew :samples:android-screenshot-test:composePreviewRenderAll` — full render suite, green.
- `./gradlew :samples:wear:testDebugUnitTest` — Wear pixel tests against the committed goldens, green.
- `./gradlew :renderer-android:testDebugUnitTest` — green, including the `WearScrollSvgGrowthTest` capsule-fixture drift guard.

No golden or fixture updates were needed, which is the evidence that capture output is unchanged; the CI visual-diff bot covers the rendered PNGs on this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V9QFVfWmNCNEWVXAf9gdPs)_

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -18,7 +18,7 @@ agp = "9.2.1"
 ktfmt = "0.26.0"
 tapmoc = "0.4.2"
 robolectric = "4.17-beta-2"
-roborazzi = "1.68.0"
+roborazzi = "1.70.0"
 # androidx.tracing — atrace-level sections for the Android daemon's render phases (see
 # `AndroidxTraceSections` in :daemon:android). Version-pinned (not BOM-managed) because the daemon
 # ships it in the standalone lib-daemon-android sidecar where no BOM applies.


yschimke:
CI blocker, not caused by this diff.

`Apply compose-preview` fails here with:

```
> composePreviewRenderAll: render produced no output file for 1 of 15 preview(s).
One or more compose-preview pipelines exited non-zero (rc=2)
```

The identical failure landed on all three open dependency PRs within minutes of each other — #2820 (AGP only), #2805 (library bumps only) and this one (roborazzi only). Those diffs are disjoint and every other check on all three is green (CI, Integration, Samples-SDK21, CodeQL, Daemon Harness, dependency-review). What the three share is the base, `main` @ 3b72252, plus the runner environment — so the render regression is upstream of the bumps.

Two candidates under investigation:

1. a preview that stopped rendering under the XR beta01 line that landed in #2819, which would reproduce on `main` untouched;
2. the `xr-composite` provisioning 404 in the same job — it fetches `xr-composite-linux-x86_64-0.18.4-SNAPSHOT.tar.gz`, gets HTTP 404 (latest published release is v0.18.3), and skips the composite stills. With `MISSING_RENDERS=fail`, a skipped still is exactly a "produced no output file" failure.

A full-workspace render is running locally to identify the module and preview. Holding this PR rather than papering over a `main` regression inside a dependency bump; will update here with the cause.

---

PR #2893, opened 2026-07-28
feat(serve): share generated documents via expiring permalinks

### LINKED ISSUE #2873: add upload mechanism for  known document formats
A way to render one of the supported doc formats

- RC doc
- Lottie

Maybe others

Not sure if it's a dedicated upload, or providing a url in a SPA?

yschimke:
Fixes #2873.

## What

A **document lane** for `compose-preview serve`, behind `--accept-docs`. The issue asked for a way to take a generated document — a Remote Compose `.rc`, a Lottie — "load [it] and get an expiring permalink to give someone else". Both shapes it wondered about are here: a **dedicated upload page** and a **URL you can hand the server**.

- `GET /docs` — drop zone (drag/drop, file picker, or a URL field when the host allows fetches).
- `POST /docs?name=<label>` — the same for a script/agent; the document is the body. `201` with `{"id","name","format","formatId","bytes","url","expiresIn","expiresAtEpochSeconds"}`.
- `GET /d/<id>` — the expiring permalink: the document played back **client-side**, plus what the server read out of it and the time left. `GET /d/<id>/raw` is the bytes.
- `GET /doc-player/<format>/bundle.js` — that format's vendored browser player.

```console
$ curl -sS --data-binary @pulse-dot.json 'http://127.0.0.1:8799/docs?name=pulse-dot.json'
{"schema":"compose-preview-serve/doc/v1","id":"q2k-e9HuvI6b45jJJTvUag","name":"pulse-dot.json",
 "format":"Lottie","formatId":"lottie","bytes":894,"url":"/d/q2k-e9HuvI6b45jJJTvUag",
 "expiresIn":"1h","expiresAtEpochSeconds":1785279488}
```

## Design

`ServeDocFormats` is the registry — per-format content sniff, summary facts, declared size, player bundle. Routes only ever look a format up by id, so adding a format is one entry plus its player, not a new branch in the server. Today: Remote Compose (`Header` op magic `0x048C`, played by the RC player the preview viewer's canvas lane already uses) and Lottie (Bodymovin shape check, played by a newly vendored MIT `lottie-web` 5.13.0 build).

`ServeDocStore` is the ephemeral side: in-memory, TTL'd (`--doc-ttl`, default 1 h), 128-bit `SecureRandom` ids, per-document (8 MB) / count (64) / total (64 MB) caps with closest-to-expiry eviction, and the same fail-closed SSRF gate the bundle store uses (`--accept-docs-from`).

Safety, since this is meant to be on a public box:

- **Data only.** The host stores and serves bytes; the player runs in the *viewer's* browser. Nothing about a document executes server-side.
- **Sniffed, not name-trusted.** An upload must *be* a known document — a zip, a script, an HTML page, or an `.rc`-named impostor is refused, so the lane can't become a general file drop or serve attacker-chosen HTML from the host origin. `?name=` is a display label only.
- **The link is the capability, and it dies.** Page is `Cache-Control: private, no-store`; after the TTL both `/d/<id>` and `/d/<id>/raw` 404 without revealing whether the id existed. Raw bytes go out with `X-Content-Type-Options: nosniff`.

Also wired: `--accept-docs` / `--doc-ttl` / `--accept-docs-from` in usage + `CliFlags`, `SERVE_ACCEPT_DOCS` / `SERVE_DOC_TTL` / `SERVE_ACCEPT_DOCS_FROM` in the deploy image entrypoint, `/status` + `status.json` config rows, and a docs section in `docs/public-preview-server.md`.

## Visual evidence

Both pages are new surfaces, so they're registered as preview-harness page fixtures (`serve-docs-upload`, `serve-doc-lottie`, `serve-doc-remotecompose`) — the CI visual-diff bot picks them up from here on with no extra wiring.

The first two shots are from a **live server** (`compose-preview serve --public --accept-docs`) with a real Lottie uploaded over `POST /docs`, not a mock:

**`GET /docs` — the drop zone (dark)**

The /docs drop zone

**`GET /d/<id>` — a shared Lottie playing at its permalink (light)**

A shared Lottie playing at its expiring permalink

**`GET /d/<id>` — the Remote Compose page (harness fixture; the stage is empty because the harness serves no player bundle)**

The Remote Compose document page

Not verified in-session: RC **playback** with a real `.rc` document — the repo carries none and this container can't fetch one. The page, the player URL, the canvas sizing from the document header, and the raw-byte lane are covered by tests; the player itself is the same bundle the preview viewer's proven canvas lane loads.

Shared-CSS side effect worth a look: `.cp-stat` tiles had no dark-mode rule, so they stayed white-on-dark. Fixed here because the document facts reuse them — which also fixes the `/status` page in dark mode (visible in the `serve-status` fixture diff).

## Test plan

- [x] `./gradlew :cli:test` — 906 tests green, including the new `ServeDocFormatsTest` (sniffing + rejection of zip/PNG/HTML/empty), `ServeDocStoreTest` (TTL, caps + eviction, label sanitisation, SSRF fail-closed, id shape), and `ServeDocRoutingTest` (real embedded server: upload → permalink → raw → expiry; player routes; routes absent without the opt-in).
- [x] `ServeWebFixtureTest` goldens regenerated; three new page fixtures committed.
- [x] `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest`.
- [x] End-to-end by hand: `serve --public --accept-docs` → `POST /docs` a Lottie → opened `/d/<id>` in headless Chromium and confirmed the animation renders (screenshot above).

https://claude.ai/code/session_01TWEZw81eSypPaixcf1RKLH

---
_Generated by [Claude Code](https://claude.ai/code/session_01TWEZw81eSypPaixcf1RKLH)_


yschimke:
Pushed `959db62` for the two code findings; the third is a false positive.

**Redirect SSRF (P1) — fixed.** Real hole: the allowlist was applied to the client-supplied URL, then OkHttp followed redirects on its own, so an allowlisted host could bounce the fetch onto a link-local/internal address. The client no longer follows redirects (`followRedirects(false)` / `followSslRedirects(false)`); `ServeDocStore.followingRedirects` walks them instead, re-checking `--accept-docs-from` **before every request** and stopping after four hops. It's pure over its transport, so the hop policy is unit-tested with no network: a `302` to `169.254.169.254` returns null with that address never sent to, an on-allowlist redirect is followed, and a redirect loop terminates.

**Token on the permalink (P1) — fixed.** Correct: `POST /docs` answers with the bare `/d/<id>` (right for a script, which already holds the token), but the upload page was linking that path verbatim, so on a token-gated host the first click 404s. The page-facing link and anchor now carry the page's own query suffix — empty in `--public` mode, `?token=…` otherwise — with a test pinning both modes.

**Commit attribution (P1) — not reproducible.** The commit's identity is the human's, not an agent's:

```
$ git show --format=fuller --no-patch fdb09dc
Author:     Yuri Schimke <yuri@schimke.ee>
Commit:     Yuri Schimke <yuri@schimke.ee>
```

No `codex@openai.com` appears anywhere in the branch, and this repo's own detector — the `Reject agent attribution` gate that scans the PR title, body, and every commit's trailers plus author/committer identity — passed on that same commit. Nothing to reattribute.

One thing this review didn't flag, worth recording since it's the same bug class: `ServeBundleStore.addFromUrl` (`--accept-bundles-from`) has the identical redirect gap and is untouched here — pre-existing, so it's left for a follow-up rather than widened into this PR.

---

PR #2895, opened 2026-07-28
fix(wear-widget): paint the widget fill as the container background, not inside content

yschimke:
## Summary

`:samples:wear-widget` rendered a square-cornered blue rectangle floating inside the host's rounded squircle, while the `WidgetContainer*Remote` stickers next door looked right.

`RemoteImageWidget` painted its own `fillMaxSize().background(gradient)` into the container's **padded content slot**, so the fill could neither reach the corners (the padding held it off) nor be clipped to the corner radius (the container clips its own background, not arbitrary content draws). The fill now goes to the document's `WearWidgetBrush` background, which is what `WearWidgetContainer` paints as the container surface — the same split upstream [`wear-os-samples/WearWidget`](https://github.com/android/wear-os-samples/tree/main/WearWidget) uses (`WearWidgetDocument(background = WearWidgetBrush.color(...))` with transparent content), and what the already-correct `WidgetContainer*Remote` stickers do.

## Visual evidence

`ImageWidgetSquirclePreview` (small squircle footprint):

| Before | After |
|---|---|
| before | after |

`ImageWidgetSquircleLargePreview` (large squircle footprint):

| Before | After |
|---|---|
| before | after |

The dark ring in "before" is the host container's default `surfaceContainerLow` surface showing through the 8dp padding the inner rectangle couldn't cross. "After" is a single corner-clipped gradient squircle.

## Every widget preview now emits its `.rc` IR

Second half of the change: make the encoded RemoteCompose document — which *is* a Wear widget's value — reach the bundle as data rather than compiled `@Preview` bytecode.

- **Promoted `CapturingWearWidgetPreview`** from `:samples:wear-widget` into `:wear-preview-runtime` so the widget-container stickers can share it. Those stickers went through upstream `WearWidgetPreview`, which builds the document internally and keeps the bytes — they rendered fine while silently riding the bundle as bytecode. They now emit `.rc` like every other sticker in the sheet (3 previously emitted **none**).
- **Coroutines is a real dependency, not `compileOnly`.** The alpha Compose/glance-wear graph resolves 1.9.0 while the module compiles against the repo's pinned 1.11.0, and `runBlocking` is not ABI-compatible across that gap (1.11 emits `BuildersKt.runBlockingK$default`, absent in 1.9.0). The resulting `NoSuchMethodError` degraded the capture to "no `.rc`" with a green build.
- **A failed capture now reports on stderr** instead of being swallowed by `runCatching`, so a missing sidecar is diagnosable from the render log rather than invisible. That silent-failure path is exactly how the skew above went unnoticed.
- **Added `WidgetContainerIrCaptureTest`**, the catalog's counterpart to `WearWidgetDocCaptureTest`. It asserts on sidecars rather than pixels on purpose — the failure mode is invisible in the PNG.

## Test plan

- `:samples:wear-widget:testDebugUnitTest` — 3 tests, green (including the `.rc` sidecar assertion, which now passes against real sidecars).
- `:samples:design-catalog-remote-m3:testDebugUnitTest` — 2 tests, green (new).
- `composePreviewRenderAll` on both modules: a `.rc` beside every widget PNG — 7 for `wear-widget`, 3 for the catalog (previously 0).
- `ktfmtCheck` / `ktfmtCheckScripts` clean on all three touched modules.
- Committed renders refreshed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L6SvmYX83WejJUsHYoCHM1)_


yschimke:
Two entries in the visual-diff comment above are **not** the widget fix — worth explaining so they don't read as collateral damage.

**`ScalingListSticker` (samples:design-catalog-wear-m3) — caused by this PR, intentional.**

The list, scaling and layout are pixel-identical; the only delta is that `TimeText` ("10:10") now renders at the top.

Cause is the coroutines change in `:wear-preview-runtime`. Moving `kotlinx-coroutines-core` from `compileOnly` to `implementation` (needed so `runBlocking`'s call site matches at render time — see the PR body) puts the repo's pinned 1.11.0 on every consumer's runtime classpath. `:samples:design-catalog-wear-m3` consumes this module for `TlcScalingHost`, so it moved 1.9.0 → 1.11.0:

```
org.jetbrains.kotlinx:kotlinx-coroutines-core:1.9.0 -> 1.11.0
```

`TimeText` sources the time from a flow, and on 1.11.0 it emits in time to be captured. So the new baseline is the *more* correct one — a Wear scaffold should show the time — and 1.11.0 is the version this repo already declares in `libs.versions.toml`.

Keeping the dependency at the source (rather than declaring coroutines in each consumer) is deliberate: a consumer that forgets it gets a `NoSuchMethodError` swallowed into "renders fine, emits no `.rc`", which is exactly the silent failure this PR is fixing.

**`lottie/spin.json` (samples:android) — unrelated.**

`:samples:android` has no path to `:wear-preview-runtime`, and the first frame is identical between before/after. This is the usual animated-preview frame-timing nondeterminism, not something this PR touches.

---

PR #2897, opened 2026-07-28
feat(serve): make the published catalog set pure config

### LINKED ISSUE #2879: Make preview server catalogs pure config
possibly even an admin url to register catalogs, save to a config DB outside the docker image?

Rather than baking in.

We should also remove any logic that special cases the existing ones.

yschimke:
Fixes #2879

## What

The catalog set lived in three places that all had to agree: a comma-separated `--catalogs` list baked into the container entrypoint, and a pair of hardcoded id/repo sets in `ServeWeb.homeSections` that decided a catalog's front-page section **by name**. Publishing a catalog meant editing the image and shipping a CLI release, and a catalog this build had never heard of could only ever land in "Other".

It's data now.

**1. `catalogs.json` — the catalog set as operator config** (`--catalogs-file`, `SERVE_CATALOGS_FILE`):

```json
{
  "groups": [
    { "id": "design-systems", "heading": "Design Systems", "noun": "design system(s)" }
  ],
  "catalogs": [
    { "system": "compose-m3", "repo": "yschimke/compose-ai-tools", "group": "design-systems" },
    { "system": "jetnews", "repo": "yschimke/compose-samples", "group": "android-samples",
      "attributionRepos": ["android/compose-samples"] },
    { "system": "cadence", "repo": "yschimke/cadence", "listed": false }
  ]
}
```

It lives on the mounted `/config` volume — **outside the image**. The image carries the standard set only as a first-boot seed (`deploy/image/catalogs.json` → `/etc/compose-preview/catalogs.default.json`), copied in when the config file is absent and never again, so an operator edit survives every pull while a bare `docker run` on an empty volume still comes up serving the standard catalogs. `SERVE_CATALOGS` / `SERVE_CATALOGS_UNLISTED` remain as *additions* for a one-off extra catalog.

**2. No per-catalog special cases left.** `homeSections` knows no catalog id at all. A declared group is a **claim checked against provenance** — the bytes must have come from the entry's `repo` or one of its `attributionRepos` (which is how Android's samples, fetched from a fork's preview branches, stay credited to `android/compose-samples`), so serving a third-party catalog as `compose-m3` still can't buy it the Design Systems section. An ungrouped catalog is sectioned by its source repo's **owner**; unattributed ones land in `Other`, pinned last. Headings are HTML-escaped now that they're operator input.

**3. `/admin/catalogs` — publish and retire on a running server**, gated by `--admin-token` (`SERVE_ADMIN_TOKEN`), which is deliberately **not** the browse token a `--public` box hands to every visitor; unset means the routes aren't registered at all.

| Route | Does |
|---|---|
| `GET /admin/catalogs` | configured catalogs + each one's load state |
| `POST /admin/catalogs` | publish one (body = a `catalogs.json` entry) |
| `DELETE /admin/catalogs/<system>` | retire one — session and live daemon dropped |

Every mutation is applied live **and** written back to `catalogs.json`, under a lock that serializes the whole load/mutate/save so concurrent admin calls can't lose each other's writes. The fetch happens *before* the persist, so a typo'd repo `502`s instead of leaving an unservable entry for every future boot to retry; malformed entries `400`, re-publishing a served catalog `409`, and an unwritable config file still serves but returns a `warning`.

To make runtime registration visible everywhere at once, `CatalogLoadTracker` became mutable and is now the live source of truth the home index, `/status`, the branch refresher, and the Wasm registry read per request instead of capturing at startup. Admin mode also builds that machinery when the configured set is *empty*, so a box pointed at a fresh `catalogs.json` can publish its first catalog.

## Visual evidence

Front page, rendered from the committed `serve-home-index` page fixture. The only pixel change: Confetti — which declares no group — is now sectioned by its publisher (`joreilly org`) instead of being dumped in `Other`. The configured `Design Systems` section is unchanged, but it now comes from `catalogs.json` rather than a hardcoded id set.

**Before**

**After**

The surface stays wired into the preview workflow: it's the same committed `vscode-extension/preview-harness/fixtures/pages/serve-home-index.html` fixture the CI visual-diff bot already renders, regenerated by `ServeWebFixtureTest`, so future changes to it are diffed automatically.

## Test plan

- [x] `./gradlew :cli:test` — green, including new `ServeCatalogsConfigTest` (parse / validate / atomic round-trip on a `FakeFileSystem`), `ServeCatalogAdminTest` (register, rollback on fetch failure, conflict, group claim scoping, persistence warning, concurrent-write safety, tracker add/remove), and `ServeAdminRoutingTest` (the routes over real HTTP: admin-token gate on a `--public` server, publish → served + indexed immediately, `400`/`409`/`502`, delete, runtime Wasm registration)
- [x] `ServeWebThumbCropTest` covers the config-driven grouping: spoofed ids fall back to their real publisher, owner-derived sections for unconfigured publishers, `Other` last, and heading escaping
- [x] `./gradlew ktfmtCheckAll` — green
- [x] `deploy/image/test-env-migrations.sh` — 11 passed
- [x] Page fixtures regenerated (`UPDATE_SERVE_WEB_FIXTURES=true`) and rendered above
- [x] Merged `main` (the `--accept-docs` document lane) and re-ran both gates green
- [ ] Not exercised here: an actual container boot (the seed-copy path in `entrypoint.sh`) — no Docker in this environment

https://claude.ai/code/session_01FdJsCmxDm9dXv8iEsxVYgs


yschimke:
All three review findings were real and are fixed in `b0d324c`.

**P1 — serialize config mutations.** Correct: `persist` did an unsynchronized load → mutate → save, so two concurrent admin requests could each read the same document and the second `atomicMove` would drop the first, with both callers told `ok`. Atomicity of the individual save never helped — the lost update happens *between* the load and the save. The whole sequence now holds a lock. The group table went with it: it was being `clear()`ed and refilled, which a concurrent `register` could observe half-empty and reject a group that does exist, so it's now a `@Volatile` list swapped wholesale and each request takes one snapshot for both its validation and its resolution. Covered by a new test that fires 8 concurrent registrations through a deliberately slow load callback and asserts all 8 survive in the file.

**P2 — admin mode couldn't bootstrap an empty catalog set.** Also correct, and the worse of the two P2s: the store and tracker were only built when the startup catalog set was non-empty, so a box pointed at a fresh `catalogs.json` had the admin routes suppressed and could never publish its first catalog — and module-less mode additionally exited with "nothing to serve". The machinery is now built whenever `--admin-token` is set, the module-less bail treats admin mode like `--accept-bundles`, and the branch refresher is built too so a runtime-added catalog starts being polled without a restart.

**P2 — Wasm registry not updated for runtime mutations.** Correct. `ServeHttpServer` held a boot-time `catalogWasm + localWasm` snapshot, so an admin-published catalog carrying `webRender` got no `/wasm/<system>/` lane and a retired one left its assets served. The registration now owns a live `ConcurrentHashMap` that the server reads through; `unload` removes the system's entry, `--wasm-dir` overrides still win over a published app (including across a branch refresh, which re-runs `registerWasm`), and the route is registered when admin mode is on even if the set starts empty. Covered by a routing test that registers an app after boot, fetches it, then unregisters and asserts the 404.

`./gradlew :cli:test ktfmtCheckAll` green.

---

yschimke:
Heads-up on the regenerated page fixtures here — this branch is based on `7e582cc`, which predates both #2891 and #2893, and both of those edited the **shared stylesheet** every serve page embeds:

- #2891 added the `.cp-theme` wrap + `.cp-theme-btn + .cp-theme-btn` divider rules;
- #2893 added the `.cp-drop` / `.cp-doc-*` rules.

Any fixture regenerated on this base (`serve-home-index.html` at minimum) is missing both blocks, so merging as-is turns `ServeWebFixtureTest` red on `main` with `… is stale vs ServeWeb`. That already happened once: #2891 and #2893 crossed and each left the other's file behind, which took four stale goldens and a fix in #2903 (`9c7f78b`) to clear.

Fix is one command **after** rebasing onto current `main`:

```
```

Worth re-running it as the last step before merge rather than at the point the fixtures were first generated, since anything landing on `main` in between can move the shared stylesheet again.

---

yschimke:
fix merge conflicts

yschimke:
## Visual evidence — front-page index

(The PR body strips image URLs when written through the API, so the before/after lives here instead.)

Rendered from the committed `serve-home-index` page fixture. The only pixel change is the last section heading: Confetti declares no group in `catalogs.json`, so it's now attributed to its publisher rather than dumped in the unattributed bucket. `Design Systems` looks identical but is now produced by config instead of a hardcoded id set.

**Before** — `Other`

home index before``

**After** — `joreilly org`

home index after``

Both PNGs are committed under `docs/evidence/serve-catalogs-config/`, and the fixture itself is the one the CI visual-diff bot renders on every PR (its comment above shows the same 2-of-82 change), so this surface stays diffed automatically.

---

yschimke:
Merge conflicts fixed and the stale-fixture risk checked — `893a90c` merges current `main` (`95a5f74`) into the branch.

**Conflicts** (three, all mechanical — this branch and the `--accept-docs` work touched adjacent lines):
- `ServeCommand.kt` imports — both sides added to the same block; kept both.
- The module-less "nothing to serve" bail — both sides added a condition. Combined rather than picked: it now stands down for `--accept-bundles`, `--accept-docs`, **and** `--admin-token`, and the error message lists all three.
- `ServeHttpServer` constructor + `docs/public-preview-server.md` — both sides appended a parameter / rewrote the same deploy paragraph; kept both parameters, and folded the `SERVE_ACCEPT_DOCS` / `SERVE_DOC_TTL` / `SERVE_ACCEPT_DOCS_FROM` env list into the new catalog-config section (the old prose it conflicted with was the baked-catalog-list paragraph this PR deletes).

**On the fixtures** — you were right to flag it, and the merge turns out to have already fixed it rather than needing a separate regeneration. I re-ran the command on the merged tree as the last step:

```
```

It produced **zero diff**, so the committed goldens already match post-#2891/#2893 `ServeWeb`. Spot-checking the two stylesheet blocks across all 26 serve page fixtures: every one carries `.cp-theme-btn + .cp-theme-btn` (2 occurrences) and `.cp-drop` (7, or 12 for `serve-docs-upload`). `serve-home-index.html` auto-merged cleanly — main's stylesheet edits and this PR's one-line section-heading change touch different parts of the file.

Worth noting the failure mode you described is one this branch can still walk into if something else lands on `main` before merge, so I'll re-run that command on any further sync rather than trusting this pass.

`./gradlew :cli:test ktfmtCheckAll` green on the merge commit.

---

yschimke:
Two bot reports landed on the merge commit. One is explained; the other I can rule out as this PR's doing but **can't** explain, so flagging it rather than waving it through.

**Accessibility report (38 errors / 259 previews, was 36 / 256)** — the delta is main's, not mine. The three extra previews are `WatchScreenRemote` (from #2892, "front the compose-m3 and remote-m3 catalogs with screen heroes") and `PlaceholderCardLoading` / `PlaceholderCardLoaded`, which arrived with the merge. `WatchScreenRemote`'s error is the standard `SpeakableTextPresentCheck` on `RemoteComposePlayer` that every Remote Compose preview reports. This PR adds no `@Preview` at all.

**Notification previews (4 changed)** — `BigTextVariantsPreview_{Arabic,German,Japanese,Large font}` differ from the `compose-preview/notifications/main` baseline. Not from this branch, and I don't think it's main either:

- My diff against the merge base touches only `cli/src/**` (serve) and `trust/producers.json`, plus docs/deploy/fixtures. No sample, renderer, daemon, or font code.
- The baseline branch head says *"Update notification baseline from 9c7f78b3"*, and `9c7f78b..origin/main` contains **exactly one commit** — #2904, `cli/src` only. So between the baseline's source and this PR's HEAD there is no renderer, `daemon/`, `data/`, or `samples/android/` change from any source.

With no render-affecting code difference on either side, the four that moved being precisely the text-heavy locale + font-scale variants points at **non-determinism in font resolution** for those renders (a downloadable/Google-font fetch resolving differently between the baseline run and this one) rather than a real regression. That's a hypothesis, not a conclusion — I haven't reproduced it, and the Android render is out of reach in this session's environment.

Nothing to fix on this PR either way. Worth a look independently if those four keep flip-flopping across unrelated PRs, since a baseline that drifts on its own quietly costs every future PR a false positive.

---

yschimke:
fix merge conflicts

PR #2904, opened 2026-07-28
fix(serve): validate every redirect hop of a client-supplied bundle URL

yschimke:
Follow-up to #2893, which fixed this for the document lane and left the bundle one noted but untouched.

## The hole

`POST /bundles/{name}?url=` checked the client's URL against `--accept-bundles-from` and then handed it to an OkHttp client with default redirect handling. The allowlist saw the *first* address and nothing after it, so an allowlisted host answering

```
302 Location: http://169.254.169.254/latest/meta-data
```

walked the server onto cloud metadata (or any internal service), with the operator's allowlist consulted exactly once and never again. Reachable on any box running `--accept-bundles --accept-bundles-from <host>`.

## The fix

Both client-supplied-URL lanes now go through one `ServeUrlFetch`:

- redirects are **off** at the HTTP client (`followRedirects(false)` / `followSslRedirects(false)`);
- `followingRedirects` walks them itself, re-checking the allowlist **before every request**, and stops after `MAX_REDIRECTS` (4), which also terminates a redirect loop;
- `isAllowedUrl` now has a single implementation instead of a copy in each store, so the two lanes can't drift on what "allowed" means.

`followingRedirects` is pure over its transport, so the whole hop policy is unit-tested with no network: an off-allowlist redirect returns null **and** the test asserts the internal address was never sent to; an on-allowlist redirect is followed; a loop terminates at the ceiling; a transport failure ends the walk. `ServeBundleStoreTest` additionally pins that a look-alike (`ci.example.com.evil.test`) and a subdomain (`sub.ci.example.com`) of an allowlisted host are refused without a fetch.

Operator-supplied URLs (`--bundle <url>`, `--catalogs`) are deliberately untouched — those addresses come from the command line, so they were never allowlist-gated in the first place.

## Drive-by: four stale goldens

`ServeWebFixtureTest` **fails on a clean `main`** — the shared-CSS change in #2892 (theme-picker chips wrapping) never had its fixtures regenerated, and three document pages inherit the same stylesheet. Regenerated here so this branch can go green; the only diff is that CSS block. Verified pre-existing by stashing my changes and re-running on `376a983`.

## Test plan

- [x] `./gradlew :cli:test` — 913 green, including the new `ServeUrlFetchTest` and the bundle-store allowlist cases.
- [x] `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest`.
- [x] Confirmed the fixture failure reproduces on `main` without this branch's changes.

No visual change: the only rendered-output diff is #2892's already-merged CSS, now reflected in the goldens.

https://claude.ai/code/session_01TWEZw81eSypPaixcf1RKLH

---
_Generated by [Claude Code](https://claude.ai/code/session_01TWEZw81eSypPaixcf1RKLH)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeBundleStore.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeBundleStore.kt
@@ -138,11 +138,16 @@ class ServeBundleStore(
    * follows a redirect on its own; [ServeUrlFetch.followingRedirects] does that, re-checking the
    * allowlist per hop.
    */
-  private fun fetchBundle(url: String): ByteArray? =
-    fetch?.invoke(url)
-      ?: ServeUrlFetch.followingRedirects(url, ::isAllowedUrl) {
-        ServeUrlFetch.sendOnce(it, maxBytes)
-      }
+  private fun fetchBundle(url: String): ByteArray? {
+    // An injected fetcher OWNS the result, including a null one — `?:` here would treat "the
+    // override reported a failure" as "there is no override" and quietly fall through to the real
+    // network.
+    val override = fetch
+    if (override != null) return override(url)
+    return ServeUrlFetch.followingRedirects(url, ::isAllowedUrl) {
+      ServeUrlFetch.sendOnce(it, maxBytes)
+    }
+  }
 
   /**
    * Extract the servable `previews/` entries (baked `<id>.png`, the `<id>.overrides.json` /
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeBundleStore.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeBundleStore.kt
@@ -2,14 +2,9 @@ package ee.schimke.composeai.cli.serve
 
 import ee.schimke.composeai.cli.BundleSigning
 import java.io.ByteArrayInputStream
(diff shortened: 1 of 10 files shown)


yschimke:
Good catch — valid, and worse than the P2 badge suggests. Fixed in `fd2920d`.

`fetch?.invoke(url) ?: ServeUrlFetch…` conflates *"the injected transport reported a failed fetch"* with *"there is no injected transport"*, so a `null` from an override fell straight through to the real HTTP client. Two consequences:

- **Tests silently went to the network.** `addFromUrl reports a fetch failure` (`fetch = { null }`, allowlist `ci.example.com`) made a live request to that host on every run and passed only because the connect failure produced the `Failed` it was asserting. Hidden network access in a unit test, plus its DNS/connect timeout.
- **The documented override wasn't actually authoritative** for any caller, so "no fetch happened" could never be asserted about a refused fetch.

The override now owns its result in both stores — the same bug existed in `ServeDocStore.fetchDocument` from #2893, so it's fixed there too rather than left to be found again:

```kotlin
val override = fetch
if (override != null) return override(url)
return ServeUrlFetch.followingRedirects(url, ::isAllowedUrl) { … }
```

Each store gains a test pinning the contract: a `null`-returning override yields a failure **and** is the only transport consulted (`assertEquals(1, calls)`), so a regression to `?:` fails rather than quietly dialling out again. `./gradlew :cli:test` green at 915.

---

PR #2908, opened 2026-07-29
feat(rc-player): render Remote Compose with the renderer's own typefaces

yschimke:
Follow-up to #2905, which noted that the residual parity gap was typeface substitution. This closes it.

before/after

Look at **Widget · large** and **Catalog · typography**: `before` flattens the type into one weight in the host's font; `after` reproduces the baked regular/bold pattern in Roboto.

## Cause

The browser lane's typeface id → CSS mapping named only the generics, with `DEFAULT` and every unrecognised id collapsing to `sans-serif`. But Android resolves those families to specific faces — `DEFAULT` and `sans-serif` are **Roboto** — while a headless container resolves `sans-serif` to DejaVu or Liberation. Every string the player drew therefore used a different typeface from the snapshot renderer. Invisible per glyph, permanent in aggregate, and it put a floor under the parity page that no layout fix could reach.

## Fix

`cssFontStackFor` names the concrete face first and keeps the generic as fallback, so a page with no faces registered renders exactly as before (viewers that ship no fonts are unaffected). `rc-compare` registers those faces from the *same vendored files the snapshot renderer rasterizes with*, inlined as data: URIs — no server, no network. Defaults on, so the parity page gets real typefaces without a workflow change.

Two things made this deceptive to get right:

- **`@font-face` is lazy and canvas does not drive it.** `ctx.font` neither triggers a load nor waits for one, and `document.fonts.ready` resolves while every face is still `unloaded`. My first measurement was of a page whose faces never loaded — it *looked* configured (faces listed, shorthand naming them, no warning) and painted the fallback. Each face is now loaded explicitly and asserted with `document.fonts.check()`, exiting non-zero rather than rendering substituted.
- **Discrete weights resolve upward.** With only 400 and 500 declared, CSS matching sends a request for 450 — which Wear M3's `bodyLarge` asks for — to Medium, rendering visibly too heavy. Contiguous ranges give every request a real file with no synthetic emboldening. This was caught from the evidence figure, not the numbers.

## Test plan

- Measured over all 30 `.rc` documents in `design-catalog-remote-m3` and `wear-widget`: mean mismatch **3.607% → 3.520%**, 17 better, 6 marginally worse, 7 unchanged.
- Gains concentrate where text dominates the frame: widget large **1.723 → 1.129**, gradient **0.868 → 0.398**, wear squircle **0.998 → 0.680**, typography **1.119 → 1.076**.
- The weight-range fix on its own accounts for widget large 1.225 → 1.129 and typography 1.104 → 1.076.
- `npm test` in `scripts/design-artifacts`: 247 pass, 0 fail (4 new in `rc-fonts.test.mjs`).

## Notes for the reviewer

**The 6 regressions are small (+0.02 to +0.14) and expected.** Changing typeface shifts glyph positions slightly; where the underlying residual is positional rather than typographic, that can cost a few pixels. None is a correctness change.

**Text-heavy previews improve less than you would expect, because typeface is not their dominant residual.** `RemoteTextSticker` barely moves (1.855 → 1.909) — its baked text starts at the left edge while the player insets it. That is a **layout/alignment** defect, unrelated to fonts, and worth a separate look; I did not chase it here.

**Named font families are still unsupported.** A named family reaches the paint layer as an id, not a name, so it silently becomes `sans-serif`. I left that alone deliberately: no `.rc` document in either catalog carries a family other than `"default"`, so plumbing the name through the paint protocol would be an unverifiable change to vendored upstream code. Worth doing when a document actually needs it.

**The new `.test.mjs` is not run by CI.** `scripts/design-artifacts` has ~15 test files and CI runs only the two shell scope tests; `npm test` covers them locally. Pre-existing gap, flagged rather than fixed here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L6SvmYX83WejJUsHYoCHM1)_


yschimke:
Not acting on the P1 "Re-author the commit with a human identity" — the premise is incorrect. `044eb49` is not authored or committed by Codex.

The commit object on GitHub:

```
author:    Yuri Schimke <yuri@schimke.ee>
committer: Yuri Schimke <yuri@schimke.ee>
```

Both map to the `yschimke` account, and the message carries no `Co-authored-by:` or other agent trailer. Same values locally (`git log -1 --format='%an <%ae> / %cn <%ce>'`).

The gate the review predicts will fail has already run against this exact head and passed: [`No Agent Attribution` run 30425568456](https://github.com/yschimke/compose-ai-tools/actions/runs/30425568456), `head_sha` `044eb49c0d66b129087894345bc6a36bc11c0f83`, conclusion **success**. That workflow is precisely the one that scans every commit's author/committer identity and trailers, so its verdict on this head is the direct answer to the concern.

Rewriting the commit would also invalidate the evidence image in the PR body, which is pinned to `044eb49` — so this would be churn with a real cost and no benefit.

Happy to be shown otherwise if there's a specific ref where the identity differs, but I can't reproduce it on the branch head, on GitHub's commit object, or in the gate's own result.

---

PR #2919, opened 2026-07-29
feat(rc-player): resolve named font families and serve them from Google Fonts

yschimke:
A Remote Compose document can name its typeface instead of picking one of the four generic families, but the browser lane never honoured it. `CoreText.updateVariables` ends `else this.mType = this.mFontFamilyId`, so a named family arrives at `PaintBundle.TYPEFACE` as the document's **text id**, which `cssFontStackFor` didn't recognise and mapped to its `default` — every branded string silently rendered Roboto.

The operand is now disambiguated by magnitude (ids are handed out from `RemoteComposeState.START_ID` = 42 upward, so no text id can collide with the generics 0..3), resolved back through the text table, and the face registered on demand.

Which families to fetch is **stated by the document, not guessed**: `google:Orbitron` is fetched from the Google Fonts CSS API; an unprefixed family is only *named* and left to the host. Treating any unrecognised family as a Google Font would turn a typo — or a name that only means something on the host, like `SF Pro` — into a network request, with no way to say "this one is local".

## Visual evidence

The catalog's real `Text/Branded` document (`RemoteFontFamily.Named("google:Orbitron")`), rendered three ways — baked snapshot, player before, player after:

named font families before/after``

Left is the baked snapshot (Roboto), middle is the player before this change (Roboto — the bug), right is the player after (real Orbitron). Committed at `docs/design/evidence/rc-named-font-families/before-after.png`.

Because headless Chromium in the dev container cannot egress to `fonts.googleapis.com` (`ERR_CONNECTION_RESET`, even through the proxy, though `curl` can), the "after" pane points the player's configurable `baseUrl` at a **local mirror of Google's genuine css2 response and woff2 binaries**. Everything but the origin is the real path: real Google CSS, real font files, real stylesheet load, real `FontFace` discovery, real canvas paint.

## Things this had to get right

- **The request enumerates weights rather than ranging them.** `wght@100..900` is rejected with HTTP 400 by any family that isn't variable — `Lobster` and `Pacifico` both do — and at the stylesheet link a 400 is indistinguishable from a network failure, so a static family would be reported unavailable and silently fall back. The enumerated `ital,wght@0,100;…;1,900` is accepted for variable and static families alike, and the API returns only the faces the family actually ships (Orbitron: six real 400–900 normal faces, no phantom italics).
- **`@font-face` is lazy and canvas does not drive it**, so the face is loaded explicitly. Note `document.fonts.check()` is *useless* as an assertion here — it answers `true` for a family that was never declared at all, so it cannot tell "registered" from "fell back". The tests measure text width instead. The load is asked for *by font shorthand*, for the (weight, style) the paint op actually carries, so the browser's own CSS matching fetches just that face: the stylesheet declares every weight, but declaring is free and fetching is not.
- **Resolution is synchronous but fetching is not.** The stack names the family immediately and the face lands later: interactive players repaint via `onFontLoaded`; single-shot renderers await `player.fontsReady()` *after* the first paint, which is what discovers the families in the first place.
- **Repaint callbacks are per waiter and fire once.** Because resolution runs per *paint*, a settled variant must never re-announce — that would schedule a repaint from inside painting, forever — while a caller arriving mid-fetch must still be recorded, or with two players on a page the second canvas sits in the fallback permanently (a static document has no later frame to recover on).
- **Family names are escaped as document data**, `\` as well as `"`. A family ending in a backslash would otherwise escape its own closing quote and swallow the rest of the stack, silently voiding the whole `ctx.font` assignment.

`configureWebFonts({enabled, baseUrl})` is the embedder's switch — off under a webview CSP that forbids the font origins or in a hermetic lane, redirected at a mirror. A family that cannot be served never throws: it degrades to the fallback generic, because a document naming a font we can't fetch is an authoring fact, not a player fault.

## Known asymmetry (documented, not hidden)

The **snapshot renderer does not yet resolve a named family** — Robolectric's `DefaultTypefaceResolver` looks the name up as a file under `/system/fonts/` and finds nothing — so the baked raster stays Roboto while the browser lane now draws the real face. Consequences to expect, both of which close together when the renderer side lands:

1. `Text/Branded` will read **high** on the PNG↔Remote-Compose parity page. That row is measuring the renderer gap, not a player regression.
2. The browser render **clips**, visible in the right pane above. An `.rc` carries geometry the *authoring* renderer measured, so while that renderer resolves the family to Roboto the player is drawing a wider face into a box measured for a narrower one.

The hook to close it exists — `RemoteDocumentPlayer` takes a `TypefaceResolver`, and `renderers/android` already downloads Google Fonts for Compose's own resolver via `GoogleFontCache`. What it needs is that downloader lifted out of `renderers/android` into a module the Remote Compose connector can also see, rather than a second copy of it: `downloadFromGoogleFonts` carries load-bearing stage-1/stage-2 reproducibility reasoning that must not be duplicated.

## Test plan

- New `scripts/design-artifacts/rc-webfonts.test.mjs` — nine cases, hermetic (serves its own stylesheet and a vendored `orbitron-400.ttf` from localhost; never touches the network): the css2 request form, multi-word family spelling, `baseUrl` redirection, the `google:` namespace parse (including that the prefix never leaks into the name or the API request), end-to-end registration proven by a text-width change, backslash/quote escaping (including that the escaped shorthand still applies), narrow per-weight fetching asserted as `["400:loaded", "700:unloaded"]`, one-shot notification of every waiter, and graceful degradation for an unservable family.
- Full `node --test scripts/design-artifacts/*.test.mjs`: **258 pass, 0 fail**.
- `:samples:design-catalog-remote-m3:composePreviewRender` succeeds; the emitted `.rc` carries `google:Orbitron` in its text table.
- `rc-compare.mjs` now awaits `player.fontsReady()` between the discovering paint and the frame it keeps.

Bundle rebuilt from the vendored source with the documented esbuild invocation; `PROVENANCE.md` records the delta and the asymmetry above.

https://claude.ai/code/session_01L6SvmYX83WejJUsHYoCHM1


yschimke:
All three review findings were real and are fixed in `141f111`. Suite is 258 pass / 0 fail.

**CodeQL (high), incomplete string escaping** — `namedFontStack` escaped `"` but not `\`. A family ending in a backslash emitted `"My Font\"`, where the backslash escapes the *closing* quote: the string runs on, swallows the rest of the stack, and the entire `ctx.font` assignment is silently voided — the exact failure the quoting exists to prevent. Escaping moved to a shared `cssQuoted`, and the test asserts the shorthand still applies rather than only checking the string.

**Over-fetching (P1)** — correct. The load now goes through the font shorthand for the (weight, style) the paint op carries, so the browser's own CSS matching fetches just that face. The stylesheet still declares every weight, since declaring is free and fetching is not. The fixture now declares two weights so the narrow fetch is asserted, not assumed: `["400:loaded", "700:unloaded"]`.

**Dropped callbacks (P2)** — correct diagnosis, but worth flagging that the suggested fix would have been worse than the bug. "Attach each caller's callback to the shared promise" notifies on *every* call, and typeface resolution runs per **paint** — so an already-settled family would schedule a repaint from inside painting, which paints, which notifies again: an endless repaint loop. What's implemented instead is one-shot listeners tracked per variant, so a caller joining an in-flight fetch is recorded (the case you correctly identified: with two players, the second canvas would otherwise sit in the fallback forever, since a static document has no later frame to recover on) while a settled variant never re-announces. Both halves are pinned by a test.

---

PR #2932, opened 2026-07-29
docs(upstream): report the Remote Compose TypefaceResolver API gap

yschimke:
Two markdown files, ready to file against AndroidX, recording why `RemoteFontFamily.Named("google:…")` still cannot be resolved in the snapshot renderer. The browser lane already resolves it (#2919); the shared font module it would resolve *through* already landed (#2927). This is the remaining blocker, and it is upstream.

## The gap

A custom `TypefaceResolver` can only **replace** the player's stock behaviour, never delegate to it — so a host that knows how to answer one kind of font request cannot add that without taking over all of them, including document-embedded fonts it knows nothing about.

Three approaches are ruled out, each for a different reason, and the report records which so nobody re-derives them:

- `AndroidRemoteContext` keeps the resolver in a plain field with **no lazy default**, so `getTypefaceResolver()` returns `null` for the whole lifetime of any player nobody has called the setter on — precisely the case where you would want to wrap it. The stock behaviour is built inside `AndroidPaintContext`, where it is not observable. **This is the root cause.**
- Reimplementing `DefaultTypefaceResolver` needs the `RemoteContext` it reads document-embedded `FontInfo` from, and `RemoteComposePlayer.getRemoteContext()` is `private`. Passing a fresh `AndroidRemoteContext` compiles but is a *different* context, so embedded fonts silently miss — a correctness regression, not a workaround.
- Secondarily, `RemoteDocumentPlayer` runs its `init` callback **before** it applies the `typefaceResolver` parameter, so a caller who passes both finds the parameter overwrites whatever `init` installed. This is a wrinkle rather than a blocker: with a lazy default in place, installing from `init` and leaving the parameter unset works.

## Files

- `docs/upstream/remote-compose-typeface-resolver-gap.md` — the report: summary, the three dead ends with quoted upstream code, impact, use case, and a three-line reproduction.
- `docs/upstream/remote-compose-typeface-resolver-proposal.md` — three fix options with compatibility trade-offs: (A) a lazy default in the getter; (B) a nullable "not mine" return so a resolver can decline; (C) applying the declared parameters before `init`. **A alone is sufficient**; C makes the fix discoverable rather than depending on the caller knowing that passing the parameter defeats wrapping; B is the better long-term shape while the API is still alpha.

Both are written for a reader with no knowledge of this repo — the use case is described generically, with no internal module names.

## Test plan

Docs only; nothing to build or render. Source claims were read at `androidx-main` and match the runtime behaviour observed while attempting the fix (a resolver installed via `init` never had its `load` reached, and no font file appeared in the cache).

Two caveats stated in the files rather than glossed: the source was read at `androidx-main` while our artifacts are `1.0.0-alpha15`, and option A narrows a `@Nullable` return to `@NonNull`, which may interact with AndroidX API-tracking rules.

Review found two errors in the first draft's analysis, both corrected in `55bf28b`: I had claimed A needed C (it doesn't — the ordering only bites a caller who also passes the parameter), and I had attributed option B's source break to implementers rather than callers (a non-null return remains a valid covariant override; it is callers who must handle `FontInstance?`).

https://claude.ai/code/session_01L6SvmYX83WejJUsHYoCHM1

### DIFF
--- a/docs/upstream/remote-compose-typeface-resolver-gap.md
+++ b/docs/upstream/remote-compose-typeface-resolver-gap.md
@@ -1,0 +1,131 @@
+# A custom `TypefaceResolver` can only replace the default, never delegate to it
+
+**Component:** `androidx.compose.remote:remote-player-core` / `remote-player-compose`
+**Versions checked:** artifacts `1.0.0-alpha15`; source read at `androidx-main`
+**Type:** API gap (no workaround available through public API)
+
+## Summary
+
+`RemoteDocumentPlayer` accepts a `typefaceResolver`, and `RemoteComposePlayer` exposes
+`get`/`setTypefaceResolver`. But there is no way to install a resolver that handles *some* requests
+and forwards the rest to the stock behaviour:
+
+- `AndroidRemoteContext.getTypefaceResolver()` returns `null` until someone calls the setter — there
+  is no default instance to read back and delegate to.
+- Reimplementing the default requires a `RemoteContext`, and the only accessor
+  (`RemoteComposePlayer.getRemoteContext()`) is `private`.
+
+So a caller who wants to add one behaviour — resolving a `RemoteFontFamily.Named` value the host
+knows how to fetch — must replace the whole resolver and cannot reproduce the part of the default
+that reads document-embedded fonts.
+
+## Why the obvious approaches don't work
+
+### 1. Decorating in the `init` callback
+
+`RemoteDocumentPlayer.kt` builds the player like this:
+
+```kotlin
+RemoteComposePlayer(it).apply {
+    doOnPreDraw { fullyDrawnReporter?.removeReporter() }
+    init(this)
+    bitmapLoader?.let(::setBitmapLoader)
+    typefaceResolver?.let(::setTypefaceResolver)
+}
+```
+
+`init(this)` runs **before** `setTypefaceResolver`, so:
+
+```kotlin
+init = { player ->
+    val current = player.getTypefaceResolver()   // always null here
+    if (current != null) player.setTypefaceResolver(MyResolver(current))
+}
+```
+
+never fires. And if it did, the `typefaceResolver` parameter would overwrite it one line later.
+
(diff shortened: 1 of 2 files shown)


yschimke:
Both P2s were right and are fixed in `55bf28b`. The P1 is factually wrong — details below.

**Option A sufficiency (P2) — correct, fixed.** I claimed A needed C because `init` runs before `setTypefaceResolver`. But that ordering only bites a caller who *also* passes the `typefaceResolver` parameter; leave it unset, install from `init`, and the trailing `?.let` is a no-op that can't overwrite anything. So A alone is the minimum. The reverse is also now stated: C alone fixes nothing, because without a lazy default there's still no stock resolver to wrap. The proposal now shows the working call shape explicitly, and the gap report says the ordering is a wrinkle rather than the root cause so the two documents agree.

**Option B compatibility (P2) — correct, fixed.** Widening a return from `@NonNull` to `@Nullable` doesn't break implementers: a non-null return remains a valid covariant override and Java implementations are unaffected. The break is on the **call** side, where Kotlin callers must now handle `FontInstance?`. Rewritten to say so, and to note the only in-tree caller is `AndroidPaintContext`, so external exposure is limited to code invoking a resolver directly.

Both of these would have sent upstream reviewers down the wrong path — one overstating the size of the change, the other pointing the compatibility question at the wrong audience. Good catches.

**Commit identity (P1) — not reproducible; no action taken.** The reviewed commit `4830ab00d2` is:

```
author=Yuri Schimke <yuri@schimke.ee>
committer=Yuri Schimke <yuri@schimke.ee>
```

with no `Co-authored-by:` or other trailers. There is no `Codex <codex@openai.com>` identity anywhere in this branch's history, so there is nothing for the attribution scanner to reject and nothing to recreate. Recreating a clean commit would be churn that loses the review history on this PR. The `No Agent Attribution` CI gate covers this independently — if it disagrees with the above I'll fix it immediately, but as of `55bf28b` it has nothing to flag.

---

PR #2958, opened 2026-07-29
refactor(rc-embedded): put the canvas text ops behind a platform seam, unchanged

yschimke:
Reworked per review — the previous revision re-expressed anchored text through Compose's text stack, which was a rendering change on Android. This one keeps Android's native text path exactly as it was and only moves it behind a seam.

## What changed

Three canvas text ops named `android.graphics` inline. Those calls now live in four functions in `RcPlayerTextPlatform.kt`, with the `android.graphics.Paint` builder private alongside them:

| | |
|---|---|
| `measureTextInkBounds` | was `Paint.getTextBounds` in `DrawTextAnchored` |
| `measureTextWidth` | was `Paint.measureText` in `DrawTextOnCircle` |
| `drawTextAtOriginPlatform` | was `nativeCanvas.drawText` in `DrawTextAnchored` |
| `drawTextOnPathPlatform` | was `nativeCanvas.drawTextOnPath` in `DrawTextOnPath` / `DrawTextOnCircle` |

**The bodies are the same framework calls, so Android's text output is unchanged.** What leaves the ops is only geometry that was never platform-specific: the anchoring arithmetic, and the arc construction, which moves from `android.graphics.Path` to Compose's `Path.addArc` — the same framework call underneath on Android.

## Why the seam sits below Compose's text APIs

Two reasons, and the second is the P1 from the last revision:

- **Ink bounds.** `DrawTextAnchored` anchors against the box the glyphs actually *mark* and reads `left`/`top` directly, mirroring `DrawTextAnchored.getHorizontalOffset`/`getVerticalOffset` in remote-core. Compose exposes *layout* bounds — side bearings and line spacing included — so substituting them shifts every anchored string.
- **Measure and draw cannot be allowed to disagree.** `toNativeTextPaint` resolves named and downloadable families (`device:` / `google:`); the `TextStyle` path maps everything outside the generics to `FontFamily.Default`. Splitting the two across those APIs means a document naming a font gets measured with one face and drawn with another. Keeping both sides on the framework `Paint` makes that impossible by construction rather than by care.

The jvm sibling replicates these four over skiko (`org.jetbrains.skia.Font`, `Canvas.drawString`, `PathMeasure` glyph placement) and is judged against Android's output — which is only a meaningful target because Android's output didn't move.

## `toTextStyle`

A pure extraction of `DrawText`'s own inline `TextStyle` construction, which is still its only caller — same generics-only family mapping, same `FontFamily.Default` fallback, same upstream `aosp/4187117` TODO. Unifying it with the native ops means teaching it the richer resolution, not pointing the native ops at it; that's a separate change that needs render verification.

## Net effect

| | before | after |
|---|---|---|
| `RcPlayerPaint.kt` | `android.graphics.Paint` | *gone* |
| `RcPlayerDrawing.kt` | `android.graphics.Rect`, 4× `nativeCanvas` | *gone* — no `android.graphics` reference at all |
| `RcPlayerDrawing.kt` Android surface | Bitmap, BitmapDrawable, Rect | Bitmap, BitmapDrawable (the image-decode seam) |
| `RcPlayerTextLayout.kt` | — | unchanged (this revision reverts the `internal` widening) |

## Test plan

- `:third-party-rc-embedded-player:check` — green (compile, Robolectric render harness, guard).
- `:third-party-rc-embedded-player-jvm:check` — green.

No visual evidence embedded, and this time that's the correct answer rather than a gap: the change is a move, not a rendering change, so there is nothing to show a before/after of. The reviewable claim is that the four extracted bodies are the same calls — visible in the diff — plus `Path.addArc` delegating to the framework on Android.

Also merges `main` (resolves the `PROVENANCE.md` conflict, where both sides appended to "Local modifications").

---
_Generated by [Claude Code](https://claude.ai/code/session_01VYTmD9R6gf92mATchtq461)_


yschimke:
As described this seems like an unacceptable regression. Can we use expect/actuals to use the normal android code, and then replicate those results with skiko?

yschimke:
Agreed, and reworked — you were right that it was a regression, and for a reason I'd underweighted.

Android's text path is now byte-for-byte what it was. `DrawTextAnchored` measures with `Paint.getTextBounds` and draws with `Canvas.drawText` again; the only change is that both calls moved into `RcPlayerTextPlatform.kt` alongside the two that were already there. Four functions, same bodies:

- `measureTextInkBounds` / `measureTextWidth` — the two `Paint` measurements
- `drawTextAtOriginPlatform` / `drawTextOnPathPlatform` — the two `Canvas` draws

That also dissolves the font-mismatch P1 rather than patching it. My previous fix resolved the named family for the `TextStyle` path so it matched the `Paint` path; with measure and draw both back on the same `Paint`, there's no second resolution to keep in sync. `resolvePaintFontFamily` is gone and `resolveFontFamily` is back to `private`.

`toTextStyle` stays, but only as a move of `DrawText`'s own inline style — same generics-only mapping, same `FontFamily.Default` fallback, same `aosp/4187117` TODO. It is deliberately *not* shared with the native ops now, and PROVENANCE says why: they resolve families differently, and unifying them is a rendering change that wants its own render verification.

So the skiko side becomes "replicate these four against Android's output", which is a target worth having precisely because Android's output didn't move. The one thing I'd flag: the arc in `DrawTextOnCircle` switched from `android.graphics.Path` to Compose's `Path.addArc` — that delegates to the framework call on Android, so it's the same geometry, but it's the only line in the diff where I'm arguing equivalence rather than showing an identical call.

Both module `check`s green. Also merged `main` and resolved the `PROVENANCE.md` conflict.

---

PR #2959, opened 2026-07-29
feat(design-artifacts): per-entry / per-axis render priority for catalog specs

### LINKED ISSUE #2950: Catalog spec: per-entry / per-axis render priority, deferring low-priority stickers to live previews
## Summary

Let a `catalog.spec.json` declare which entries must be baked into the published bundle and which can be produced on demand by the preview server later.

Catalogs that publish with `publish-live-bundle: true` + `split-per-preview: true` + `split-mode: full` already carry the classpath to re-render any preview on the serve host. So a deferred entry is not lost coverage — it is simply not rasterised in CI, and the server produces it when a viewer asks.

Split out of #2948 (semantics collector stops at a fixed 180s deadline), where this came up as a mitigation. It is complementary to that issue rather than a substitute — see Caveats.

## Proposal

```jsonc
{
  "componentId": "Buttons/Row button",
  "preview": "RowButtonLightPreview",          // priority: required (default)
  "variants": [
    { "preview": "RowButtonDarkPreview", "priority": "deferred" },
    { "preview": "RowButtonDisabledPreview", "priority": "deferred" }
  ]
}
```

- **`required`** (default) — rendered, joined, and subject to the strict completeness gate exactly as today.
- **`deferred`** — recorded in `catalog.json` with its preview id and flagged live-only; not rendered in CI, and not counted as a missing render or missing-semantics failure.

An axis-level form is probably the more valuable one (see measurements below):

```jsonc
"modes": ["light", "dark"],
"modePriority": { "light": "required", "*": "deferred" }
```

Bake one sticker per component, let the serve host produce the remaining palettes live. Given the theme switcher on the serve host is already interactive, that is arguably the better product as well as the cheaper build.

## Why not just `--allow-incomplete`

`--allow-incomplete` is all-or-nothing. Turning it on to tolerate a late semantics capture also means a genuinely broken *required* render publishes unnoticed — the gate stops being a gate. Priority keeps it strict over the core inventory while letting a catalog explicitly opt the long tail out.

## Measurements

Against `yschimke/pocket-casts-android@feffc631` `catalog.spec.json` — 61 componentIds, 139 preview references, nine user-selectable themes fanned out via `@PreviewParameter(ThemePreviewParameterProvider::class)`:

| Strategy | Renders | Change |
| --- | ---: | ---: |
| Everything (today) | 339 | — |
| Defer all variants | 213 | −37% |
| Keep all entries, defer themes beyond the primary | 139 | −59% |
| Both (base entries, primary theme only) | 61 | −82% |

The distribution is the interesting part, and it is the opposite of what I assumed before measuring: the fan-out lives in the **base** entries, not the variants — **19 of 61** bases fan out across nine themes, versus only **6 of 78** variants. So a purely per-entry `priority` aimed at variants is the weaker lever. If only one form is implemented, the mode/axis form is worth more.

## Caveats

- **Requires a live bundle.** Deferred entries only resolve where the serve host can re-render. `validateSpec` should reject `deferred` when `publish-live-bundle` is off, rather than silently dropping coverage from the published sheet.
- **Static consumers see less.** The Figma import path, `images/`, and `figma/*.svg` would only carry `required` entries. For an importable kit that is a real reduction, so the default must remain `required` and deferral must always be explicit.
- **Mitigation, not a fix, for #2948.** This reduces exposure to the fixed semantics deadline by shrinking the render set; a large enough `required` set still hits it. The collector should still follow `--timeout`.
- **Interaction with #2946.** Worth deciding whether a `deferred`-style escape also covers previews that *cannot* produce a static PNG at all (`AndroidView`, horologist `ScalingLazyColumn`). Those are a different problem — impossible rather than merely expensive — so they may warrant a distinct marker (`"capture": "animated"`) rather than being folded into priority.

yschimke:
Closes #2950. Follow-ups: #2965 (serve-side lane — the proper fix), #2966 (per-preview-id render filter).

## Summary

A `catalog.spec.json` can now declare which entries must be baked into the published bundle and which the preview server produces on demand.

```jsonc
// per axis — takes effect today, and the bigger lever per #2950's measurements
"modes": ["light", "dark"],
"modePriority": { "light": "required", "*": "deferred" }

// per entry / per variant — authorable now, inert until #2965
{ "componentId": "Buttons/Row", "preview": "RowButton",
  "variants": [{ "preview": "RowButtonDisabled", "state": "disabled", "priority": "deferred" }] }
```

- **`required`** (default, and what every spec that says nothing gets) — rendered, joined, gated exactly as today.
- **`deferred`** — recorded in `catalog.json` under a new top-level `deferred[]` array, with the `@Preview` name and the daemon `previewIds` its function produces. Exempt from the missing-render / missing-semantics gate.

The point being that the gate **stays strict over the required inventory** instead of being switched off wholesale with `--allow-incomplete`, which is all-or-nothing and would let a genuinely broken required render publish unnoticed.

## Staged against the serve-side gap

Codex flagged (correctly) that `ServeCatalogStore` never decodes `deferred[]`: `previewAliasFor` builds the catalog-id → daemon-id alias solely from `components[].images[].previewId`, and the load loop registers only baked images. #2965 is the proper fix. Until it lands, this PR is staged so the gap is **tracked but harmless**:

| | render time | published bundle | served catalog | status |
| per axis (`modePriority`) | no — needs #2966 | thinned | component stays, one fewer palette | **active** |
| per entry (`priority`) | wired, behind the switch | unchanged | unchanged | **recorded, not acted on** |

An entry/variant deferral is honoured as `required`: it renders and bakes exactly as it would without the annotation, so the published catalog is unchanged and nothing disappears from `serve --catalogs`. Both the driver and the pre-flight name the entries whose deferral is pending, so it can't go unnoticed:

```
note:    8 entry/variant `priority: "deferred"` annotation(s) are recorded but NOT yet acted on —
         rendered and baked as `required` until the preview server can route them (#2965):
         Button/Text (entry), Button/Group (entry), Card (entry), …
```

Refusing the publish and honouring the deferral were both worse — one blocks a catalog for annotating ahead of the server, the other quietly drops entries from the served sheet.

A single constant, `ENTRY_DEFERRAL_SERVED` in `catalog-priority.mjs`, drives this through `effectivePriority`, read by every consumer: the spec→candidate join, the variant split, `specDefersAnything` (so the live-path requirement isn't demanded for a deferral that does nothing), and the render-filter derivation. **That last one is load-bearing** — if the filter kept dropping entry-deferred functions while the driver baked them, a "required" entry would render no PNG and trip the completeness gate. Flipping the constant is the last step of #2965.

`modePriority` degrades gracefully by construction: the component keeps its untagged primary sticker in `components[]`, so the served catalog browses as before with one fewer baked palette — a reduction the catalog asked for, not a missing entry.

Also, per the issue's caveats:

- **Deferral requires a live path.** The driver refuses unless `--publish-live-bundle` or `--source-module` is set; the pre-flight mirrors it via `--live-bundle` / `--no-live-bundle`, which the reusable workflow derives from its own inputs. Keyed on *effective* deferral, so an inert annotation doesn't impose the constraint.
- **The primary sticker is never mode-deferrable.** Only a render that *names* a theme is eligible; a component whose every render is mode-deferred fails the gate rather than publishing empty.
- **The default stays `required`,** so static consumers only ever lose coverage explicitly.

Deferred entries still have their `preview` resolved against the module's discovered `@Preview` functions — the serve host renders them from the same module, so a name matching nothing is just as broken, only later.

## Files

- `catalog-priority.mjs` (new, pure, no deps) — declared vs effective priority, the serve-side switch, required/deferred preview split, render-filter derivation, image/variant splitting.
- `catalog-spec.mjs` — validates `priority` and `modePriority` (an unrecognised value is an **error**, since `entryPriority` fails closed and a typo would otherwise silently bake the entry), plus the live-path check.
- `generate-design-catalog.mjs` — the join skips deferred entries/variants, thins mode-deferred images out of the baked set, records `manifest.deferred[]`, enforces the live-path rule, warns on pending deferrals.
- `catalog.spec.schema.json`, `docs/design/DESIGN_CATALOGS.md`, both design-artifacts workflows.

## Test plan

- `node --test` in `scripts/design-artifacts` — **303 tests, 0 failures**. `catalog-priority.test.mjs` covers both sides of the switch (so the tests keep asserting the right thing when #2965 flips it), the shared-function rule, the untagged-sticker guarantee, wildcard expansion, and variant→`@Preview` resolution; plus new cases in `catalog-spec.test.mjs` and `bundle-previews.test.mjs`.
- Pre-flight against the real in-repo specs: unchanged, empty render filter — no behaviour change for today's catalogs. With `priority: "deferred"` injected into a third of `design-catalog-remote-m3`'s 24 entries it emits the pending-deferral note, an **empty** filter, and exits 0 even with `--no-live-bundle`. A `modePriority` spec with `--no-live-bundle` still fails with the live-path error.
- Both workflow YAMLs parse. Full CI green on the first commit (46 checks).

No visual surfaces touched — spec/export/build plumbing, so text-only evidence is the right form here.

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -256,12 +256,15 @@ jobs:
       # module's @Preview functions BEFORE the (long) render, so a mistyped or
       # renamed name fails here in seconds instead of as a late "missing" entry.
       - name: Validate catalog spec
+        id: validate-spec
         env:
           SPEC: ${{ inputs.spec }}
           MODULE: ${{ inputs.module }}
           EXTRA_MODULE: ${{ inputs.extra-module }}
           PREVIEW_ANNOTATIONS: ${{ inputs.preview-annotations }}
           WORKDIR: ${{ inputs.working-directory }}
+          LIVE_BUNDLE: ${{ inputs.publish-live-bundle }}
+          LIVE_SOURCE: ${{ inputs.live-rerender-source }}
         run: |
           set -euo pipefail
           # Gradle path (":app", "a:b") → directory ("app", "a/b"), under the
@@ -277,7 +280,21 @@ jobs:
           # annotation module). Without them the scan finds nothing and every
           # entry fails — which reads as a broken spec, not a missing flag.
           for ann in $PREVIEW_ANNOTATIONS; do args+=(--preview-annotation "$ann"); done
-          node compose-ai-tools/scripts/design-artifacts/validate-catalog-spec.mjs "${args[@]}"
+          # Render priority (issue #2950): a spec may defer entries/modes to the live
+          # preview server. Tell the pre-flight whether this publish HAS a live path, so
+          # a deferral with no way to produce the deferred pixels fails in seconds here
+          # rather than at the end of the render.
+          if [ "$LIVE_BUNDLE" = "true" ] || [ "$LIVE_SOURCE" = "true" ]; then
+            args+=(--live-bundle)
+          else
+            args+=(--no-live-bundle)
+          fi
+          node compose-ai-tools/scripts/design-artifacts/validate-catalog-spec.mjs "${args[@]}" \
+            --render-filter-out "$GITHUB_WORKSPACE/render-filter.txt"
+          # The `--preview` patterns that render just the required entries. EMPTY for every
+          # spec that defers no entry — i.e. all of them today — in which case the render
+          # step below sees an empty property and renders everything, exactly as before.
+          echo "render-filter=$(cat "$GITHUB_WORKSPACE/render-filter.txt")" >> "$GITHUB_OUTPUT"
 
       # Stage bundled fallback fonts into fontconfig before any render, so a
       # desktop (Skiko) render resolves glyphs a preview's FontFamily doesn't
@@ -330,6 +347,17 @@ jobs:
         # compose-preview finds the Gradle root by walking up for gradlew, so a
         # monorepo of separate builds must render from the build's own dir.
         working-directory: ${{ inputs.working-directory || '.' }}
+        env:
+          # Render only the previews this catalog still needs baked, when its spec defers
+          # entries to the live preview server (issue #2950). Read by the render task as the
+          # `composePreview.filter` project property — Gradle sources project properties from
+          # ORG_GRADLE_PROJECT_* env vars, so this reaches `composePreviewRender` through
+          # `bundle pack` with no CLI flag. Empty ⇒ no filter ⇒ render everything.
+          #
+          # Scoped to the PRIMARY render on purpose: the patterns name preview functions from
+          # this catalog's spec, and the filter fails fast when it matches nothing — an extra
+          # module whose previews are all deferred would sink the render rather than skip it.
+          ORG_GRADLE_PROJECT_composePreview.filter: ${{ steps.validate-spec.outputs.render-filter }}
         run: |
           set -euo pipefail
           run() { if [ "${{ inputs.desktop-render }}" = "true" ]; then xvfb-run -a -s "-screen 0 2000x1400x24" "$@"; else "$@"; fi; }
--- a/.github/workflows/design-artifacts.yml
+++ b/.github/workflows/design-artifacts.yml
@@ -278,7 +278,16 @@ jobs:
       # errors (duplicate componentId, malformed variant) fail here too. Uses only
       # node built-ins (no npm deps). Coverage warnings don't fail the job.
       - name: Validate catalog spec
-        run: node scripts/design-artifacts/validate-catalog-spec.mjs --spec "${{ matrix.spec }}"
+        id: validate-spec
+        run: |
+          set -euo pipefail
+          node scripts/design-artifacts/validate-catalog-spec.mjs --spec "${{ matrix.spec }}" \
+            --render-filter-out "$GITHUB_WORKSPACE/render-filter.txt"
+          # Render priority (issue #2950): the `--preview` patterns that render just the
+          # entries this spec still needs baked. EMPTY unless the spec marks entries
+          # `priority: "deferred"` — no catalog here does today, so the render set is
+          # unchanged. Consumed by the render step below.
+          echo "render-filter=$(cat "$GITHUB_WORKSPACE/render-filter.txt")" >> "$GITHUB_OUTPUT"
 
       # Published previews pin their preview-runtimes to a RELEASED version
       # (`composeaiReleasedRuntimeVersion`, bumped by release-please on each release). Sonatype →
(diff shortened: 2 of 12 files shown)


yschimke:
The P1 on `generate-design-catalog.mjs` is correct, and I've acted on it in e18fcf1.

Verified all three claims: `ServeCatalogStore.Catalog` (the `@Serializable` model, ~line 872) has no `deferred` field so the array is dropped at parse; `previewAliasFor` (126–135) builds the alias solely from `components[].images[].previewId`; and `load`'s registration loop (223–263) walks `components[].images` only, keying each preview off `previewIdFor(image.path)` — which a deferred record has no `path` for.

Rather than land a half-wired path, I've split the feature by **what a viewer actually loses**, because the two forms are not equivalent here:

- **`modePriority` (per axis) stays enabled.** A mode-deferred image leaves its component in `components[]` with the untagged primary sticker, so the served catalog browses exactly as before with one fewer baked palette. Nothing disappears, nothing 404s. This is also the form #2950's measurements say is worth more (−59% vs −37%), so the useful half ships.
- **Entry- and variant-level `priority: "deferred"` is now refused** — the driver exits non-zero, naming the offending componentIds and pointing at the follow-up. Those genuinely would vanish from `serve --catalogs`, and "silently reduced coverage dressed up as an optimisation" is the exact failure mode the `--allow-incomplete` argument in #2950 is against, so it fails closed like the completeness gate immediately above it.

Follow-ups filed rather than left implicit:

- **#2965** — serve-side consumption of `deferred[]`. Includes the design question your comment surfaces but doesn't have an obvious answer to: `ServeBundleHost` serves `previews/<id>.png` off disk and `load` hard-fails on `count == 0`, so a preview with no baked PNG needs either a live-only placeholder card or to be hidden when the session has no daemon — plus fail-soft behaviour on a baked-only server. Closing it means deleting the guard.
- **#2966** — the per-preview-id render filter, which is what would make `modePriority` save render time rather than only publish size (today's `PreviewNameFilter` matches `functionName`, and a theme fan-out shares one function). It also notes that `packSemanticsBlob` needs the same filtering or the axis saving stays partial.

Worth noting the render-side half of entry deferral is already built and tested here (`renderFilterPatterns` → `validate-catalog-spec --render-filter-out` → the workflows' `ORG_GRADLE_PROJECT_composePreview.filter`), so #2965 is the only thing between this and the full feature — no rework, just removing one guard.

---

yschimke:
Reworked in 160878e — the hard refusal from my previous comment was the wrong shape. Replaced with a **tolerated, tracked, harmless** gap:

An entry/variant `priority: "deferred"` is now **recorded but not acted on**: it renders and bakes as `required`, so the published catalog is byte-for-byte what it was without the annotation and nothing disappears from `serve --catalogs`. Both the driver and the build-free pre-flight name the pending entries, so the gap is visible rather than a surprise when the render bill doesn't drop:

```
note:    8 entry/variant `priority: "deferred"` annotation(s) are recorded but NOT yet acted on —
         rendered and baked as `required` until the preview server can route them (#2965):
         Button/Text (entry), Button/Group (entry), Card (entry), …
```

Refusing the publish (what I had) blocks a catalog for annotating ahead of the server; honouring the deferral drops entries from the served sheet. Neither is a good trade for a gap we're tracking — a spec can now be annotated today, causes no change in the meantime, and starts deferring when #2965 lands.

One constant drives it, `ENTRY_DEFERRAL_SERVED`, read through `effectivePriority` by every consumer — the spec→candidate join, the variant split, `specDefersAnything` (so the live-path requirement isn't imposed for a deferral that does nothing) and the render-filter derivation. That last one is the subtle part your comment's logic implies: had the filter gone on dropping entry-deferred functions from the render while the driver baked them, a "required" entry would have had no PNG and tripped the completeness gate. One switch means the render set and the published set cannot disagree. Flipping it is the last step of #2965.

`modePriority` is unchanged and active — its component keeps the untagged primary sticker in `components[]`, so the served catalog browses as before with one fewer baked palette.

Tests assert **both** sides of the switch, so they keep checking the right thing when #2965 flips it rather than needing a rewrite. 303 tests, 0 failures.

---

Issue #2965, opened 2026-07-29
serve: consume catalog.json `deferred[]` so deferred entries have a live lane


yschimke:
Follow-up to #2950 / #2959.

#2959 added per-entry / per-axis render priority to `catalog.spec.json`. The export records deferred coverage in a new top-level `catalog.json` `deferred[]` array — each record carrying its `componentId`, group/section, axes (`state` / `props` / `theme` / `size`), the `@Preview` function name, and the daemon `previewIds` that function produces.

**Nothing consumes it yet.** In `ServeCatalogStore`:

- `Catalog` (the `@Serializable` model, ~line 872) has no `deferred` field, so the array is dropped at parse.
- `previewAliasFor` (126–135) builds the catalog-id → daemon-id alias solely from `components[].images[].previewId`.
- `load`'s registration loop (223–263) walks `components[].images` only, and keys each preview off `previewIdFor(image.path)` — a deferred record has no `path`.

Consequence: were the export to act on an entry-level deferral today, a wholly-deferred **entry** would be absent from `compose-preview serve --catalogs` entirely, and a deferred **variant** would have no route a viewer could request from either a carried live bundle or a buildable source. So the on-demand path the feature advertises can't work.

Because of that, #2959 ships entry- and variant-level deferral as **declared but inert**: `ENTRY_DEFERRAL_SERVED = false` in [`catalog-priority.mjs`](https://github.com/yschimke/compose-ai-tools/blob/main/scripts/design-artifacts/catalog-priority.mjs) makes `effectivePriority` answer `required`, so such an entry renders and bakes exactly as it would without the annotation — the published catalog is unchanged and nothing disappears from the served sheet. The driver and the build-free pre-flight both name the pending entries so the gap stays visible. `modePriority` is unaffected and active, since it keeps every component in `components[]` with its primary sticker.

**Closing this issue means flipping that constant to `true`.** One switch drives every consumer (the spec→candidate join, the variant split, `specDefersAnything`, and the render-filter derivation), so the render set and the published set can't disagree about which entries are baked. The unit tests assert both sides of it already, so they don't need rewriting when it flips.

## What's needed

1. `Catalog.deferred: List<DeferredEntry> = emptyList()` in the serve model.
[...]

PR #2972, opened 2026-07-29
fix(design-artifacts): make entry-level deferral inert until the serve host can route it

yschimke:
Follow-up to #2959, which **merged before this landed on its branch** — so `main` currently has the unguarded version. Refs #2965.

## The trap

As merged, an entry/variant `priority: "deferred"` is acted on immediately. But nothing consumes the `deferred[]` records the export writes: `ServeCatalogStore` builds both its preview registration (`load`, lines 223–263) and its catalog-id → daemon-id alias (`previewAliasFor`, 126–135) from `components[].images` alone, and its `Catalog` model has no `deferred` field at all, so the array is dropped at parse.

A wholly-deferred entry has no `images[]` record — so it would be **absent** from `compose-preview serve --catalogs` rather than rendered on demand. That's silently reduced coverage presented as an optimisation, which is exactly the `--allow-incomplete` failure mode #2950 set out to avoid.

No catalog sets `priority` yet, so nothing is broken today. This closes the trap before one does.

## The fix

An entry/variant deferral is now **recorded but not acted on**: it renders and bakes as `required`, so the published catalog is byte-for-byte what it was without the annotation and nothing disappears from the served sheet. Both the driver and the build-free pre-flight name the pending entries:

```
note:    8 entry/variant `priority: "deferred"` annotation(s) are recorded but NOT yet acted on —
         rendered and baked as `required` until the preview server can route them (#2965):
         Button/Text (entry), Button/Group (entry), Card (entry), …
```

Inert rather than fatal, on purpose. Refusing the publish would block a catalog for annotating ahead of the server; honouring the deferral drops coverage. Neither is a good trade for a gap we're tracking — this way a spec can be annotated now, changes nothing meanwhile, and starts deferring when the serve lane lands.

One constant drives it — `ENTRY_DEFERRAL_SERVED` in `catalog-priority.mjs` — read through `effectivePriority` by every consumer: the spec→candidate join, the variant split, `specDefersAnything` (so the live-path requirement isn't demanded for a deferral that does nothing), and the render-filter derivation.

**That last one is load-bearing.** With the filter still dropping entry-deferred functions from the render while the driver bakes them, a "required" entry would render no PNG and trip the completeness gate. One switch means the render set and the published set cannot disagree. Flipping it is the last step of #2965.

`modePriority` is unaffected and takes effect today: a mode-deferred image leaves its component in `components[]` with the untagged primary sticker, so the served catalog browses as before with one fewer baked palette — a reduction the catalog asked for, not a missing entry.

## Test plan

- `node --test` in `scripts/design-artifacts` on top of `main` — **303 tests, 0 failures** (the one skip is the pre-existing `rc-compare-pixels.test.mjs`, which needs `pngjs` from `npm ci`).
- Tests assert **both** sides of the switch, so they keep checking the right thing when #2965 flips it instead of needing a rewrite.
- Pre-flight against a deferring copy of `design-catalog-remote-m3` (8 of 24 entries): emits the pending-deferral note, an **empty** render filter, and exits 0 even with `--no-live-bundle`. A `modePriority` spec with `--no-live-bundle` still fails with the live-path error. Real in-repo specs unchanged.

Non-visual (spec/export plumbing), so text-only evidence is the right form here.

---
_Generated by [Claude Code](https://claude.ai/code/session_016azcVhRhQze6Y91gMT4BeF)_


yschimke:
Correct — fixed. The `priority` schema description and the `catalog-priority.mjs` header still described the end state (entry omitted from CI, live path required), which is exactly what the validator and generator no longer do for an entry-only deferral. Editor tooling reads that schema description, so it was the most likely place to mislead someone.

Both now distinguish the two forms explicitly: the axis form (`modePriority`) as active today, the entry form as recorded-but-not-acted-on with a pointer to #2965, plus the note that an inert annotation imposes no live-path requirement and is safe to author ahead of the server. The schema also steers anyone wanting deferral *now* to `modePriority`.

303 tests still pass; schema still parses.

---

Issue #2948, opened 2026-07-29
Semantics collector still stops at the fixed 180s deadline despite --timeout 600 (second repo hitting #2857, which was closed via workaround)


yschimke:
## Summary

#2857 is closed as completed, but it was resolved by scoping `allow-incomplete: true` to Jetsnack in compose-samples rather than by changing the collector. The root cause described there — a semantics subscriber deadline that is separate from, and unaffected by, `--timeout` — is still present.

A second, independent repository now hits it on its first adoption, which suggests this is a general scaling limit rather than anything specific to Jetsnack.

## Evidence

Repository/ref: `yschimke/pocket-casts-android@9d838203e887f7e203857fae9034b3be623fe0c3` (branch `previews`), workflow run [30471225420](https://github.com/yschimke/pocket-casts-android/actions/runs/30471225420), job 90641670656 (`pocketcasts`, module `:modules:services:compose`).

The workflow passes `--timeout 600`:

```text
compose-preview bundle pack --module ":modules:services:compose" --with-semantics \
  -o "$GITHUB_WORKSPACE/bundle.png" --timeout 600
```

Render step succeeded and took **13m 10s** (16:33:23 → 16:46:33), well beyond 180s. Generation then reported:

```text
[pocketcasts] no semantics for: Buttons/Small icon button, Buttons/Toggle button group,
  Buttons/Time play button, Inputs/Search bar, Inputs/Selected option column,
  Content/Explicit icon, Content/Podcast subscribe row, Content/Bookmark row,
  Feedback/Banner, Feedback/No content banner, Feedback/Dialog, Feedback/Tooltip,
  Feedback/Sparkle image, Chrome/Top app bar, Chrome/Bottom sheet content,
  Settings/Setting section, Identity/User avatar, Ads/Ad banner,
  Ads/Ad decisions column, Ads/Ad report reasons, Insights/Calendar heat map,
  Insights/Player header row
[pocketcasts] incomplete render — refusing to publish.
```

22 of 61 componentIds — roughly a third of the catalog — published without an a11y tree once `allow-incomplete` was set.

## Why this catalog is a useful data point

[...]

Issue #2066, opened 2026-06-25
composePreviewRender: filter to a single preview (or subset) by name


Wavesonics:
## Context

The `composePreviewRender` Gradle task (`RenderPreviewsTask`, from the `ee.schimke.composeai.preview` plugin) renders **every** discovered preview in a module. `gradle :module:help --task composePreviewRender` lists only `--rerun` — there's no way to narrow the run to a single preview function.

Two pain points in a plain Gradle-task workflow (no daemon / no MCP):

1. **Tight iteration loop is slow.** When I'm iterating on one screen's `@Preview`, re-rendering the whole module on every edit is a lot of wasted work in a module with many previews.
2. **One unrelated broken preview poisons the batch.** An unrelated preview that fails to render shows up as a failure in the run even when the preview I actually care about renders perfectly, adding noise to the output I have to grep past.

The MCP/CLI daemon already renders a single preview (`render_preview`), but for a CI-free, daemon-free Gradle-task setup there's no equivalent filter. In practice people wrap the bulk task in a local script that runs the full render and then copies out the one PNG they want — which is the gap this issue is about.

## Proposal

Add a filter option to `composePreviewRender`:

- `--preview <name>` (repeatable) and/or `-Pcompose.preview.filter=<glob>` matching preview functions by simple or fully-qualified name, with glob/substring support (e.g. `*ExportHelpDialogPreview` or `com.example.preview.ExportHelpDialogPreview`).
- When set, only matching previews are discovered + rendered; non-matching previews (and their failures) are skipped.
- No match → fail fast with a clear message (ideally listing nearby/available names, or pointing at `composePreviewDiscover`).

## Acceptance criteria

- `./gradlew :module:composePreviewRender --preview '*ExportHelpDialogPreview'` renders only that preview's PNG to `build/compose-previews/renders/`.
- An unrelated broken preview in the same module does **not** fail a filtered run that doesn't match it.
- An empty match exits non-zero with a helpful message.

## Out of scope

- Daemon/MCP single-preview rendering (already exists via `render_preview`).
- Watch mode / "only render previews whose source changed" incremental detection.

---

(Filed as a user of the Gradle plugin — happy to test a snapshot. Thanks for the tool!)

PR #2973, opened 2026-07-29
feat(plugin): per-preview-id render filter, so a deferred catalog palette skips its render

### LINKED ISSUE #2966: plugin: per-preview-id render filter, so `modePriority` skips renders instead of only trimming the publish
Follow-up to #2950 / #2959.

#2959 added `modePriority` to `catalog.spec.json`, which per #2950's measurements is the bigger lever — against `yschimke/pocket-casts-android@feffc631` (61 componentIds, nine `@PreviewParameter` themes) deferring every palette beyond the primary is **−59%** of 339 renders, versus −37% for deferring all variants.

Today it only thins what is **published** (`images/`, `figma/*.svg`, the Figma import). It does not shrink the render, because:

- the theme fan-out lives *inside* one `@Preview` function — a multipreview member (`@CatalogModes` → `Foo_Light` / `Foo_Dark`) or a `@PreviewParameter` row — and each combination is its own `PreviewInfo` with a distinct `id` but the **same** `functionName`;
- the only render-side selector is `PreviewNameFilter`, which matches `functionName` / package-qualified name (`RenderPreviewsTask.selectNamedPreviews`). It can drop a whole function, never one of its fan-out members.

So entry-level `priority: "deferred"` can skip renders (#2959 wires that through `--render-filter-out` → `ORG_GRADLE_PROJECT_composePreview.filter`) but `modePriority` cannot.

## What's needed

1. An **id-level** filter on `RenderPreviewsTask`, alongside `previewFilters` — e.g. `previewIdFilters` / `--preview-id` + `-PcomposePreview.idFilter`, matched against `PreviewInfo.id` with the same glob semantics `PreviewNameFilter` already implements. Exclusion polarity is probably right for deferral (a typo then renders too much rather than nothing).
2. The same caching treatment as the name filter: `outputs.cacheIf` must refuse to cache a filtered run, since `outputDir` is then not the module's complete render set (see the comment at `RenderPreviewsTask.kt:132-148`).
3. A way for the export driver to know the ids **before** the render. They come from discovery (`previews.json`), which the build-free pre-flight doesn't have — so either the filter accepts globs the spec can express (`*_dark`, fragile) or the workflow runs `composePreviewDiscover` first and derives exact ids. The latter is more robust and worth the extra Gradle invocation.
4. Plumbing: `bundle pack` passes `-P`s it constructs itself, so either a new pass-through flag or (as #2959 does for the name filter) rely on `ORG_GRADLE_PROJECT_*` reaching `composePreviewRender` through the subprocess.
5. **Semantics capture is the other half.** `bundle pack --with-semantics` starts the daemon and captures per preview (`packSemanticsBlob`), which is where #2948's fixed 180s deadline bites. Filtering the render alone doesn't reduce that pass, so the deferred previews must be excluded there too or the axis saving is partial.

Note the interaction with #2965: deferred previews must stay listed in the bundle's `previews.json` (the bundle task already carries every selected preview and simply omits the PNG for one that didn't render) — that listing is what keeps them addressable for the serve host's live lane. So this must filter the **render**, not the pack.

yschimke:
Closes #2966 (plugin half). Follow-up to #2950 / #2959.

## Why

`--preview` / `-PcomposePreview.filter` matches `functionName`, so it can only keep or drop a whole `@Preview`. A multipreview member or a `@PreviewParameter` row is its own `PreviewInfo` with a distinct `id` but the **same** `functionName` — which is why a catalog's per-theme `modePriority` could thin its published sticker set but not its build. Per #2950's measurements that axis is the bigger lever: against `yschimke/pocket-casts-android@feffc631` (61 componentIds, nine themes) deferring every palette beyond the primary is **−59%** of 339 renders, vs −37% for deferring all variants.

## What

Id-level granularity, in **both** polarities — they are not interchangeable:

| flag | property | semantics |
| --- | --- | --- |
| `--preview-id` | `composePreview.idFilter` | select by id; select-or-fail |
| `--exclude-preview-id` | `composePreview.idExclude` | skip by id, keep the rest |

Exclusion is the one a deferral needs. A positive `*_light` **cannot** express "bake one palette per component": it would also drop every preview whose id carries no theme suffix — the untagged primary stickers, i.e. most of the catalog. There's a test pinning exactly that asymmetry.

The two polarities also fail in opposite directions, deliberately:

- a positive filter that matches nothing **throws**, listing available ids — a stale spec must be loud, not a silent zero-preview render that resurfaces later as a bundle of missing stickers (same policy as `selectNamedPreviews`);
- an exclusion that matches nothing is a **no-op** — it renders more than intended, which the publish catches. Excluding *every* preview still throws, since that packs a bundle of missing stickers.

Both compose with the name filter, which runs first: `--preview FilledButton --preview-id *_Light`.

Both are `@Input`, and both disqualify build-cache storage for the same reason the name filter does — `outputDir` is then not the module's complete render set, so a partial snapshot must never be restorable over a full one (see the existing comment block at `RenderPreviewsTask.kt:132-148`).

The property forms are what make this reachable from the design-artifacts pipeline: Gradle sources project properties from `ORG_GRADLE_PROJECT_*`, so `ORG_GRADLE_PROJECT_composePreview.idExclude` reaches `composePreviewRender` through `bundle pack` with no new CLI flag — the same route #2959 already uses for the name filter.

## Not in this PR

- **Deriving the exclusions from `modePriority`.** The ids come from discovery (`previews.json`), which the build-free pre-flight doesn't have, so the driver side needs either an extra `composePreviewDiscover` pass or spec-expressible globs. Left for the export-side half of #2966 so this is reviewable and testable on its own.
- **Semantics capture.** `bundle pack --with-semantics` captures per preview via `packSemanticsBlob`, which is where #2948's fixed 180s deadline actually bites. Filtering the render alone doesn't shrink that pass, so the axis saving stays partial until it's filtered too. Noted on #2966.
- Deferred previews must stay listed in the bundle's `previews.json` (the bundle task already carries every selected preview and omits only the PNG) — that listing is what keeps them addressable for the serve host's live lane. So this filters the **render**, never the pack.

## Test plan

- `:gradle-plugin:test` and `:gradle-plugin:preview-discovery:test` — **BUILD SUCCESSFUL**, full suites, not just the new classes.
- `./gradlew ktfmtCheckAll` — **BUILD SUCCESSFUL** (the CI format gate).
- New `SelectPreviewIdsTest` (13 cases): glob/substring/OR matching, splitting one function's fan-out where a name filter can't, composition with the name filter, both failure policies, and the polarity asymmetry above. New `matchesId` cases in `PreviewNameFilterTest`, including that a `.` in an id stays literal.

Non-visual (build wiring), so text-only evidence is the right form here.

---
_Generated by [Claude Code](https://claude.ai/code/session_016azcVhRhQze6Y91gMT4BeF)_

### DIFF
--- a/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewNameFilter.kt
+++ b/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewNameFilter.kt
@@ -48,6 +48,25 @@ object PreviewNameFilter {
     return if (pkg.isEmpty()) functionName else "$pkg.$functionName"
   }
 
+  /**
+   * True when a preview **id** matches at least one of [patterns], or when [patterns] is empty.
+   *
+   * The id-level counterpart of [matches], for selecting individual members of one `@Preview`
+   * function's fan-out (issue #2966). A multipreview member or `@PreviewParameter` row is its own
+   * `PreviewInfo` with a distinct `id` (`FilledButton_Light` / `FilledButton_Dark`) but the *same*
+   * `functionName`, so [matches] can only ever keep or drop the whole function — which is why a
+   * catalog's `modePriority` could thin its published sticker set but not its render.
+   *
+   * Ids are opaque strings rather than named code, so only one candidate is compared (no FQN
+   * variant), but the glob-vs-substring rules are deliberately identical to [matches] so a single
+   * documented pattern syntax covers both filters.
+   */
+  fun matchesId(patterns: Collection<String>, id: String): Boolean {
+    val cleaned = patterns.map(String::trim).filter(String::isNotEmpty)
+    if (cleaned.isEmpty()) return true
+    return cleaned.any { matchOne(it, id, id) }
+  }
+
   private fun matchOne(pattern: String, simpleName: String, fqName: String): Boolean =
     if (pattern.any { it == '*' || it == '?' }) {
       val regex = globToRegex(pattern)
--- a/gradle-plugin/preview-discovery/src/test/kotlin/ee/schimke/composeai/discovery/PreviewNameFilterTest.kt
+++ b/gradle-plugin/preview-discovery/src/test/kotlin/ee/schimke/composeai/discovery/PreviewNameFilterTest.kt
@@ -99,4 +99,31 @@ class PreviewNameFilterTest {
   fun `fqName falls back to the bare function name in the default package`() {
     assertThat(PreviewNameFilter.fqName("FooKt", "BarPreview")).isEqualTo("BarPreview")
   }
+
+  @Test
+  fun `matchesId with an empty filter keeps every id`() {
+    assertThat(PreviewNameFilter.matchesId(emptyList(), "FilledButton_Dark")).isTrue()
+    assertThat(PreviewNameFilter.matchesId(listOf("  ", ""), "FilledButton_Dark")).isTrue()
+  }
+
+  @Test
+  fun `matchesId globs a fan-out suffix`() {
+    // The catalog case: keep one palette per component, defer the rest to the live preview server.
+    assertThat(PreviewNameFilter.matchesId(listOf("*_Light"), "FilledButton_Light")).isTrue()
+    assertThat(PreviewNameFilter.matchesId(listOf("*_Light"), "FilledButton_Dark")).isFalse()
+  }
+
+  @Test
+  fun `matchesId treats a glob-free pattern as equality-or-substring, like matches`() {
+    assertThat(PreviewNameFilter.matchesId(listOf("FilledButton_Dark"), "FilledButton_Dark"))
+      .isTrue()
+    assertThat(PreviewNameFilter.matchesId(listOf("Filled"), "FilledButton_Dark")).isTrue()
+    assertThat(PreviewNameFilter.matchesId(listOf("Outlined"), "FilledButton_Dark")).isFalse()
+  }
+
+  @Test
+  fun `matchesId does not treat a dot in an id as a wildcard`() {
+    // Ids can carry package-qualified forms; `.` must stay literal, as it does for fqName matching.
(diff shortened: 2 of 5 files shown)


yschimke:
Verified and correct — thanks, this is the useful kind of finding. `composePreviewRender` is registered twice: `ComposePreviewTasks.kt:241` as a `RenderPreviewsTask` (desktop) and `AndroidPreviewSupport.kt:1903` as a Robolectric `Test` task. `previewFilterProperty` has exactly one call site — line 259, the desktop registration — so the Android task reads none of these.

Two things follow, and the second is the more important one:

1. **The pre-existing #2066 name filter has the same limitation.** So this PR isn't introducing an inconsistency; it inherits one. Worth stating plainly rather than leaving implied.
2. **#2959's already-merged workflow wiring is therefore desktop-only.** Both design-artifacts workflows export `ORG_GRADLE_PROJECT_composePreview.filter` from the pre-flight, and for an Android catalog (wear-m3, remote-m3, pocket-casts) nothing reads it. Harmless today, since entry deferral is inert pending #2965 — but it must not stay silent once that flips. And the motivating pocket-casts measurement is an *Android* module, so the −59% needs the Android path, not just this PR.

I've **not** wired the Robolectric path here. It's a different subsystem — the render runs inside the test JVM, so the patterns have to reach the manifest loader as system properties, with matching cache gating and the same treatment for the sibling `composePreviewRenderAndroidResources` / `composePreviewRenderXr` tasks. That's its own change with its own tests, not a small in-scope fix.

Instead, pushed in this PR: the desktop-only scope is now documented on `previewIdFilters`, `previewIdExcludes`, and both property resolvers, so the limitation is visible at the point of use rather than discovered from a render bill that didn't drop. The filters remain **inert** on Android (absent filter ⇒ render everything, the historical behaviour), so nothing regresses there.

Tracked as #2977, which also records the #2959 consequence above so it isn't lost.

`ktfmtCheckAll`, `:gradle-plugin:test` and `:gradle-plugin:preview-discovery:test` all still pass.

---

Issue #2954, opened 2026-07-29
rc-embedded: AGSL shaders need a platform seam for the jvm/desktop target


yschimke:
Tracking the one part of the embedded player's paint path that **cannot** become a single multiplatform implementation. Punted deliberately; everything around it is being made single-impl.

## What's Android-only

`RcPlayerPaint.kt`, the `PaintBundle.SHADER` path:

- **`android.graphics.RuntimeShader`** — AGSL. Desktop Compose exposes SkSL `RuntimeEffect` through skiko: same idea, different shading language *and* different uniform-binding API. Not a swap.
- **`android.graphics.BitmapShader`** (line ~182) — binding a bitmap uniform via `shader.setInputShader(name, …)`. Only reachable from inside `buildRuntimeShader`, so it lives or dies with AGSL. (The *other* `BitmapShader`, on the `PaintBundle.TEXTURE` path, is already gone — replaced by multiplatform `ImageShader` over an `ImageBitmap`.)
- **`android.graphics.Matrix`** — one call, `shader.setLocalMatrix(...)` in `applyShaderMatrix` for `PaintBundle.SHADER_MATRIX`. Compose exposes no portable shader local-matrix API.
- **`android.os.Build`** — just the API-33 guard on `RuntimeShader`.

## Suggested shape

One narrow seam rather than a parallel implementation:

```kotlin
// returns the *multiplatform* androidx.compose.ui.graphics.Shader
fun buildRuntimeShader(shaderId: Int, context: RemoteContext): Shader?
fun applyShaderMatrix(paintState: ComposeLocalPaint, matrixWord: Int, context: RemoteContext)
```

Android keeps today's `RuntimeShader` implementation; desktop gets a skiko `RuntimeEffect` one. Everything else in the paint path — brushes, tile modes, blend modes, colour filters, images — stays single-impl and shared.

Note `androidx.compose.ui.graphics.Shader` is already a typealias for `android.graphics.Shader` on Android, so the seam's *signature* is portable even though its body isn't.

## Why this is not urgent

Shaders are one of three deliberate deferrals for the desktop lane (with particles and downloadable fonts) — see the sequencing in `third_party/rc-embedded-player/PROVENANCE.md`. Nothing else in the port is blocked on it.

[...]

PR #2982, opened 2026-07-29
feat(rc-embedded): implement the canvas text seam's jvm half over skiko

yschimke:
#2958 declared four canvas text functions for Android and left the jvm side owing a skiko replication. This writes it.

## What had to move first

The four took `ComposeLocalPaint` — which stays Android-coupled through its framework `Shader` (the AGSL path, #2954) — so **no jvm sibling could have compiled against them**. They now take a `TextPaintSpec`: the six fields the text ops actually read (size, family id, weight, slant, whether a typeface was set, alpha-folded ARGB), projected out of the paint state. That plus `TextInkBounds` lives in a new shared, platform-neutral file, so both halves name the same values and neither signature names a platform.

**Android is a pure refactor** — the same framework `Paint` built from the same values, so text output is unchanged. One thing got *stronger*: the two call sites that measure **and** draw (`DrawTextAnchored`, `DrawTextOnCircle`) now build one spec and hand it to both, so a measure/draw face mismatch is impossible per call rather than merely unlikely.

## The jvm half

**Everything goes through the shaper, not `SkFont` directly** — the single decision the rest of the file turns on. `Font.measureText` / `Canvas.drawString` are the obvious one-line counterparts to `Paint.getTextBounds` / `Canvas.drawText`, and they are correct only for plain Latin: they map code points to glyphs in one typeface with no shaping, so kerning and ligatures are skipped, Arabic and Indic come out unjoined, RTL is not reordered, and anything the face lacks becomes a missing-glyph box instead of falling back. Android's `Canvas.drawText` does all of it via Minikin, so the direct calls would not be a *metrics* difference of the kind `PROVENANCE.md` records as a parity limit — they would be visibly wrong text. (This PR's first revision used them; see the review thread.)

| | how |
|---|---|
| `measureTextInkBounds` | `TextBlob.tightBounds` of the shaped blob — the union of positioned glyph outlines, the same construction as `Paint.getTextBounds`. Android rounds out to whole pixels (integer `Rect` out-param); this does not, so bounds are fractionally tighter. |
| `measureTextWidth` | `TextLine.width` — the shaped advance |
| `drawTextAtOriginPlatform` | `drawTextBlob` of the same shaped blob the measurement comes from |
| `drawTextOnPathPlatform` | no single call exists; built on Skia's own primitive for it |

Shaping is `Shaper.make(FontMgr)` — HarfBuzz, ICU bidi, fallback through the font manager. One wrinkle worth knowing: a shaped blob's origin is **not** its baseline (`shape` puts the first baseline an ascent below the offset it is given), so both sides correct by `TextBlob.firstBaseline`. That is what keeps measure and draw on the same origin.

Text-on-path is **not** hand-rolled glyph pushing: it builds a `TextBlob` of per-glyph `RSXform`s, which is what the framework assembles internally too — so placement is computed here but drawing is still one Skia call per face. It reproduces the framework's behaviour: each glyph centred half an advance along the path and rotated to the tangent there, `hOffset` along and `vOffset` perpendicular, glyphs past the end dropped, and the run continuing onto the next contour of a multi-contour path.

Font resolution mirrors `EmbeddedPlayerTypefaceResolver` branch for branch, with two divergences, both in `PROVENANCE.md`:

- `google:` is a `FontsContractCompat` download on Android with no JVM equivalent, so the name is tried locally and substituted if absent — a substitution rather than an error.
- Skia has no generic families, so core ids 0–3 map through a candidate list: CSS-style names first (what fontconfig resolves on Linux), concrete faces after, for hosts where those mean nothing.

Two traps that cost debugging rounds, recorded so the next person doesn't pay again: `matchFamilyStyle(null, …)` — the obvious way to ask Skia for "the default face" — returns **null** on Linux, and `Font(null, size)` then maps everything to the missing glyph and measures **zero**. Not a fallback, a silently blank render. And `RunInfo.font` is only borrowed for its callback, so retaining it past the shaper is a use-after-free that surfaces as a native SIGABRT.

## The one place the seam is not fully shaped

`drawTextOnPathPlatform` resolves glyphs per character (with fallback) rather than from a shaped run, so cross-character kerning and joining are not applied along a path. The reason is mechanical, not principled: per-glyph placement needs each glyph's *font* — an `RSXform` blob carries a single face, and a fallback glyph id drawn against the primary face renders unrelated glyphs, worse than a missing one — and the only skiko API exposing per-glyph fonts is the `RunHandler` callback, which **segfaults** (use-after-free inside skiko's own ICU run iterator, reproducible with a minimal handler). Every other shaping entry point works, which is how the other three are shaped. Drawing from a flattened shaped line instead would reintroduce the wrong-face problem. Documented as a limit with the diagnosis attached, for a follow-up once the handler is usable.

## Test plan

`DesktopTextPlatformTest` — 18 tests, rasterizing for real through skiko into an `ImageBitmap` and reading pixels back. Both modules' `check` green, `ktfmtCheckAll` green.

It asserts **relationships, not numbers**: the font stacks differ across the seam, so a pinned width would pin whatever fonts the host has. Three carry most of the weight:

- `inkBoundsPredictWhereTheGlyphsLand` — measured ink bounds predict where drawn glyphs actually land, which is exactly the invariant `DrawTextAnchored` rests on, and the one a face or origin mismatch would break while every other test still passed.
- `shapesAKernedPairTighterThanTheSumOfItsParts` — `AV` must measure narrower than `A` + `V` apart. Exactly equal is the signature of no shaping, so this **fails** against the unshaped implementation.
- `fallsBackToAFaceThatHasTheGlyphs` — a CJK string must measure about an em per ideograph, not the narrower missing-glyph box. Skips itself where the host has no CJK face, since then the assertion would be about fontconfig; CI installs one so it actually runs.

Also covered: advance scaling with text size, ink bounds being tight rather than a layout box, pen/baseline origin semantics, rotation along a vertical path and an arc, `hOffset`/`vOffset`, clipping when the path runs out, continuing onto a second contour, every family id resolving to a usable face, and degenerate inputs (empty text, empty path, zero-length path) drawing nothing without throwing.

**It needs skiko's natives**: the per-OS `skiko-awt-runtime-*` artifact (added as `testRuntimeOnly(compose.desktop.currentOs)`, test-only because a library has no business choosing its host) *and* a loadable GL library, since `libskiko` links one even for raster-only drawing. Where that is missing the class skips **loudly** rather than reporting eighteen cryptic `NoClassDefFoundError`s — and CI fails the build on that, see below.

## CI never ran this module

Not from the review, but found while checking that the tests above actually execute anywhere: `:third-party-rc-embedded-player-jvm` was not merely untested by CI but **entirely unreferenced** — every Gradle test task in `ci.yml` is named per-module and there is no aggregate `check` — so neither its tests nor its *compile* ran anywhere but a laptop. That module exists to make the "platform-neutral" claim falsifiable; a claim nothing on CI evaluates is not one.

Added a leg that runs its `check`, plus a step that reads the JUnit XML and fails if the **class** did not execute — zero tests, or all of them skipped. That is the hazard worth failing on, because the skiko skip above still reports green: a runner image that quietly dropped libGL would otherwise turn this PR's only verification into eighteen no-ops without failing anything. Individual skips warn rather than fail, so the build isn't hostage to the runner image's font list — and the step also installs a GL library and a CJK face when the host lacks either, so the fallback discriminator runs rather than opting out.

No visual evidence, and that is the correct answer rather than a gap: Android's rendering is unchanged by construction, and the jvm half has no `@Preview`-reachable surface — its callers still live in `RcPlayerDrawing.kt`, which needs `Bitmap`/`BitmapDrawable`. The pixels it does produce are asserted on directly.

## What this does not finish

No draw op runs on the JVM yet — the ops calling these four are still blocked on the image-decode seam. The seam is ready ahead of its callers, deliberately: it was the step with an actual unknown in it (*does Skia answer the same four questions?*), so `PROVENANCE.md`'s sequencing step 4 is spent early instead of leaving the only real risk in the plan until last.


yschimke:
Both P1s were right, and one of them made a claim in my own comment false. Fixed in `471fa6b`.

**On shaping and fallback (the `drawString`/`SkFont` P1).** Correct, and the important half of the finding is the part about the tests: measurement and drawing both went through the same unshaped path, so they agreed with each other while both disagreed with Android, and every cross-check in the suite stayed green. A self-consistent seam that is wrong on both sides is exactly the failure the suite was blind to.

Measurement and the origin draw now go through `Shaper.make(FontMgr)` — HarfBuzz, ICU bidi, fallback through the font manager — reading `TextBlob.tightBounds` for ink bounds and `TextLine.width` for the advance. One wrinkle: a shaped blob's origin is *not* its baseline (`shape` puts the first baseline an ascent below the offset it is given), so both sides correct by `TextBlob.firstBaseline`. That is what keeps `inkBoundsPredictWhereTheGlyphsLand` passing, and it is the test that would have caught getting it wrong.

Two tests now **discriminate** rather than tolerate, aimed at precisely what was unobservable before:

- `shapesAKernedPairTighterThanTheSumOfItsParts` — `AV` must measure narrower than `A` + `V` measured apart. Exactly equal is the signature of no shaping, so this fails against `Font.measureTextWidth` and passes against the shaper.
- `fallsBackToAFaceThatHasTheGlyphs` — a CJK string must measure about an em per ideograph, not the much narrower missing-glyph box. Skipped where the host has no CJK face, since then there is nothing to fall back *to* and the assertion would be about fontconfig rather than the code.

**On shaping the glyph run before laying it on the path.** Also right, and this one I could only fix halfway — flagging that rather than implying otherwise.

Per-glyph placement needs each glyph's *font*, because an `RSXform` blob carries a single face and a fallback glyph id drawn against the primary face renders unrelated glyphs — worse than a missing one. The only skiko API exposing per-glyph fonts is the `RunHandler` callback, and that path **segfaults**: a use-after-free inside skiko's own ICU run iterator (`FontMgrRunIterator` → `makeHbIcuScriptRunIterator`), reproducible with a minimal handler and nothing to do with this code. `RunInfo.font` is additionally only borrowed for the callback's duration, so it needs `makeWithSize` to copy — I fixed that first and still crashed. Every other shaping entry point (`shape`, `shapeLine`, with and without a `FontMgr`) works, which is how the other three functions are shaped.

So glyphs on a path are resolved per character *with* fallback (glyph id 0 means the face cannot draw it — the signal to ask the font manager) but without cross-character shaping. Placing from a flattened shaped line instead would reintroduce the wrong-face problem, which is the worse of the two failures. Recorded in `PROVENANCE.md` as a limit with the diagnosis attached, alongside the existing metrics and downloadable-font limits, for a follow-up once skiko's handler is usable. Curved text is where it matters least — per-glyph rotation dominates sub-pixel kerning — but it is a real gap, not a non-issue.

Separately, and not from the review: CI never ran this module at all — it was unreferenced, so neither its tests nor its compile ran anywhere but a laptop. `3242571` adds the leg, plus a step that fails the build if the skiko tests *skip*, since a skip still reports green.


---

PR #2985, opened 2026-07-30
fix(preview-diff): guard preview/resource removals against partial renders

### LINKED ISSUE #2949: fix(preview-diff): a partial render is reported as mass "Removed", not as a failure
Seen on #2945: the preview-diff bot commented **"Removed (315 variants)"** on a PR that deleted no previews at all. The same job's log said `Copied 0 changed/new preview(s)` — zero pixels differed. The change was a behaviour fix in `rc-embedded-player` plus the deletion of a dead `autoUpdate` parameter; nothing renderable was removed.

**Not reproduced.** The PR merged before a rebased run finished, so there is exactly one observation. Filing it anyway because the failure mode is bad regardless of trigger: a run that renders *less* than the baseline is indistinguishable, in the comment, from a PR that deliberately deletes 315 previews. A reviewer either believes it and blocks a correct PR, or learns to ignore the Removed section — both bad.

## Where it comes from

`removed` is a pure set difference: [`.github/actions/lib/compare-previews.py:1056-1059`](.github/actions/lib/compare-previews.py#L1056-L1059) — every baseline key absent from `current` is a removal. There's no notion of "absent because it wasn't rendered".

The composable path guards exactly one cause of that: module-level scoping, via `--scope-modules` / `_baseline_module` (:1049-1055, docstring at :931-941). Any *other* reason an in-scope preview is missing from `current` — a truncated render, a cancelled or partially-uploaded job, a manifest written before the render finished — falls straight through to "removed". `Copied 0 changed/new` alongside 315 removals is consistent with the render simply not having produced those entries.

**Concrete defect found while writing this up:** the resource comparison, `cmd_compare_resources` at [:1605-1607](.github/actions/lib/compare-previews.py#L1605-L1607), has **no scope guard whatsoever** — not even the module one. Its `removed` loop is the bare `if key not in current`. So on any change-scoped run, every out-of-scope resource baseline is reported as removed. That one doesn't need the #2945 mystery solved to be worth fixing; it's the composable-path bug that :931-941 already documents, left unfixed in the sibling code path.

## What to do

1. **Fix the resource path** — give `cmd_compare_resources` the same `--scope-modules` / `_baseline_module` filtering the composable path has. Straightforward, independently verifiable.
2. **Sanity-check the removal set before reporting it.** A removal count that's a large fraction of the baseline is far more likely a partial render than a real deletion. Either cross-check against the PR's actual deleted files, or refuse to render the section and fail loudly instead — "expected N previews for the in-scope modules, got M" is a useful CI failure; "Removed (315 variants)" on a PR that deleted nothing is not.
3. **Make the render assert its own completeness**, so the manifest can't be consumed as authoritative when it's short. That's the root fix — 1 and 2 are containment.

## Acceptance

- A deliberately truncated render (drop entries from the CLI JSON, keep the modules in scope) produces a failure or an explicit incompleteness warning, never a Removed section.
- A change-scoped resource run reports zero removals for out-of-scope resources — the case that's broken today.
- A genuine preview deletion still reports correctly under both paths.
- Both covered by the comparison script's existing tests.

Related: #2945 (where it was observed).

yschimke:
Closes #2949.

## Problem

`removed` in the preview-diff comparator is a pure set difference: every baseline key absent from the current render is reported as a deletion. A run that renders *less* than the baseline — a change-scoped render, a truncated/cancelled job, a manifest written before the render finished — is then indistinguishable, in the PR comment, from a PR that deliberately deleted that many previews. This is what produced `Removed (315 variants)` on #2945, a PR that deleted nothing (`Copied 0 changed/new preview(s)` in the same run).

Two concrete defects:

1. **The resource path had no scope guard at all.** `cmd_compare_resources`' removal loop was the bare `if key not in current`, so on any change-scoped run *every* out-of-scope resource baseline was reported as removed — the composable-path bug already documented at `_parse_scope_modules`, left unfixed in the sibling code path.
2. **Neither path sanity-checked the removal set**, so a partial render surfaced as a mass "Removed" section rather than a completeness failure.

## Changes

- **Resource scope filtering.** `cmd_compare_resources` now takes the same `--scope-modules` / module filtering the composable path has. The module comes from the baseline's `module` field, or is parsed from the `<module>::<id>::<basename>` key when the field predates it (`_resource_baseline_module`). Wired through `apply/action.yml`'s resource compare step.
- **Partial-render guard (both paths).** When the removal set is a large fraction of the in-scope baseline (≥50%, above an absolute floor so tiny modules deleting their only previews aren't caught), the misleading `### Removed` section is suppressed and replaced with a loud `> [!CAUTION]` incompleteness warning ("Expected N in-scope previews… got M; K are missing… **not** reported as Removed"). The warning always posts — it never collapses to the no-op sentinel.

## Acceptance criteria (all covered by `test_compare_previews.py`)

- ✅ A truncated render (fewer in-scope entries than the baseline) produces an explicit incompleteness warning, never a Removed section — `test_truncated_render_warns_instead_of_removing`, `test_truncated_resource_render_warns_instead_of_removing`.
- ✅ A change-scoped resource run reports zero removals for out-of-scope resources — `test_out_of_scope_resource_baselines_are_not_removed`, `test_out_of_scope_module_parsed_from_key_when_field_absent`.
- ✅ A genuine deletion still reports correctly under both paths — `test_genuine_single_deletion_still_reports_removed`, `test_in_scope_resource_deletion_still_reported`, plus a small-baseline floor check (`test_small_baseline_does_not_trip_the_guard`).

## Scope note

This implements the two containment fixes from the issue (resource scope guard + removal sanity-check). Item 3 — making the render assert its own manifest completeness at the source (CLI/gradle plugin) — is the deeper root fix and is left for a follow-up; the comparator-level guard here catches the same failure at report time, which is what the acceptance criteria specify.

## Test plan

- `python3 -m unittest test_compare_previews` → 98 tests OK.
- Full `.github/actions/lib` suite → 164 tests OK.

Non-visual change (CI comparison-script logic + tests), so no rendered before/after evidence applies.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WpLcKjJsRVnNxSMMUsq5uu)_

### DIFF
--- a/.github/actions/apply/action.yml
+++ b/.github/actions/apply/action.yml
@@ -842,6 +842,7 @@ runs:
         REPO: ${{ github.repository }}
         BASELINE_BRANCH: 'compose-preview/main'
         RESOURCE_HEAD_BRANCH: 'compose-preview/resources/pr'
+        SCOPE_MODULES: ${{ steps.scope.outputs.modules }}
       run: |
         set -euo pipefail
         if [ ! -f _resources.json ]; then
@@ -856,6 +857,14 @@ runs:
         [ -z "$RESOURCE_BASE_REF" ] && RESOURCE_BASE_REF="$BASE_REF"
         RESOURCE_HEAD_REF=$(cat _resource_head_sha 2>/dev/null || echo "")
         [ -z "$RESOURCE_HEAD_REF" ] && RESOURCE_HEAD_REF="$RESOURCE_HEAD_BRANCH"
+
+        # Change-scoped runs rendered a module subset — tell the comparator so
+        # out-of-scope resource baselines read as unchanged, not removed.
+        SCOPE_ARGS=()
+        if [ -n "${SCOPE_MODULES:-}" ] && [ "$SCOPE_MODULES" != "full" ]; then
+          SCOPE_ARGS=(--scope-modules "$SCOPE_MODULES")
+        fi
+
         python3 "$ACTION_PATH/../lib/compare-previews.py" compare-resources \
           _resources.json \
           --baselines _baselines/resource-baselines.json \
@@ -863,6 +872,7 @@ runs:
           --repo "$REPO" \
           --base-ref "$RESOURCE_BASE_REF" \
           --head-ref "$RESOURCE_HEAD_REF" \
+          "${SCOPE_ARGS[@]}" \
           > _comment_body_resources.md
 
     - name: Post or update compose PR comment
--- a/.github/actions/lib/compare-previews.py
+++ b/.github/actions/lib/compare-previews.py
@@ -1009,18 +1009,21 @@ def _suspected_partial_render(removed_count: int, in_scope_total: int) -> bool:
 def _partial_render_warning(
     removed_count: int, in_scope_total: int, kind: str = "preview"
 ) -> str:
-    """The CAUTION block that replaces the Removed section when a run looks
-    truncated. Frames the shortfall as a completeness problem, not a deletion."""
+    """The CAUTION block that reframes the Removed section when a run looks
+    truncated. Frames the shortfall as a completeness question, not confirmed
+    deletions — but the missing entries are still listed (unverified) below it,
+    so a genuine large deletion is never permanently hidden."""
     rendered = in_scope_total - removed_count
     return (
         "> [!CAUTION]\n"
         f"> Expected {in_scope_total} in-scope {kind}(s) from the baseline but "
         f"this run produced only {rendered}; {removed_count} are missing. A render "
-        f"that emits far fewer entries than the baseline is far more likely "
-        f"truncated or incomplete than a PR that deliberately deleted "
-        f"{removed_count} {kind}(s), so they are **not** reported as Removed. "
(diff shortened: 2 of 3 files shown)


yschimke:
Good catch on the false-hidden case. You're right that a *complete* render legitimately deleting ≥50% of an in-scope module's baseline can't be told apart from a truncated render by set-difference size alone — that discrimination needs an explicit render/manifest completeness signal, which is issue #2949's item 3 (the deeper root fix, deliberately out of scope for this containment PR).

What I've changed in b5f0cf9 so this PR no longer *hides* those removals: when the guard trips, the missing entries are no longer dropped — they're listed under the caution as **"Missing this run — unverified"** (a `<details>` block, not the confirmed `### Removed` heading a reviewer has learned to trust). The warning wording is also fixed to stop falsely promising a rerun will bring them back: it now says *"if this run was truncated, re-run and they'll return; if the render was complete, they are genuine deletions."* So a reviewer always sees exactly what went missing and can make the call, while a truncated run no longer reads as `Removed (315 variants)`.

The explicit-signal approach you suggest is the right long-term fix and is tracked as item 3 on #2949.

---

PR #2984, opened 2026-07-30
docs: design proposal for a Compose playground over the preview server

yschimke:
## Summary

A design proposal (`docs/design/PLAYGROUND.md`) for a hosted Kotlin/Compose **playground** built on top of the `serve` preview host, rather than as a new stack.

The core idea is a **two-stage, compile-then-permalink** flow that keeps a stock one-shot editor frontend usable while still delivering a live, interactive preview:

- **Stage 1 (playground)** — the editor POSTs source to a `kotlin-compiler-server`-compatible `/run` shim. We compile via the shipped `BtaCompileSession` (no Gradle), return diagnostics + a first-frame PNG + an expiring **preview token**.
- **Stage 2 (preview)** — an "Open live preview →" link redeems `/pg/<token>`, standing up (or resuming) the existing daemon-backed streaming session with the existing `input` protocol (click / pointer / rotary / key).

This split shrinks the frontend change from *rewrite the transport* to *surface one additive field*, and makes session creation a deliberate, rate-limitable, TTL'd act instead of a per-keystroke one.

### Mode matrix

| Mode | Stage-2 target | Per-interaction cost | Seat cost |
|---|---|---|---|
| Compose (CMP) | live streaming session (desktop/Skiko daemon) | server re-render | 1 |
| Compose (Android) | live streaming session (Robolectric daemon) | server re-render | 2 |
| Remote Compose | `/d/<id>` document permalink, played client-side | **none** | 0 |

Remote Compose is the standout — its serving side is already public-safe, and the browser `RcdPlayer` is platform-neutral, so RC-in-browser works now even though CMP *authoring* of RC is future.

### What already exists vs. what's new

Most of the hard parts are shipped: in-process compile (`BtaCompileSession` / `compileSources`), headless render (both backends), the live streaming session, the `input` protocol, and the expiring-permalink capability (`ServeDocStore`, the literal template for the preview-token store). New work is the `PlaygroundSession`, the REST shim, the preview-token store + `/pg/<token>` route, and — the real gate to going public — per-session **isolation** (the doc is explicit that running a stranger's code inverts the serve host's "never run untrusted code on the server" invariant, and that `--live-seats` is admission control, not a sandbox).

### Phased plan

1. Token-gated, CMP, single snippet (**in progress**).
2. Android backend (Robolectric).
3. Remote Compose mode (reuses `ServeDocStore` / `rc-player`; first mode safe for a wider gated audience).
4. Per-session sandbox — the gate to `--public`.

[...]

PR #2989, opened 2026-07-30
feat(serve): playground preview-token store and REST DTOs

yschimke:
## Summary

First implementation slice for the Compose playground designed in [`docs/design/PLAYGROUND.md`](https://github.com/yschimke/compose-ai-tools/blob/main/docs/design/PLAYGROUND.md) (merged in #2984). Two self-contained, unit-tested foundation pieces — no route wiring yet, so nothing is reachable from a running server until the follow-ups land.

- **`PlaygroundApi.kt`** — the Stage-1 run request/response wire types. A deliberate **superset** of the `kotlin-compiler-server` `/compiler/run` contract so a stock `kotlin-playground` frontend can POST to us unmodified (`{ args, files, confType }` → `{ text, exception, diagnostics }`), plus the two **additive** handoff fields it ignores and ours surfaces: `image` (first-frame `data:` PNG) and `previewToken`. Includes `PlaygroundMode` (CMP / Compose-Android / Remote-Compose) with `confType` resolution that tolerates the stock frontend's target ids.
- **`PlaygroundTokenStore.kt`** — the expiring `/pg/<token>` capability that hands a compiled snippet off to a live session, modelled directly on `ServeDocStore`: 128-bit `SecureRandom` `pg_`-prefixed ids, TTL expiry, a count cap with nearest-expiry eviction. Unlike the doc store it owns **on-disk** state, so every drop path (expiry, overflow, `remove`, `clear`) deletes the snippet's temp work dir through an **injected Okio `FileSystem`** (per the repo's Okio-everywhere rule). Best-effort delete so one undeletable dir can't wedge a purge.

### Where this sits in the plan

Phase 1 (token-gated, CMP, single snippet) from the design doc. Still to come as follow-ups: the `/api/{v}/compiler/run` route + compile handler (over the shipped `BtaCompileSession`), the first-frame render, and the `/pg/<token>` redemption into the existing live streaming session. Isolation (Phase 4) remains the gate to any `--public` exposure and is untouched here.

## Test plan

- [x] `./gradlew :cli:test --tests "ee.schimke.composeai.cli.serve.PlaygroundTokenStoreTest"` — green (6 cases: redeem-until-expiry, mode/preview-id round-trip, overflow eviction + work-dir cleanup, explicit remove, clear, confType mapping). Work-dir deletion is asserted against a `FakeFileSystem`.
- Non-visual change (server-side data types + capability store); no rendered surface touched, so no visual evidence applies.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundApi.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundApi.kt
@@ -1,0 +1,108 @@
+package ee.schimke.composeai.cli.serve
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * Wire types for the **playground** REST surface — the Stage-1 "compile a snippet, get a result +
+ * an expiring preview token" contract described in
+ * [docs/design/PLAYGROUND.md](../../../../../../../../docs/design/PLAYGROUND.md).
+ *
+ * The request/response shapes are deliberately a **superset** of the `kotlin-compiler-server`
+ * `/api/{version}/compiler/run` contract that a stock `kotlin-playground` frontend speaks, so an
+ * unmodified editor can POST to us: it sends `{ args, files, confType }` and reads `{ text,
+ * exception, errors }`. The two fields it does **not** know about — [PlaygroundRunResponse.image]
+ * and [PlaygroundRunResponse.previewToken] — are additive; a stock frontend ignores them and ours
+ * surfaces them as the still frame + the "Open live preview →" handoff.
+ *
+ * These are pure data types with no server behaviour; the route handler and the compile/render
+ * plumbing live elsewhere.
+ */
+
+/** Which renderer + permalink target a snippet compiles for. See PLAYGROUND.md §3. */
+@Serializable
+enum class PlaygroundMode {
+  /** Compose Multiplatform, desktop (Skiko) daemon → live streaming session. */
+  @SerialName("compose-cmp") CMP,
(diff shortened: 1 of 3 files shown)


yschimke:
Re: the P1 on `PlaygroundApi.kt:102` (serialize diagnostics under `errors`) — good catch on the mismatch with the doc, but the fix is bigger than a field rename, so I'm handling it in the route-handler PR rather than here.

The reason a straight `diagnostics` → `errors` rename (or a both-fields alias) wouldn't actually work: a stock `kotlin-playground` frontend doesn't read `errors` as a flat list. Upstream `errors` is a **map keyed by filename** — `{ "File.kt": [ { interval: { start: {line, ch}, end: {line, ch} }, message, severity, className } ] }` — and the frontend iterates it per file and reads `error.interval.start`. My `PlaygroundDiagnostic` is a flat object with top-level `line`/`ch`. So renaming the field would *look* stock-compatible while still rendering nothing in the stock editor — worse than an honestly-named internal field.

So the real shape:
- These DTOs are our **own** internal response shape (`diagnostics` is fine there).
- True stock compatibility needs an **adapter at the route boundary** that projects our diagnostics into upstream's `errors` map + `interval` nesting. Per `docs/design/PLAYGROUND.md` §4, hosting a stock frontend is explicitly deferred, and that adapter is where it belongs.

What I'll do in the `/api/{v}/compiler/run` handler PR: add the `errors`-map projection alongside our `diagnostics`, and tighten the §4 line that currently says the response "reuses the upstream shape `{errors, …}`" — that wording is what's actually imprecise (it conflates our response with the stock wire format), and it's the source of the mismatch you flagged. Leaving this foundation PR as-is since the DTOs aren't wired to any route yet, so nothing renders either way.

---

PR #2990, opened 2026-07-30
fix(daemon): degrade missing drawables read via getDrawableForDensity to a placeholder

### LINKED ISSUE #2976: render failure
https://preview.coo.ee/pocketcasts-wear/p/playback-marquee-media-display__ideal__default__compact

render failed: NotFoundException: Drawable au.com.shiftyjelly.pocketcasts.debug:drawable/splash_background with resource ID #0x7f0806b5

yschimke:
## Summary

Closes #2976.

The live serve daemon rendered `pocketcasts-wear/playback-marquee-media-display` with:

```
render failed: NotFoundException: Drawable au.com.shiftyjelly.pocketcasts.debug:drawable/splash_background with resource ID #0x7f0806b5
```

`PlaceholderFallbackResources` is meant to turn a **missing** app-resource lookup into an obvious magenta placeholder rather than letting `Resources.NotFoundException` abort the whole preview render. It already guards the string/color/dimension families and `getDrawable(id)` / `getDrawable(id, theme)` — but **not** the density-aware `getDrawableForDensity(...)` overloads.

That exact `"Drawable <pkg>:drawable/<name> with resource ID #0x…"` message is produced only by `ResourcesImpl.loadDrawable`. The plain `getDrawable(...)` overloads funnel into `loadDrawable` (and were caught), but a caller that reaches `getDrawableForDensity(...)` **directly** — density-aware image loaders, `ResourcesCompat.getDrawableForDensity`, RemoteViews / `ImageView` inflation — enters `loadDrawable` without passing through the guarded methods, so the miss still threw and aborted the render even though the fallback was enabled (serve sets `composeai.render.placeholderMissingResources=true`).

## Change

- Override both `getDrawableForDensity(id, density)` and `getDrawableForDensity(id, density, theme)` in `PlaceholderFallbackResources` to catch `NotFoundException` and return the same magenta placeholder, so **all four** `loadDrawable` entry points degrade gracefully.
- Update the class doc's "not (yet) covered" note: drawables are now guarded on every `loadDrawable` path; only raw `getValue` / `openRawResource` decode paths (which can't be substituted at the `Resources` seam without fabricating a `TypedValue` / stream) remain a theoretical gap.
- Add unit coverage for the density-aware misses in `PlaceholderFallbackResourcesTest`.

## Test plan

- `./gradlew :daemon:android:testDebugUnitTest --tests "ee.schimke.composeai.daemon.PlaceholderFallbackResourcesTest"` — **BUILD SUCCESSFUL** (new density-variant tests pass alongside the existing coverage).
- `./gradlew :daemon:android:ktfmtCheck` — **BUILD SUCCESSFUL**.

Non-visual change (resource-fallback plumbing in the render engine), so no rendered before/after evidence applies.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WTUqnZfpeTGS82tWZFpfQc)_

### DIFF
--- a/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/PlaceholderFallbackResources.kt
+++ b/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/PlaceholderFallbackResources.kt
@@ -35,11 +35,18 @@ import android.graphics.drawable.Drawable
  * a later `String.format` with args leaves it unchanged. Value resources ([getColor], the dimension
  * family, [getDrawable]) don't route through a shared accessor, so each is guarded directly. The
  * base class's `AssetManager` / `DisplayMetrics` / `Configuration` are reused, so every *resolvable*
- * resource still takes the normal Android path.
+ * resource still takes the normal Android path. Drawables are guarded on **all four**
+ * `loadDrawable` entry points — `getDrawable(id)`, `getDrawable(id, theme)`, and both
+ * `getDrawableForDensity(...)` overloads — because a density-aware caller
+ * (`ResourcesCompat.getDrawableForDensity`, RemoteViews / ImageView inflation, image loaders)
+ * reaches the throwing `ResourcesImpl.loadDrawable` without funnelling through the plain
+ * `getDrawable`, so guarding only the latter left the miss to abort the render (issue #2976).
  *
- * Not (yet) covered: `getValue`-based paths such as Compose `painterResource(...)` — a missing
- * drawable id read that way still throws. Tracked as a follow-up; the string family is the common
- * crash the live server hits (`wear-m3` stickers use `stringResource`).
+ * Not (yet) covered: raw `getValue` / `openRawResource` decode paths (e.g. a bitmap whose id
+ * resolves but whose backing file is absent) can't be substituted at the `Resources` seam without
+ * fabricating a `TypedValue` / stream, so a miss read that way still throws. The value families
+ * above (`getText`, `getColor`, the dimension family) and every `getDrawable*` overload are the
+ * common crashes the live server hits.
  */
 @Suppress("DEPRECATION")
 internal class PlaceholderFallbackResources(base: Resources) :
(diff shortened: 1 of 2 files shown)


yschimke:
should we also fix the issue so the resource is found as well?

yschimke:
Good question — I'd keep them as separate concerns, and here's the reasoning.

The bundle already packs the full merged resource APK (`android/resources.ap_`), which carries both the resource table **and** the compiled drawable files. That's consistent with the error: `getResourceName(0x7f0806b5)` resolved to `…:drawable/splash_background`, so the id *is* in the packed table. So this isn't the classic "resource table is absent" case that `stringResource` misses hit — the entry exists, but `ResourcesImpl.loadDrawable` still couldn't materialize a `Drawable` from it.

That points at **rendering** the drawable, not finding it. `splash_background` is a splash/windowBackground drawable, and those are commonly a `<layer-list>` / theme-attr / 9-patch / raster that headless Robolectric's graphics path can't decode, or that references a sub-resource. "Make it actually load" would therefore mean reproducing against the real `pocketcasts-wear` bundle and figuring out *why* that specific drawable is undrawable — and it may turn out to be genuinely un-renderable here rather than a packing bug. That's a bigger, bundle-specific investigation, and the bundle isn't in this repo.

Meanwhile, the graceful-degradation fix in this PR is the right thing at the render layer regardless of the outcome of that investigation: a preview shouldn't hard-fail because of an incidental app resource (a splash background) it doesn't actually care about — the magenta placeholder keeps the *rest* of the UI visible and legible.

So my recommendation: land this as-is to stop the render abort, and I'm happy to open a follow-up issue for "resolve `splash_background` for real" if you can point me at the `pocketcasts-wear` bundle (or the design-catalog entry that produces it) so I can reproduce and pin down whether it's a packing gap or a Robolectric-decode limit. Want me to file that follow-up?

---

yschimke:
Is that ci builds failure ours?

PR #2994, opened 2026-07-30
feat(plugin): forward preview filters to the Android/Robolectric render path

### LINKED ISSUE #2977: plugin: preview filters (name + id) don't reach the Android/Robolectric render path
Found while reviewing #2973. Affects the pre-existing #2066 name filter as well as the new id filters, and it silently scopes the render-time saving #2959 wired into the design-artifacts workflows.

## The gap

`composePreviewRender` is registered **twice**, by backend:

- **desktop/JVM** — `ComposePreviewTasks.kt:241`, a `RenderPreviewsTask`. This is where `previewFilters` (#2066) and, per #2973, `previewIdFilters` / `previewIdExcludes` (#2966) are wired and applied.
- **Android** — `AndroidPreviewSupport.kt:1903`, a Robolectric `Test` task. It reads **none** of them.

`previewFilterProperty` has exactly one call site (`ComposePreviewTasks.kt:259`), the desktop registration, so:

- `--preview` / `-PcomposePreview.filter` narrows a desktop render only;
- `--preview-id` / `--exclude-preview-id` likewise;
- an Android module renders every discovered preview regardless.

They're **inert** rather than wrong — an absent filter means "render everything", the historical behaviour — so nothing is broken. But the KDoc implied otherwise, and it means a documented saving doesn't land where several catalogs actually live.

## Consequences worth naming

- **#2959's workflow wiring is desktop-only.** Both design-artifacts workflows now export `ORG_GRADLE_PROJECT_composePreview.filter` from the pre-flight's `--render-filter-out`. For an Android catalog (`design-catalog-wear-m3`, `design-catalog-remote-m3`, pocket-casts) that property is read by nothing, so entry-level deferral would trim the published bundle without trimming the build. Harmless today — entry deferral is inert pending #2965 — but it should not stay silent once that flips.
- **#2966's axis saving is desktop-only too**, and the motivating measurement (`pocket-casts-android`, nine `@PreviewParameter` themes) is an Android module. So the −59% needs this issue, not just #2966.

## What's needed

1. Forward the filters to the Robolectric render. The Test task drives the render inside the test JVM, so the patterns have to reach the manifest loader there — most likely as system properties (`composeai.preview.filter` / `.idFilter` / `.idExclude`) applied where the Android path reads `previews.json`, reusing `PreviewNameFilter.matches` / `matchesId` so the semantics can't drift from the desktop path.
2. Expose the same `--preview` / `--preview-id` / `--exclude-preview-id` CLI options on the Android task, so `:app:composePreviewRender --preview Foo` behaves the same on both backends (today `help --task composePreviewRender` shows them only for desktop modules).
3. Mirror the cache gating: the Test task must not store a build-cache entry for a filtered run, for the same reason `RenderPreviewsTask.outputs.cacheIf` refuses one — `outputDir` is then not the module's complete render set.
4. Decide the same question for the sibling Android render tasks (`composePreviewRenderAndroidResources`, `composePreviewRenderXr`), which have the same shape.
5. `bundle pack --with-semantics` starts the daemon and captures per preview (`packSemanticsBlob`); that pass needs the same filtering or the saving stays partial on either backend. Tracked on #2966 too — worth doing together.

## Meanwhile

#2973 documents the desktop-only scope on the task properties and both property resolvers, so the limitation is visible at the point of use rather than discovered from a render bill that didn't drop.

yschimke:
## The gap (#2977)

`composePreviewRender` is registered twice — a desktop `RenderPreviewsTask` and an Android Robolectric `Test` task. Only the **desktop** one applied the `--preview` name filter (#2066) and `--preview-id` / `--exclude-preview-id` id filters (#2966). The Android task read **none** of them, so a filter was *inert* on the backend where several design catalogs (`design-catalog-wear-m3`, `design-catalog-remote-m3`, pocket-casts) actually live — stranding the render-time saving #2959 wired into the design-artifacts workflows, and the −59% axis saving #2966 measured on the Android `pocket-casts-android` module.

## What this does

Forwards all three filters to every Android render as `composeai.preview.*` system properties and applies them where the manifest is read, sharing **one** matcher across backends so the semantics can't drift.

- **`PreviewFilter` (`:data-render-core`)** — a renderer-side mirror of the plugin's `PreviewNameFilter` (identical glob / substring / FQN rules, plus the compose + fail-fast policy). Applied in:
  - `PreviewManifestLoader.loadShard` (the image render) — filters the discovered entries before `@PreviewParameter` expansion; the **full** manifest is still used for stale-fan-out protection, so a filtered-out preview keeps its PNG on disk exactly as on desktop.
  - `XrSubspaceRenderTest` (XR subspace render) — XR previews are ordinary `@Preview` composables, full name/id parity.
  - `ResourcePreviewRenderTest` (XML resources) — a resource has no `@Preview` function name, so all three match against its `<type>/<name>` id.
- **`composePreviewRender` is now a `RobolectricRenderTask`** (a `Test` subclass) carrying the `--preview` / `--preview-id` / `--exclude-preview-id` options, so `:app:composePreviewRender --preview Foo` behaves the same on both backends. The sibling XR / resource tasks honour the `composePreview.*` property conventions (what the design-artifacts wiring and CLI already set) without their own CLI options.
- **Build-cache gating** — a filtered run is partial, so it no longer stores a build-cache entry, mirroring the desktop `RenderPreviewsTask.cacheIf`.
- **Docs** — refreshes the "inert on Android" KDoc that #2973 added on both property resolvers and `RenderPreviewsTask`.

### Issue checklist

1. ✅ Forward the filters to the Robolectric render (system properties, reusing the shared matcher).
2. ✅ Expose `--preview` / `--preview-id` / `--exclude-preview-id` on the Android task.
3. ✅ Mirror the cache gating (no cache entry for a filtered run).
4. ✅ Sibling tasks: `composePreviewRenderXr` (full parity) and `composePreviewRenderAndroidResources` (id-based, name filter matches the resource id) both honour the property conventions.
5. ✅ `bundle pack --with-semantics` already scopes its per-preview capture via `PackPreviewIdExclusions.retain(...)` — so with this change both halves of the axis saving land on both backends.

## Test plan

- [x] `:data-render-core:test` — new `PreviewFilterTest` (glob/substring/FQN, compose, fail-fast, sysprop parsing).
- [x] `:gradle-plugin:test` — new `PreviewFilterSystemPropsProviderTest` pins the `-Dcomposeai.preview.*` wire format.
- [x] `:renderer-android` / `:renderer-xr` compile; `ktfmtCheck` clean on all touched modules.
- [x] **End-to-end on the Android sample:** `:samples:android:composePreviewRender --preview BlueBoxPreview` executes exactly **1 of 120** previews (`BlueBoxPreview_Blue Box`) and `BUILD SUCCESSFUL`; an unfiltered run schedules all 120. This is a render-selection/build-wiring change with no pixel output change, so no visual before/after applies.

Closes #2977

---
_Generated by [Claude Code](https://claude.ai/code/session_012rwS5pf4TEGFJZkPn8Bms9)_

### DIFF
--- a/data/render/core/src/main/kotlin/ee/schimke/composeai/data/render/PreviewFilter.kt
+++ b/data/render/core/src/main/kotlin/ee/schimke/composeai/data/render/PreviewFilter.kt
@@ -1,11 +1,13 @@
 package ee.schimke.composeai.data.render
 
 /**
- * Renderer-side `@Preview` selector shared by every Robolectric render entry — the image render
- * ([ee.schimke.composeai.renderer.PreviewManifestLoader]), the XR subspace render, and the XML
- * resource render — so `--preview` / `--preview-id` / `--exclude-preview-id` (and their
+ * Renderer-side `@Preview` selector shared by the Robolectric render entries that render actual
+ * `@Preview` composables — the image render ([ee.schimke.composeai.renderer.PreviewManifestLoader])
+ * and the XR subspace render — so `--preview` / `--preview-id` / `--exclude-preview-id` (and their
  * `composePreview.filter` / `.idFilter` / `.idExclude` property conventions) narrow an Android
- * render the same way they already narrow the desktop one (issues #2066, #2966, #2977).
+ * render the same way they already narrow the desktop one (issues #2066, #2966, #2977). The XML
+ * resource render is a separate manifest of assets with no `@Preview` function and is deliberately
+ * left unfiltered — see `ResourcePreviewRenderTest`.
  *
  * The matching rules ([matches] / [matchesId] / [globToRegex]) are a **deliberate mirror** of the
  * plugin's `ee.schimke.composeai.discovery.PreviewNameFilter`, which drives the desktop
@@ -85,8 +87,16 @@ object PreviewFilter {
    * produce zero output is a typo or stale spec, not a silent no-op (the failure the desktop path
    * guards against too). Empty filter lists pass [items] through unchanged.
    *
+   * [failOnNoMatch] gates that throw. `true` (the default) is for the authoritative render that
+   * sees every discovered preview — the Android image render, whose manifest carries all kinds
+   * (COMPOSE / XR / LOTTIE / SVG / catalog), so a global no-match really is a typo. `false` is for
+   * a **kind-restricted sibling** view — the XR-only render — where a filter naming a preview of a
+   * different kind legitimately matches nothing *here* while the image render matches it; failing
+   * would sink `composePreviewRenderAll`. There, a no-match yields an empty list and the sibling
+   * simply renders nothing.
+   *
    * Type-agnostic via the accessor lambdas so the image-render [ee.schimke.composeai.renderer]
-   * entry, the XR entry, and the resource entry can each pass their own row type.
+   * entry and the XR entry can each pass their own row type.
    */
   fun <T> select(
     items: List<T>,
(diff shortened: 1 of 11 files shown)


yschimke:
Thanks for the review — addressed in 81f6d96:

- **Lottie/SVG (P2):** ✅ Forwarded the three filter conventions to `composePreviewRenderLottie` / `composePreviewRenderSvg`. Both are `RenderPreviewsTask`s that apply the name/id filter over the *full* manifest before the `includeKinds` restriction, so a filter naming a non-Lottie/SVG preview is dropped by the kind filter (no fail-fast) while a global typo still fails.
- **XR fail-fast (P1):** ✅ Added `failOnNoMatch` to `PreviewFilter` (default `true`). The XR render — whose reader has already narrowed to `XR_SUBSPACE` — passes `false`, so a filter naming a non-XR preview renders nothing there instead of throwing and sinking `composePreviewRenderAll`. The image render (which sees every kind and keeps fail-fast) stays the authority on a genuine typo.
- **Resource sidecar (P2):** ✅ Stopped filtering the resource render entirely. Resource previews are a separate manifest of XML assets with no `@Preview` function, and the task owns one shared `resource-render-errors.json` keyed across the whole set — a partial run would overwrite it and erase diagnostics for skipped resources. Resources now always render in full; the composable render is where a filter narrows output.
- **`@PreviewParameter` ids (P2):** Intentional parity, not changing. The desktop `RenderPreviewsTask` also filters the *base* manifest ids and expands `@PreviewParameter` values later (in the renderer, not the task), so on both backends the filterable granularity is the multipreview member (which *is* a distinct manifest entry, e.g. `_light`/`_dark`), not the per-parameter value suffix. Filtering the expanded rows here would make Android diverge from desktop — the "semantics can't drift" requirement this PR is built around. The design-catalog deferral (`--exclude-preview-id *_dark`) targets those multipreview members, which this handles.

---

PR #3000, opened 2026-07-30
feat(serve): project playground diagnostics into the stock kotlin-playground errors map

yschimke:
## Summary

Delivers the follow-up I committed to on the #2989 review, plus the doc precision fix. A stock `kotlin-playground` frontend reads compile errors from an `errors` **map keyed by file name**, each entry nesting its position under `interval: { start, end }` with an uppercase `severity` — **not** a flat list. So `diagnostics` → `errors` couldn't be a rename (it would look compatible and render nothing); it has to be a real projection.

- **`PlaygroundStockError` / `PlaygroundInterval` / `PlaygroundPosition`** wire types, and an `errors` field on `PlaygroundRunResponse` alongside our own flat `diagnostics`. Both are populated, so neither a stock nor our own frontend is second-class.
- **`PlaygroundErrorsWire.project`** groups diagnostics by file, nests positions under `interval`, and maps severity to the upstream uppercase spelling + the CodeMirror gutter class. Warnings are carried too (upstream's `errors` includes `WARNING` descriptors). A file-less diagnostic lands under a `File.kt` default at the origin; a diagnostic with no end collapses to a zero-width caret.
- **`PLAYGROUND.md` §4** tightened — it previously said the response "reuses the upstream shape", which conflated our response with the stock wire format and was the source of the review mismatch.

Still pure data + a projection function — not wired to any route yet, so nothing is reachable from a running server until the compile-handler follow-up lands.

## Context on the classpath question (for reviewers)

Settled the design-doc open question on classpath source: the playground compile classpath will be the **liveBundle's resolved `userClassPath`** (`ServeBundleDaemon` already extracts `classes/app.jar` + resolves the Maven coordinates), so a snippet can `import` both the library and the catalog's own composables with no new resolver. First catalog is `compose-m3` (CMP), whose component surface is a complete, un-minimized Google-Maven jar. That work lands with the compile handler, not here.

## Test plan

- [x] `./gradlew :cli:test --tests "ee.schimke.composeai.cli.serve.PlaygroundErrorsWireTest"` — green (grouping by file, interval nesting + upstream severity/className, caret collapse, file-less default, empty→empty).
- [x] `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest` — clean.
- Non-visual change (server-side wire types + a pure projection); no rendered surface touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #3004, opened 2026-07-30
feat(serve): Stage-1 playground compile orchestrator

yschimke:
## Summary

The Stage-1 brain for the Compose playground (`docs/design/PLAYGROUND.md` §2), building on the merged foundation (#2989) and stock-`errors` wire (#3000). `PlaygroundCompileService` turns a `PlaygroundRunRequest` into a `PlaygroundRunResponse`:

1. Stage the snippet files to a temp work dir (path-traversal-safe, deduplicated `.kt` names).
2. Resolve the mode's catalog classpath (unavailable mode → refused before any work dir is created).
3. Compile against it.
4. **Compile error** → diagnostics under both our `diagnostics` shape and the stock `errors` map, **no token**, work dir cleaned up.
5. **Clean compile** → discover the `@Preview`, mint an expiring preview token (the token store now owns the work dir), optionally attach a first-frame `data:` PNG, and return `previewToken` + `previewUrl` for the Stage-2 handoff.

Every daemon/compiler/catalog collaborator is an **injected seam** (the same split `DefaultBtaCompileService.forSession` uses), so the orchestration is unit-tested with fakes + a `FakeFileSystem` — no BTA classloader or running daemon.

### The classpath seam is bundle-backed (per discussion)

`catalogClasspath` is the one knob with design weight. Its production implementation is documented as a thin adapter over the **liveBundle** resolution the serve host already runs (`ServeBundleDaemon.materialize`): extract the catalog's packed `.png`/zip bundle to `classes/app.jar` and resolve its `manifest.classpath` Maven coordinates to jars — the same `userClassPath` the live daemon runs on. So a snippet can `import` the resolved library (complete, since it comes from the unminimized library jar) plus whatever catalog composables survived bundle minimization. `compose-m3` is the clean first catalog because its component surface *is* that resolved jar. Written up in PLAYGROUND.md §8; the real resolver + route wiring + Stage-2 redemption are the next slices.

## Test plan

- [x] `./gradlew :cli:test --tests "…PlaygroundCompileServiceTest" --tests "…PlaygroundErrorsWireTest" --tests "…PlaygroundTokenStoreTest"` — green (19 tests). Covers: clean compile → token + render classpath (catalog jars + snippet classes), compile error → both diagnostic shapes + cleanup + no token, warnings pass through, missing `@Preview`, unavailable mode refused pre-workdir, blank request, first-frame `data:` URI, filename sanitisation/dedup.
- Non-visual change (server-side orchestration + capability plumbing); no rendered surface touched.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundCompileService.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundCompileService.kt
@@ -1,0 +1,200 @@
+package ee.schimke.composeai.cli.serve
+
+import java.util.Base64
+import okio.FileSystem
+import okio.Path
+
+/**
+ * Stage-1 orchestrator for the playground (`docs/design/PLAYGROUND.md` §2). Turns a
+ * [PlaygroundRunRequest] into a [PlaygroundRunResponse]: stage the snippet to a temp dir, compile
+ * it against the mode's catalog classpath, and — on a clean compile — mint an expiring preview
+ * token that Stage 2 redeems into a live session. Compile errors return diagnostics and **no**
+ * token.
+ *
+ * Every collaborator that touches the daemon, the compiler, or the catalog is an **injected seam**,
+ * so the orchestration (staging, gating, cleanup, token minting, response shaping) is unit-testable
+ * without a real BTA classloader or a running daemon — the same split
+ * `DefaultBtaCompileService.forSession` uses. The route handler (a follow-up) constructs this with
+ * the real backends.
+ *
+ * The single knob with real design weight is [catalogClasspath]. Its production implementation is a
+ * thin adapter over the **liveBundle** resolution the serve host already runs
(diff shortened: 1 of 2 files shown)


yschimke:
Both P2s addressed in `c8464d2` — good catches, both were real.

- **Work-dir allocation outside the error boundary** — `newWorkDir()` now runs inside the guard, so a full/unwritable temp volume returns the `PlaygroundRunResponse.exception` JSON contract instead of escaping as a throwable. Cleanup runs only once a path exists (tracked as a nullable), so a `newWorkDir()` failure has nothing to clean up and a later failure still clears its own dir.
- **Case-sensitive dedup keys** — collision keys are now case-folded, so two case-only-distinct sources (`A.kt` / `a.kt`) can't map to the same on-disk file on a case-insensitive FS and silently overwrite each other while both paths reach the compiler. Folding is universally safe — on a case-sensitive FS it only ever renames a name that would otherwise have been kept, never causing an overwrite.

Added a test for each (temp-volume failure → JSON contract; `A.kt`+`a.kt` → `A.kt`, `a_1.kt`).

---

PR #3005, opened 2026-07-30
ci: rotate the local Maven (~/.m2) cache monthly

yschimke:
## Summary

Caches the **whole local Maven repository** (`~/.m2/repository`) in the integration workflow with a **UTC year-month** key component, so the cache **rotates monthly instead of accumulating forever**.

`~/.m2` is populated at render/daemon time by Robolectric's own Maven resolver — its instrumented `android-all` runtime (~150–250 MB, the dominant entry) plus its other runtime artifacts — none of which Gradle's `setup-gradle` cache covers. It was already cached in `integration.yml`, but only the `android-all` subtree, keyed by `runner.os` + slug (or `agp8-min`) + coordinate; immutable per-coordinate entries lingered in the repo's 10 GB budget until LRU eviction.

## Changes (`.github/workflows/integration.yml` only)

Two commits:
1. Add a UTC year-month (`M2_CACHE_MONTH`, computed once per job — Actions expressions have no date function) to every cache `key` + `restore-keys`, in both the external-repo matrix job and `agp8-min`. It runs early so the early `cache/restore` and the late `always()` `cache/save` key to the same month.
2. Broaden the cached path from the `android-all` subtree to the entire `~/.m2/repository`; key renamed `m2-repo-<os>-<slug|agp8-min>-<YYYY-MM>-<coordinate>`.

Preserved (what keeps whole-repo caching safe at this scope): per-cell keying (cells resolve different SDK coordinates; a shared key would ping-pong — immutable keys mean only the first cell saves, the rest re-download every run), the `android-all` coordinate as the version discriminator, the coordinate prune (`android-all-cache-key.py`), and the corrupt-jar drop. So the whole-repo entry stays immutable per month and size-bounded.

Net effect: each month opens a fresh key prefix; a cell's entry falls out of the active month's prefix and LRU-evicts instead of persisting. Cost: one cold fetch per cell on its first run of a new month. Trade-off: a third-party artifact newly resolved mid-month isn't folded in until the next month or a coordinate bump — the price of an immutable, budget-safe entry (a `run_id`-append key would blow the budget at this size).

## Test plan

- `python3 .github/ci/test_android_all_cache_key.py` → 11 tests pass (the retention-policy script is unchanged).
- `integration.yml` parses as valid YAML; all four cache `key`/`restore-keys` carry `${{ env.M2_CACHE_MONTH }}` over `path: ~/.m2/repository`.
- Full verification lands on push to `main` — this workflow's `paths-ignore` intentionally skips re-running the heavy matrix for edits to `integration.yml` itself on PRs.

Non-visual change (CI cache keys), so no preview/visual evidence applies.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -750,6 +750,15 @@ jobs:
         with:
           cache-provider: basic
 
+      # The android-all cache key carries the current UTC year-month (see the
+      # keys below) so entries rotate monthly instead of accumulating forever.
+      # Actions expressions have no date function, so resolve it once here into
+      # the job env — the early Restore and the late always() Save both read it,
+      # so a run always keys restore and save to the same month.
+      - name: Resolve android-all cache month
+        if: matrix.render || matrix.render_xr || matrix.daemon
+        run: echo "ANDROID_ALL_CACHE_MONTH=$(date -u +%Y-%m)" >> "$GITHUB_ENV"
+
       # Robolectric's instrumented `android-all` runtime (~150-250 MB) is fetched
       # at RUN time by Robolectric's own Maven resolver into ~/.m2/repository —
       # not by Gradle — so setup-gradle's cache doesn't cover it and every render
@@ -765,6 +774,13 @@ jobs:
       # dead jar forever. (A `github.run_id` key like the font cache uses would
       # be simpler, but at this size it would save a fresh quarter-gig per cell
       # per run and evict everything else in the repo's 10 GB budget.)
+      #
+      # The key is additionally namespaced by UTC year-month (`ANDROID_ALL_CACHE_MONTH`,
+      # resolved above): each month opens a fresh key prefix, so a coordinate that
+      # stops being fetched — after a `robolectric` bump, or a cell being retired —
+      # falls out of the active month's prefix and LRU-evicts instead of lingering
+      # in the budget indefinitely. The cost is one cold ~200 MB fetch per cell on
+      # its first run of a new month; every run after that hits the month's entry.
       #
       # Keyed PER CELL (`matrix.slug`), not globally. Cells don't all want the
       # same coordinate: the SDK Robolectric synthesizes depends on the cell's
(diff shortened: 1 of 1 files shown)


yschimke:
The **Apply compose-preview** check is red here, but it's a **pre-existing failure on `main`**, not something this PR introduced:

- This PR only changes cache **keys** in `.github/workflows/integration.yml` — it adds a `YYYY-MM` component to the Robolectric `android-all` `.m2` cache so those entries rotate monthly. It touches nothing in the `samples/**` render path, the renderer, or the `apply` action this job runs.
- The failure is in the heavy full `samples/**` compose render (`SCOPE_MODULES: full`) — a Gradle build exception after ~13 min, with the real stacktrace going to the uploaded reports rather than the aggregated log.
- The same job has failed on the most recent `push`-to-`main` runs (`eb2e4d6d`, `6d9d700f`, `32167d45`, `87ee1d24`); the last green push was `c59bfad8`. So it's red on the base branch independent of this change.

This is base-branch drift, not a regression from this cache-key-only change, and fixing the sample-render breakage is out of scope here — so I'm not touching it in this PR. It should clear once the underlying render failure is fixed on `main` and this branch is updated against it. Happy to dig into the render failure separately if you'd like.

---

yschimke:
I wanted all artifacts not just android-all

yschimke:
Done — broadened in `3b7d1b7`. The cache now covers the whole local Maven repo (`~/.m2/repository`), not just the `android-all` subtree, so every artifact Robolectric's Maven resolver pulls at render/daemon time is cached. Key/path renamed to `m2-repo-<os>-<slug|agp8-min>-<YYYY-MM>-<coordinate>` over `~/.m2/repository`, in both the external-repo matrix and the `agp8-min` job.

Kept the parts that make it safe at this scope:
- **Per-cell keys** — cells resolve different `android-all` SDK coordinates, so a shared key would ping-pong (immutable keys → only the first cell saves, the rest re-download every run).
- **android-all coordinate as the version discriminator + the prune** — android-all is the dominant, version-bearing entry, so keying and pruning on it keep the whole-repo entry immutable per month and size-bounded (a `run_id`-style key would save a fresh quarter-gig-plus per cell per run and blow the 10 GB budget).

One documented trade-off: because the tree is keyed on the android-all coordinate, a third-party artifact newly resolved *mid-month* isn't folded in until the next month or a coordinate bump — the price of keeping the entry immutable. If you'd rather it pick up new artifacts within a month, the alternative is a `run_id`-append key with a monthly restore-prefix, at meaningfully higher cache-budget cost; happy to switch if you prefer that.

Note: the integration matrix (the job these keys live in) doesn't run on this PR — `integration.yml` is in its own `paths-ignore`, so the keys get their first live exercise on push-to-`main` after merge.

---

PR #3008, opened 2026-07-30
feat(serve): resolve a catalog liveBundle into a playground compile classpath

yschimke:
## Summary

The keystone toward "build a snippet against a catalog's components": `PlaygroundCatalogClasspath` is the production backing for `PlaygroundCompileService`'s `catalogClasspath` seam (#3004). It turns a catalog's packed **liveBundle** into the classpath a playground snippet compiles against, so `import`s resolve to the real catalog components.

It's the compile-time twin of `ServeBundleDaemon.materialize`'s classpath resolution, **minus the daemon launch**, reusing the exact same helpers (`BundleReader.readMetadata` / `extractZipBytes` / `extractEmbeddedLibs`, `extractBundleClassesAndManifest`, `CoordinateResolver`):

1. Extract the bundle's `classes/app.jar` — the catalog's own composables.
2. Resolve `manifest.classpath` Maven coordinates to jars (Central + Google Maven + `extraMavenRepos`).
3. Assemble a **flat** compile classpath: catalog classes first, then embedded libs, then all resolved dependency jars.

The flatness is deliberate and documented: `ServeBundleDaemon.bundleDaemonClasspaths` splits jars into a daemon-parent overlay and a user-child loader, but that partition is a *render-time* classloader-delegation concern. For *compiling* the snippet, every jar belongs on one classpath. A snippet compiled against this can `import androidx.compose.material3.*` (complete, from the unminimized library jar) plus whatever catalog composables survived bundle minimization — which is why `compose-m3` is the clean first catalog.

Structured so the pure assembly (ordering, dedup) is unit-tested, while the IO orchestration is thin reuse of already-tested bundle helpers — the same split the existing `ServeBundleDaemonTest` uses (it unit-tests `bundleDaemonClasspaths` with fake paths, not `materialize`'s IO).

## Test plan

- [x] `./gradlew :cli:test --tests "…PlaygroundCatalogClasspathTest"` — green (3 cases: catalog-classes-first ordering, dedup with first-seen order preserved, empty-deps still yields catalog classes).
- Non-visual change (classpath resolution plumbing).

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundCatalogClasspath.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundCatalogClasspath.kt
@@ -83,7 +83,7 @@ object PlaygroundCatalogClasspath {
 
     val libJars = BundleReader.extractEmbeddedLibs(zipBytes, libsDir, fileSystem)
     val mavenCoords = manifest.classpath.filterIsInstance<BundleReader.ClasspathEntry.Maven>()
-    val resolvedJars =
+    val resolutions =
       CoordinateResolver(
           warn = { onLog("playground $system: $it") },
           networkEnabled = if (offline) false else CoordinateResolver.defaultNetworkEnabled(),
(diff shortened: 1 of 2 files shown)


yschimke:
Good catch — fixed in `c8d6e87`. `mapNotNull { it.file }` silently dropped unresolved coordinates and returned a partial classpath, which would surface a misleading `unresolved reference` to the user instead of the honest mode-unavailable response.

Added `requireAllResolved`: if any declared coordinate resolves to a null file (offline, missing cache, unreachable repo), it logs the misses and returns null so the caller reports the mode unavailable, rather than compiling against an incomplete classpath. This deliberately differs from the live-daemon path, which tolerates partial resolution and falls back to baked PNGs — a playground *compile* can't fall back, so it fails closed. Unit-tested both ways (all-resolved → jar list; one miss → null + logged coordinate).

---

PR #3010, opened 2026-07-30
feat(serve): the playground — compile a snippet against a catalog's components

yschimke:
## Summary

Completes Phase 1 of the Compose playground (`docs/design/PLAYGROUND.md`): a `compose-preview serve` host can now **compile a snippet against a catalog's real components** and hand back diagnostics + an expiring preview token. Builds on the merged pieces (design doc, token store + DTOs, stock-`errors` wire, Stage-1 orchestrator, catalog classpath resolver).

**End to end:** `serve --playground-bundle <compose-m3 liveBundle>` → `POST /api/1/compiler/run` with `{files:[{name,text}], confType:"compose-cmp"}` → the snippet compiles against compose-m3's resolved classpath (so `import androidx.compose.material3.*` resolves), returning diagnostics and, on a clean compile, a `pg_…` preview token.

### The pieces

- **`lib-bta/` staging** — stage the BTA impl jars + Compose compiler plugin into the CLI install, mirroring `lib-daemon-desktop`. Verified: `installDist` stages 13 jars.
- **`PlaygroundBtaCompiler`** — in-process BTA compile over the shipped `BtaCompileSession`. Uses the **non-incremental** path (directory-safe classpath — the catalog classpath leads with an extracted `classes/` dir) on a **retained** session (bootstrap paid once). Diagnostics mapped 1-based kotlinc → 0-based CodeMirror.
- **`PlaygroundPreviewDiscoverer`** — scoped ClassGraph scan of the snippet's `classesDir` for `@Preview` methods → `<class>.<method>` ids.
- **`PlaygroundCatalogClasspath`** (merged #3008) resolves the catalog liveBundle → compile classpath.
- **The run route + `--playground-bundle` gate** — `POST /api/{version}/compiler/run` wired into `ServeHttpServer` (mirrors the `docStore` lane); `ServeCommand` resolves the classpath at startup and constructs the service. **Refused under `--public`** — the compile runs user-supplied code in-process, which the public server's "never run untrusted code" model forbids; fail-soft otherwise.

### Scope / follow-ups

Phase-1 minimum to reach "build against catalog components." Not yet here (design doc phases 2–4): the `/pg/<token>` Stage-2 **redemption** into the live interactive session, the Android + Remote-Compose modes, and the per-session **sandbox** that gates any `--public` exposure. The lane stays token-gated and off under `--public` until that sandbox lands.

## Test plan

- [x] `./gradlew :cli:test --tests "…Playground*"` — green: routing test over real HTTP (clean compile → `pg_` token; route 404s when the lane is off), compiler diagnostic mapping, discoverer real-bytecode scan, catalog classpath assembly + fail-closed, token store, orchestrator.
- [x] `./gradlew :cli:installDist` → `lib-bta/` staged (13 jars); `./gradlew --no-daemon :cli:installDist --stacktrace` (the serve-lanes CI command) builds clean.
- [x] `./gradlew :cli:compileKotlin :cli:ktfmtFormat` — clean.
- Non-visual (server-side compile plumbing + HTTP route); no rendered surface touched.

### DIFF
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -193,6 +193,10 @@ dependencies {
   // Renderer-agnostic daemon core helpers that are safe to use as a local library from CLI
   // commands. Keep renderer backends (`:daemon:android`, `:daemon:desktop`) out of this module.
   implementation(project(":daemon:core"))
+  // ClassGraph for the `serve --playground` preview scan: a scoped `@Preview` enumeration of a
+  // just-compiled snippet's classes dir (mirrors `:daemon:core`'s IncrementalDiscovery, which keeps
+  // classgraph as its own `implementation` and so doesn't leak it here).
+  implementation(libs.classgraph)
   // Wire-shape of the `compose/overrides` data product (`PreviewOverrideDeclaration`) — the
   // editable
   // knobs `compose-preview serve` reads from a bundle's `previews/<id>.overrides.json` sidecar to
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -110,7 +110,39 @@ val composePreviewDaemonAndroid =
     }
   }
 
+// BTA (Kotlin Build Tools API) *implementation* classpath for the `serve --playground` in-process
+// compile: `kotlin-build-tools-impl` (transitively `kotlin-compiler-embeddable` + runtime) plus the
+// Compose compiler plugin. These are the jars the gradle plugin's `DaemonBootstrapTask` supplies to
+// the editor daemon via sysprops — the serve host has no gradle plugin, so it stages them into the
+// CLI install (`lib-bta/`) and loads them into BTA's isolated classloader at compile time. Never on
+// the CLI's own classpath (a whole compiler frontend); resolved into
+// `cli/build/install/compose-preview/lib-bta/`, located at runtime via `APP_HOME/lib-bta/` (or
+// `-Dcomposeai.cli.libBtaDir`).
+val composePreviewBta =
+  configurations.create("composePreviewBta") {
+    isCanBeResolved = true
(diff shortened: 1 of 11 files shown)


yschimke:
Both compiler findings fixed in `df044ef` — the P1 especially was a real latent bug, thanks.

- **P1 (directory classpath entries snapshotted as JARs)** — correct: `compileIncremental` snapshots every classpath entry (`sha256OfFile` opens it as a stream + `classpathSnapshottingOperationBuilder`), and the catalog compile classpath leads with the extracted `classes/` **directory** (`PlaygroundCatalogClasspath.assemble`), so it would have thrown before compiling. Switched to the **non-incremental** `compile()` path, which doesn't snapshot the classpath and accepts directory entries. This is also the *right* mode for a playground: it compiles a fresh snippet against a stable catalog classpath, so IC's per-entry snapshotting was pure overhead. Added a per-call `diagnosticListener` to `BtaCompileSession.compile` so the non-incremental path still collects structured diagnostics.
- **P2 (retain the session)** — now one `BtaCompileSession` per `PlaygroundBtaCompiler` instance (lazy), so the BTA-impl bootstrap is paid once and the session's built-in serialization covers concurrent requests; no per-request lock/cache race, and the non-incremental path writes no `icWorkingDir` state anyway.

Separately, I saw the **`Serve render lanes honor overrides`** CI failure on the prior SHA — a Gradle configuration-phase failure whose root cause I couldn't pin from the truncated log. My local `:cli:installDist` (which that job runs) succeeds with the `lib-bta` staging, and the staging mirrors the passing `lib-daemon-desktop` pattern exactly, so I suspect a flake; this push re-runs it. I'll dig in if it recurs on `df044ef`.

---

PR #3009, opened 2026-07-30
refactor(rc-embedded): gather bitmap decode/blit behind a platform seam

yschimke:
## Summary

First increment of porting the embedded Remote Compose player's **draw path** to CMP Desktop / Skiko (so the `cmp-jvm` render lane, currently shown-disabled in the serve viewer, can eventually render). This lands the **image-decode seam** — the last `android.graphics` coupling on the draw path's two shared files — mirroring exactly how the text seam was built.

Before, the draw path reached `android.graphics.Bitmap` inline in three shapes:
- `resolveBitmap` / `resolveCanvasBitmap` returned a framework `Bitmap` that the blit ops (`DrawBitmap`/`DrawBitmapScaled`/`DrawBitmapInt`, the three bitmap-font ops) converted with `asImageBitmap()`;
- `DrawToBitmap`'s offscreen target did a mutable-copy + `eraseColor` + `Canvas(target.asImageBitmap())` dance;
- `RcPlayerPaint`'s `TEXTURE` path wrapped the decode in an `ImageShader`.

Those framework touches now live in one written-here seam, **`RcPlayerImagePlatform.kt`** (`resolveImage` / `resolveCanvasImage` / `prepareOffscreenTarget`, with the framework `resolveBitmap` kept `internal` beside them). **This is a move, not a port** — the bodies are the same decode/lookup/copy, so Android's pixels are unchanged. The seam hands back Compose's multiplatform `ImageBitmap` (the type every draw-path caller already converted to), so **`RcPlayerDrawing.kt` and `RcPlayerPaint.kt` no longer name `android.graphics` at all**. With framework `Paint` already gone via the text seam, the draw path's two shared files are now import-clean.

`resolveBitmap` stays framework-typed and `internal` because the surfaces that still genuinely need a `Bitmap` — the AGSL `BitmapShader` (`RcPlayerShaders.kt`), the reactive `rememberRemoteBitmapAsState` (`state/RcPlayerBitmapState.kt`), and the host `RcImageLoader` — are all Android-only anyway (each separately deferred). The seam removes `android.graphics` from the two files otherwise ready to compile off Android; it doesn't try to make those portable.

## Why the jvm sibling isn't in this PR

[...]

PR #3012, opened 2026-07-30
feat(rc-embedded-jvm): add JvmRemoteContext with a skiko bitmap decode

yschimke:
## Summary

Step 3 of porting the embedded Remote Compose player's **draw path** to CMP Desktop / Skiko (toward the `cmp-jvm` render lane): the desktop/JVM draw `RemoteContext` — the counterpart of `remote-player-core`'s `AndroidRemoteContext` for the pixel path.

As `PROVENANCE.md`'s *"What a `JvmRemoteContext` actually costs"* predicted, it's `StoreBackedRemoteContext` (the whole neutral variable/state store, already shared to JVM) **plus the one genuinely platform-bound member** — `loadBitmap`, decoding a document's encoded bytes to a raster — over skiko:
- **PNG types** (`TYPE_PNG` / `TYPE_PNG_8888` / `TYPE_PNG_ALPHA_8`) → `org.jetbrains.skia.Image.makeFromEncoded`.
- **Raw types** (`TYPE_RAW8888` straight RGBA, `TYPE_RAW8` single-channel alpha) → a raster built straight from the pixel bytes.
- A buffer too short for the declared `width × height × stride` is rejected (no OOB read), and any decode failure is swallowed to a blank rather than crashing the render — the text seam's never-throw posture.
- Only `ENCODING_INLINE` decodes; `ENCODING_URL` / `ENCODING_FILE` are the (deferred) host image loader's job.

Where Android caches an `android.graphics.Bitmap` under the id, this caches a Compose `ImageBitmap`; the store is untyped, so the JVM image seam's `resolveImage` (a later increment) reads its own concrete type back.

Two documented parity nuances: `TYPE_PNG_ALPHA_8` decodes to RGBA here rather than an alpha mask, and URL/FILE encodings are host-fetched.

## Sequencing (per `third_party/rc-embedded-player/PROVENANCE.md`)

- ✅ text seam (both halves)
- ✅ image seam (Android half) — #3009
- ✅ draw dispatcher out of `RcPlayer.kt` — #3011
- ✅ **JVM draw `RemoteContext` + bitmap decode** — this PR (step 3's `RemoteContext` half)
- ⏭ next: the JVM `resolveImage`/`resolveCanvasImage` seam siblings (which read this decode back) + splitting the `Drawable`-typed host loader and googlefonts text layout, then the draw files onto the JVM source list → a JVM `ImageComposeScene` raster harness that finally renders an `.rc` to PNG

## Test plan

- [x] `:third-party-rc-embedded-player-jvm:check` — compiles the new context against skiko + shares the Android sources; green.
- [x] `JvmRemoteContextBitmapTest` — PNG + raw round-trip to an `ImageBitmap` of the declared size, plus reject-short-buffer / reject-malformed / skip-non-inline.
- ⚠️ **skiko-native-gated:** like the existing `DesktopTextPlatformTest`, the decode tests need `skiko-awt-runtime-*` + a loadable GL library. In this PR's build container the loader couldn't bring skiko up, so the 5 tests **skip loudly** there (compile-verified); they **execute on CI**, which has the natives. Not a silent skip — the class prints why and `Assume`-skips.
- [x] `ktfmtCheck` clean.

Text-only PR: a new JVM-only building block, not yet wired into any render, so no rendered-output change and no visual evidence applies.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EXNXNpWA4JUpN3m4mVD8zT)_

### DIFF
--- a/third_party/rc-embedded-player-jvm/build.gradle.kts
+++ b/third_party/rc-embedded-player-jvm/build.gradle.kts
@@ -85,7 +85,12 @@ val sharedPlayerSources =
  * be told apart — including the jvm one by name would silently pull in the Android one as well.
  */
 val jvmPlayerSources =
-  listOf("androidx/compose/remote/player/compose/embedded/RcPlayerTextPlatformJvm.kt")
+  listOf(
+    "androidx/compose/remote/player/compose/embedded/RcPlayerTextPlatformJvm.kt",
+    // The jvm draw RemoteContext — StoreBackedRemoteContext + a skiko `loadBitmap` decode, the one
+    // platform-bound member of the contract. Written here, so it names skiko rather than the SDK.
+    "androidx/compose/remote/player/compose/embedded/JvmRemoteContext.kt",
+  )
 
 kotlin {
   sourceSets["main"].kotlin.apply {
--- a/third_party/rc-embedded-player-jvm/src/main/kotlin/androidx/compose/remote/player/compose/embedded/JvmRemoteContext.kt
+++ b/third_party/rc-embedded-player-jvm/src/main/kotlin/androidx/compose/remote/player/compose/embedded/JvmRemoteContext.kt
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2026 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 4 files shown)


yschimke:
Good call — fixed in 60df6d7. The decode `catch` is now `Exception`, not `Throwable`: skiko throws an ordinary exception on undecodable bytes (which stays non-fatal → blank), while a `LinkageError` (mispackaged skiko) or `OutOfMemoryError` now propagates rather than being hidden behind a blank image. The explicit short-buffer guard already handles the undersized-raster case ahead of the decode.

---

PR #3013, opened 2026-07-30
fix(design-artifacts): reject an all-deferred catalog instead of rendering everything

### LINKED ISSUE #2993: design-artifacts: an all-deferred catalog renders everything, because the empty render filter means "render all"
Raised by Codex review on #2991 (P2), confirmed in the code but deliberately left out of that PR — it needs a decision about whether an all-deferred catalog is a legal config at all.

## The overload

`renderFilterPatterns` returns `[]` for two different situations:

```js
const { required, deferred } = previewNamesByPriority(spec);
if (deferred.length === 0) return [];   // "nothing deferred → render everything"
return required;                        // ← also [] when EVERYTHING is deferred
```

The pre-flight writes that as an empty `composePreview.filter`, and `design-artifacts.yml` / `design-artifacts-reusable.yml` both read an empty filter as *render everything*. So a catalog that defers every entry renders every single preview — precisely the renders deferral exists to skip. The saving inverts exactly where it should be largest.

Costs render time, never correctness: the export keys deferral off the spec, not off whether a PNG happens to exist, so nothing is mis-published. And it is unreachable today — no committed catalog declares `priority` yet.

#2991 widened the case slightly: with variants now inheriting their component's deferral, a catalog of all-deferred components-with-variants lands on `[]` where its required-default variants previously kept the list non-empty.

## The decision to make first

Is an all-deferred catalog legal?

- **No** → the spec validator should reject it ("a catalog cannot defer every entry — it would publish no stickers at all"). Small, self-contained, and matches the positive-list philosophy already documented on `renderFilterPatterns`: name what must render so a stale spec fails loudly instead of silently rendering wrong.
- **Yes** (e.g. a wasm system served entirely on demand) → the empty-vs-none collision has to be broken with an explicit render-none sentinel plumbed through the pre-flight and both workflows, since an empty string can't carry the distinction.

Worth resolving before the first catalog actually declares `priority`, since that's when it becomes reachable.

yschimke:
Closes #2993.

## The bug

`renderFilterPatterns` returns `[]` for two different situations, and both design-artifacts workflows read an empty `composePreview.filter` as **render everything**:

```js
const { required, deferred } = previewNamesByPriority(spec);
if (deferred.length === 0) return [];   // "nothing deferred → render everything" (the legit case)
return required;                        // ← also [] when EVERYTHING is deferred
```

So a catalog that defers *every* entry (`deferred` non-empty, `required` empty) collapses to an empty filter and rasterises the whole sheet — precisely the renders deferral exists to skip — while the published bundle carries no baked stickers at all. The saving inverts exactly where it should be largest. It's unreachable today (no committed catalog declares `priority`), but #2991 widened the case, so it's worth closing before the first catalog declares `priority`.

## The decision

The issue asks whether an all-deferred catalog is a legal config. This PR takes the **"No"** answer it recommends: it isn't. That matches the positive-list philosophy already documented on `renderFilterPatterns` (name what must render, fail loudly on a stale spec) and avoids plumbing a render-none sentinel through the pre-flight and both workflows for a config that publishes an empty sticker sheet.

## Changes

- **`catalog-priority.mjs`** — add `defersEveryPreview(spec)`: true when at least one preview is deferred and none is left required (keyed off the same `previewNamesByPriority` split the render filter derives from, so it flags exactly the specs that would empty the filter). Mode-only deferral and no-deferral catalogs are not flagged. Expanded the `renderFilterPatterns` doc to note the all-deferred empty is now rejected upstream.
- **`catalog-spec.mjs`** — `validateSpec` rejects an all-deferred catalog structurally, independent of the live-path gate (so it fires even on the lenient CLI path), with a message explaining the empty-filter render-all trap.
- Tests in both `*.test.mjs`, plus adjusting one existing name-resolution test that happened to use an all-deferred spec.

No workflow / sentinel plumbing needed — the render-all-via-empty-filter case is now unreachable.

## Test plan

- `node --test catalog-priority.test.mjs` → 15 pass
- `node --test catalog-spec.test.mjs` → all pass (new all-deferred rejection test included)
- `node --test validate-samples.test.mjs` → 3 pass (no committed catalog trips the new gate)

Non-visual change (spec validation + pure helper), so text-only evidence per the repo's visual-evidence rule.

---
_Generated by [Claude Code](https://claude.ai/code/session_012edJaPyTm9fygk37ZhwaXP)_

### DIFF
--- a/scripts/design-artifacts/catalog-priority.mjs
+++ b/scripts/design-artifacts/catalog-priority.mjs
@@ -134,11 +134,19 @@ export function isImageDeferred(spec, image) {
   return modePriority(spec, image?.theme) === DEFERRED;
 }
 
-/** Every component entry in the spec, with its group, flattened. */
+/**
+ * Every component entry in the spec, with its group, flattened. Tolerant of a malformed spec — a
+ * non-array `groups`, or a group whose `components` isn't an array — so the pure priority readers
+ * (`previewNamesByPriority`, `specDefersAnything`, `defersEveryPreview`, `deferralPlan`) can run on a
+ * structurally-invalid spec without throwing. `validateSpec` reports the shape error separately; this
+ * must not pre-empt it with a `TypeError`.
+ */
 function components(spec) {
-  return (spec?.groups ?? []).flatMap((group) =>
-    (group?.components ?? []).map((component) => ({ group, component })),
-  );
+  const groups = Array.isArray(spec?.groups) ? spec.groups : [];
+  return groups.flatMap((group) => {
+    const entries = Array.isArray(group?.components) ? group.components : [];
+    return entries.map((component) => ({ group, component }));
+  });
 }
 
 /**
@@ -218,6 +226,13 @@ export function previewNamesByPriority(spec) {
  * exactly the specs that would otherwise empty the filter through deferral. Mode-level deferral
  * can't trigger it — it never empties the preview list — and a catalog that defers nothing is the
  * ordinary render-all case, also not flagged.
+ *
+ * Reflects only the spec's OWN `groups`, which is all the pre-flight render filter sees too. A hybrid
+ * catalog can leave its required entries in `@CatalogComponent` annotations and place only deferred
+ * overrides in `spec.groups`; those annotation entries are merged in as `required` by the driver
+ * (`mergeCatalogGroups`), so such a spec is not truly all-deferred. The caller must therefore only act
+ * on a `true` here when the spec's groups are known to be the complete inventory — `validateSpec`
+ * gates on `annotatedInventory === false` for exactly that reason.
  */
 export function defersEveryPreview(spec) {
   const { required, deferred } = previewNamesByPriority(spec);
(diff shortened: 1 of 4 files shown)


yschimke:
Both Codex points addressed in `2c49d8e`:

- **Annotation-only required components** — the all-deferred rejection now only fires when `annotatedInventory === false` (module scanned, no `@CatalogComponent`), i.e. `spec.groups` is the complete inventory. A hybrid catalog that keeps its required entries in annotations and places only deferred overrides in `spec.groups` is no longer blocked — those entries merge in as `required` and an empty pre-flight filter correctly renders them. Unknown annotation state stays lenient, matching the existing `groups`-omitted and `liveBundle` gates. Test: *does not flag an all-deferred spec that may have annotation-supplied required entries*.

- **Malformed `groups` shape** — `components()` now tolerates a non-array `groups` (or a group whose `components` isn't an array), so `defersEveryPreview` can't throw a `TypeError` before `validateSpec` reports the intended `` `groups`, when present, must be a non-empty array `` error. Test: *reports the groups-shape error, not a TypeError, on malformed groups*.

---

PR #3018, opened 2026-07-30
feat(vscode): navigate previews to their component source on title click

### LINKED ISSUE #2940: VS code plugin click to navigate to component

yschimke:
Closes #2940.

## Summary

Clicking a preview card's title in the VS Code panel already opened source, but it (a) reconstructed the file path from the compiled class name and (b) always landed on the `@Preview` **function stub**. Discovery already records each preview's authoritative, module-relative `sourceFile` **and** infers the composable a preview renders (`PreviewInfo.targets[]`, via `PreviewTargetInference`) — neither was being used for navigation.

This wires both into the existing title-click affordance so a click takes you to the **component** the preview shows, which is what the issue asks for:

- **`previewSourceTarget(p)`** (pure, in `cardData.ts`) picks the navigation destination: the inferred component (`targets[0]`, most-confident first) when discovery found one, else the preview function itself. It carries the destination's `sourceFile` along.
- The `openFile` webview→host message gains an optional **`sourceFile`**; the card sends the resolved destination's coordinates.
- **`resolvePreviewSourceUri`** resolves the file from the manifest's `sourceFile` first (a `src/…/Foo.kt` glob, robust when the file name differs from the class), falling back to the old class-derived path (`com.example.FooKt` → `com/example/Foo.kt`) for older manifests or stale paths. The `build` output tree is excluded from both globs.
- The title tooltip now reads **`Open component: <FQN>`** when the destination is an inferred component, and keeps `Open source: <FQN>` for the preview-function fallback.

Behaviour is backward compatible: previews with no inferred target and no recorded `sourceFile` (older manifests) resolve exactly as before.

## Test plan

- [x] `npm run compile` — clean (host + webview)
- [x] `npm test` — **1610 passing**, including new `previewSourceTarget` cases and the `Open component:` tooltip assertion in `cardData.test.ts`
- [x] `npm run format` check (`prettier --tab-width 4`) — all touched files clean

## Visual evidence

Not applicable — this is a navigation/behaviour change. The card renders identically; the only user-visible surface is the click destination and the hover-tooltip string (`Open component: …`), neither of which the `@Preview` PNG / preview-harness pipeline captures. Covered by the unit tests above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FR2akUjBgsAj6cZrsRpmnU)_

### DIFF
--- a/vscode-extension/src/extension.ts
+++ b/vscode-extension/src/extension.ts
@@ -5778,9 +5778,28 @@ function handleWebviewMessage(msg: WebviewToExtension) {
             );
             break;
         }
-        case "openFile":
-            openPreviewSource(msg.className, msg.functionName, msg.sourceFile);
+        case "openFile": {
+            // Resolve the destination under the preview's owning module first,
+            // so a module-relative source path that also exists in a sibling
+            // module can't open the wrong module's file.
+            let moduleDirAbs: string | undefined;
+            if (msg.previewId && gradleService) {
+                const mod = previewModuleIndex.get(msg.previewId);
+                if (mod) {
+                    moduleDirAbs = path.join(
+                        gradleService.workspaceRoot,
+                        mod.projectDir,
+                    );
+                }
+            }
+            openPreviewSource(
+                msg.className,
+                msg.functionName,
+                msg.sourceFile,
+                moduleDirAbs,
+            );
             break;
+        }
         case "selectModule":
             selectedModule = msg.value || null;
             logInfo(`[panel] module selected: ${selectedModule ?? "<none>"}`);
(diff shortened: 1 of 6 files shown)


yschimke:
Both review findings are addressed in `5a2fa67`:

- **Stale navigation target on reseed** — the title's destination now lives on the card's dataset (`writeNavDataset`), restamped by `refreshCardMetadata` on every manifest reseed, and the click handler reads it at click time. Tooltip and click destination can no longer drift when a reused card's inferred component changes.
- **Cross-module ambiguity** — the click now carries `previewId`; the host resolves the preview's owning module and searches under its directory first (`RelativePattern`), falling back to the workspace-wide glob only for genuinely cross-module targets, then the class-derived path. So a module-relative path that also exists in a sibling module resolves to the right module's file.

Added unit tests for `writeNavDataset` (including the stale-`sourceFile` clear). `npm run compile` clean, `npm test` 1612 passing, format check clean.

---

PR #3023, opened 2026-07-30
fix(figma-svg): clip children to a Modifier.clip shape (#2852)

### LINKED ISSUE #2852: figma-svg loses clip shapes and brush backgrounds in real Compose components
## Reproduction

The six compose-samples catalogs expose one shared modifier-capture gap.

1. Jetcaster `PodcastDetails/Subscribe`:
   https://preview.coo.ee/jetcaster/render/podcastdetails-subscribe__ideal__default__compact.svg?mode=web

   The PNG has a rounded yellow toggle and a circular notification toggle. The SVG emits plain `<rect>` nodes with no `rx`, `clipPath`, or mask. The source uses Material 3 expressive `ToggleButtonShapes` with `RoundedCornerShape(15.dp)` / `RoundedCornerShape(60.dp)`.

2. Jetsnack `Tokens/Gradients`:
   https://preview.coo.ee/jetsnack/render/tokens-gradients__ideal__default.svg?mode=web

   The PNG has eight gradient ramps. The SVG keeps the labels but drops every ramp. The source uses `Modifier.background(Brush.horizontalGradient(ramp), RoundedCornerShape(4.dp))`.

3. The same loss appears in Jetcaster podcast cards/rows and Jetsnack gradient buttons/icon buttons.

The PNGs are correct. The generated SVG model either separates the fill from its clip shape or treats the brush as no paint, then emits an incomplete editable subtree instead of a faithful hybrid fallback.

## Expected

- Resolve shape and paint from the effective drawing node even when Material 3 implements them through separate/internal modifier nodes.
- Preserve rounded/circle clipping for expressive `ToggleButton` containers.
- Emit supported Compose brushes as SVG gradients, or raster the complete affected draw layer when the brush cannot be vectorized.
- Add Android end-to-end fixtures for both examples and compare PNG versus browser-rendered SVG.

yschimke:
Closes #2852.

## Problem

The layered `compose/figma-svg` export had **no per-layer clip concept**. A `Modifier.clip(shape)` container rounded its own fill but never masked its children, so a child placed beyond the clip — the residual named in [comment 3](https://github.com/yschimke/compose-ai-tools/issues/2852#issuecomment-5123795726), Jetsnack **Search/Categories'** minimum-size image under `.clip(CategoryShape)` — was emitted as an unclipped `<image>`/`<rect>` and **grew the exported canvas past the card viewport** the render actually shows.

The `FigmaSvgLayer` model had `raster`, `vector`, `cornerRadiiPx`, `circle` … but nothing that said "this node clips its subtree", and the only `<clipPath>` in the whole document was the device-screen mask. That is the "loses clip shapes" half of this issue's title.

## Fix

Thread a real clip through the pipeline:

- **Capture** — `ModifierTokenResolver` reads the `clip = true` flag off the `graphicsLayer` a `Modifier.clip(shape)` lowers to and records a new `clipsContent` token (distinct from the paint-shape tokens: a `background(color, shape)` rounds its own fill but does **not** clip an overflowing child).
- **Model** — the token becomes `FigmaSvgLayer.clipChildren`; `extent()` clamps a clipping ancestor's descendants to its box so an intentionally-oversized child no longer grows the canvas.
- **Serialize** — `FigmaLayeredSvg` emits an SVG `<clipPath>` (rect / rounded-rect / circle / cut path, matching the layer's own corners) on the group **only when a descendant actually overflows**. A clip whose children all sit inside changes no pixels, so the overwhelmingly common case stays byte-for-byte unchanged — no fidelity regression.

Schema bumped additively (`layout/inspector` v12, `compose/semantics` v11); older files decode with `clipsContent = null`.

## Visual evidence

The overflowing child (blue) now clips to the rounded card instead of bleeding past it, and the canvas clamps to the card box:

before/after clip mask

*(Rendered from the exporter's actual output — the same `<rect x="70" width="70">` child under the same rounded `<rect rx>` card, with vs. without the emitted `clip-path`. Rasterised with the sandbox Chromium since these are unit-level fixtures, not catalog `@Preview`s the CI diff-bot renders.)*

## On the icon-button residual

The [comment 2](https://github.com/yschimke/compose-ai-tools/issues/2852#issuecomment-5123791082) residual (gradient ring drawn at the padded root instead of the inner control) **no longer reproduces on current `main` (0.19.9)** — that comment predates the `paintInset` growth-suppression that landed for 0.19.x. Reproducing the exact `padding → clip → animateColorAsState brush border → background → drawWithContent BlendMode.Plus tint` chain in-repo, the gradient ring already lands on the inner 61px control, not the 83px padded root. This PR captures that as a passing regression fixture (`FigmaSvgGradientTintedIconButtonRenderTest`) so it can't silently regress. No exporter change was needed for the ring bounds.

## Test plan

- [x] `FigmaSvgChildClipTest` (pure model) — extent clamps to the clip; `<clipPath>` emitted on overflow; **no** clipPath when children fit.
- [x] `FigmaSvgClipOverflowRenderTest` (Android/Robolectric e2e) — a real `.clip(RoundedCornerShape)` with an offset child: clipPath emitted end-to-end (modifier → token → model → serializer) and the canvas clamps to the card, not the overflow bbox.
- [x] `FigmaSvgGradientTintedIconButtonRenderTest` (Android/Robolectric e2e) — the exact chained path from the issue; ring lands on the inner control.
- [x] `:data-layoutinspector-core:test` + `:data-layoutinspector-connector:test` — full suites green (no regressions).
- [x] `:renderer-android:testDebugUnitTest --tests "*FigmaSvg*"` — full FigmaSvg suite green.
- [x] `ktfmtFormat` on all touched modules.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TgU5dMHzpGyrCaZsSJonxT)_


yschimke:
Thanks — both addressed in f47b076:

- **P1 (shaped clips):** the emit gate no longer keys on bbox overflow alone. For a rounded/circle/cut clip it now fires when a descendant's nearest in-box point falls **outside** the arc/chamfer, so a full-bleed child under `.clip(CircleShape)` (the ToggleButton / podcast-card case) is masked. It's a point-vs-arc test rather than a corner-bounding-box test, so inset content that stays inside the arc (a padded card label) still emits no mask — verified the Wear-scroll drift fixture is unchanged, i.e. no catalog churn. New tests cover the full-bleed rounded and circle cases.
- **P2 (clip position):** the mask now wraps only the children, not the whole named group, so the node's own background/border and drop shadow stay unclipped — matching `background(color).clip(circle)` keeping a square background.

---

PR #3033, opened 2026-07-30
feat(serve): link each preview to its source file on GitHub

yschimke:
Follow-up to the VS Code "navigate to component" work (#3018): previews should link to their source code everywhere, not just in the editor. This adds a per-preview **"source" link** under the title on the preview server's viewer page (`/p/<id>`) that opens the file on GitHub.

## What it took (cross-layer)

Review (Codex) correctly caught that a naive viewer-only version renders nowhere on the catalog viewer and would point at the wrong branch. The catalog carries no per-preview source path, and the delivery branch (`design-artifacts/<system>`) holds generated assets, not Kotlin. So the fix spans the export → serve → URL:

- **Export (`scripts/design-artifacts`)** — `apply-source-files.mjs` stamps each component's module-relative `sourceFile` (from discovery's `previews.json`, via the generator's existing `sourceByFunction` lookup) onto `catalog.json`, in the same post-process pass as `applySpecSections`. New helper + unit tests.
- **Serve store (`ServeCatalogStore`)** — parses `Component.sourceFile`, carries it into `previews/variants.json` (`VariantMeta.sourceFile`), and retains the catalog's real **source** `{repo, ref, module}` as `ServeWeb.CatalogSource` on `ServeBundleHost`. The host reads `sourceFile` from `variants.json` first (the catalog path), falling back to a root `previews.json` for plain uploaded bundles.
- **URL (`ServeUrls.githubBlobUrl`)** — now takes `repo/ref/module/sourceFile` and builds `blob/<ref>/<module>/<sourceFile>` (module-prefixed, per-segment encoded). `ServeHttpServer.handleViewer` builds the link from the catalog **source**, not delivery provenance.

The link is additive and gated: it renders only when the session carries a catalog source and the preview recorded a path; local/unprovenanced sessions and older catalogs render nothing.

## Rollout note

Existing catalogs need a re-export (the `design-artifacts` render) before their `catalog.json` carries `sourceFile` — until then the link is simply absent, no regression. Per the repo's delivery-branch policy, dispatch `design-artifacts.yml` after this merges so live catalogs pick it up.

## Visual evidence

The full viewer page needs a running server + render backend, so the `@Preview` PNG pipeline doesn't reach it. Below is the new header element rendered from the **real `ServeWeb` CSS + markup**, light and dark — the muted `source` link with the GitHub mark sits directly under the preview label:

preview-server viewer header with the new source link, light and dark

The `serve-viewer.html` golden fixture now renders a real source link too (module-prefixed blob URL), so future changes to this surface are diffed automatically. _(The screenshot above is committed only as a review artifact and dropped in a later commit; the squash-merge never lands it on `main`, and the pinned URL stays valid.)_

## Test plan

- [x] `./gradlew :cli:test` — **BUILD SUCCESSFUL** (full suite, incl. the regenerated `ServeWebFixtureTest` goldens)
- [x] New `ServeUrlsTest` cases for the 4-arg `githubBlobUrl` (module join, `/`-bearing ref, no-module, per-segment encoding, null-when-any-part-missing) and `ServeWebTest` link render/hide
- [x] `node --test` in `scripts/design-artifacts` — new `apply-source-files.test.mjs` passes (the one unrelated failure is a missing `pngjs` dev-dep in the sandbox; present in CI)
- [x] `ktfmtFormat` clean


yschimke:
Both review findings were correct — the original version rendered no link on the catalog viewer and would have used the wrong branch. Fixed in `1457cc8` across the whole path:

- **"Populate source paths for catalog previews"** — catalogs write no root `previews.json`, so `readSourceFiles()` always resolved empty. The exporter now records each component's `sourceFile` into `catalog.json` (`apply-source-files.mjs`, from discovery's `previews.json`), `ServeCatalogStore` carries it into `previews/variants.json`, and `ServeBundleHost` reads it from there (with the root-`previews.json` fallback kept for uploaded bundles).
- **"Build source links from source provenance"** — the URL is now built from the catalog's `source = {repo, ref, module}` (its actual Kotlin), not the `design-artifacts/<system>` delivery branch, with the `module` prefixed ahead of the module-relative `sourceFile`: `blob/<ref>/<module>/<sourceFile>`.

Existing catalogs need a re-export before their `catalog.json` carries `sourceFile`; the link is additive and absent until then, so no regression. `:cli:test` (incl. regenerated goldens) and the new export/serve unit tests pass.

---

PR #3064, opened 2026-07-31
fix(cli): only report show's partial results when this run produced them

yschimke:
Follow-up to #3059, addressing a P1 the Codex reviewer raised on that PR after it had already merged.

## The problem

#3059 made `show` report partial results when the render task failed, gated only on "the result list is non-empty". That is not evidence the renders belong to the invocation that just failed.

On a warm checkout `build/compose-previews/` still holds the previous run's `previews.json` and PNGs. If gradle fails during **configuration or compilation** — before the renderer rewrites anything — then:

1. `renderAllModules` reads the stale manifest,
2. `PreviewResultBuilder` accepts the stale PNGs,
3. `show` emits the previous run's images, hashes and `changed` flags as if they were this run's.

Worse, every `changed` reads `false`, because the shas still match the state file the last *successful* run wrote. So an agent iterating on a composable gets back "nothing changed" for a build that never actually ran — a more dangerous answer than the silence #3059 replaced. The motivating case for #3059 (one broken preview among many) is unaffected; this is the *other* failure mode.

## The fix

Require positive evidence that this invocation produced renders: at least one capture PNG whose mtime is at or after the invocation start.

- A render that ran and partially succeeded leaves fresh PNGs → report, exactly as #3059 intended.
- A build that died before the render task leaves only stale ones → fall back to the pre-#3059 bail-out (exit 2, no output).

The start stamp is floored to whole seconds (`floorToSecond`), because filesystem mtimes are commonly second-granular and a PNG written milliseconds after an unfloored start would otherwise look stale.

`FileSystem` is injected rather than reaching for the global, per the `:common-io` convention in `docs/AGENTS.md`, which is also what makes the policy testable in memory.

**Exit codes are unchanged.**

## Test plan

- [x] `./gradlew :cli:test` — green
- [x] `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest` — applied
- [x] `ShowPartialResultsTest` extended against a `FakeFileSystem`, covering: fresh render reported despite the failed build; stale leftovers **not** reported; PNG missing from disk; empty results; a null-`pngPath` preview vouching for nothing on its own; and `floorToSecond` truncation
- [x] Existing exit-code assertions retained

Visual evidence: N/A — this only affects CLI stdout and exit codes, not rendered output.

---
_Generated by [Claude Code](https://claude.ai/code/session_014NDRHMxYyfVYUF3pih7LjL)_


yschimke:
Moved to draft — the three review findings on this PR are correct, and together they show the mtime-after-the-fact approach here is unsound. Flagging rather than letting it merge as-is.

**1. Freshness must be per-result, not global (P1).** `results.any { fresh }` passes as soon as *one* PNG is fresh. In a multi-module run where module A renders and module B fails to compile, A's fresh PNG green-lights emitting B's stale rows too. Correct as written.

**2. `buildResults` can launder stale output (P2).** This is the one that sinks the approach. Under the Claude/Codex/Antigravity image-size override, `applyImageSizeOverride` rewrites oversized PNGs *inside* `buildResults` — which runs before the metadata lookup. That rewrite advances the mtime, so a stale PNG that gradle never touched can be presented as fresh. Sampling mtimes after `buildResults` cannot distinguish "the renderer wrote this" from "the CLI resized it".

**3. Same-second start (P2).** `floorToSecond` was meant to absorb second-granular filesystem timestamps, but it also widens the window: a PNG written by a previous successful run in the same wall-clock second as the new invocation's start satisfies `mtime >= start` even on a precise filesystem. It trades a false negative for a false positive.

**Corrected design:** snapshot the render outputs (path → mtime/size) per module *immediately before* `runGradle` inside `renderModules`, carry that snapshot on `RenderModulesOutcome`, and compute freshness by diffing against it *before* `buildResults` gets a chance to rewrite anything. Freshness then becomes a per-result property — stale rows are excluded or marked individually rather than the whole set being admitted on one fresh file. That removes all three failure modes: it is per-result (1), it observes the filesystem before CLI post-processing (2), and it compares against concrete prior state rather than a wall-clock threshold (3).

That reshapes `RenderModulesOutcome` and the render pipeline, so I'd rather have a maintainer's read on it than land it unilaterally.

**Note on the red check:** `Module Unit Tests` fails on `:daemon:desktop:test` and `:daemon:android:testDebugUnitTest` (SIGABRT, exit 134, native `Typeface` crash in `libandroid_runtime.so`). Same signature on `main` and on #3059 before it merged, so it predates this branch — this diff is two files under `cli/`, and `Build CLI`, `Plugin Unit Tests` and `Plugin Functional Tests` are all green here.

In the meantime `main` carries the #3059 behaviour, which reports partial results gated only on "results is non-empty" — i.e. finding 1 and the stale-leftover case are live there. Worth deciding whether to revert that half of #3059 until this lands.

---

yschimke:
Update: I built the snapshot design out, then ran it end-to-end against `meshcore-mobile:wear` — and the e2e killed it. Recording that here because it rules out a whole family of approaches, including the one I proposed in the previous comment.

## What was built

Snapshot every file under each module's `renders/` before `runGradle`, diff after, per-result filtering, all computed before `buildResults` can rewrite anything. That does fix the three review findings: per-result rather than per-batch, immune to the image-size override, and no start-second window (size is carried alongside mtime, so a same-second rewrite still registers).

## What the e2e showed

First run, warm build dir, one broken preview (`activity__WearMainActivity`, ViewModel can't be constructed):

```
Reporting the 28 preview(s) this run rendered, omitting 1 left over from a previous run.  EXIT=2
```

Correct-looking. Second run, no source change:

```
No preview was rendered by this invocation — gradle failed before the render task wrote anything…  EXIT=2
```

**That is wrong, and the message is a lie.** `:wear:composePreviewRenderAll` did execute; it failed on the same broken preview and re-rendered the other 28 to byte-identical PNGs. Nothing changed on disk, so the diff saw zero fresh files and suppressed all 29 rows — strictly worse than the behaviour on `main`.

The flaw is not in the details, it's the premise: **"no files changed" does not mean "the renderer did not run."** An up-to-date task, a from-cache task, and an idempotent re-render all write nothing while their outputs are perfectly current — and in an iterate loop that is the *common* case, not the corner. Any filesystem-observation scheme inherits this, including content hashing.

## The signal has to come from the build

Whether `:<module>:composePreviewRenderAll` actually executed this invocation is a build fact, and the Tooling API already reports it. `GradleConnection`'s progress listener (`GradleConnector.kt:165`) sees `TaskOperationDescriptor` with `FinishEvent`; the result distinguishes `TaskSuccessResult` (with `isUpToDate` / `isFromCache`), `TaskFailureResult` and `TaskSkippedResult`. Plumbing per-task outcomes through `runTasks` into `RawRenderOutcome` gives per-module currency directly, with no filesystem heuristic:

- Task success, up-to-date, or from-cache → that module's renders are current; report all its rows.
- Task never started (upstream compile failure, configuration failure, skipped) → stale; drop the module.
- Task started and failed → it ran, so report its rows; the per-preview `.error.json` sidecars already say which previews didn't produce output.

That last bucket is the one to think hardest about — a task that dies midway leaves some captures from this run and some from the last, and task outcome alone can't separate them. The pre-run snapshot is still useful *there* as a secondary signal, but only inside a module whose task actually ran, never as the primary gate.

## State of this branch

Pushed as WIP (`f903903`), still draft, not for merge. It carries the snapshot machinery plus unit tests, and one improvement worth keeping regardless: rows whose render *threw* are no longer dropped. They hold no image, sha or `changed`, so nothing stale can mislead, and they are exactly the "which preview is broken?" signal this behaviour exists to deliver — my first cut silently discarded them.

I'd rather hand this over than take a third swing at it in one sitting. `main` still has the #3059 behaviour (partial output gated only on "results is non-empty"), so the stale-leftover case is live there and reverting that half remains worth considering.

---

yschimke:
CI note on `82ed8ba` (the merge of `main` into this branch). None of the four failures come from this branch — its diff is two files under `cli/`, and `:cli:test` runs in the `Build CLI` job, which is green.

```
:renderer-xr:testDebugUnitTest
:daemon:core:test
:daemon:desktop:test
:daemon:android:testDebugUnitTest
```

Two of these are worth flagging because they're *new information*, not the standing red:

**`:daemon:android` now runs 212 tests, 43 failed, 10 skipped.** Previously 70 completed / 22 failed. That's #3079 landing — the suite no longer aborts mid-run, so roughly three times as many tests actually execute. Worth reading carefully when the final numbers get written up: the jump from 22 to 43 failures is mostly **tests that never ran before**, not new breakage, since the old run's JVM died partway and everything after it was silently never counted. Pass/fail should be quoted against the 212, not compared to the 70.

**`:daemon:desktop:test` failed here but passes on `main`.** I ran it locally against `b9fd678` (current `main`, which includes #3081) and it's **BUILD SUCCESSFUL**. This branch adds nothing that `:daemon:desktop` depends on, so this looks like either a flake or order/isolation sensitivity newly exposed by #3079's change to test forking — unconfirmed either way, I couldn't reproduce it. Flagging rather than diagnosing, since if it *is* order-dependent it will resurface on unrelated PRs and be blamed on them.

`:renderer-xr:testDebugUnitTest` and `:daemon:core:test` remain undiagnosed and untracked by any issue.

Separately, this PR is still draft and still carries a design I disproved end-to-end (see the comment above) — the filesystem-diff approach can't distinguish "renderer didn't run" from "renderer re-ran and reproduced identical bytes". It needs the Tooling API task-outcome signal before it's mergeable.

---

yschimke:
Closing. The approach on this branch is disproven and the right one is a different shape, so this is not worth rebasing forward.

Summary of what the branch established, so it isn't lost:

**Two implementations, both wrong, both disproven by end-to-end runs rather than review.**

1. *mtime threshold* — reports anything whose PNG mtime is at or after the invocation start. Broken three ways, all raised in review and all correct: freshness was decided per batch rather than per result (module A's fresh render vouched for module B's stale rows); `buildResults` can rewrite PNGs via the image-size override **after** gradle, laundering a stale file into a fresh-looking one; and flooring the start stamp to whole seconds admitted anything a previous run wrote inside that same second.

2. *pre-run filesystem snapshot* — diff `renders/` before and after the gradle run, per result, taken before `buildResults` can touch anything. Fixes all three of the above, and still fails. Run against `meshcore-mobile:wear`: on a second invocation the render task executes, fails on the same broken preview, and reproduces **byte-identical** PNGs. Nothing changes on disk, the diff sees zero fresh files, and all 29 rows get suppressed — strictly worse than the behaviour it replaced.

The premise both share is the bug: **"no files changed" does not mean "the renderer did not run."** An up-to-date task, a from-cache task, and an idempotent re-render all write nothing while their outputs are perfectly current — and in an iterate loop that is the common case, not an edge case. Every filesystem-observation scheme inherits this, content hashing included.

**The signal has to come from the build.** Whether `:<module>:composePreviewRenderAll` actually executed is a build fact, and the Tooling API already reports it: `GradleConnection`'s progress listener (`GradleConnector.kt:165`) sees `TaskOperationDescriptor` with `FinishEvent`, whose result distinguishes success, up-to-date, from-cache, failed and skipped. Plumbing per-task outcomes through `runTasks` into `RawRenderOutcome` gives per-module currency with no heuristic.

**One piece worth salvaging** if someone picks this up: rows whose render *threw* must stay in the output. They carry no image, sha or `changed`, so nothing stale can mislead, and they are the "which preview is broken?" signal the whole behaviour exists to deliver. My first cut silently dropped them.

**Still live on `main`:** #3059's partial reporting is gated only on "results is non-empty", so on a warm checkout a configuration or compile failure will emit the previous run's images, shas and `changed: false` — "nothing changed" about a build that never ran. That is the original motivating bug and it is currently untracked. Reverting that half of #3059 until a correct implementation lands remains my recommendation.

---

Issue #2883, opened 2026-07-28
figma-svg reuses one captured tree across multi-annotation preview variants


yschimke:
## Summary

Figma SVG data products are being reused across distinct `@Preview` annotations on the same composable. The PNGs correctly reflect each annotation's `uiMode`, device, and font scale, but the per-variant SVGs contain the same captured tree and raster crops.

This is separate from #2855 / #2858: output paths and compare-page pairing are now collision-free, but the content written to those distinct SVG paths is still cross-contaminated.

## Reproduction

`FeedScreenPreview` in compose-samples Jetsnack has default compact, dark compact, and large-font medium annotations.

Published report, generated by compose-preview 0.19.2:

- [Jetsnack compare page](https://github.com/yschimke/compose-samples/blob/6f510ae898ba1a6a5d3fbc1172da9f8c22df8dff/compare.html)
- `Screens/Feed` default medium scores **14.7%**

The following SVGs are structurally identical after normalizing only their relative `.figma-raster` directory names:

- [`ideal__default__compact.svg`](https://github.com/yschimke/compose-samples/blob/6f510ae898ba1a6a5d3fbc1172da9f8c22df8dff/figma/screens-feed/ideal__default__compact.svg)
- [`ideal__default__dark__compact.svg`](https://github.com/yschimke/compose-samples/blob/6f510ae898ba1a6a5d3fbc1172da9f8c22df8dff/figma/screens-feed/ideal__default__dark__compact.svg)
- [`ideal__default__medium.svg`](https://github.com/yschimke/compose-samples/blob/6f510ae898ba1a6a5d3fbc1172da9f8c22df8dff/figma/screens-feed/ideal__default__medium.svg)

All three normalized files have SHA-256:

`1049f63168a5b52edf225b6b7e8fed96e9a3d794b6d5de3dcf8a0865abe5fc10`

Their referenced raster-crop files also have matching hashes across all three directories. All SVGs contain the dark palette (`#121212`, `#232323`, white text), while the default and large-font PNGs are visibly light and have distinct dimensions/content.

The same cluster appears in JetNews:

- [JetNews compare page](https://github.com/yschimke/compose-samples/blob/15962f3fbd29a5268f12b9d99fd9868488145a74/compare.html)
- `Screens/Home feed`: **34.0%**
- `Interests/Screen`: **55.5%**
- `Screens/Article`: **56.7%**

[...]

Issue #2615, opened 2026-07-21
Viewport figma-svg export doesn't apply TransformingLazyColumn edge scaling — items overlap/merge


yschimke:
## Summary

The plain viewport `.svg` export (`compose/figma-svg`, served at `GET /render/<id>.svg`) does not honour Wear `TransformingLazyColumn` (TLC) item scaling. Near the round face's edges, TLC shrinks items via a draw-time `graphicsLayer` scale and compresses the space each item occupies (`transformedHeight`). The layered-SVG exporter walks the `LayoutNode` tree and emits each item's **intrinsic, un-transformed bounds** while honouring the **compressed** placement offsets — so edge-scaled items are drawn at full height at compressed positions and **overlap / visually merge**. The PNG raster is correct because the `graphicsLayer` scale is applied at draw time.

## Repro

`confetti-wear` conferences list, light mode:

- SVG (bug): https://preview.coo.ee/confetti-wear/render/conferences__ideal__default__compact.svg?uiMode=light
- PNG (correct): https://preview.coo.ee/confetti-wear/render/conferences__ideal__default__compact.png?uiMode=light

Rects pulled straight from the emitted SVG:

| Chip | `y` | height | spans |
|------|-----|--------|-------|
| KotlinConf 2023 | 161 | 104 | 161 → 265 |
| Fosdem 2023 | 273 | 104 | 273 → 377 |
| droidcon London | 350 | 104 | 350 → 454 |

Top gap (265→273) is a normal 8px, but `droidcon` starts at y=350 — **27px before** Fosdem ends at 377, so the bottom two chips overlap. Note all three rects are height **104**: the per-item scale is not baked into the emitted geometry, only the compressed offsets are.

## Root cause

This is the exact case already documented in [`docs/design/SCROLLING_SVG.md`](../blob/main/docs/design/SCROLLING_SVG.md) (Wear section, reason #3): *"TransformingLazyColumn scales items toward the edges … Captured in place they'd be at varying, wrong-for-a-flat-list scales."* The `?scroll=long` path solves it by rendering with `LocalReduceMotion(true)` (via the `WearReduceMotionLocal` seam), which flattens the transform so items stack at natural size. The **plain viewport** `.svg` lane does not apply reduce-motion, so it inherits the scaled/compressed geometry and the items overlap.

## Possible directions

[...]

PR #3065, opened 2026-07-31
fix(figma-svg): clip to the rendered box and keep text on clip-edge nodes

### LINKED ISSUE #3056: figma-svg uses shared-element lookahead bounds instead of the rendered scroll viewport
## Summary

`figma-svg` exports the lookahead/shared-element height instead of the rendered `heightIn`/scroll viewport for a `sharedBounds` container. Below-fold children are therefore visible in the SVG even though the PNG clips them.

## Reproduction

Jetsnack `Catalog/Filter screen`, compose-preview 0.19.12:

- compare: https://htmlpreview.github.io/?https://github.com/yschimke/compose-samples/blob/design-artifacts/jetsnack/compare.html
- direct SVG: https://github.com/yschimke/compose-samples/blob/96d6e55157b69cb1df629664feb569c7ce87a53c/figma/catalog-filter-screen/ideal__default.svg
- score: 64.7%
- run: https://github.com/yschimke/compose-samples/actions/runs/30614185745

The production component applies this chain to the centered filter column:

```kotlin
padding(16.dp)
    .clip(MaterialTheme.shapes.medium)
    .sharedBounds(
        ...,
        resizeMode = SharedTransitionScope.ResizeMode.RemeasureToBounds,
        clipInOverlayDuringTransition = OverlayClip(MaterialTheme.shapes.medium),
    )
    .wrapContentSize()
    .heightIn(max = 450.dp)
    .verticalScroll(rememberScrollState())
    ...
    .skipToLookaheadSize()
```

At density 2.625, the Android PNG's visible filter surface is about 1181px high, exactly 450dp, from roughly y=460 to y=1641. The SVG instead emits the owning clip as:

```svg
<clipPath id="clip-0">
  <rect x="105" y="502" width="840" height="1712" rx="52.5" ry="52.5"/>
</clipPath>
```

1712px is about 652dp. As a result, Category chips, the slider, and Lifestyle chips that are below the rendered scroll viewport remain visible in the SVG. The outer canvas clamp from #3043 cannot fix this because the leaked children are still inside the 1050×2100 rendered frame.

## Expected

The SVG clip must use the final rendered scroll-container bounds (the 450dp `heightIn` result), not the lookahead/shared-element content bounds. Please add an Android end-to-end fixture combining `SharedTransitionLayout`, `sharedBounds`/`skipToLookaheadSize`, a height-limited `verticalScroll`, and below-fold children.

### LINKED ISSUE #3057: figma-svg drops visible text from lazy-list items clipped at the viewport edge
## Summary

In `figma-svg`, a partially visible lazy-list item can retain its surface and raster/image leaves while its visible text nodes are emitted as empty groups. The Android PNG correctly paints the clipped portion of the text.

## Reproductions

Fresh compose-preview 0.19.12 artifacts from:

- `previews` commit: `3642d5d680a5dc3436466be9586ea1ca4a767216`
- run: https://github.com/yschimke/compose-samples/actions/runs/30614185745

### Jetchat

Compare: https://htmlpreview.github.io/?https://github.com/yschimke/compose-samples/blob/design-artifacts/jetchat/compare.html

Both `Conversation/Screen` (90.5%) and `Messages/List` (91.9%) start with a message clipped by the top of the `LazyColumn`. In the PNG, the visible lower lines of that message are painted. In the SVG, the lavender message bubble survives but is blank.

In the `Conversation/Screen` SVG, the partly visible bubble is retained from y=-134 through y=378, but its text group is empty:

```svg
<path d="M204.5,-134 ... V325.5 ... 378 ..." fill="#E2E1EC"/>
<g id="Empty">
</g>
```

Artifact: https://github.com/yschimke/compose-samples/blob/8bf0cb61ef7ba9be41423fe06235e47a22155f39/figma/conversation-screen/ideal__default__light__compact.svg

### JetNews

Compare: https://htmlpreview.github.io/?https://github.com/yschimke/compose-samples/blob/design-artifacts/jetnews/compare.html

`Screens/Home feed` (88.5%) shows the complementary bottom-edge case: the next post's thumbnail/bookmark reaches the viewport edge in both outputs, and its title is visible in the PNG, but the SVG retains empty text groups for that row.

Artifact: https://github.com/yschimke/compose-samples/blob/bb2b76bad8f51ecb637e0757edc7ce2ee73775a1/figma/screens-home-feed/ideal__default__compact.svg

## Expected

Visibility filtering should intersect each text/draw node with the rendered viewport. A node that straddles an edge must be retained and clipped, just like the item's surface and image; it must not be discarded merely because its origin or un-clipped bounds fall outside the frame.

Please add an Android end-to-end fixture with first/last lazy-list items deliberately straddling the viewport edge and assert that visible text remains in the SVG.

yschimke:
Fixes #3057, fixes #3056. Two of the four compose-samples preview reports in this batch; see **Not fixed** below for #2615 and #2883.

## Summary

**#3057 — clipped lazy-list items lost their text.** The two producers disagree about clipping: the layout-inspector records every node's **unclipped** box (`localBoundingBoxOf(clipBounds = false)`), while a semantics node's `boundsInRoot` is **clipped** by its ancestors. A fully visible row presents identical boxes, so `assignTextToLayers` matches. A row straddling the viewport edge presents boxes that differ by the whole clipped-away strip — far past the 2px matching tolerance — so its text attached to nothing and the row exported as an empty `<g>` while the PNG painted the visible lines. Each candidate now also offers its clip-intersected box (clips threaded through the walk and intersected as they nest, the same way the layers themselves are clipped), so an edge row matches without loosening the tolerance for anything else.

**#3056 — shared-element lookahead bounds leaked scroll content.** A node's `bounds` come from its innermost coordinator, and under `sharedBounds(… RemeasureToBounds) … .heightIn(max = …).verticalScroll(…).skipToLookaheadSize()` that coordinator reports the lookahead/content extent, not the box the frame was drawn in. Captured in the new Android fixture, the node reports 480px tall while every modifier coordinator *outside* the scroll — the `clip` among them — still reports the real 240px viewport. The clipping `graphicsLayer`'s own box now wins where it is a strict subset of the node's bounds, and the measured-`size` fill growth is suppressed for such a node so it cannot grow straight back.

Also repairs a pre-existing compile break in `FigmaLayeredSvgTest` (positional `ComposeSemanticsTextSpan(…, "16.0sp", "monospace", 400)` args landing on `fontSizePx`/`fontFamily`/`fontWeight`), which had `:data-layoutinspector-core:test` red on `main` before this change.

## Visual evidence

Rendered from the new Android end-to-end fixtures' actual `compose-figma.svg` output (base vs. this branch), rasterised with the bundled Chromium.

### #3057 — first `LazyColumn` row straddling the top edge

Before: the lavender row survives but is blank. After: "Alpha row" is painted and clipped at the edge, exactly as the PNG paints it.

| before | after |
|---|---|
| before | after |

### #3056 — `heightIn(max = 240.dp)` scroll surface under `sharedBounds`

Before: the surface is exported 480px tall (the scroll content), with the below-fold rows filling the extra space. After: it is the 240px rendered viewport, and the three visible rows keep their labels while the rest is clipped away.

| before | after |
|---|---|
| before | after |

## Not fixed in this PR

- **#2615 (Wear transformed content nearly transparent)** — not reproduced, so not fixed. A new Android fixture (`FigmaSvgWearTlcAlphaTest`) drives the production shapes the issue names — `ButtonGroup` + `FilledIconButton`s, a transformed `Button`, transformed `TitleCard` rows inside `AppScaffold`/`ScreenScaffold`/`TransformingLazyColumn`, at Wear's 2x density — and the reflective `graphicsLayer` alpha evaluation produces sane edge fades throughout (~0.53–0.98), never the `0.0`/`0.04` Jetcaster Wear reports. The fixture is committed as the standing guard the issue asks for (it asserts no group is exported below 0.1 opacity), but the Jetcaster-specific trigger is still unidentified. Reproducing it likely needs the real Jetcaster Wear composition rather than a synthetic list.
- **#2883 (breakpoint SVGs reuse default geometry)** — not fixed. The per-breakpoint vector is chosen by `image.previewId`, which `bridgeLivePreviewIds`/`pickVariantId` resolve by scoring a candidate's `@Preview(widthDp = …)` against the spec breakpoint. When a function has only one candidate `pickVariantId` returns it unconditionally, so a breakpoint render generated from a single preview candidate is handed the default annotation's vector. Confirming whether the bundle's candidates genuinely carry no `widthDp` for these Jetsnack rows (versus the breakpoint render coming from the private `@design-parity/catalog-export` engine with no vector of its own) needs a real catalog bundle, which isn't reachable from this repo alone.

## Test plan

- [x] `:data-layoutinspector-core:test` — 232 tests green, including the new `FigmaSvgClippedNodeTest` unit coverage for both fixes (and the repaired `FigmaLayeredSvgTest`, which did not compile before)
- [x] `:data-layoutinspector-connector:test` — green
- [x] `:renderer-android:testDebugUnitTest` — green, including the three new end-to-end fixtures (`FigmaSvgClippedLazyItemTextTest`, `FigmaSvgSharedBoundsScrollClipTest`, `FigmaSvgWearTlcAlphaTest`)
- [x] `ktfmtCheckAll` — green

https://claude.ai/code/session_013QDREaaoYqh2SKdz6aELeB

---
_Generated by [Claude Code](https://claude.ai/code/session_013QDREaaoYqh2SKdz6aELeB)_


yschimke:
Pushed `b014680` — both Codex review points were right and are fixed:

- **Reflectively detected clip flags** — `toWireModifier` now persists a reflected `clip` field on a `graphicsLayer` when the inspector element is absent, alongside the `BackgroundElement`/`brush` carry directly above it. Without it `clipModifierBounds` finds no clipping coordinator on desktop/release captures and the lookahead-inflated box survives on exactly the builds the fix targets.
- **Rendered clip box for descendant text** — text candidate collection now derives its child clip from `clipModifierBounds() ?: n.bounds`, the same effective box `toLayer` clips with, so a descendant row straddling the real viewport edge doesn't get its candidate intersected against the larger lookahead box. New `textOnARowStraddlingALookaheadInflatedNodesRealViewportStillAttaches` covers exactly that combination — it fails without either half.

## `Module Unit Tests` failure

Not caused by this branch. That job went red four ways, none of them in the changed code path:

- `:daemon:android:testDebugUnitTest` — 27 failures, every one `NullPointerException at Typeface.java:928` / `Typeface.java:1005`, and the executor then died with `SIGSEGV in libandroid_runtime.so` (exit 134). A native font-init crash, not an export change.
- `:renderer-xr:testDebugUnitTest` — `SubspaceWorldIntegrationTest`, `XrSubspaceRendererTest`. Unrelated subsystem.
- `:daemon:core:test` — executor exit 1.
- `:daemon:desktop:test` — `DesktopLayoutInspectorTest.captures_an_imagevector_icon_as_editable_vector_paths` and `TcpPanelFigmaSvgTest.the tcp connect panel exports no zero-area layers or raster targets`. These two are in figma-svg territory, so I checked them directly: both reproduce locally **with `FigmaSvgModel.kt` restored from `origin/main`**, identically. Pre-existing on `main`.

Separately, `FigmaLayeredSvgTest` did not compile on `main` before this branch (positional `ComposeSemanticsTextSpan(…, "16.0sp", "monospace", 400)` args landing on `fontSizePx`/`fontFamily`/`fontWeight`), so `:data-layoutinspector-core:test` was red there too; that is repaired here.

Green locally on this branch: `:data-layoutinspector-core:test`, `:data-layoutinspector-connector:test`, `:renderer-android:testDebugUnitTest`, `ktfmtCheckAll`.

---

yschimke:
`Module Unit Tests` on `b014680`: same four-way failure as on `e64cb72`, nothing new from this branch.

Full failing set, compared run-to-run:

| task | failures | changed by this branch? |
|---|---|---|
| `:daemon:android:testDebugUnitTest` | 27, all `NullPointerException at Typeface.java:928`/`:1005`, then the executor dies `SIGSEGV in libandroid_runtime.so` (exit 134) | no — identical count and set on both commits |
| `:renderer-xr:testDebugUnitTest` | `SubspaceSceneRecorderTreeTest`, `SubspaceWorldIntegrationTest`, `XrSubspaceRendererTest` | no — unrelated subsystem |
| `:daemon:core:test` | executor exit 1 | no |
| `:daemon:desktop:test` | `DesktopLayoutInspectorTest.captures_an_imagevector_icon_as_editable_vector_paths`, `TcpPanelFigmaSvgTest.the tcp connect panel exports no zero-area layers or raster targets`, `RenderEngineLottieScrubTest.scrubPersistsAcrossRenderWithoutOverride` | no — see below |

The two figma-svg desktop tests reproduce with `FigmaSvgModel.kt` restored from `origin/main`, so they're pre-existing. The third, `RenderEngineLottieScrubTest`, is new to this run only: it passed on `e64cb72` (that run was 2 failures, this one 3), it passes locally on this branch (`:daemon:desktop:test` → 202 tests, the same 2 failures), and Lottie frame scrubbing shares no code with clip geometry or modifier serialisation. Reading it as a flake.

I don't think there's a green path for this job from this branch — the `daemon:android` half is a native font-init crash that predates it. Happy to be pointed at it separately if it's actually expected to be green on `main`.

---

Issue #3015, opened 2026-07-30
Playground Phase 2 — Android/Robolectric backend


yschimke:
**Epic — playground roadmap ([`docs/design/PLAYGROUND.md §7` item 2](../blob/main/docs/design/PLAYGROUND.md)).** Phase 1 shipped in 0.19.9 (#3010).

## Goal

The same compile→token→redeem loop as Phase 1, but against the **Android** backend: a Robolectric daemon (2 permits) compiling Compose Android snippets against an Android catalog classpath (e.g. `wear-m3`).

## Shape (small stacked PRs, mirroring Phase 1)

- [x] Resolve an Android catalog liveBundle → compile classpath (Android variant of `PlaygroundCatalogClasspath`). _(#3040 — `--playground-android-bundle` resolves the Android compile classpath via `PlaygroundCatalogClasspath.resolve`.)_
- [x] Compile the snippet for the Android target and render via the Robolectric daemon path rather than the desktop `ImageComposeScene` path. _(#3044 — the **first-frame** render on the Android/Robolectric daemon; the held **live-session** render is the Stage-2 follow-up below.)_
- [x] Admission control: 2 permits via the existing `LiveSeatLimiter` (admission control, **not** a sandbox). _(#3054 — automatic via the registered `ServeSessionState.liveSeatWeight` = `ANDROID_LIVE_SEAT_WEIGHT` (2); no route changes.)_
- [x] Wire the Android mode into `PlaygroundApi` `confType` resolution (`compose-android`). _(#3040; #3044 gates it on the render backend actually coming up.)_
- [x] Tests: Android-mode compile + render + token round-trip. _(#3044 — compile + first-frame render against a fake session; #3054 — `PlaygroundRedeemServiceTest` covers the token **redeem** round-trip against a real registry; #3066 — a browser e2e drives editor→compile→first-frame→`/pg/` redeem end to end.)_
- [x] **Stage-2 live redemption:** `/pg/<token>` redeems an ANDROID snippet into a held Robolectric streaming session (the `input` protocol), not just a still first frame. _(#3054 — `PlaygroundRedeemService` + `ServeBundleDaemon.materializePlaygroundSnippet` register the snippet as a live session and hand off to the existing viewer/WS/input lane.)_

## Dependencies / relationship to other phases

- Builds only on the merged Phase 1 loop; **independent of Phase 3** (RC mode) — the two can proceed in parallel.
[...]

PR #3066, opened 2026-07-31
fix(playground): redeem /pg/ on token-gated hosts, with a browser e2e

yschimke:
## Summary

Adds the first **browser end-to-end** coverage of the Kotlin playground (`docs/design/PLAYGROUND.md`, epic #3015) — and that e2e immediately caught a real bug, now fixed here.

### The bug it caught

Stage-2 redemption was mounted at `GET /pg/{token}`, but on a token-gated host the access token also rides as `?token=…`. Ktor's `call.parameters` merges path + query, so `call.parameters["token"]` resolved to the **access token** instead of the `pg_…` path segment, failed the shape check, and 404'd **every** redemption as NotFound. Because the playground lane is refused under `--public`, a token-gated host is the *only* way it runs — so live redemption was broken for all real deployments. The existing `serve-lanes` suites never caught it: they run `--public`, so there's no `?token=` to collide.

Fix: rename the path segment to `{pgToken}` (can't be shadowed) and read it explicitly. Plus a token-gated `PlaygroundRoutingTest` case (mint → redeem with `?token=`) that asserts the 302 to the viewer — it 404'd before the fix.

### The e2e (`playground.spec.mjs`)

Drives the whole chain in Chromium against a live, daemon-backed serve:

1. **Editor page** serves its controls and offers the Android compile mode.
2. **Compile → first frame** — running the default `@Preview` snippet reaches `Done.` via `POST /api/1/compiler/run` (BTA compile → Android/Robolectric first frame) and the response carries an inline `data:image/png` still (asserted, not treated as best-effort).
3. **Live redemption** — the returned `/pg/<token>` capability 302-redirects into the ordinary viewer (`/<sessionId>/p/<previewId>`), not the styled 404.

Reuses the Android/Robolectric backend (the default snippet declares an Android `@Preview`): `playground-boot.sh` packs `:samples:android-live-lane` and boots `serve --playground-android-bundle` **token-gated** (so the spec carries `?token=` — which is exactly what surfaced the bug). A `playground-e2e` CI job mirrors the Android render lane; on failure it dumps the serve log into the job log so a redemption reason is diagnosable without the artifact.

### Review follow-ups (Codex)

- Assert the first-frame `#pg-image` is a real PNG data URI (not just that a token minted).
- The redemption test **fails** rather than green-skips on a non-`Done.` compile.
- Trigger the workflow on `ServeCommand.kt` (owns `openPlaygroundService` + the daemon-opener wiring, outside `serve/**`).

## Test plan

- [x] `./gradlew :cli:test --tests "*PlaygroundRoutingTest*"` — passes, incl. the new gated-redemption case (which 404'd before the route-param fix).
- [x] `node --check` / `prettier --check` clean on the harness files; YAML + `package.json` parse.
- [ ] The `playground-e2e` CI job drives the full chain green (compile + first-frame already pass; redemption fixed by this PR).

Visual note: no repo-owned surface changes — the rendered PNG is the user's snippet. Non-visual for repo purposes.

---

### DIFF
--- a/.github/workflows/serve-lanes-e2e.yml
+++ b/.github/workflows/serve-lanes-e2e.yml
@@ -29,6 +29,9 @@ on:
       - "vscode-extension/preview-harness/serve-lanes.*"
       - "vscode-extension/preview-harness/serve-lanes-boot.sh"
       - "vscode-extension/preview-harness/serve-android-lane-boot.sh"
+      # The playground e2e (editor → compile → first-frame → /pg live redemption).
+      - "vscode-extension/preview-harness/playground.*"
+      - "vscode-extension/preview-harness/playground-boot.sh"
       # The Android lane's fixture and the detached-render Application pin it guards.
       - "samples/android-live-lane/**"
       - "cli/src/main/kotlin/ee/schimke/composeai/cli/AndroidBundleResources.kt"
@@ -48,6 +51,9 @@ on:
       - "vscode-extension/preview-harness/serve-lanes.*"
       - "vscode-extension/preview-harness/serve-lanes-boot.sh"
       - "vscode-extension/preview-harness/serve-android-lane-boot.sh"
+      # The playground e2e (editor → compile → first-frame → /pg live redemption).
+      - "vscode-extension/preview-harness/playground.*"
+      - "vscode-extension/preview-harness/playground-boot.sh"
       # The Android lane's fixture and the detached-render Application pin it guards.
       - "samples/android-live-lane/**"
       - "cli/src/main/kotlin/ee/schimke/composeai/cli/AndroidBundleResources.kt"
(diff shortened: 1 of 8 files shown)


yschimke:
The e2e paid off immediately — it caught a **real redemption bug**, now fixed in `2e01642`.

**Root cause:** the route was `GET /pg/{token}`, but on a token-gated host the access token also rides as `?token=…`. Ktor's `call.parameters` merges path + query, so `call.parameters["token"]` resolved to the *access* token, failed the `pg_…` shape check, and 404'd **every** redemption as NotFound. Since the playground lane is refused under `--public`, a token-gated host is the only way it runs — so live redemption was broken for all real deployments. The `serve-lanes` suites never caught it because they run `--public` (no `?token=` to collide). Fix: rename the path segment to `{pgToken}` and read it explicitly; plus a token-gated `PlaygroundRoutingTest` case that asserts the 302 (it 404'd before the fix).

Diagnosis was driven entirely from the e2e failure — the reason (a 404 with no `[playground live]` log ⇒ NotFound, not Unavailable) is why the redemption test now hits `/pg/` raw and prints the status + body, and the CI job dumps the serve log on failure.

Codex suggestions, all folded in:
- **#107 (assert the first frame):** the first-frame test now asserts `#pg-image` is a `data:image/png` URI — proving the daemon actually rendered (it does; that test passes green in CI), not just that a token minted.
- **#122 (fail, don't skip):** the redemption test now asserts `Done.` instead of green-skipping a non-`Done.` compile.
- **#34 (trigger on `ServeCommand.kt`):** added to both the push and pull-request path filters.

---

yschimke:
Status: green except one known flake. The new **`playground-e2e`** job passes (all three tests — editor, compile→first-frame, `/pg/` redemption→viewer), and `Serve render lanes`, `ktfmt`, `Build CLI`, `Build Samples`, and the daemon harnesses are all green on `2e01642`.

The only red is **`Module Unit Tests`** — the pre-existing SIGABRT flake in `:daemon:core`/`:daemon:desktop`/`:daemon:android`/`:renderer-xr`, none of which this `cli`+test+harness diff touches. Same flake #3050/#3052/#3054 merged through. I've re-run it three times; it's an environment flake, not this change. Not something I can fix from here — flagging it as the merge blocker rather than chasing it further.

---

PR #3068, opened 2026-07-31
feat(discovery): warn when a preview installs a theme under declared theme catalogs

yschimke:
## The gap

A `@ThemeCatalog` / `@WearThemeCatalog` provider is applied by **wrapping** the preview function. A theme installed *inside* the preview body composes within that wrapper and shadows it, so every entry in the viewer's **Theme** select renders byte-identical pixels — the catalog looks like it has a live theme axis and doesn't.

That is exactly what shipped in `confetti-wear` and `confetti-mobile` (fixed in joreilly/Confetti#1753). All five conference themes returned the same PNG, including the *Typography* specimen that is meant to show a different type family per conference, because every catalog preview opened with `ConfettiThemeFixed { … }`.

Nothing in this pipeline noticed. The one surface that would have shown it is the one surface the existing checks don't cover: the synthetic per-theme specimen sheets wrap a **canned grid** rather than an app preview, so they kept differing correctly while every real preview was pinned.

## The check

Discovery already walks a preview's bytecode for `PreviewTargetInference`, so this reuses that walker rather than adding a render pass or a manifest field:

- Follow the preview body (the existing walker already descends into its own lambdas, so `AppTheme { … }` is seen from the preview's root method), then into module-local methods, bounded to 6 hops. Dependency JARs are never followed.
- Look for a `MaterialTheme` install — `material`, `material3`, and both Wear variants.
- Report only when the module declares theme providers. Without them there is no theme axis, and an app preview installing its own theme is just an app preview.

Reported through the existing `Outcome.warnings` channel, so no schema change and no new task: it surfaces on `composePreviewDiscover` like `CheckDebugPreviewsTask`'s warning.

**Warning, never an error.** Pinning a preview to one theme is legitimate — a per-identity theme-foundation sticker documents exactly one theme on purpose. The check can't distinguish that from the bug, so it reports and lets the author decide.

Two precision details, both pinned by tests:

- `MaterialTheme.colorScheme` **reads** compile to `…material3.MaterialTheme.getColorScheme` — different owner (no `Kt` facade) and method. Reading the ambient theme is what a well-behaved preview does; only installing one shadows the wrapper.
- `MaterialTheme` has defaulted parameters, so all but the fully-specified call site compiles to the synthetic **`MaterialTheme$default`** bridge — and `AppTheme { content }` is by far the common shape. Matching the bare name fired on almost nothing; the functional test caught this before it shipped.

## Test plan

- `:gradle-plugin:preview-discovery:test` — 8 new unit tests, all pass. Cover the install-vs-read distinction, the `$default` bridge, the "no themes declared → silent" gate, and the warning's content/truncation.
- `:gradle-plugin:functionalTest --tests "*DiscoveryFunctionalTest*"` — 37/37 pass, including two new end-to-end cases that compile a real Compose project against real `compose.material3`:
  - **positive** — two `@ThemeCatalog` providers plus `ShadowedPreview → AppTheme → MaterialTheme`; asserts the warning names `test.ShadowingPreviewsKt.ShadowedPreview` and the `AppTheme` hop, and that discovery still succeeds and still publishes the preview.
  - **negative control** — the identical body-installed theme with *no* theme catalog declared; asserts the warning does not appear.
- ktfmt clean.

Text-only PR: this changes a discovery warning, not rendered output.

https://claude.ai/code/session_01LhWssk26QYPNjkx7vcbKC6

---
_Generated by [Claude Code](https://claude.ai/code/session_01LhWssk26QYPNjkx7vcbKC6)_

### DIFF
--- a/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewDiscovery.kt
+++ b/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewDiscovery.kt
@@ -406,6 +406,10 @@ object PreviewDiscovery {
     val rawShapeCatalogTokens = mutableListOf<RawCatalogToken>()
     // `@ThemeCatalog`-annotated `PreviewWrapperProvider` classes → one theme catalog sheet each.
     val rawThemeCatalogs = mutableListOf<RawThemeCatalog>()
+    // (declaring class, method) for every method that produced at least one preview. Fed to
+    // [PreviewThemeShadowing] after the class walk, which needs both the theme providers (collected
+    // in the same pass) and the previews before it can say anything.
+    val previewMethods = mutableListOf<Pair<ClassInfo, MethodInfo>>()
 
     if (classpath.isNotEmpty()) {
       ClassGraph()
@@ -478,6 +482,7 @@ object PreviewDiscovery {
               for (ann in annotations) {
                 annotationFqnCounts.merge(ann.name, 1, Int::plus)
               }
+              val previewCountBefore = previews.size
               discoverFromMethod(
                 classInfo,
                 method,
@@ -489,6 +494,12 @@ object PreviewDiscovery {
                 warnings,
                 catalogGroupsByFile,
               )
+              // Remember the methods that actually yielded previews, so the theme-shadowing check
+              // below walks exactly those rather than re-deriving "is this a preview?" (which
+              // multi-preview meta-annotations make non-trivial).
+              if (previews.size > previewCountBefore) {
+                previewMethods += classInfo to method
+              }
             }
             // `@ColorCatalog` / `@TypographyCatalog` / `@ShapeCatalog` design tokens: an annotated
             // `Color` / `TextStyle` / `Shape` (single token) or `ColorScheme` / `Typography` /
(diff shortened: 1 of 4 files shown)


yschimke:
CI red on **Module Unit Tests** — failing on `main` too, and not caused by this change. Will re-run when the base branch recovers.

Same job is the only failed job on `main`'s last completed run ([30630708627](https://github.com/yschimke/compose-ai-tools/actions/runs/30630708627), `16e4e130`), with an identical signature:

| | main `16e4e130` | this PR `b270efe` |
|---|---|---|
| `:daemon:android:testDebugUnitTest` | 70 tests, **23 failed**, 1 skipped | 70 tests, **23 failed**, 1 skipped |
| Root cause | `NullPointerException` @ `Typeface.java:928` ← `RobolectricHost.kt:1820` | identical |
| JVM | `SIGSEGV` in `libandroid_runtime.so+0x7b6ba9`, process abort (exit 134) | identical |
| `:daemon:desktop:test` | `TcpPanelFigmaSvgTest > …no zero-area layers or raster targets` | identical |
| `:daemon:core:test` | exit 1 | identical |
| `:renderer-xr:testDebugUnitTest` | failing tests | identical |
| `:data-layoutinspector-core:compileTestKotlin` | **compile error** | not present |
| Total | 5 failures | 4 failures |

`main` has one failure this branch doesn't. Every failing test is in `:daemon:*`, `:renderer-xr:*` or `:data-layoutinspector-core` — none of which this PR touches. The diff is confined to `preview-discovery` (one new file plus a warning append in `PreviewDiscovery`) and a functional test; it adds no API, no schema field, and nothing on any daemon code path. A composition of `Typeface` NPEs and a native `libandroid_runtime` segfault has no route back to a ClassGraph/ASM bytecode walk in discovery.

The parts this PR does own are green locally and in CI:

- `:gradle-plugin:preview-discovery:test` — 8/8, including the new `PreviewThemeShadowingTest`
- `:gradle-plugin:functionalTest --tests "*DiscoveryFunctionalTest*"` — 37/37, including the two new theme-shadowing cases
- ktfmt clean

Not pushing a fix: the breakage is on the base branch and repairing `:daemon:android`'s Robolectric font init / the XR and figma-svg tests is well outside this change. Happy to look at the Typeface crash separately if it isn't already known.

---

PR #3071, opened 2026-07-31
fix(daemon): restore the capture overload the forensic dump looks up reflectively

yschimke:
`ClassloaderForensicsDaemonTest > captureDaemonClassloaderForensics` has never passed. It's one of the failures in the `Module Unit Tests` job that is currently red on `main`.

## Root cause

`RobolectricHost.runForensicDump` resolves the forensics library reflectively — it lives on the sandbox runtime classpath, not the host's compile classpath — and asks for a 4-arg method:

```kotlin
forensicsClass.getMethod(
  "capture",
  List::class.java,
  Class.forName("…RobolectricConfigSnapshot", true, effectiveLoader),
  String::class.java,
  java.io.File::class.java,
)
```

But `capture` carries a defaulted trailing parameter, and **Kotlin does not emit an overload for a defaulted parameter**. Before this change the compiled class had exactly two entry points:

```
public final void capture(List, RobolectricConfigSnapshot, String, File, okio.FileSystem);
public static void capture$default(ClassloaderForensics, List, …, FileSystem, int, Object);
```

No 4-arg `capture` — so the lookup threw at runtime, with nothing at compile time to catch it:

```
java.lang.NoSuchMethodException: …ClassloaderForensics.capture(
  java.util.List, …RobolectricConfigSnapshot, java.lang.String, java.io.File)
	at java.base/java.lang.Class.getMethod(Class.java:2229)
	at …RobolectricHost$SandboxRunner.runForensicDump(RobolectricHost.kt:2212)
```

The library, the reflective caller and the test all landed together in #2985 (`git log -S` shows no later edit to either side), so this shipped broken rather than regressing later — which also means #2985 merged with this test red.

## Fix

`@JvmOverloads` on `capture`. It materialises exactly the trailing-`fileSystem`-omitted overload the reflective caller names, and keeps doing so if further defaulted parameters are appended later:

```
public final void capture(List, RobolectricConfigSnapshot, String, File, okio.FileSystem);
public final void capture(List, RobolectricConfigSnapshot, String, File);   ← restored
public final void capture(List, String, File);
public static void capture$default(…);
```

Chosen over patching the reflective call to name the 5-arg method, because it pins the ABI the reflective consumer depends on instead of leaving it implicit — the next signature change would otherwise re-break it silently.

The KDoc now records why the annotation is load-bearing. The underlying trap is worth naming: `fileSystem` was added to satisfy the injected-`FileSystem` convention in `docs/AGENTS.md`, which is source-compatible for ordinary Kotlin callers but **silently binary-incompatible for a reflective one**. Any future convention sweep over this class needs `ClassloaderForensicsDaemonTest` re-run, since it is the only thing exercising the reflective path.

## Test plan

- [x] Reproduced the failure locally on `main` before the change
- [x] `javap` before/after confirms the 4-arg overload is absent, then present
- [x] `./gradlew :daemon:android:testDebugUnitTest --tests "*ClassloaderForensicsDaemonTest*"` — **BUILD SUCCESSFUL** (failed before)
- [x] `./gradlew :daemon:core:compileKotlin` — green
- [x] ktfmt applied

## Scope

This fixes **one** of the failures in `Module Unit Tests`. The larger group in that job — ~23 of 70 in `:daemon:android`, plus `:daemon:desktop`, `:daemon:core` and `:renderer-xr` — is a separate crash: `NullPointerException at Typeface.java:928` during Robolectric native-runtime font loading. Notes on that in the comment below; it is not addressed here.

Visual evidence: N/A — diagnostic library, no rendered output.

---
_Generated by [Claude Code](https://claude.ai/code/session_014NDRHMxYyfVYUF3pih7LjL)_

### DIFF
--- a/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/forensics/ClassloaderForensics.kt
+++ b/daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/forensics/ClassloaderForensics.kt
@@ -65,7 +65,24 @@ object ClassloaderForensics {
    * @param contextHint short string identifying the configuration ("standalone-control" /
    *   "daemon-subject" / etc.). Recorded verbatim in the dump.
    * @param out target file. Parent directories are created if absent. Existing file is overwritten.
+   *
+   * **`@JvmOverloads` is load-bearing — do not remove it.** `RobolectricHost.runForensicDump` calls
+   * this method *reflectively* (`getMethod("capture", List, RobolectricConfigSnapshot, String,
+   * File)`), because the forensics library is on the sandbox runtime classpath rather than the
+   * host's compile classpath. Kotlin does not emit an overload for a defaulted parameter: without
+   * this annotation the only JVM entry points are the full 5-arg `capture` and the synthetic
+   * `capture$default`, so that lookup throws `NoSuchMethodException` at runtime and nothing at
+   * compile time catches it. `@JvmOverloads` materialises the trailing-`fileSystem`-omitted
+   * overload the reflective caller names, and keeps doing so if further defaulted parameters are
+   * appended.
+   *
+   * The same trap applies to the `fileSystem` parameter itself: adding it to satisfy the injected-
+   * `FileSystem` convention in `docs/AGENTS.md` is source-compatible for ordinary Kotlin callers
+   * but silently binary-incompatible for a reflective one. Any future signature change here needs
+   * `ClassloaderForensicsDaemonTest` re-run, since that is the only thing that exercises the
+   * reflective path.
    */
+  @JvmOverloads
   fun capture(
     surveySet: List<String>,
     robolectricConfig: RobolectricConfigSnapshot? = null,


yschimke:
Notes on the *other* `Module Unit Tests` failure, which this PR does not fix. Recording what's been ruled out so the next person doesn't repeat it.

## The crash

```
java.lang.NullPointerException: Cannot read field "mStyle" because "family" is null
	at org.robolectric.nativeruntime.DefaultNativeRuntimeLoader.lambda$ensureLoaded$0(…:253)
	at org.robolectric.nativeruntime.DefaultNativeRuntimeLoader.injectAndLoad(…:178)
	at org.robolectric.android.internal.AndroidTestEnvironment.setUpApplicationState(…:162)
```

`setSystemFontMap` calls `Typeface.create(familyName, style)` for the default families; `family` is null, so the system font map came back without them. It surfaces as `RobolectricHost.bootSlotWithRetries` giving up after 3 attempts, and further downstream as the `SIGSEGV in libandroid_runtime.so` / exit-134 that kills the `:daemon:android` test JVM.

Reproduces locally, deterministically — not a CI-only flake:

```
./gradlew :daemon:android:testDebugUnitTest --tests "*AndroidRecordingSessionTest*"
→ 34 tests completed, 3 failed   (every test that boots a sandbox)
```

## Ruled out

**Not the environment.** `:samples:android:composePreviewRenderAll` succeeds in the same container, same Robolectric 4.17-beta-2, and produces **124 PNGs** through the native path. Robolectric native graphics and font loading work fine here.

**Not the SDK level.** The working sample render resolves to `sdk=35` — the same level the daemon pins in `RobolectricHost.ANDROID_SDK`. Bumping the daemon to 36 doesn't help and makes it worse: Robolectric 4.17-beta-2 rejects 36 outright with `UnsupportedOperationException at DefaultSdkProvider.java:174`.

**Not the `--add-opens` set.** Applying `AndroidPreviewClasspath.buildJvmArgs()`'s five flags to the daemon's unit-test JVM leaves the NPE unchanged.

**Not the Robolectric system properties.** Adding `robolectric.graphicsMode=NATIVE`, `looperMode=PAUSED`, `conscryptMode=OFF`, `pixelCopyRenderMode=hardware` — the "third independent config channel" from `buildSystemProperties` — also leaves it unchanged.

## Where the difference actually lives

Two configurations boot a Robolectric sandbox. One works, one doesn't:

| | sandbox config | native fonts |
|---|---|---|
| `:samples:android:composePreviewRenderAll` | generated `renderer/robolectric.properties`: `application`, `sdk=35`, `graphicsMode=NATIVE`, `shadows=…`, `instrumentedPackages=…` | ✅ |
| `:daemon:android:testDebugUnitTest` | generated `daemon/robolectric.properties`: **`application` only** — `sdk`/`graphicsMode` come from `@Config`/`@GraphicsMode` on `SandboxHoldingRunner` | ❌ |

Note also that the **`Daemon Harness (android, real renderer)` CI job passes**. That runs the daemon with a real renderer, but launches it as its **own JVM process** configured by the plugin, whereas these unit tests boot `RobolectricHost` **in-process** in Gradle's plain test JVM. So the daemon sandbox is fine when launched the way production launches it; it's the in-process unit-test JVM that fails.

Two threads worth pulling, in order:

1. **Boot thread.** `RobolectricHost.bootSlotWithRetries` boots the sandbox on a bare `Thread` (`bootSlotWithRetries$lambda$1` → `Thread.run`). The render path boots on the JUnit runner thread. Native runtime loading is a plausible candidate for thread-context sensitivity, and this is the sharpest structural difference between the two.
2. **The remaining properties gap** — the daemon's file omits `shadows=`/`instrumentedPackages=`; worth confirming empirically rather than by inspection, since the two config channels (properties vs annotations) are not obviously equivalent in Robolectric's bootstrap ordering.

The `GenerateRobolectricPropertiesTask` comment (issue #142) says `sdk`/`graphicsMode` deliberately live in the properties file rather than on annotations, to avoid `AnnotationParser` resolving `@Config.application()` during discovery. The daemon does the opposite. That divergence is deliberate and documented, but it means the two paths bootstrap differently, and only one of them loads fonts.

---

yschimke:
CI confirms the fix works, even though `Module Unit Tests` is still red for the unrelated reason.

`:daemon:android:testDebugUnitTest` on this branch:

```
70 tests completed, 22 failed, 1 skipped
```

versus the same job on `main` and on the previous head:

```
70 tests completed, 23 failed, 1 skipped
```

One fewer, and it's the right one. The previous run listed:

```
ClassloaderForensicsDaemonTest > captureDaemonClassloaderForensics FAILED
    java.lang.NoSuchMethodException at Class.java:2229
```

That entry is **absent** from this run, and `NoSuchMethodException` no longer appears anywhere in the log. Every remaining failure in the list is `IllegalStateException at RobolectricHost.kt:1820` → `NullPointerException at Typeface.java:928` — the native font-loading crash described in the comment above, which this PR does not claim to fix.

So the job stays red on merge, and that is expected: it is red on `main` for the same four tasks (`:renderer-xr:testDebugUnitTest`, `:daemon:core:test`, `:daemon:desktop:test`, `:daemon:android:testDebugUnitTest`). This change strictly reduces the failure count and introduces none.

---

Issue #3072, opened 2026-07-31
Sandbox pool needs a JVM per sandbox — Robolectric's native runtime serves one classloader per process


yschimke:
## Summary

The in-process sandbox pool (`RobolectricHost(sandboxCount > 1)`) cannot boot. Robolectric's native-graphics runtime binds to a **single classloader per JVM**, and the pool gives every worker its own sandbox on purpose, so the second sandbox's native init fails and takes the JVM down.

`sandboxCount` is now capped to 1 (`RobolectricHost.MAX_SANDBOXES_PER_JVM`) so a daemon asking for a pool degrades to a working single-sandbox host instead of crashing at boot. This issue tracks lifting that cap by giving each sandbox its **own process**.

## What breaks today

```
NullPointerException: Cannot read field "mStyle" because "family" is null
  at android.graphics.Typeface.create(Typeface.java:928)
  at android.graphics.Typeface.setSystemFontMap(Typeface.java:1410)
  at android.graphics.Typeface.loadPreinstalledSystemFontMap(Typeface.java:1550)
  at org.robolectric.nativeruntime.DefaultNativeRuntimeLoader.ensureLoaded(…)
```

`RobolectricHost` retries the slot 3×, throws out of `start()`, and the JVM then dies with `SIGSEGV` in `libandroid_runtime.so` (exit 134). This was the standing red in the `Module Unit Tests` CI job: every failing test was a `sandboxCount = 2` one, every passing Robolectric test was single-sandbox.

## Mechanism

`DefaultNativeRuntimeLoader` keeps its `loaded` flag in a static owned by the *sandbox's* classloader. The pool deliberately breaks Robolectric's sandbox cache key so each worker gets its own classloader (the `doNotAcquireClass("…Runner${identityHashCode}")` discriminator in `SandboxHoldingRunner.createClassLoaderConfig`, per SANDBOX-POOL.md Layer 2). So sandbox #2 sees `loaded == false`, re-runs `Typeface.loadPreinstalledSystemFontMap()` against a native library that is already loaded process-wide, and gets back a font map with no default family.

Verified: two *sequential single-sandbox* hosts in one JVM both boot fine — they share one `InstrumentationConfiguration`, so Robolectric hands back the same sandbox and the native runtime initialises exactly once.

## Rejected workarounds (don't re-try these)

| attempt | result |
|---|---|
[...]

PR #3079, opened 2026-07-31
fix(daemon): stop the Robolectric test JVM aborting mid-suite

yschimke:
`:daemon:android:testDebugUnitTest` has been dying partway through every run — the test executor takes a `SIGSEGV` inside `libandroid_runtime.so` (exit 134), so it aborts around test 70 of 222 and the rest never run at all. That is the standing red in `Module Unit Tests`.

> **Scope corrected mid-PR.** The first version of this branch also capped `RobolectricHost`'s `sandboxCount` to 1. That was wrong and is reverted in `89a9dcd` — see "What I got wrong" below.

## The crash

A second Robolectric sandbox in the same JVM comes up with a `Typeface` whose system font map never populated, because the native-graphics runtime loads `libandroid_runtime.so` once and registers its JNI natives against whichever sandbox's instrumented framework classes got there first:

```
NullPointerException: Cannot read field "mStyle" because "family" is null
  at org.robolectric.nativeruntime.DefaultNativeRuntimeLoader.ensureLoaded(…)
```

The `SIGSEGV` is the same missing font map one step later: `AndroidSemanticsWireframe.render` calls `Paint.getFontMetrics()`, which dereferences that null typeface natively and kills the process.

This module runs two runners whose `InstrumentationConfiguration`s necessarily differ — plain `RobolectricTestRunner` for the direct tests, and `SandboxHoldingRunner` (which excludes the bridge package) for anything driven through `RobolectricHost` — so a shared JVM always allocated at least two sandboxes.

## What this PR does

- **`forkEvery = 1`** on the module's test tasks, so each JVM sees one sandbox configuration. The suite completes instead of aborting: **212 of 222 tests now execute, up from ~70.**
- **Repairs a drifted reflective call.** `runForensicDump` pinned the 4-parameter shape of `ClassloaderForensics.capture`, which broke when it gained the injected Okio `fileSystem` parameter (`NoSuchMethodException`, which killed the held-sandbox render loop mid-dispatch). It now resolves by name and routes through Kotlin's `$default` bridge so later defaults can't break it again.

## What I got wrong

I first concluded that Robolectric's native runtime can only ever serve one sandbox per JVM, and capped `sandboxCount` accordingly. CI disproved it: `Daemon Harness (android, real renderer)` — green on `main` — went red, because `RecompositionAttachmentsAndroidRealModeTest` drives recomposition attachments through the held interactive loop, which needs `sandboxCount >= 2`. A **spawned daemon process boots a two-sandbox pool fine**; it is specifically the Gradle unit-test JVM where the second sandbox can't initialise. Capping the product to accommodate a test-harness constraint was the wrong trade, so the cap, the `@Ignore`s it forced onto ~30 interactive/pool tests, and the contract test that pinned it are all reverted.

Rejected workarounds, recorded in #3072 so they aren't re-attempted: dropping the pool's cache-key discriminator (slot 1 then never registers; `start()` hangs to the full boot timeout) and parent-loading `org.robolectric.nativeruntime` (font crash replaced by `UnsatisfiedLinkError` from `RenderNode`). Reproduces identically on the pinned Robolectric 4.17-beta-2 and on 4.16.1, so it is not a version regression.

## Where that leaves the job

| | `main` | this branch |
|---|---:|---:|
| tests that execute | ~70 of 222 | **212** |
| executor aborts mid-suite | yes | **no** |
| reported failures | 27 (+ ~150 never run) | 43 |

The failure count goes *up* because ~30 pool tests that the abort used to hide now run and fail honestly. They fail in the Gradle test JVM for the reason above, and `forkEvery` alone doesn't fix them — understanding why a spawned daemon succeeds where the test JVM doesn't is the open question, tracked in #3072. The remaining 13 are pre-existing and unrelated to sandbox lifecycle: downloadable-font resolution (`FigmaSvgDownloadableFont*`, `ScaledTypographyExport`, `TalkBackHostNavigation`), `ExceptionInInitializerError` in `RenderEngineSizeBounds`, and one test requesting SDK 36 on the JDK 17 toolchain.

So this does not make `Module Unit Tests` green. It fixes the crash, stops the suite hiding ~150 tests, and leaves the remaining failures visible and attributable.

## Test plan

- [x] `:daemon:android:testDebugUnitTest` — 212 execute, no executor crash (was: aborted at ~70)
- [x] `ktfmtCheckAll`, `:daemon:android:compileDebugUnitTestKotlin` — green
- [ ] `Daemon Harness (android, real renderer)` — expected back to green now the cap is reverted; confirming on this run

Non-visual change (daemon test wiring + a reflective call), so no rendered evidence applies.

https://claude.ai/code/session_013QDREaaoYqh2SKdz6aELeB


yschimke:
Pushed `89a9dcd`, which reverts the `sandboxCount` cap. CI caught what I got wrong.

**`Daemon Harness (android, real renderer)` went red** — it was green on `main`. `RecompositionAttachmentsAndroidRealModeTest` drives recomposition attachments through the held interactive loop, which needs `sandboxCount >= 2`; the cap made `supportsInteractive` false and took the lane out.

That failure also falsifies my diagnosis. I had concluded Robolectric's native runtime can only ever serve one sandbox per JVM, full stop. But a **spawned daemon process boots a two-sandbox pool fine** — it is specifically the Gradle unit-test JVM where the second sandbox can't initialise. Capping the product to work around a test-harness constraint was the wrong trade, so the cap, the ~30 `@Ignore`s it forced onto the interactive/pool tests, and `RobolectricHostSandboxCapTest` are all gone.

What stays is `forkEvery = 1` (the suite completes instead of aborting at ~test 70 — 212 of 222 now execute) and the `ClassloaderForensics.capture` reflective-lookup repair.

Both Codex review points were about the cap and are moot with it reverted:

- **Skip the capped four-sandbox benchmark** — `SandboxPoolMemoryBench` is back to its unmodified state and again measures a real four-sandbox pool, so its output isn't corrupted. Good catch on the divide-by-four; it would have silently poisoned the SANDBOX-POOL.md figures.
- **Update the pool guide before linking capped users to it** — also correct, and no longer applicable: nothing now contradicts SANDBOX-POOL.md, and the warning that pointed at a nonexistent "In-process pool is disabled" section is gone with the cap.

I have not made `Module Unit Tests` green. The remaining 43 failures are ~30 pool tests (which the abort used to hide, and which `forkEvery` alone doesn't fix) plus 13 pre-existing font/SDK-environment failures. Why a spawned daemon succeeds where the Gradle test JVM doesn't is the open question — tracked in #3072.

---

yschimke:
CI on `89a9dcd` confirms the local numbers exactly:

```
212 tests completed, 43 failed, 10 skipped
```

Before this change the executor aborted around test 70 of 216 and the run ended there — the remaining ~150 tests never reported at all. Now every test in the module executes and reports; the per-fork `SIGSEGV` in `libandroid_runtime.so` still appears in the log, but it happens on JVM teardown *after* the test has run and reported, so it no longer truncates the suite.

The 43 failures split into the two groups already described in the PR body:

- **~30 pool / interactive tests** — `RobolectricHostPoolTest`, `SandboxPoolMemoryBench`, `AndroidInteractiveSessionTest`: `IllegalStateException` at `RobolectricHost.kt:1820` → `NullPointerException` at `Typeface.java:928`. This is #3072 (a second sandbox in the same JVM can't get native graphics initialised). `forkEvery = 1` does not fix it — one fork still hosts two sandboxes.
- **13 pre-existing** — `RenderEngineSizeBoundsTest` / `ScaledTypographyExportTest` / `TalkBackHostNavigationTest` ("Could not load font", `Typeface` NPE, `ExceptionInInitializerError`) and `SandboxHoldingRunnerApplicationOverrideTest` (`UnsupportedOperationException` at `DefaultSdkProvider.java:174`).

So this does not make `Module Unit Tests` green, and it isn't meant to. It fixes the crash that was aborting the executor, stops the suite from hiding ~150 tests, and leaves the remaining failures visible and attributable to #3072 / the pre-existing font-init issues rather than buried behind a `SIGABRT`.

(The same job also reports failures in `:renderer-xr:testDebugUnitTest`, `:daemon:core:test` and `:daemon:desktop:test` — those are outside this change and unaffected by it.)

---

Issue #874, opened 2026-05-06
UIAutomator extension: stretch goals after the v1 vertical


yschimke:
Tracking issue for follow-on work after the end-to-end UIAutomator-shaped dispatch shipped across #832, #864, and #872.

## What's already in

- `:data-uiautomator-core` — `Selector` DSL, `By` factory, `NodeBacking` matcher, `ViewBacking` + `SemanticsBacking`, `UiObject` sealed result, `SelectorJson` wire format. Tested against both View trees and Compose `SemanticsOwner` (merged + unmerged).
- `:data-uiautomator-connector` — `uiautomator` `DataExtensionDescriptor` advertising 9 `uia.*` script-event ids.
- `:daemon:android` bridge — `InteractiveCommand.DispatchUiAutomator` envelope, `RobolectricHost.performUiAutomatorAction`, `AndroidRecordingSession` per-id handlers.
- `:daemon:core` `InteractiveSession.dispatchUiAutomator` interface.
- MCP `record_preview` schema gains `selector` / `useUnmergedTree` / `inputText` fields.
- End-to-end Robolectric integration tests.

Agents can now post `{kind: "uia.click", selector: {text: "Submit"}}` and the click fires inside the held composition.

### 1. `data/uiautomator/hierarchy-android` producer

A `uia/hierarchy` data product that walks the held composition's `SemanticsOwner` and emits the available nodes (text, contentDescription, testTag, exposed actions, bounds) as a typed JSON product. Pairs with `:data-a11y-hierarchy-android`'s shape; same producer pattern.

**Why:** Agents currently have to guess at selector targets (or read the screenshot). A `uia/hierarchy` payload lets them inspect what's clickable / scrollable / has-text *before* dispatching, so a `uia.click` with a bad selector becomes rare.

**Scope:** new `:data-uiautomator-hierarchy-android` module + producer + `DataProductRegistry` + connector wiring + ~5 tests. Mirrors `:data-a11y-hierarchy-android` closely.

### 2. Typed evidence shapes for unsupported `uia.*` events

Today `AndroidRecordingSession.uiAutomatorActionHandler` reports unsupported as a free-form string (`"no node matched selector=… for uia 'click'"`). Should be a structured `UnsupportedReason` carrying the matched-count, the available text/desc/testTag values on the closest near-match node, and the actions that node *does* expose.

[...]

PR #3091, opened 2026-07-31
feat(daemon): give each sandbox its own process so the pool can boot

### LINKED ISSUE #3072: Sandbox pool needs a JVM per sandbox — Robolectric's native runtime serves one classloader per process

yschimke:
Closes #3072.

## Why

Robolectric's native-graphics runtime binds to a **single classloader per process**. `libandroid_runtime.so` loads once and registers its JNI natives against the first sandbox's instrumented framework classes; its `loaded` flag lives in a static owned by *that sandbox's* classloader. A second sandbox in the same JVM sees `loaded == false`, re-runs `Typeface.loadPreinstalledSystemFontMap()` against an already-loaded library, and gets a font map with no default family:

```
NullPointerException: Cannot read field "mStyle" because "family" is null
```

…then the next native text call kills the process with `SIGSEGV` (exit 134). So `RobolectricHost(sandboxCount > 1)` could never boot — which blocked interactive sessions, scripted recording, and concurrent multi-slot renders. Reproduced on `main` in this session: **all 7** `RobolectricHostPoolTest` cases fail with that NPE.

## What

Scale the pool by **processes**, one sandbox each.

- **`SandboxProcessPool`** (parent side) spawns one worker JVM per slot beyond the first — same classpath, the daemon's JVM flags minus what must not be duplicated (agents, JDWP), and every `composeai.*` / `robolectric.*` / `android.*` / `roborazzi.*` system property, so a worker's render is configured identically to the daemon's own (including `composeai.daemon.userClassDirs` for hot reload). The parent listens; the worker dials back.
- **`SandboxWorkerMain`** (child side) is a plain `RobolectricHost(sandboxCount = 1)` plus a socket loop — deliberately *not* a daemon. `RobolectricHost.submit` resolves `previewId` → full spec payload before dispatch, so only a payload in and a `RenderResult` out cross the boundary.
- **The result survives intact.** The host already reduced sandbox-side results to plain data before handing them to callers (`copyPreviewContextAcrossClassloaders` copies device context + theme payload and drops the live `slotTables` / `rootForTest` handles). Every surviving field is a String/number/Map, so the wire DTO is faithful and host-side data products see the same input either way.
- **Interactive sessions move to slot 0**, the in-process sandbox: the held rule, bridge queues and frame latches are live object handles that can't be proxied. The *renders* are what serialize cleanly, so they move out of process while a session holds the slot. `supportsInteractive` still reads `sandboxCount >= 2` — one to pin, one left to render — plus a new clean `Unsupported` while no worker is ready yet.
- **The per-worker sandbox-cache discriminator is gone** from `SandboxHoldingRunner`. It existed to force N sandboxes into one JVM, i.e. the exact arrangement that breaks; without it, sequential hosts in a JVM share one cached sandbox.
- Hot-reload swaps, background boot, boot-time warm renders, affinity dispatch and shutdown all cross the process boundary; a worker that dies mid-request is dropped from the pool and logged with its pid.
- `SandboxPoolMemoryBench` rewritten: it now reports the daemon JVM plus each worker's RSS from `/proc/<pid>/status`, since reading only this JVM would under-report the pool by a factor of N.
- `docs/daemon/SANDBOX-POOL.md` rewritten around the process model, including the rejected in-JVM workarounds so they don't get re-tried.

## Trade-off, stated plainly

A slot beyond the first is now a whole JVM again (~1–2 GB), not the ~75 MB per extra classloader the in-JVM pool claimed. That claim was never realisable — the second in-JVM sandbox cannot boot at all. Sizing knobs are unchanged (`replicasPerDaemon`, `composeai.daemon.sandboxCount`), so a constrained host turns the pool down the same way it always did.

## Test plan

Run in this session on `:daemon:android` (JDK 17 / Robolectric SDK 35):

- [x] `RobolectricHostPoolTest.twoSlotsServeRendersFromDistinctProcesses` — 20 renders spread across both slots; the worker's pid is present and differs from the daemon JVM's. **This is the case that could not boot before.**
- [x] `swapUserClassLoadersBroadcastsToTheInProcessHolderAndTheWorkers` — in-process holder observes the swap, the worker survives it and keeps serving.
- [x] `samePreviewIdAlwaysLandsOnSameSlot`, `normalRendersAvoidInteractiveSlotWhenHeldSessionIsPinned`, `rejectsLegacyHolderPlusFactory`, `rejectsLegacyHolderWithSandboxCountAboveOne`
- [x] `RobolectricHostTest` (single sandbox) and `ThemeConsumerCaptureRenderTest` (real render, plain runner) still pass — the `sandboxCount = 1` path is untouched.
- [x] `./gradlew :daemon:android:ktfmtFormat :mcp:ktfmtFormat`, `:mcp:compileKotlin`, `:daemon:android:compileDebugKotlin` / `compileDebugUnitTestKotlin`

**Not verifiable in this container — needs CI:** the three pool cases that require real Compose *text* rendering through the daemon host path (`realRendersOnBothSlots…`, `interactiveSessionRunsOnTheInProcessSandbox`, `backgroundBootServesRendersBeforePoolCompletesAndWarmsTheLateSlot`). This sandbox's daemon-host sandbox comes up with a broken font map and any real render aborts the JVM — and that is **pre-existing, not from this change**: `RenderEngineTest` on unmodified `main` dies the same way here (exit 134). Please watch those three on CI; I'll follow the run and fix what it reports.

Text-only PR (no rendered-surface changes), so no visual evidence applies.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MpgFQcgv5s3P25GGJxNGv6)_

### DIFF
--- a/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidInteractiveSession.kt
+++ b/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidInteractiveSession.kt
@@ -434,8 +434,7 @@ internal constructor(
     lastUsedAtMs.set(System.currentTimeMillis())
     val replyLatch = CountDownLatch(1)
     val replyError = AtomicReference<Throwable?>(null)
-    val replyReason =
-      AtomicReference<ee.schimke.composeai.daemon.protocol.UiAutomatorUnsupportedReason?>(null)
+    val replyReasonJson = AtomicReference<String?>(null)
     slot.interactiveCommands.put(
       InteractiveCommand.FindUiAutomatorEvidence(
         streamId = streamId,
@@ -445,7 +444,7 @@ internal constructor(
         inputText = inputText,
         replyLatch = replyLatch,
         replyError = replyError,
-        replyReason = replyReason,
+        replyReasonJson = replyReasonJson,
       )
     )
     if (!replyLatch.await(DISPATCH_TIMEOUT_SEC, TimeUnit.SECONDS)) {
@@ -456,7 +455,8 @@ internal constructor(
       )
     }
     replyError.get()?.let { throw it }
-    return replyReason.get()
+    // Re-parse into the host's own class — the sandbox only ever hands back a String.
+    return replyReasonJson.get()?.let { UiAutomatorEvidence.decode(it) }
   }
 
   /**
--- a/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidInteractiveSession.kt
+++ b/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidInteractiveSession.kt
@@ -26,10 +26,13 @@ import kotlinx.serialization.json.Json
  * sandbox-side [RobolectricHost.SandboxRunner.runHeldInteractiveSession] loop.
  *
  * **Sandbox pinning** (INTERACTIVE-ANDROID.md § 2). Each session pins exactly one sandbox slot for
- * its lifetime. v3 ships with `INTERACTIVE_SLOT_INDEX = 1` so slot 0 stays the always-on
- * normal-render slot — that's enforced host-side in [RobolectricHost.acquireInteractiveSession]
- * (the constraint that `sandboxCount >= 2`). The session itself is slot-agnostic; it only carries a
- * reference to the slot it was constructed against.
+ * its lifetime. Since #3072 that is `INTERACTIVE_SLOT_INDEX = 0`, the sandbox in the daemon JVM:
+ * the held rule and its bridge queues are live object handles that can't be proxied to a worker
+ * process, so the session stays local and the *normal* renders move to slots 1..N-1 (worker JVMs)
+ * for its lifetime. Enforced host-side in [RobolectricHost.acquireInteractiveSession] (the
+ * constraint that `sandboxCount >= 2` — one sandbox to pin, at least one left to render). The
+ * session itself is slot-agnostic; it only carries a reference to the slot it was constructed
+ * against.
  *
  * **Threading.** Per the [InteractiveSession] contract, callers are serialised — `JsonRpcServer`
  * dispatches one input per session at a time. Each public method enqueues an [InteractiveCommand]
@@ -39,7 +42,7 @@ import kotlinx.serialization.json.Json
  * the composition.
  *
  * **Lifecycle** (INTERACTIVE-ANDROID.md § 7).
- * - **Allocate** at `interactive/start`. The host enqueues [InteractiveCommand.Start] onto slot 1's
+ * - **Allocate** at `interactive/start`. The host enqueues [InteractiveCommand.Start] onto slot 0's
  *   [SandboxSlot.interactiveCommands]; the sandbox's idle loop picks it up and enters
  *   `runHeldInteractiveSession`, which constructs the rule + ActivityScenario + paused-clock
  *   `setContent`, then counts down [InteractiveCommand.Start.replyLatch] before draining further
@@ -63,8 +66,8 @@ internal constructor(
    */
   internal val streamId: String,
   /**
-   * The sandbox slot this session was acquired against — almost always slot 1 in v3 (slot 0 is the
-   * always-on normal-render slot). Carried explicitly rather than re-derived because the host's
+   * The sandbox slot this session was acquired against — slot 0, the in-process sandbox, since
+   * #3072. Carried explicitly rather than re-derived because the host's
    * routing code (`acquireInteractiveSession`) already chose it; passing it in keeps the session
    * itself slot-policy-agnostic so v4 (multi-target Android) can reuse this class unchanged.
    */
@@ -82,7 +85,7 @@ internal constructor(
    * tests can drive a sub-second lease without sleeping CI for a minute.
    *
    * Without this lease, a panel that crashes or a websocket that drops mid-session would leak a
-   * pinned sandbox slot for the rest of the daemon's lifetime — slot 1 stays held with no
+   * pinned sandbox slot for the rest of the daemon's lifetime — slot 0 stays held with no
    * `interactive/stop` ever arriving. v3 supports one held session at a time per host, so a single
    * zombie burns the whole interactive capacity. Symmetry with the desktop story comes via the
    * panel's auto-stop on editor change/scroll (#427); the lease is the daemon-side belt-and-braces.
--- a/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/DaemonMain.kt
+++ b/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/DaemonMain.kt
@@ -78,9 +78,10 @@ fun main(args: Array<String>) {
   // The plugin has exposed `composeai.daemon.warmSpare=true` by default since the daemon launch
   // descriptor was introduced, but the Android daemon previously ignored it and came up with a
   // single sandbox unless the experimental sandbox-count property was set manually. Held
-  // interactive sessions need slot 1 pinned while slot 0 continues normal renders, and a typical
-  // preview grid wants extra slots for parallel renders, so default the pool to five sandboxes
-  // when warmSpare is on. Explicit `composeai.daemon.sandboxCount` still wins.
+  // interactive sessions pin the in-process sandbox (slot 0) and need at least one worker process
+  // left to serve normal renders, and a typical preview grid wants extra slots for parallel
+  // renders, so default the pool to five sandboxes when warmSpare is on — one here plus four
+  // worker JVMs (SANDBOX-POOL.md). Explicit `composeai.daemon.sandboxCount` still wins.
   val warmSpareEnabled = System.getProperty(WARM_SPARE_PROP)?.toBooleanStrictOrNull() ?: true
   val defaultSandboxCount = if (warmSpareEnabled) 5 else 1
 
--- a/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/RobolectricHost.kt
+++ b/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/RobolectricHost.kt
@@ -923,8 +923,52 @@ open class RobolectricHost(
     // Throwable path still classifies it into a typed `renderFailed`.
     if (slotIdx > 0) {
       val pool = processPool ?: error("slot $slotIdx dispatched with no process pool configured")
-      return pool.submit(slotIdx - 1, typed, timeoutMs)
+      try {
+        return pool.submit(slotIdx - 1, typed, timeoutMs)
+      } catch (t: Throwable) {
+        // A render failure *inside* a live worker is the caller's to see — re-throw it. A worker
+        // that died taking the request with it is different: the pool has already dropped it, so
+        // re-dispatch once across the remaining slots rather than failing a render the pool can
+        // still serve. Only one retry: if the replacement slot dies too, something systemic is
+        // wrong and the caller should hear about it.
+        if (pool.isReady(slotIdx - 1)) throw t
+        val fallback = chooseSlotIndex(typed)
+        System.err.println(
+          "compose-ai-daemon: sandbox worker $slotIdx died holding render ${typed.id}; " +
+            "re-dispatching to slot $fallback (${t.javaClass.simpleName}: ${t.message})"
+        )
+        if (fallback > 0) return pool.submit(fallback - 1, typed, timeoutMs)
+        return submitInProcess(typed, timeoutMs)
+      }
     }
+    return submitInProcess(typed, timeoutMs)
+  }
+
+  /**
+   * Blocks until at least one sandbox worker is routable (so a held session on slot 0 leaves
+   * something to serve normal renders), or [timeoutMs] elapses.
+   *
+   * Throws the same clean `UnsupportedOperationException` the caller already handles with a v1
+   * fallback if the wait times out — that is the "the pool never produced a worker" case (every
+   * worker's boot failed permanently), not the ordinary warm-up window.
+   */
+  private fun awaitWorkerForInteractive(timeoutMs: Long) {
+    val deadline = System.nanoTime() + timeoutMs * 1_000_000
+    while (routableSlots().size < 2) {
+      if (System.nanoTime() >= deadline) {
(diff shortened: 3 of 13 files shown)


yschimke:
Pushed `c9d2037` — three CI/review items addressed, plus one I found while reading the CI logs.

**`Reject agent attribution` (red → fixed).** The first commit landed with the container's default git identity (`Claude <noreply@anthropic.com>`), which is exactly what the gate exists to stop. Both commits are re-authored to `Yuri Schimke <yuri@schimke.ee>` and force-pushed; `agent-attribution-scan.sh --range origin/main..HEAD` now passes locally.

**Codex P1 — dead workers stayed in affinity routing.** Correct, and it was a real stranding bug: `SandboxProcessPool` marked a dead worker's entry, but `chooseSlotIndex` kept hashing over the contiguous ready range, so every preview pinned to that slot failed forever. Dispatch now builds a routable-slot list (slot 0 plus workers that are both booted *and* alive) and hashes into that, so a dead worker re-keys its previews onto live slots. On top of that, a render whose worker died mid-flight is re-dispatched once across the survivors rather than surfacing the socket error — the pool degrades slot by slot instead of losing a share of the catalog.

**Codex P2 — interactive start during the warm-up window.** Also correct. `handleInteractiveStart` sees `supportsInteractive == true` and turns the `UnsupportedOperationException` into an internal error that the panel doesn't retry, so with `backgroundSandboxBoot=true` an ordinary early `interactive/start` would just fail. It now waits for a routable worker within the existing `INTERACTIVE_START_TIMEOUT_MS` budget and only throws if none ever appears (every worker's boot failed permanently) — matching the pre-#3072 behaviour, which waited on slot 1's readiness.

**Found while reading the logs — workers could outlive a crashed daemon.** Socket EOF only reaps a worker that already reached its serve loop; a parent dying during the worker's minutes-long Robolectric bootstrap would strand a whole JVM. Workers now watch the parent process handle and halt when it exits. Relevant here because the daemon test JVM aborting mid-suite is a live condition in this repo.

**On `Module Unit Tests`:** it is red, and it is red on `main` too — this is the standing failure #3072 describes, not a regression from this PR. Comparing the `:daemon:android:testDebugUnitTest` runs:

| | tasks failed | Android tests |
|---|---|---|
| `main` @ 72facbde ([run](https://github.com/yschimke/compose-ai-tools/actions/runs/30649378242)) | 4 (`renderer-xr`, `daemon:core`, `daemon:desktop`, `daemon:android`) | 212 completed, **43 failed** |
| this PR @ 24e0b72 ([run](https://github.com/yschimke/compose-ai-tools/actions/runs/30652968417)) | 3 (same minus `daemon:desktop`) | 205 completed, **16 failed** |

The surviving 16 are the native font-stack / `DefaultSdkProvider` family (`Typeface.java:928`, `FontFamilyResolver.kt:214`, `AndroidParagraphIntrinsics.android.kt:100`, `UnsupportedOperationException at DefaultSdkProvider.java:174` — SDK 36 wants JDK 21 and the job runs Temurin 17). That looks like separate toolchain drift worth its own issue; I don't think this PR should try to absorb it. I'll keep watching the run on the new head and report what moves.

---

yschimke:
CI on `c9d2037`: **`Reject agent attribution` is green**, and every other check passes except `Module Unit Tests`.

Two things worth pulling out of that run:

**`Daemon Harness (android, real renderer)` passed.** That is the real-render daemon path end to end, which is precisely the coverage I said in the PR body I couldn't produce in my container. It's the strongest available evidence that the out-of-process pool serves genuine renders, not just the fixture ones I ran locally.

**`Module Unit Tests` is the same pre-existing red, now slightly smaller.** `:daemon:android:testDebugUnitTest` — 204 completed, **15 failed** (43 on `main`, 16 on the previous head). Every named failure is the native font stack or the SDK provider:

- `RenderEngineSizeBoundsTest` ×3 — NPE at `AndroidParagraphIntrinsics.android.kt:100`
- `TalkBackHostNavigationTest` ×4 — NPE at `Typeface.java:928`
- `ScaledTypographyExportTest` — `IllegalStateException` at `FontFamilyResolver.kt:214`, caused by the same `Typeface` NPE
- `SandboxHoldingRunnerApplicationOverrideTest$DummyTest` — `UnsupportedOperationException` at `DefaultSdkProvider.java:174`

plus repeated `SIGSEGV` in `libandroid_runtime.so` aborting whole test executors (exit 134). The same abort signature is on `main`, so this is environmental, not from the diff.

One caveat I want to state rather than gloss: Gradle only prints *failing* tests, and a SIGABRT takes its whole executor's results with it — so "no `RobolectricHostPoolTest` line in the failure list" means those tests didn't fail, but I can't prove from the log alone that all of them ran. The passing real-renderer harness job is better evidence than the unit-test log here.

The `DefaultSdkProvider` throw is Robolectric refusing SDK 36 on a JDK 17 runner, and the `Typeface` NPEs are the font-stack symptom of the same environment. Fixing it means moving the CI toolchain (or pinning the SDK) across every module — a repo-wide change, well outside a sandbox-pool PR, and it would make this diff impossible to review. I'd rather file it separately; happy to take it on if you want it chased now.

---

yschimke:
Check-in. **The three pool tests are answered — they run and pass on CI.** I couldn't download the test-results artifact (the proxy blocks the blob-storage redirect), so I reconstructed the per-test picture from the job logs instead, and the numbers reconcile exactly.

On `main`, **all 7** `RobolectricHostPoolTest` cases fail:

```
twoSandboxesServeDistinctClassloaders            interactiveStartUsesTheCurrentSlotsChildLoader
swapUserClassLoadersBroadcastsToEverySlot        perSlotHoldersHaveDistinctChildLoadersParentedToTheirSandbox
samePreviewIdAlwaysLandsOnSameSlot               realRenderUsesTheCurrentSlotsChildLoader
backgroundBootServesRendersBeforePoolCompletesAndWarmsTheLateSlot
```

On `c9d2037` the string `RobolectricHostPoolTest` does not appear anywhere in the `:daemon:android` output — **zero pool failures**, including `realRendersOnBothSlots…`, `interactiveSessionRunsOnTheInProcessSandbox` and `backgroundBootServesRendersBeforePoolCompletesAndWarmsTheLateSlot`. The 15 reported failures account for exactly the 15 I can name, so nothing is hidden in a lost executor:

| | `main` | `c9d2037` |
|---|---|---|
| `RobolectricHostPoolTest` | **7** | **0** |
| `AndroidInteractiveSessionTest` | **16** | **2** |
| `AndroidRecordingSessionTest` | **3** | **0** |
| font stack / SDK (`Typeface`, `FontFamilyResolver`, `DefaultSdkProvider`, `TalkBack…`, `Figma…`) | 17 | 13 |
| **total** | **43** | **15** |

Interactive sessions and scripted recording — the two things #3072 said the broken pool blocked — go from fully red to essentially green.

**Fixed one of the two remaining interactive failures in `a083a3a`.** `uiAutomatorClickReportsUnmatchedWhenNoNodeSatisfiesTheSelector` was failing on a `ClassCastException`, not a font error, so I chased it: `findUiAutomatorEvidence` handed a **typed** `UiAutomatorUnsupportedReason` back from the sandbox loop, and the object is built under the instrumenting classloader —

```
class …UiAutomatorUnsupportedReason cannot be cast to class …UiAutomatorUnsupportedReason
(… loader AndroidSandbox$SdkSandboxClassLoader …; … loader 'app')
        at AndroidInteractiveSession.findUiAutomatorEvidence(AndroidInteractiveSession.kt:459)
```

It now serialises inside the sandbox and re-parses host-side, which is the rule the sibling commands already follow and document in their own KDoc (`CaptureProbeSemantics` → `replyNodesJson`, `CaptureA11yFindings` → `replyFindingsJson`). `FindUiAutomatorEvidence` was the one that deviated.

Worth being precise about ownership: this bug is latent since #874 and not caused by this PR — but it is *exposed* by it, because the pool booting is what lets the test reach the evidence call instead of dying at session start. It reproduces locally, and the fix is verified locally (the test passes; the six runnable pool cases and `RobolectricHostTest` still pass).

That leaves **one** known `:daemon:android` failure that is mine to care about — `nonComposableNotificationKindRendersInHeldSessionInsteadOfErroring` — and it bottoms out in `NullPointerException at Typeface.java:928` inflating a notification View: the same pre-existing font-map problem as the other 13, red on `main`, and not fixable without the toolchain work I flagged earlier.

I'll watch the run on `a083a3a`.

---

yschimke:
CI on `a083a3a` (run [30658602335](https://github.com/yschimke/compose-ai-tools/actions/runs/30658602335)) is red in three tasks. None of them is caused by this diff, and I'm not fixing them here — details and the evidence for that claim below.

**What moved in the right direction:** `:daemon:android` went **15 → 14** failures. `uiAutomatorClickReportsUnmatchedWhenNoNodeSatisfiesTheSelector` is gone — that's the `ClassCastException` fix in `a083a3a` confirmed on CI. The 14 that remain are the font-map / SDK failures already red on `main`, including `nonComposableNotificationKindRendersInHeldSessionInsteadOfErroring` (`NullPointerException at Typeface.java:928`). #3094 is addressing that family.

**Two tasks failed that I hadn't called out before**, and both are base-branch failures:

| task | failures |
|---|---|
| `:renderer-xr:testDebugUnitTest` | 3 — `SubspaceSceneRecorderTreeTest.recordsThreeDimensionalTreeWithPerPanelTwoDimensionalContent`, `SubspaceWorldIntegrationTest.producesCompleteSceneForFloatingWindowWorld`, `XrSubspaceRendererTest.rendersSceneJsonFromPreviewFunction` (all `AssertionErrorWithFacts`) |
| `:daemon:core:test` | test worker exits with status 1 mid-run — no test failure recorded, the JVM just goes away |

Why they aren't mine:

- This branch's tree under `daemon/core/` and `renderers/xr/` is **byte-identical to `origin/main`** (`git diff --name-only origin/main..HEAD -- daemon/core renderers/xr` → empty), and `origin/main` is an ancestor of the head commit. The diff is confined to `daemon/android/`, `mcp/`, and docs.
- `:daemon:core` has no project dependency on `:daemon:android`, so nothing in this PR is even on its test classpath.
- Both reproduce locally on this checkout, with exactly the CI symptoms: the same three XR assertions, and `:daemon:core:test` → `Process 'Gradle Test Executor 1' finished with non-zero exit value 1`.

I did narrow the `:daemon:core` one while confirming it, in case it's useful to whoever picks it up: the worker dies inside `JsonRpcServerIntegrationTest`, during `fileChanged_source_emits_discoveryUpdated` — the log's last line from that test is `read loop entering`, and the test is then reported SKIPPED because the JVM is gone. It passes when run alone, so it's an interaction with an earlier test in the class, not that test itself. Separately in that class, `full_lifecycle_renders_one_preview_and_emits_finished_notification` fails on its own at `JsonRpcServerIntegrationTest.kt:124` — the `capabilities.androidSdk == 35` assertion — which looks like the same SDK-level drift behind the font failures rather than a distinct bug. Exit status 1 (not 134) means a deliberate `System.exit`, not the native `SIGSEGV` that kills `:daemon:android`; every `JsonRpcServer` in that class overrides `onExit`, so I didn't find the caller and stopped there rather than chase it inside this PR.

Happy to open a separate issue or PR for either if you'd like — say the word.

---

Issue #3092, opened 2026-07-31
fix(renderer-desktop): `showBackground` leaves the fixed axis unpainted when the other axis wraps


yschimke:
Found by the new preview-mode matrix (#3082), which renders the same `@Preview` set through both backends and compares.

## Symptom

A preview with **one axis fixed and the other wrapping** — `@Preview(widthDp = 240, showBackground = true)` — comes back from the desktop renderer with the harness background painted only across the *content's* measured width; the rest of the fixed 240dp frame is transparent. The Android renderer (and Android Studio) fill the whole fixed frame.

From `:samples:cmp` vs `:samples:android`, same annotation, same fixture (`MatrixFixedWidthOnlyPreview`, a 160×80dp probe in a 240dp-wide frame):

| backend | PNG | pixel at `(width-2, 2)` |
|---|---|---|
| Android (Robolectric) | 630×210 | `#FFFFFFFF` — opaque white |
| Desktop (CMP) | 630×210 | `#00000000` — transparent |

Both axes fixed is fine (`MatrixFixedBothAxesPreview` is opaque white to the corners on both backends); it is specifically the mixed fixed/wrap case.

## Cause

[`ComposePreviewContentBox`](``https://github.com/yschimke/compose-ai-tools/blob/main/renderers/desktop/src/main/kotlin/ee/schimke/composeai/renderer/ComposePreviewScene.kt#L85-L116``) takes the wrap branch whenever *either* axis wraps, and that branch lays the box out at `placeable.width × placeable.height`. On the **fixed** axis the incoming root constraint has `minWidth = 0`, so the content measures at its intrinsic size rather than the fixed frame, and `.background(backgroundColor)` only covers that measured box. The PNG is still emitted at the fixed size (the crop only trims wrapped axes), so the difference shows up as transparency.

The likely fix is to pin the non-wrapping axis in the constraints the layout block builds — `minWidth = constraints.maxWidth` when `!wrapWidth` (and the height equivalent) — since on a fixed axis the scene is already exactly the fixed size. Worth checking the daemon's desktop `RenderEngine` at the same time: it shares `ComposePreviewContentBox`, so `compose-preview serve` and bundle re-renders have the same gap.

## Notes

Not fixed in #3082 — that PR only adds the fixtures and the assertions that surfaced this. The desktop matrix test documents the current (divergent) pixel so this issue's fix will trip it and the expectation gets updated in the same change.

PR #3093, opened 2026-07-31
test(samples): add a preview-mode matrix for both renderers

### LINKED ISSUE #3082: Add tests demonstrating all the preview modes for different renderers
- component preview 
- device preview (wear, mobile, other? foldable?)
- with different typical preview annotation params set
- using multipreviews
- using device ids

yschimke:
Closes #3082.

## Summary

The issue asked for tests demonstrating every preview mode across the renderers, with the goal being *"to know we match Android Studio by default"*. Nothing covered that end to end — `DeviceDimensions` has unit tests for the catalog, but no test tied a **rendered PNG** back to the geometry Studio resolves for the same annotation.

This adds `PreviewModeMatrix.kt` (fixtures) + `PreviewModeMatrixTest` (assertions) to **`:samples:android`** (Robolectric backend) and **`:samples:cmp`** (Compose Multiplatform desktop backend), covering per backend:

| axis from the issue | fixtures |
|---|---|
| component preview | bare `@Preview` (both axes wrap to intrinsic size) |
| device preview | `id:pixel_5` (mobile), `id:wearos_small_round` (wear), `id:pixel_fold` (foldable), `id:tv_1080p` (TV, Android only) |
| device ids | the id catalog above, plus the `spec:width=…,height=…,dpi=…` grammar |
| typical annotation params | `widthDp`/`heightDp` (both axes and one-axis), `fontScale`, `locale`, `uiMode`, `showBackground` + `backgroundColor`, `showSystemUi` |
| multipreviews | `@PreviewLightDark`, `@PreviewFontScale`, `@PreviewScreenSizes`, and an app-declared meta-annotation |

The expected pixel sizes are written out as `dp × density` **from Studio's own device catalog**, not read back from our resolver — so the tests assert Studio parity rather than self-consistency. Both backends assert the *same numbers*, and today every shared mode lands on a byte-identical canvas, so drift between the two renderers now fails a build too.

Beyond geometry each param's observable outcome is pinned: night is darker than day, `backgroundColor` reaches all four corners, the light/dark fan-out produces visibly *different* pixels (the probe colours itself from `isSystemInDarkTheme()`, so a fan-out that rendered twice identically would fail), and the meta-annotation resolves each nested device independently.

`:samples:cmp` had no test source set — `composePreview { renderBeforeUnitTests }` only chains AGP's `test<Variant>UnitTest` tasks, which a JVM/CMP module doesn't have — so its plain `test` task is wired onto `composePreviewRenderAll`. Every new preview renders in the existing sample render jobs, so future changes to these surfaces get before/after diffs automatically.

## It already found something

Rendering the same fixture through both backends surfaced a real divergence, filed as **#3092**: on desktop, `showBackground = true` with **one axis fixed and the other wrapping** paints only the measured content, leaving the rest of the fixed frame transparent. Android and Studio fill the whole frame.

| backend | `MatrixFixedWidthOnlyPreview` | pixel at `(width-2, 2)` |
|---|---|---|
| Android | 630×210 | `#FFFFFFFF` opaque white |
| Desktop | 630×210 | `#00000000` transparent |

Not fixed here (out of scope for a test-coverage PR, and it changes rendered output): the Android test carries the reference expectation, and the desktop test pins the current pixel as a change detector so #3092's fix trips it instead of silently changing renders.

## Visual evidence

Every mode, rendered in-session, with the actual pixel size under each cell. Contact sheets committed under `renders/preview-mode-matrix/`.

**Android (Robolectric) renderer**

Android preview-mode matrix

**Desktop (Compose Multiplatform) renderer**

Desktop preview-mode matrix

Note the third cell in each sheet — that transparent right-hand region on desktop vs. white on Android is #3092, visible in the pixels rather than described.

## Test plan

All run in-session on this container:

- `./gradlew :samples:android:composePreviewRenderAll` — BUILD SUCCESSFUL, 31 new matrix PNGs
- `./gradlew :samples:cmp:composePreviewRenderAll` — BUILD SUCCESSFUL, 23 new matrix PNGs
- `./gradlew :samples:android:testDebugUnitTest --tests "*PreviewModeMatrixTest*"` — BUILD SUCCESSFUL (7 tests)
- `./gradlew :samples:cmp:test --tests "*PreviewModeMatrixTest*"` — BUILD SUCCESSFUL (7 tests)
- `ktfmtFormat` applied to both samples

Every expectation in both tests was verified against a real render, not derived: the geometry table was written first, then checked against the rendered PNGs, and the two multipreview ladder assertions (`@PreviewFontScale` → 7 stops, `@PreviewScreenSizes` → 6 screens) were pinned from the actual fan-out at the sample's compose BOM.


yschimke:
CI note on **Module Unit Tests** ([run](https://github.com/yschimke/compose-ai-tools/actions/runs/30654402628/job/91235142403)) — red, and red on the base branch too. This PR's merge base is `72facbd`, whose own [`main` push run](https://github.com/yschimke/compose-ai-tools/actions/runs/30649378242/job/91218616574) fails the same job; so does the run before it (`b9fd678`).

Four tasks failed here:

| task | on base `72facbd`? |
|---|---|
| `:renderer-xr:testDebugUnitTest` | yes |
| `:daemon:core:test` (test JVM exits 1) | yes |
| `:daemon:android:testDebugUnitTest` (SIGABRT, exit 134) | yes |
| `:daemon:desktop:test` | no — extra on this run |

The dominant signature is runner-level, not diff-level: `:daemon:android` reports **43 of 212 tests failed**, nearly all rooted in `java.lang.NullPointerException at Typeface.java:928`, plus three `SIGSEGV` aborts in `libandroid_runtime.so` and a `DefaultSdkProvider` `UnsupportedOperationException`. That's the Robolectric/native font stack coming up broken on the runner, which is also what makes the base branch red.

On `:daemon:desktop:test` — the one failure not present on the base run — I checked whether this diff could reach it. It doesn't: the change is `samples/android` + `samples/cmp` fixtures/tests, committed contact-sheet PNGs, and one line of `samples/cmp` build wiring. The only `daemon:desktop` tests that touch the CMP sample at all are `TouchOverlayPinchRecordingTest` and `TouchOverlayDrawingCanvasGesturesRecordingTest`, which drive their own existing previews, not the new ones. I ran the task locally on this branch:

- `./gradlew :daemon:desktop:test` — **BUILD SUCCESSFUL**

The truncated log doesn't name which desktop test failed, so I can't rule out a genuine flake there versus more of the same font breakage; I'll re-check on the next run rather than push a no-op commit.

The change's own tests pass locally, on this exact tree:

- `:samples:android:testDebugUnitTest --tests "*PreviewModeMatrixTest*"` — BUILD SUCCESSFUL
- `:samples:cmp:test --tests "*PreviewModeMatrixTest*"` — BUILD SUCCESSFUL

---

Issue #3097, opened 2026-07-31
feat(render): clip round wear devices to their circle, as Android Studio does


yschimke:
Found by the new Studio-parity gate, which diffs our renders against Layoutlib — the engine Android Studio's preview pane uses.

## Symptom

`@Preview(device = "id:wearos_small_round")`:

- **Layoutlib / Studio**: the capture is clipped to the watch face's circle — corners transparent.
- **Ours**: the full 384×384 square is drawn.

Canvas size (384×384) and the resolved configuration (192×192dp @2.0×) agree on both sides; only the mask differs, on ~22% of pixels.

## Why it matters

Round Wear surfaces are exactly where the corners lie to you: content that a real watch clips away looks fine in a square render. We already track roundness — `DeviceDimensions.DeviceSpec.isRound` is resolved from `id:…_round`, `isround=true` and `shape=round` — so the information is present and simply isn't applied at capture time.

## Fixing it

Apply a circular clip to the captured bitmap when the resolved `DeviceSpec.isRound` is true, in both renderers (and the daemon `RenderEngine`), leaving the corners transparent to match Layoutlib. Then update `ParityWearDevicePreview`'s entry in `StudioParityTest`, which currently pins the divergence, and re-render the Wear sample — `:samples:wear` has committed renders that would change.

Worth deciding at the same time whether the round mask should also apply to the Wear tile / notification captures, or only to Compose previews with a round `device`.

Issue #3096, opened 2026-07-31
fix(render): a fixed @Preview frame should measure content with tight constraints, as Android Studio does


yschimke:
Found by the new Studio-parity gate, which diffs our renders against Layoutlib — the engine Android Studio's preview pane uses.

## Symptom

Given a `@Preview` with an explicit frame and a child whose *preferred* size is smaller:

```kotlin
@Preview(name = "fixed-width", widthDp = 240, showBackground = true)
@Composable fun P() { Box(Modifier.size(160.dp, 80.dp).background(Blue)) { Text("fixed-w") } }
```

- **Layoutlib / Studio**: the blue box fills the full 240dp width.
- **Ours**: the blue box is drawn at its intrinsic 160dp and the remaining 80dp is harness background.

Same canvas (630×210 both), ~35% of pixels differ. Also visible on `@Preview(widthDp = 200, heightDp = 100)`.

## Cause

Layoutlib measures the preview content with **tight** constraints equal to the frame. `Modifier.size` is a *preferred* size that yields to tight incoming constraints, so the child stretches. Our renderers measure with loose constraints (`min = 0`, `max = frame`) and letterbox the result:

- desktop: [`ComposePreviewContentBox`](https://github.com/yschimke/compose-ai-tools/blob/main/renderers/desktop/src/main/kotlin/ee/schimke/composeai/renderer/ComposePreviewScene.kt) builds `Constraints` with `minWidth = constraints.minWidth` (0 at the root) on a fixed axis;
- the Android renderer's fixed-frame path has the same shape.

Note this is the same code path as #3092 (`showBackground` not filling a mixed fixed/wrap frame on desktop) — a fix that pins the fixed axis to `min = max = frame` likely closes both, since the child then covers the frame it is supposed to paint.

## Why it matters

This is the difference between "the preview shows what Studio shows" and "the preview shows something plausible". A component sized by its parent in the real app renders at the wrong size in our preview, and any layout that depends on fill behaviour (weights, `fillMaxWidth`, alignment inside the frame) is drawn against the wrong constraint set.

## Fixing it

Pin the non-wrapping axis to tight constraints in both renderers (and the daemon's `RenderEngine`, which shares the desktop box), then update the pinned expectations in `StudioParityTest` — `ParityFixedPreview` and `ParityFixedWidthPreview` currently record the divergence, so closing it fails the test until the table is updated. Re-render the `:samples:android` / `:samples:cmp` matrices and check the committed renders for fallout.

Issue #3095, opened 2026-07-31
fix(sizing): fixed preview axes truncate where Android Studio rounds up (1px short on half pixels)


yschimke:
Found by the new Studio-parity gate, which diffs our renders against Layoutlib — the engine Android Studio's preview pane uses.

## Symptom

A fixed axis whose `dp × density` lands on a half pixel comes out **1px shorter** than Studio's.

| `@Preview` | dp × density | Layoutlib (Studio) | ours |
|---|---|---|---|
| `heightDp = 100` @2.625× | 262.5 | **263** | 262 |
| `heightDp = 60` @2.625× | 157.5 | **158** | 157 |

Whole-pixel frames agree exactly — `id:pixel_5` is 1080×2340 on both, `spec:…dpi=320` is 720×1280 on both — so this is purely the rounding rule at the half.

## Cause

Layoutlib rounds a fixed axis up (`ceil`); we truncate. Our side resolves in `DeviceDimensions.resolveForRender` and is then applied as whole-pixel sandbox dimensions by each renderer, so the `.5` is dropped on the floor.

## Scope

Affects 5 of the 11 parity fixtures today. It is invisible on device previews (whole pixels) and only bites `widthDp`/`heightDp` at densities where the product is fractional — which includes Studio's default 420dpi (2.625×), so it is the common case for hand-sized component previews.

## Fixing it

Switch the fixed-axis conversion to round-half-up and update the pinned expectations in `StudioParityTest` (the table records both engines' sizes deliberately, so the fix flips it) plus the `:samples:android` / `:samples:cmp` `PreviewModeMatrixTest` geometry tables, which currently document `floor`. Both renderers need the same change so the two backends stay identical to each other.

PR #3098, opened 2026-07-31
test(samples): diff our renders against Android Studio's Layoutlib output

yschimke:
## Summary

The preview-mode matrix (#3082) pinned our geometry to numbers **written out from** Studio's device catalog. That answers *"do we match the spec we believe Studio implements"* — not *"do we match Studio"*. Nothing rendered the same `@Preview` through Studio's own engine and compared.

Google ships that engine headlessly: `com.android.compose.screenshot` renders `@Preview`s through **Layoutlib**, which is exactly what Studio's preview pane draws with. So the reference images here are *generated*, not hand-captured — no IDE, no GUI, and no screenshot that silently rots.

- **`StudioParityMatrix`** declares the modes once, in `:samples:android-screenshot-test`'s `screenshotTest` source set, where both engines can reach them: Layoutlib via Google's plugin, ours via the discovery this module already exercises.
- **The Layoutlib references are committed** (the plugin's own convention), so the gate runs anywhere without regenerating anything.
- **`StudioParityTest`** pairs them by preview, pins *both* engines' sizes exactly, and compares pixels where the sizes agree. It emits side-by-side composites for review, uploaded as a CI artifact.
- **`docs/STUDIO_PARITY.md`** documents the setup, the regenerate command, and the live divergences.

Sizes are pinned per engine rather than compared loosely: where they agree that pins the parity, and where they disagree the divergence is written down with its cause, so closing one **fails** the test instead of silently changing output. Rasterisation differs between Layoutlib (native Skia) and Robolectric, so exact equality is unreachable — the tolerance is on the *fraction* of differing pixels, which cleanly separates AA noise (~0.1%) from real divergence (20%+).

## It found three real divergences

| # | What | Impact |
|---|---|---|
| [#3095](https://github.com/yschimke/compose-ai-tools/issues/3095) | **Half-pixel rounding** — a fixed axis is `ceil(dp × density)` in Layoutlib, `floor(...)` for us. `heightDp = 100` at 2.625× is **263** in Studio, 262 for us. | 5 previews, 1px |
| [#3096](https://github.com/yschimke/compose-ai-tools/issues/3096) | **Fixed-frame constraints** — Layoutlib measures content with *tight* constraints equal to the frame, so a `Modifier.size` child stretches to fill it. We measure loose and letterbox. Same canvas, ~35% of pixels differ. Probably the same root cause as #3092. | 2 previews |
| [#3097](https://github.com/yschimke/compose-ai-tools/issues/3097) | **Round device mask** — Layoutlib clips a round wear device to its circle; we render the full square. | 1 preview, ~22% |

The good news is in the same table: device previews and wrapped previews already match Studio **exactly** on geometry, with only sub-percent AA noise.

## Visual evidence

Left = Layoutlib (Studio's renderer), right = ours. Generated by the test itself; the full set of 11 is uploaded as a CI artifact.

Layoutlib vs our renderer, side by side

## Wiring notes

Two things were needed to make Google's plugin run at all, both gated on the existing `-Pandroid.experimental.enableScreenshotTest` property so the default build is unchanged:

- the per-module `experimentalProperties["android.experimental.enableScreenshotTest"] = true` — the gradle property alone fails configuration with *"Please enable screenshotTest source set in module first"*;
- `@PreviewTest` on every fixture (plus `screenshot-validation-api` on that classpath) — from alpha15 the plugin discovers nothing without the marker and fails with *"did not discover any tests"*.

The CI leg that already renders this module now passes the flag and runs the gate, so the comparison happens on every PR rather than on request.

## Test plan

- `./gradlew -Pandroid.experimental.enableScreenshotTest=true :samples:android-screenshot-test:updateDebugScreenshotTest` — BUILD SUCCESSFUL, 11 Layoutlib references generated
- `./gradlew -Pandroid.experimental.enableScreenshotTest=true :samples:android-screenshot-test:testDebugUnitTest --tests "*StudioParityTest*"` — BUILD SUCCESSFUL (2 tests)
- `./gradlew :samples:android-screenshot-test:testDebugUnitTest` (no flag) — BUILD SUCCESSFUL; the gate skips itself rather than failing when the source set is absent
- **Verified the gate bites**: flipping one expected height by 1px fails the test with the exact mismatch; reverted and green again
- `ktfmtFormat` applied

Rebased onto `5009aa1` so it builds on the Robolectric font fix rather than the red base.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fuba6kiB52jrnnuaxc8JtR)_


yschimke:
**Module Unit Tests** is red here, but the failure is on `main`, not from this diff — filed as #3099.

One test of 239 fails:

```
RenderEngineTest > figmaSvgExportRastersBrushBackgroundWithItsCompleteSubtree FAILED
    java.lang.AssertionError: brush-backed layer must use the hybrid raster path
```

I checked out this PR's base `5009aa1` **with no local changes** and ran that single test — it fails identically there. `main`'s own [push run for `5009aa1`](https://github.com/yschimke/compose-ai-tools/actions/runs/30672067290) is red on the same job. The test lives in `:daemon:android`; this PR touches only `ci.yml`, the version catalog, `:samples:android-screenshot-test`, docs and committed PNGs, so it has no path to it.

Worth noting the good news underneath: this is now a *single* failing test rather than the 43-failure Robolectric/Typeface collapse that made this job red all evening — the font fix in #3094 cleared that, and (per #3099) most likely caused this one.

This PR's own gate is green, on this exact tree:

- `-Pandroid.experimental.enableScreenshotTest=true :samples:android-screenshot-test:testDebugUnitTest --tests "*StudioParityTest*"` — BUILD SUCCESSFUL
- same task without the flag — BUILD SUCCESSFUL (skips itself, as designed)

I'll re-run once #3099 is fixed rather than push a no-op commit.

---

PR #3103, opened 2026-08-01
fix(daemon-tests): assert the gradient export contract the exporter actually implements

### LINKED ISSUE #3099: fix(daemon-tests): main is red — RenderEngineTest brush-background raster assertion fails since #3094
`main` is currently red. The [push run for `5009aa1`](https://github.com/yschimke/compose-ai-tools/actions/runs/30672067290) fails **Module Unit Tests**, and it is now a single test rather than the wholesale Robolectric breakage that preceded it:

```
RenderEngineTest > figmaSvgExportRastersBrushBackgroundWithItsCompleteSubtree FAILED
    java.lang.AssertionError: brush-backed layer must use the hybrid raster path
    at RenderEngineTest.kt:398
```

239 tests run, 1 fails.

## Reproduced on an unmodified base

Checked out `5009aa1` with no local changes and ran the single test:

```
./gradlew :daemon:android:testDebugUnitTest \
  --tests "*RenderEngineTest.figmaSvgExportRastersBrushBackgroundWithItsCompleteSubtree*"
→ FAILED, same assertion
```

So it is not environment-specific to CI and not caused by any open PR — it is on `main` itself.

## Likely cause

The failing assertion is `svg.contains("``&lt;image ")' — the exported 'compose-figma.svg' for a brush-backed layer no longer takes the hybrid raster path.``

That points at [#3094](https://github.com/yschimke/compose-ai-tools/pull/3094) (`stop a test fonts/ resource shadowing Robolectric's system fonts`), the commit this run is on. That change altered which fonts the Robolectric render sees, and this test's subject is exactly the raster-vs-editable-text decision in the figma-svg exporter: the layer is rastered when its subtree can't be represented as editable SVG text. With different fonts resolving, the exporter can plausibly now decide the label *is* representable and emit editable text instead of an `<image>`, which is precisely the shape of this failure (the companion `assertFalse(svg.contains(">Gradient<"))` on the next line asserts the opposite of that outcome).

Worth confirming against the sibling test `FigmaSvgDownloadableFontEmbedTest`, which #3094 touched directly.

## Not blocking, but worth knowing

This is unrelated to #3098 (Studio/Layoutlib parity), which is where I hit it — that PR's own gate passes. Flagging separately because a red `main` silently becomes "expected red" for every PR that follows.

yschimke:
Fixes #3099 — `main` is red on `RenderEngineTest.figmaSvgExportRastersBrushBackgroundWithItsCompleteSubtree`.

## It wasn't the font fix

I originally suspected #3094, which landed alongside it. That was wrong, and I've corrected the issue: **the test fails identically on `a30b0f9`, the commit before the font fix**. #3094 only made the failure *visible*, by clearing the 43-failure Robolectric/Typeface collapse this one was buried inside.

## What's actually happening

The fixture uses `Brush.horizontalGradient`. Here is what the exporter produces today:

```svg
<defs>
  <linearGradient id="gf-0" x1="0" y1="0" x2="1" y2="0">
    <stop offset="0" stop-color="#FF0000"/>
    <stop offset="1" stop-color="#0000FF"/>
  </linearGradient>
</defs>
...
<rect x="0" y="0" width="128" height="48" rx="8" ry="8" fill="url(#gf-0)"/>
<text x="37" y="29.02" font-size="14" fill="#FFFFFF">Gradient</text>
```

No `<image>`, no `figma-raster/` crop — fully vector, gradient preserved exactly, label still editable in Figma. The test demands the opposite ("brush-backed layer must use the hybrid raster path").

That output is **correct and better**, and it's the documented contract. #2852 added `ModifierTokenResolver.linearGradient` to capture a linear brush as a real gradient def, and its KDoc is explicit that it returns null for radial/sweep/shader brushes *"so those keep the raster fallback rather than being emitted as a gradient they aren't"*. `FigmaLayeredSvgTest.aGradientBackgroundIsEmittedAsALinearGradientFill` already pins the same rule at unit level and passes. The end-to-end test just never got updated when the feature landed, and has been failing since — masked.

## The fix

Rather than relax the assertion and lose what it was guarding, the contract is split across two tests, so "gradients no longer raster" can't quietly become "nothing rasters":

| test | asserts |
|---|---|
| `figmaSvgExportEmitsALinearBrushBackgroundAsAGradientWithEditableText` | `<linearGradient>` def emitted, rect references it via `fill="url(#…)"`, `>Gradient<` **stays** editable, and **no** `<image >` or `figma-raster/` directory is produced |
| `figmaSvgExportRastersARadialBrushBackground` *(new)* | a radial brush still bakes to exactly one raster crop, is **not** emitted as a linear gradient, and the crop is red at the centre → blue at the edge |

The new `RadialGradientBackgroundCard` fixture is deliberately identical to `GradientBackgroundCard` apart from the brush, so the pair isolates the one variable that decides the path.

The old test's end-to-end colour claim is kept rather than dropped — red on the left, blue on the right — now read off the gradient's stops instead of sampled crop pixels, since there are no longer any crop pixels to sample.

## Test plan

- `:daemon:android:testDebugUnitTest --tests "*figmaSvgExportEmitsALinearBrushBackgroundAsAGradientWithEditableText*" --tests "*figmaSvgExportRastersARadialBrushBackground*"` — **BUILD SUCCESSFUL**, both pass
- Verified the failure predates #3094 by running the original test on `a30b0f9` — fails there too
- Captured the exported SVG directly to confirm the emitted shape quoted above, rather than inferring it
- `ktfmtFormat` applied
- Full `:daemon:android:testDebugUnitTest` was still running locally when this was pushed; CI's Module Unit Tests leg is the authority and should now go green. I'm tracking it and will follow up here if anything else in that module surfaces.

No visual surface changes — this is test/fixture only, so there is nothing to render for before/after evidence.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fuba6kiB52jrnnuaxc8JtR)_

### DIFF
--- a/daemon/android/src/test/kotlin/ee/schimke/composeai/daemon/RenderEngineTest.kt
+++ b/daemon/android/src/test/kotlin/ee/schimke/composeai/daemon/RenderEngineTest.kt
@@ -373,8 +373,22 @@ class RenderEngineTest {
     }
   }
 
+  /**
+   * A **linear** brush background exports as a real `<linearGradient>` fill with its label left
+   * editable — no raster at all.
+   *
+   * This test used to assert the opposite ("brush-backed layer must use the hybrid raster path"),
+   * which was correct when any brush defeated flat-colour resolution and forced the whole subtree
+   * to be baked. #2852 superseded that for linear gradients: the brush is now captured and emitted
+   * as an SVG gradient def, so the export stays vector and the text stays editable in Figma. The
+   * assertion was simply never updated, and the failure sat masked inside a larger Robolectric
+   * breakage on `main` until that was fixed (#3099).
+   *
+   * [figmaSvgExportRastersARadialBrushBackground] holds the other half of the contract, so
+   * "gradients no longer raster" can't quietly become "nothing rasters".
+   */
   @Test
-  fun figmaSvgExportRastersBrushBackgroundWithItsCompleteSubtree() {
+  fun figmaSvgExportEmitsALinearBrushBackgroundAsAGradientWithEditableText() {
     val outputDir = tempFolder.newFolder("renders-figma-gradient")
     System.setProperty(RenderEngine.OUTPUT_DIR_PROP, outputDir.absolutePath)
     System.setProperty("roborazzi.test.record", "true")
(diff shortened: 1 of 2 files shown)


yschimke:
Closing the loop on the one hedge in the test plan: the full module suite finished locally after I pushed —

```
./gradlew :daemon:android:testDebugUnitTest
BUILD SUCCESSFUL in 15m 36s
```

That's the whole module green, not just the two targeted tests, so nothing else in `:daemon:android` was depending on the old raster behaviour.

---

PR #3146, opened 2026-08-01
feat(remotecompose): add animated Material 3 previews

yschimke:
## Summary

- add standard and embedded-player previews for Remote Material 3 indeterminate and click-driven circular progress animations
- add an embedded preview wrapper so the same captured Remote Compose document can be exercised through both replay implementations
- make continuous time, float overrides, and nested animated expressions reactive in the embedded player
- add focused regression coverage for continuous-time and interaction-driven animations

## Why

The Remote Material 3 samples previously had no animated preview coverage, and the embedded replay lane could freeze document-driven time or flatten nested `animateRemoteFloat` expressions to their target values. That made comparisons with the standard View-backed RC player incomplete.

## Validation

- rendered all four circular-progress previews with `compose-preview`; 11 total module previews, 0 missing
- recorded the standard and embedded click-driven previews
- `:third-party-rc-embedded-player:testDebugUnitTest` focused animation tests
- scoped `ktfmtCheck` for `:samples:remotecompose`, `:data-remotecompose-connector`, and `:third-party-rc-embedded-player`
- `git diff --check`

### DIFF
--- a/data/remotecompose/connector/src/main/kotlin/ee/schimke/composeai/daemon/RemoteComposeIrReplay.kt
+++ b/data/remotecompose/connector/src/main/kotlin/ee/schimke/composeai/daemon/RemoteComposeIrReplay.kt
@@ -85,7 +85,7 @@ class RemoteComposeIrReplay {
  *
  * Invalid hex is skipped rather than thrown, matching [applyConnectorOverrides].
  */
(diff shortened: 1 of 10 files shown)


yschimke:
The animated one doesn't look animated

yschimke:
Make further fixes in a follow up PR. Going to merge

PR #3153, opened 2026-08-01
fix(daemon): size a device frame from the catalog on the live lane too

yschimke:
Combines this PR with #3151 (now merged), which fixed the same bug from the other end. #3151 owns the batch resolver; this PR carries what it didn't reach — the **live/stream lane** — and lifts its precedence rule so all four resolvers share one implementation.

## What #3151 already landed

`PreviewManifestEntry.resolved()` on both daemons resolves a manifest-declared `@Preview(device = …)` against `DeviceDimensions`, with annotation dp overriding the catalog when both axes are set, truncating dp→px, and falling back to the device's density. That fixed the bake/route path and the `figma-svg` canvas (#2615 / #2883).

## What this PR adds on top

- **`renderSpecFromInfo()` on both daemons** — the held-interactive / stream / serve lane, which #3151 didn't touch. It made the same decision independently and still fell through to the fixed frame, so a `@Preview(device = "id:wearos_small_round")` streamed at 320×320 (android) or the 400×800 dp sandbox bound → 1050×2100 (desktop) instead of its 384² device box. A held session therefore drifted from the preview's own baked snapshot.
- **One shared precedence rule.** #3151 put `frameDpOverriddenBy` on `DeviceDimensions` but applied it in two places; the two `renderSpecFromInfo` copies were still free to disagree. It's now the single source for all four sites (`resolved()` × 2, `renderSpecFromInfo()` × 2), so annotation dp displace the catalog only when *both* axes are set — on every lane, not just the bake.
- **Device density on the stream lane too** — a bare `spec:width=340dp,height=800dp,dpi=160` (Jetchat `Profile/Screen`) now streams at 1.0, matching what the bake resolves, instead of the session default.
- **Routed-payload coverage**, which the Codex review on #3149 asked for: assertions on the wire payload the router actually emits, not just on the resolver's return value.

## Test plan

- `PreviewManifestRouterRoutingTest` (new) — `routePayload` emits `widthPx=384;heightPx=384`, the device id, and no wrap token for a manifest-declared Wear device; a second case pins that an inbound explicit `widthPx=600` still wins while the untouched axis stays device-derived.
- `RenderSpecFromInfoTest` — the stream lane frames `id:pixel_5` at the catalog's 393×851 dp and **2.75** density (1080 px wide, truncated not rounded), and honours the same both-axes-set annotation override (`200²` when both are set, `454²` when only one is).
- `PreviewManifestEntryResolveTest` (both daemons) — keeps #3151's catalog / precedence / dpi cases and adds the single-vs-both-axis pair on desktop.
- `./gradlew :daemon:core:test :daemon:desktop:test :daemon:android:testDebugUnitTest` → green. Two unrelated flakes surfaced in this container — `RenderEngineLottieTest.lottieAssetRendersBlueSquare` and `CancellationInvariantTest.shutdownDoesNotInterruptInFlightRender` — both pass on re-run and neither is touched by this diff.
- Merged onto `main` after #3151 landed; the shared helper is deduplicated to a single definition.

## Visual evidence

No render output changes here. The geometry this fixes is already shown in #3151's before/after (the Wear `TransformingLazyColumn` SVG lane going `352×352 → 486×486`) — this PR moves the *live* lane onto that same frame, and the live lane emits no committed artifact of its own. The CI preview-diff bot reports every device preview unchanged, which is the intent: the daemon's stream geometry moves onto the PNG's without moving the PNG.

(The one variant the bot keeps flagging, `RemoteIndeterminateCircularProgressIndicatorStandardPreview`, is a nondeterministic animated-GIF capture — it has flagged identically on three unrelated PRs, including ones that touch no sample code.)

---
_Generated by [Claude Code](https://claude.ai/code/session_01WPgP8v6tA9nMoisxKqjU3V)_


yschimke:
Rounding review point is valid — fixed in `d45ee01`.

`RenderPreviewsTask`'s device-frame branch converts with `(dp * density).toInt()` (the `roundHalfUpPx()` branch only applies when `explicitWidthDp != null`, which is null for a device frame), and the router's existing inbound-`device` path truncates too. My four new paths rounded half-up, so `id:pixel_5` — 393dp × 2.75 = 1080.75 — would have rendered 1081 px live against a 1080 px baked snapshot. That's precisely the parity this PR is meant to restore, so: truncate on all four, comment the rule where it's easy to get wrong, and pin the case in `routePayload truncates a fractional device frame exactly like the bake` (1080 / 2340).

`RenderSpecFromInfoTest` moved from `roundToInt()` to `toInt()` for the same reason.

---

Separately, the red **`Serve render lanes honor overrides`** check on this PR is not from this change — it's the `Wasm iframe re-renders on knob override` breakage that has been failing on `main` and every branch since #3127, fixed by #3150. This branch only touches daemon Kotlin. Once #3150 lands I'll merge `main` in here so the check re-runs against the fix.

---

PR #3163, opened 2026-08-02
fix: preserve Wear TimePicker SVG fidelity

yschimke:
## Goal

Make Horologist TimePicker SVG rendering match its PNG output. Prefer true SVG, but detect and rasterize complex screen regions when necessary.

## Summary

- Always captures the visible string and authored typography from Compose text modifiers; the rendered `TextLayoutResult` remains authoritative for resolved pixels, wrapping, baselines, and spans.
- Uses plain accessibility semantics text only as a compatibility fallback for older captures that lack modifier text, so semantic overrides cannot replace visible SVG copy.
- Preserves Android 14+ nonlinear font scaling when the measured layout-result hook has been cleared.
- Detects visibility-changing `drawWithContent` behavior by replaying it with two near-identical descendant fills. Clips, masks, fades, transforms, clears, opaque replacement, and omitted content use a narrow frame crop—with no component-name checks.
- Keeps pass-through/background draws and overlays whose output remains sensitive to descendant content—such as Remote Compose and additive gradient tints—layered with editable descendants.
- Adds regressions for the legacy Wear `PickerGroup`, generic content clipping, gradient-tinted controls, text-source precedence, nonlinear font scaling, and Remote Compose custom-draw content.

### Approaches considered

- Fully reconstructing the Picker’s dynamic scaling alpha and read-only mask as SVG was rejected because those effects are executed inside an opaque `drawWithContent` block and are not represented in the captured layout or semantics trees.
- Rasterizing the whole TimePicker was unnecessary; only the two complex picker columns use frame crops. The separator and surrounding UI remain editable SVG.

### Known gaps

- Picker-column text is intentionally raster-backed in hybrid mode. A future draw-scope clip/alpha recorder could make those columns fully editable without sacrificing fidelity.

## Visual evidence

The top row shows the supplied catalog PNG target against the released SVG. The bottom row shows the focused Wear Picker PNG capture against the local hybrid SVG output.

TimePicker PNG and SVG comparison

- [Open the generated hybrid SVG](https://raw.githubusercontent.com/yschimke/compose-ai-tools/e636738ac55d4474aef8faae2853742ed0672a9d/renders/timepicker-svg-fidelity/compose-figma.svg)
- [Open the matching renderer PNG](https://raw.githubusercontent.com/yschimke/compose-ai-tools/e636738ac55d4474aef8faae2853742ed0672a9d/renders/timepicker-svg-fidelity/timepicker.png)

The SVG bundle includes two relative `figma-raster` crops for the opaque Picker draw effects; the colon remains a normal `<text>` element.

## Verification

- [x] `./gradlew :data-layoutinspector-connector:ktfmtFormat :data-layoutinspector-core:ktfmtFormat :renderer-android:ktfmtFormat`
- [x] `./gradlew :daemon:android:compileDebugKotlin`
- [x] `./gradlew :data-layoutinspector-core:test`
- [x] `./gradlew :renderer-android:testDebugUnitTest --rerun-tasks` (172 tests, 1 skipped)
- [x] `./gradlew :third-party-rc-embedded-player:testDebugUnitTest --tests ee.schimke.composeai.rcembedded.RcFigmaSvgExportTest`
- [x] Rebased onto current `origin/main` (`415153c35`)


yschimke:
Is there a rendered svg of the final result that matches the png? I don't see one.

PR #3165, opened 2026-08-02
fix(serve): prioritize visible theme renders

### LINKED ISSUE #3160: Changing theme on https://preview.coo.ee/confetti-mobile/ doesn't always refresh
Expect when cahnging to Kotlinconf dark that image requests will fire, but seemingly doesn't trigger anything.

No changed images,but changing Dark/Light does work.

yschimke:
## Summary

- prioritize cards visible in the active tab/search when applying a declared theme
- preserve the daemon's serialized render queue, delayed retry, and completion of hidden cards
- update the generated landing-page fixture and add a regression assertion

## Root cause

Declared-theme renders are intentionally serialized because the preview daemon renders one image at a time. The queue followed document order, so on later tabs it rendered hidden Theme and Component cards before the cards the visitor could see. The selected theme changed immediately, but visible image requests could be delayed long enough to look like no refresh happened.

## Validation

- `./gradlew :cli:ktfmtCheck :cli:test --tests ee.schimke.composeai.cli.serve.ServeWebFixtureTest`
- reproduced against the live Confetti catalog by selecting KotlinConf Dark and observing the document-order queue delay

There is no static layout/pixel change: this only changes request priority after an existing theme control is selected.

Closes #3160

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeWeb.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeWeb.kt
@@ -963,6 +963,13 @@ object ServeWeb {
     // Under a DECLARED theme a swap card keeps its server-side default variant's metadata (label /
     // id / viewer link / stage backing, from `data-def`) — only the pixels come from the themed
     // render — so picking a theme never silently flips the light/dark axis too.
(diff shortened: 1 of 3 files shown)


yschimke:
How confident and how do you know this is the cause?

PR #3282, opened 2026-08-04
feat(catalog): select one breakpoint of a multipreview without splitting the @Preview

yschimke:
## Goal

Make a catalog able to say "this card is the large-round render" without the module hand-splitting its multipreview into per-device `@Preview` siblings — the workaround [joreilly/Confetti#1778](https://github.com/joreilly/Confetti/pull/1778) reaches for, which costs 111 lines, five shared-fixture indirections, and the `@WearPreviewFontScales` coverage on all five screens.

## Summary

A multipreview (`@WearPreviewDevices`, a local `@CatalogWearBreakpoints`) renders one function at several device sizes, and the candidate join keys on **function name** — so every size folds onto one spec entry. Right when the entry means "this component, at every breakpoint we document"; wrong when the catalog wants a card per breakpoint with its own id and caption.

**`select`** — a spec component or variant may pick one value of its function's fan-out, so two entries can share one `preview` and stay separate cards:

```jsonc
{ "componentId": "Home/SmallRound", "preview": "HomeListViewPreview",
  "select": { "size": "smallRound" }, "caption": "Home — small round." },
{ "componentId": "Home/LargeRound", "preview": "HomeListViewPreview",
  "select": { "size": "largeRound" }, "caption": "Home — large round." }
```

- `selectComponentImages` / `selectImages` narrow an entry to one axis value. A selection matching nothing is a **missing render that names what the function did produce** (`Home/XL [select size=xlRound; HomeListViewPreview renders size ∈ {largeRound, smallRound}]`), not a silent fold of everything.
- `foldVariants` accepts `select` on a variant, counts it as a distinguishing axis, and still recognises a same-function variant now that the fold can be handed a filtered view.
- The validator checks the shape, rejects an unknown axis or an undeclared breakpoint name (with a did-you-mean), and stands down the existing "these fold into one sticker" warning when every reference selects a distinct value — it still fires when one reference selects nothing, since that one really does fold in its sibling's renders.
- Only `size` is selectable today: it is the one axis the serve grid keeps as separate cards (theme/state/props fold into one card with a switcher), so it is the only one an author had to reach for Kotlin to control. The object shape means a second axis needs no spec migration.

**The size axis becomes exact rather than width-derived.** A `breakpoints` entry may now name the `@Preview(device = …)` id, matched ahead of `widthDp`:

```jsonc
{ "size": "smallRound", "device": "id:wearos_small_round", "widthDp": 192 }
```

The device is the expansion's own identity (carried verbatim from `BUILT_IN_MULTIPREVIEW_EXPANSIONS`); a width is a fingerprint two devices can share, and a device no breakpoint names silently keeps the generic `compact` class — indistinguishable from the sibling that did match. Alongside that:

- the live-preview bridge scores the same device id, and reads `catalogBreakpoints(spec)` rather than `spec.breakpoints`, so a Wear catalog declaring no breakpoints no longer leaves every size unconstrained and resolves both stickers to the first-listed annotation;
- the Wear default table gains device ids plus `xlRound` / `smallSquare`;
- the export **warns** about any device id no breakpoint claims, instead of collapsing it onto `compact` and surfacing it much later as an overwritten sticker or a duplicate-axis failure.

## Scope notes

- `select` applies to spec-authored `groups`. An annotation-led inventory (`@CatalogComponent`, which is what this repo's own catalogs use) has no equivalent yet — it fans out one card per breakpoint, which the serve grid disambiguates by appending the breakpoint to a *colliding* card label (`Edgebutton · Small Round`, #3268). Noted in the docs; a `@CatalogComponent` selector is the natural follow-up.
- No committed catalog changes shape here, so no rendered output moves — this is opt-in spec vocabulary plus a strictly-better axis resolution. Text-only PR: nothing visual changes until a catalog adopts `select`.

## Test plan

- `node --test` in `scripts/design-artifacts`: **422 pass, 1 fail**, the failure being the pre-existing `rc-compare-pixels.test.mjs` `ERR_MODULE_NOT_FOUND: pngjs` in this container (no `npm ci`); it fails identically on a clean tree.
- New coverage: `catalog-select.test.mjs` (12 tests — selection semantics, the untagged-image strictness, the failed-selection report, and the `foldVariants` interactions), plus device-matching / undeclared-device tests in `catalog-breakpoints.test.mjs` and two Wear resolution tests in `bridge-live-preview-ids.test.mjs` (both fail before this change — the two annotations tie and the first-listed id wins for both stickers).
- `validate-samples.test.mjs` still validates all three committed sample specs unchanged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mwx5Kh8gyDeZ5Ua2Yukx46)_

### DIFF
--- a/docs/design/DESIGN_CATALOGS.md
+++ b/docs/design/DESIGN_CATALOGS.md
@@ -340,6 +340,59 @@ The spec shape is described by
 [`scripts/design-artifacts/catalog.spec.schema.json`](../../scripts/design-artifacts/catalog.spec.schema.json)
 (referenced via `$schema` in each sample spec for editor validation).
 
+### Breakpoints and multipreviews: `select`, not a split `@Preview`
+
+A multipreview annotation (`@WearPreviewDevices`, a local `@CatalogWearBreakpoints`)
+renders one function at several device sizes. The candidate join keys on **function
+name**, so all of those renders fold into one spec entry carrying one image per size,
+tagged from the spec's `breakpoints`:
+
+```jsonc
+"breakpoints": [
+  { "size": "smallRound", "device": "id:wearos_small_round", "widthDp": 192 },
+  { "size": "largeRound", "device": "id:wearos_large_round", "widthDp": 227 }
+]
+```
+
+Declare each breakpoint by `device` (the `@Preview(device = …)` id, matched first) as
+well as by `widthDp` (the fallback, and what the live-preview bridge scores a
+candidate annotation against). The device is the expansion's own identity; a width is
+a fingerprint two devices can share, and a device no breakpoint names keeps the
+generic Material width class — which makes two distinct renders indistinguishable on
+one axis. The export warns when it sees one. A Wear catalog that declares no
+`breakpoints` inherits the standard round table (`smallRound` / `largeRound` /
+`xlRound` / `smallSquare`, by device id and width).
+
+Folding every size onto one entry is right when the entry means "this component, at
+every breakpoint we document". When a catalog wants a **card per breakpoint** — its
+own id and caption — use `select` rather than splitting the `@Preview` function in
+the module:
+
+```jsonc
+{ "componentId": "Home/SmallRound", "preview": "HomeListViewPreview",
+  "select": { "size": "smallRound" }, "caption": "Home — small round." },
+{ "componentId": "Home/LargeRound", "preview": "HomeListViewPreview",
+  "select": { "size": "largeRound" }, "caption": "Home — large round." }
+```
+
+Two entries may share one `preview` as long as each selects a different value; the
+validator's "these fold into one sticker" warning stands down for that case, and
+fires as before when one of the references selects nothing (it would fold in the
+renders its sibling selected). A variant may `select` too, and a selection counts as
+its distinguishing axis. A selection that matches no render is reported as a missing
+render naming what the function *did* produce, so a typo — or an undeclared
+breakpoint — reads as itself rather than as a broken `preview`.
+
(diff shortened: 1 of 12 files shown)


yschimke:
The 29 changed variants above are **inherited baseline drift, not this PR** — it touches no Kotlin at all (`git diff --name-only origin/main...HEAD` is 10 `.mjs` files, the spec schema, and one `.md`).

The bot's before/after are `ca006d5` ("Update preview baselines from `4822114`", rendered 09:23Z) and `b74d235` (this PR, 11:30Z). Everything that landed on main in between is folded into the "after" side, and one of those commits is exactly the wear renderer change:

- `6e23440` — `fix(wear): measure device-less Wear previews against the watch screen, don't pin it` (#3278), merged 11:57 local / after the baseline render.

That's what moved all 26 wear stickers (23 × `samples:design-catalog-wear-m3`, 3 × `samples:wear`). It's the phantom-diff mode `compose-preview.yml`'s own concurrency comment documents — a stale baseline makes the next PR report the skipped commits' output as its own.

I checked #3278 against this PR specifically, since both touch Wear sizing, and they don't interact: the retarget sets `PreviewParams.wrapSandboxWidthDp` / `wrapSandboxHeightDp` and explicitly **not** `widthDp` / `heightDp`, and it stamps no `device`. Device-bearing previews (`id:wearos_*_round`) take the earlier `resolveForRender` branch, unchanged. So `params.device` and `params.widthDp` — the only two fields this PR's breakpoint matching reads — are identical before and after it. If anything the two changes are complementary: keying breakpoints on the device id rather than a width makes the axis independent of exactly the measurement policy #3278 adjusts.

No action taken. The next baseline render on main should clear the list.

---

PR #3285, opened 2026-08-04
feat(catalog): declare a component's breakpoints on @CatalogComponent

yschimke:
## Goal

Close the gap left by #3282. `select` let a **spec** entry pick one breakpoint out of a multipreview's fan-out — but only for spec-authored `groups`. An annotation-led inventory (this repo's own catalogs, `wear-m3` and its three round breakpoints included) had no way to say it, so a card per breakpoint meant either reintroducing `groups` or splitting the `@Preview` in the module.

## Summary

`@CatalogComponent(sizes = […])` and `@CatalogVariant(size = "…")` — the annotation form of the spec's `select`, meaning exactly the same thing to the export:

```kotlin
@CatalogComponent(id = "Layout/List", sizes = ["smallRound", "largeRound"])
@CatalogWearBreakpoints @Composable fun ListLayout() = FullScreenWear { … }
```

`inventoryFromPreviews` expands that into one component per breakpoint carrying the `select` the export already understands, so **one annotation produces exactly what two hand-written spec entries would** and nothing downstream needs to change.

| `sizes` | Result |
|---|---|
| absent (default) | today's behaviour — every render folds onto one component |
| one name | narrows without splitting, **plain id kept** (suffixing would move a published sticker's URL to say what the id already says) |
| several | one component each, id suffixed `<id>/<size>` — the breakpoint name verbatim, so there's no casing convention to get wrong in one place and not the other |

Details worth a reviewer's eye:

- **Variant attachment.** `@CatalogVariant(of = …)` names the parent as the annotation spells it, which the fan-out has suffixed — so the first expansion keeps the plain id registered. Without that, every variant on a fanned-out component silently orphans. A variant may also name a suffixed id (`of = "Layout/List/largeRound"`) to target one breakpoint.
- **`display.hero`.** The build-free pre-flight resolves the hero against ids scanned from source, so `discoverComponentIds` now predicts the same expansion — through one shared `catalogSizeExpansion` helper, so the suffix can't tag stickers one way and validate the hero another. Without it, a hero naming a fanned-out id would be a false pre-flight failure.
- **Blank names dropped** at discovery, so a stray comma in `sizes = ["", …]` can't mint a component whose `select` matches nothing.
- A mistyped breakpoint isn't caught by the pre-flight (it reads source, not the render); it fails the publish with #3282's selection report, naming the sizes the function actually rendered.

## Scope notes

- **No catalog adopts it here.** Adopting `sizes` on an already-published catalog moves those components' sticker URLs, since the id is part of the path — `wear-m3`'s six `@CatalogWearBreakpoints` components are the obvious candidates, but that's a migration call, not a drive-by. #3268 already gives those cards distinct labels (`Edgebutton · Small Round`), so the fan-out is for when you want authored ids and captions per breakpoint, not merely readable ones.
- Nothing visual changes and no rendered output moves: this is opt-in vocabulary with an inert default. Text-only PR.

## Test plan

- `./gradlew :preview-annotations:compileKotlinJvm :gradle-plugin:preview-discovery:compileKotlin` — BUILD SUCCESSFUL
- `./gradlew :gradle-plugin:preview-discovery:test :preview-annotations:jvmTest` — BUILD SUCCESSFUL
- `./gradlew ktfmtCheckAll` — BUILD SUCCESSFUL
- `node --test` in `scripts/design-artifacts` — **433 pass, 1 fail**, the failure being the pre-existing `rc-compare-pixels.test.mjs` `ERR_MODULE_NOT_FOUND: pngjs` in this container (no `npm ci`); it fails identically on a clean tree.
- New coverage: 7 inventory tests (fan-out, single-size narrowing, the unchanged default, duplicate-name collapse, variant attachment by both plain and suffixed id, variant `size` → `select`, spec override) and 2 in `catalog-spec.test.mjs` (id prediction from the source scan, hero resolution through a fan-out). `DiscoveryFunctionalTest` extended to assert both new annotation arguments survive into `previews.json`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mwx5Kh8gyDeZ5Ua2Yukx46)_


yschimke:
This feels confusing to developers, and stringly typed.

Can we find a better expression of it

yschimke:
Agreed, and there was a worse problem underneath the stringly-typed one: it was **redundant**. The multipreview on the line below already decides which devices the function renders at, so `sizes` was a second source of truth that could disagree with the render — plus a plural/singular rule (one name narrows, several fan out) you had to learn.

Reworked to a flag that says *split this per breakpoint*, never *which*:

```kotlin
@CatalogComponent(id = "Layout/List", perBreakpoint = true)
@CatalogWearBreakpoints @Composable fun ListLayout() = FullScreenWear { … }
```

The names come from the renders — each `@Preview`'s `device` / width resolved through the same matcher that tags the baked stickers — ordered by the catalog's `breakpoints` table. Nothing to misspell, nothing to keep in sync, and adding a device to the multipreview adds its card for free. `@CatalogVariant.size` went with it: a variant narrowing to one breakpoint is a subset selection, which the spec's `select` already expresses.

Behaviour now derives rather than being configured:

| Function renders | Result |
|---|---|
| flag unset | one component, every render folded — unchanged for every existing catalog |
| one breakpoint | one card, **plain id** (the suffix would move a published URL to say what the id says) |
| several | `<id>/<size>` each, in `breakpoints` order so they read small→large |
| none resolvable | component kept whole + a warning naming it, so an undeclared device reads as itself |

Also picked up both Codex findings in c54b58c, and the second one changed shape under this redesign:

- **Mirrored DTO** — `sizes`/`size` were missing from `api/preview-data-api`'s `CatalogEntry`, whose readers use `ignoreUnknownKeys`; now `perBreakpoint` is mirrored with wire coverage for both the set and absent cases. Good catch, that was a real silent-drop.
- **Unsuffixed hero** — right that the plain id shouldn't validate as a hero. But the fix no longer works as proposed: the ids depend on what *rendered*, which the build-free source scan can't see, so predicting them there isn't possible. Instead the pre-flight resolves a `<id>/<breakpoint>` hero on its parent id (documented as deliberate leniency), and a **new export-time check** resolves `display.hero` against the merged inventory — where every id is known, and which covers authored and derived ids alike. That's the exact check the source-scan prediction was standing in for.

The earlier `Android (Robolectric) serve lane` failure was an infrastructure flake — `SSLHandshakeException` fetching `lifecycle-viewmodel-wasm-js-2.9.6.klib` from Maven Central on `:rc-player-wasm`, nothing to do with this diff. This push re-runs it.

---

PR #3312, opened 2026-08-04
fix(renderer): keep smallestScreenWidthDp in step with the preview viewport

### LINKED ISSUE #3309: Preview device override leaves smallestScreenWidthDp inconsistent with viewport
## Problem

A Compose preview rendered with `device = "id:wearos_large_round"` gets a 227 × 227dp viewport, but `LocalConfiguration.current.smallestScreenWidthDp` remains approximately 320dp instead of matching the installed device configuration.

That produces internally inconsistent configuration values for composables that read Android's `sw<N>dp` field. The live Horologist reproduction is:

https://preview.coo.ee/horologist/render/layoutfillmaxrectangle__ideal__default__compact.png?themeProvider=com.google.android.horologist.catalog.HorologistBlueThemeCatalog

The preview manifest reports:

- device: `id:wearos_large_round`
- widthDp: `227`
- heightDp: `227`
- density: `2.0`

Horologist's current calculation is:

```kotlin
val screenHeightDp = LocalConfiguration.current.screenHeightDp
val screenWidthDp = LocalConfiguration.current.smallestScreenWidthDp
val maxSquareEdge = sqrt((screenHeightDp * screenWidthDp / 2).toDouble())
```

The rendered rectangle is about 190.6dp wide, which is exactly:

```
sqrt(227 × 320 / 2) ≈ 190.6
```

rather than the 160.5dp square inscribed in a 227dp circle.

## Expected

When compose-preview installs a device override, all related `Configuration` dimensions should describe that device consistently. For this round 227dp preview, `screenWidthDp`, `screenHeightDp`, and `smallestScreenWidthDp` should not disagree.

## Notes

Horologist will separately fix `fillMaxRectangle()` to use actual layout constraints, since `smallestScreenWidthDp` is a resource qualifier and should not be used as live geometry. This issue is for the renderer inconsistency itself, which can affect other previews that legitimately inspect the full Android configuration.

## Reproduction source

- Preview: `catalog/src/main/java/com/google/android/horologist/catalog/LayoutCatalog.kt`
- Modifier: `compose-layout/src/main/java/com/google/android/horologist/compose/layout/FillMaxRectangle.kt`
- compose-ai-tools: `0.19.32`

yschimke:
Fixes #3309.

## The bug

`RuntimeEnvironment.setQualifiers("+…")` **merges** with the current state rather than replacing it — a token we don't emit keeps whatever was there before. All three render paths emitted `w…dp`/`h…dp` but never `sw…dp`, so Robolectric's baseline `sw320dp` survived every device override.

A `device = "id:wearos_large_round"` preview therefore rendered into a 227 dp viewport while still reporting `smallestScreenWidthDp == 320`. Anything reading the full `Configuration` disagreed with the pixels it was drawing into — the issue's repro being Horologist's `fillMaxRectangle()`, which inscribed itself in a `sqrt(227 × 320 / 2) ≈ 190.6` dp square instead of the 160.5 dp one that fits a 227 dp circle.

## The fix

Emit `sw…dp` — the narrower axis, matching what Studio derives for the device a preview is pinned to — ahead of `w`/`h`, which is where the [resource qualifier grammar](https://developer.android.com/guide/topics/resources/providing-resources#QualifierRules) requires it.

The three qualifier builders were hand-copied and had already drifted from each other, so the token comes from one shared `previewSizeQualifiers` helper in `:data-render-core` rather than being pasted a fourth time. All three now call it:

- `RobolectricRenderTest.applyPreviewQualifiers` — standalone renderer
- `RenderEngine.applyPreviewQualifiers` — daemon one-shot
- `RobolectricHost` — daemon held session

## Visual evidence

`:samples:wear`'s device-matrix preview prints the configuration it was handed, so the PNG is its own evidence. Added `sw … dp` to it — before/after over all four Wear devices, same render command:

smallestScreenWidthDp before and after, across four Wear devices

Every device reported `sw 320 dp` before; each now reports its own width (192 / 227 / 240 / 385). This is also the "wire it into the pipeline" half — the field is in a committed `@Preview`, so any future regression shows up in the CI visual diff without anyone remembering to check for it.

## Test plan

- `:data-render-core:test` — new `PreviewSizeQualifiersTest` pins the emitted triple and its order (portrait, landscape, single-axis, empty). ✅
- `:renderer-android:testDebugUnitTest` — new `PreviewSizeQualifiersConfigurationTest` pins what Robolectric does with it: `sw`/`w`/`h` all land on 227 for a Wear viewport, a landscape phone reports its narrow axis, **and** the pre-fix qualifier set (`w227dp-h227dp` alone) still yields `smallestScreenWidthDp == 320` — the regression itself, asserted. ✅ 3/3
- Full `:renderer-android:testDebugUnitTest` + `:daemon:android:testDebugUnitTest` — BUILD SUCCESSFUL, no baseline or pixel regressions from the added qualifier. ✅
- `:samples:wear:composePreviewRenderAll` rendered clean on both sides of the change (the images above).
- `./gradlew ktfmtCheckAll` ✅

CI on this PR: every check green except the two serve-lane jobs, which fail identically on this PR's base commit — see the comments below for the base-branch evidence and root cause.

Note for consumers: this also makes `values-sw…dp/` resource selection follow the previewed device, which it previously never did. That's the intended Studio-parity behaviour, but it is a behaviour change for any preview that ships smallest-width resource buckets. The full-corpus preview diff moved only the two `DeviceSpecScreen` callers (8 variants), leaving 524 variants across 324 functions byte-identical, so nothing in the sample set was relying on the stale bucket.


yschimke:
**`Serve render lanes honor overrides` is red on the base branch too — not this PR.**

Failing case: `serve-lanes.spec.mjs:162 › viewer wires the knob into every render lane`, at line 186:

```
Received string:  "blob:http://127.0.0.1:8725/ead347f3-…"
```

Both the initial attempt and the retry fail identically, so it isn't a flake.

**Why it can't pass.** `refreshSnapshot()` in `viewer.js` no longer assigns the render URL to `#cp-img` at all — it `fetch`es the bytes and hands the image an object URL (`img.src = objectUrl; img.setAttribute("data-cp-blob", objectUrl)`). That landed in a8cc181 (#3301, "keep themes across snapshot overrides") to stop the double-render race on `no-store` override renders. `#cp-img`'s `src` is now *always* `blob:…`, and no attribute carries the originating request URL, so the two `toHaveAttribute(/\.png…/)` / `/\.svg…/` assertions can never match. The spec hasn't been touched since 117c512, i.e. it predates that change.

**Base-branch history** for this workflow:

| run | head | conclusion |
|---|---|---|
| 70463dec (#3298) | main | success |
| a7166dcf (#3302) | main | failure |
| **52fb54b (#3310)** | main — *this PR's base* | **failure** |

It went red between #3298 and #3302, which brackets a8cc181.

**Not fixing it here.** This PR touches Robolectric resource qualifiers, one Wear sample, and docs — nothing under `cli/serve` or `vscode-extension`. The fix is also a design call rather than a mechanical one: either the viewer stashes the request URL in a `data-cp-src` attribute for the spec to assert on, or the spec drops the `src` assertions and leans on `#cp-url-png` / `#cp-url-svg` (which already assert the knob is folded into the URLs) plus the `naturalWidth > 0` load check. Worth doing deliberately in its own PR rather than guessing inside an unrelated one.

Every other check on this PR is green or still running.

---

yschimke:
Addendum: `Android (Robolectric) serve lane honors overrides` just went red with the **same** failure — `serve-lanes.spec.mjs:186`, `#cp-img` src is `blob:…`. Both serve-lane jobs drive the same spec, so it's one root cause, not two.

Verified against this PR's base commit directly — run [30954555194](https://github.com/yschimke/compose-ai-tools/actions/runs/30954555194) on `main` @ 52fb54b, before this branch existed:

| job | conclusion on base |
|---|---|
| Android (Robolectric) serve lane honors overrides | **failure** |
| Serve render lanes honor overrides | **failure** |
| Bundle upload returns a browsable preview URL | success |
| Playground compiles + redeems a snippet end to end | success |

Same two jobs, same spec, already red without this change.

---

yschimke:
Preview diff is exactly the expected shape — worth recording why.

**Both changed functions are the same composable.** `WearDeviceMatrixPreview` and `WearRealDeviceMatrixPreview` both just call `DeviceSpecScreen`, which is where the `sw … dp` line was added. 8 variants = the 4 catalog devices + the 4 real-device `spec:` entries. Nothing else changed.

**324 functions / 524 variants unchanged** is the useful signal here. The PR body flags that this change also makes `values-sw…dp/` resource selection follow the previewed device, which it never did before — that's the one part of this with blast radius beyond the reported symptom. A full-corpus render moving *zero* other previews is direct evidence no existing sample was relying on the stale `sw320dp` bucket.

The real-device variants confirm the fix generalises past the catalog ids — Galaxy Watch 5 renders `211 × 211 dp` / `sw 211 dp` / `2.125× · 340 dpi`, all three agreeing where `sw` previously read 320.

No action needed.

---

PR #3318, opened 2026-08-05
fix(serve): make the playground repo gate visibility-aware

yschimke:
Follow-up to #3313, which merged before this refinement was ready — so it's a fresh branch off `main` rather than more commits on the merged one.

## Two problems with #3313

**1. It over-corrected.** #3313 required write access on every repo. That closed the real hole — GitHub reports `read` for any authenticated user on a **public** repository, so a read-level gate on a public `--github-auth-repo` admitted every GitHub account to a surface that compiles and runs submitted Kotlin.

But `read` means the opposite thing on a **private** repo: somebody deliberately granted this person access to a repository nobody else can see. Requiring write there locked out read-only collaborators in the one case that was never broken.

**2. It documented an escape hatch that doesn't exist.** Caught by Codex on #3313, after it had merged. #3313's docs and kdoc said `SERVE_GITHUB_AUTH_USERS` was the way to admit named logins without write access. It isn't — `allowedUsers` is checked in `GitHubOAuthVerifier.verify` to gate *sign-in* and never touches `repositoryAccess`, so `rejectMissingGithubRepoAccess` still 403s such a login. The claim was false as written.

## The change

Split the verdict on repository visibility:

| repo | accepted |
| --- | --- |
| public | `admin` / `maintain` / `write` |
| private | anything other than `none` |
| visibility unreadable | `admin` / `maintain` / `write` (strict branch) |

The unknown case going strict is deliberate: a token that can't read the repo metadata tells us nothing that should *widen* a gate on code execution. Costs one extra API call on a sign-in path that already makes two.

For the second problem, the false sentence is gone from both the docs and the kdoc, replaced with an explicit statement that the allowlist only ever **restricts** — naming a login does not grant playground access. That's the "remove the promise" half of Codex's suggestion rather than the "fold the allowlist into the verdict" half, because the latter would change `SERVE_GITHUB_AUTH_USERS` from a restriction into a grant, and an operator using it purely to narrow sign-in would not expect it to hand out playground rights.

## Test plan

```
./gradlew :cli:test --tests "…GitHubOAuthVerifierTest" \
                    --tests "…ServeAuthTest" \
                    --tests "…PlaygroundRoutingTest" :cli:ktfmtCheck
```
BUILD SUCCESSFUL — 12 tests in `GitHubOAuthVerifierTest`, 0 failures.

Five new cases on top of #3313's seven, all driving the real `verify()` path through an interceptor that now serves repo visibility as well as the permission payload:

- `read` on a private repo → access (the #3313 regression)
- `triage` on a private repo → access
- `none` on a private repo → still no access
- `write` on a private repo → access
- unreadable repo → `read` denied, `write` allowed (strict fallback)

The seven public-repo cases from #3313 are unchanged and still pass, so the original hole stays closed.

Auth verdict plumbing and docs — no visual surface, so no before/after render to embed.

## Unrelated: `main` is currently red

While checking CI on the earlier batch I found **Serve render lanes honor overrides** and **Android (Robolectric) serve lane honors overrides** failing on `main` itself — run [30976700883](https://github.com/yschimke/compose-ai-tools/actions/runs/30976700883) @ `45e2981`. `serve-lanes.spec.mjs:186` expects `#cp-img` to carry the override PNG URL but gets `blob:…`, so the viewer's move to blob-URL swapping (#3274 / #3301 look like the candidates) outran the harness assertion. Not from this PR; flagged in #3316 too.

---
_Generated by [Claude Code](https://claude.ai/code/session_0184U4iKHdJDcdTcC1Tm1UJA)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeGithubAuth.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeGithubAuth.kt
@@ -294,19 +294,23 @@ class GitHubOAuthVerifier(private val client: OkHttpClient = OkHttpClient()) {
   }
 
   /**
-   * Whether [login] has **write** access to [repository] — the gate on the playground, which
-   * compiles and runs a stranger's Kotlin.
+   * Whether [login] has access to [repository] that means something — the gate on the playground,
+   * which compiles and runs a stranger's Kotlin.
    *
-   * Write, not merely "not none". GitHub reports `read` from this endpoint for *any* authenticated
-   * user on a **public** repository, because reading is what public means — so the old `permission
-   * != "none"` test admitted every GitHub account whenever the configured `--github-auth-repo` was
-   * public, which is the documented setup. The legacy `permission` field collapses the finer roles
-   * onto `admin` / `write` / `read` / `none`, so `maintain` arrives as `write`; it is accepted by
-   * name too in case that ever changes.
+   * What "means something" is depends on the repository's visibility, and #3313 got this half
+   * right. On a **public** repo, `read` is what GitHub reports for *every* authenticated user,
+   * because reading is what public means — so a read-level gate there admits the whole of GitHub,
+   * which is the hole #3313 closed. On a **private** repo, `read` is the opposite: somebody
+   * deliberately granted this person access to a repository nobody else can see. Requiring write
+   * everywhere, as #3313 did, locked out read-only collaborators in the one case that was never
+   * broken.
    *
-   * This is deliberately narrower than before: a read-only collaborator on a private repo no longer
-   * reaches the playground. Live preview, which only needs a signed-in user, is unaffected — and
-   * `--github-auth-users` remains the way to admit named logins.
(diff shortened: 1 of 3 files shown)


yschimke:
CI red on **Serve render lanes honor overrides** — pre-existing on the base branch, not from this diff.

Same assertion, same shape as before (`1 failed / 5 passed`):

```
serve-lanes.spec.mjs:186
Received string:  "blob:http://127.0.0.1:8725/6d70cf6c-…"
```

This PR changes `ServeGithubAuth` (the repo-access verdict), its test, and docs. The serve-lanes harness boots serve with no GitHub auth configured, so `githubAuth` is null and none of this code executes on that path. It's failing on `main` too — run [30976700883](https://github.com/yschimke/compose-ai-tools/actions/runs/30976700883) @ `45e2981`.

Rather than just report it: I've opened #3321, which fixes it. The viewer's blob-URL swap left the logical `/render` URL nowhere in the DOM, so the harness had nothing to assert on; #3321 records it as `data-cp-src`. Verified by booting the real daemon-backed serve and running the spec — 6 passed.

Once #3321 lands I'll merge `main` in here so this PR's CI re-runs against a green base.

---

Issue #3215, opened 2026-08-03
Tracking: the playground is not enabled on preview.coo.ee — what it would take


yschimke:
## Current state (measured 2026-08-03, against the live box)

The playground is **off**. `preview.coo.ee` is running `0.19.20`, `public: true`, 18 catalogs loaded, 0 degraded.

| Endpoint | Result |
|---|---|
| `GET /playground` | **503** — the styled "Playground unavailable" page |
| `POST /api/1/compiler/run` | 404 — route never registered |
| `GET /pg/<token>` | 404 — route never registered |
| `GET /auth/github` | 404 — **GitHub auth is not configured on this box either** |

Root cause is the first branch of the lane builder: with neither bundle flag set, the lane is never constructed and only the disabled page is mounted.

```kotlin
// cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt:994-996
val cmpBundle = playgroundBundlePath
val androidBundle = playgroundAndroidBundlePath
if (cmpBundle == null && androidBundle == null) return null
```

```kotlin
// cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt:440-444
} else {
  // Reserve the well-known page path even when the compile lane is disabled.
  get("/playground") { handlePlaygroundDisabledPage() }
}
```

`SERVE_PLAYGROUND_BUNDLE` / `SERVE_PLAYGROUND_ANDROID_BUNDLE` are unset in that deployment's `.env`.

**The feature itself is built.** Phases 1–4 of [`docs/design/PLAYGROUND.md`](docs/design/PLAYGROUND.md) all shipped — the compile shim, the preview-token store, `/pg/` redemption, the per-session sandbox, the startup containment probe, and the jailed compiler. This is a deployment gap, not a missing implementation.

## What stands between here and enabled

Four things, in rough dependency order:

### 1. Something to compile against — #3212
`--playground-bundle` takes a local path only, and nothing on the box fetches one. Enabling means hand-placing a `.bundle` on the config volume, which then goes stale as catalog auto-refresh moves the real catalog underneath it. Accepting a served system id (`SERVE_PLAYGROUND_BUNDLE=compose-m3`) removes both problems.

### 2. Admission — #3210 and #3211, pick one
[...]

Issue #3214, opened 2026-08-03
security(serve): the playground lanes have no rate limiting, only concurrency slots


yschimke:
## Observation

There is no rate limiting anywhere in the serve host — a search across `cli/src/main/kotlin/ee/schimke/composeai/cli/serve/` finds no per-IP, per-user, or per-window limiter. The playground's only bounds are concurrency and capacity:

| Bound | Value | Where |
|---|---|---|
| Concurrent compiles | 2 (`--playground-compile-slots`) | `PlaygroundJailedCompiler.kt:227` |
| Wait for a slot before "busy" | 30 s | `PlaygroundJailedCompiler.kt:230` |
| Compile timeout | 180 s | `PlaygroundJailedCompiler.kt:219` |
| Request body | 256 KB | `ServeHttpServer.kt:3082` |
| Live tokens | 64, 600 s TTL | `PlaygroundTokenStore.kt:190-192` |
| Live seats | 8 on `preview.coo.ee` | `--live-seats` |

Those bound *simultaneous resource use*. They do not bound *request rate*, and they are not a per-caller budget: two clients issuing back-to-back 180-second compiles hold both slots indefinitely and every other user gets "the playground is busy compiling". Each mint also allocates a work dir under the playground work root, and each redemption pins a jailed daemon JVM for its token's TTL.

## Why it matters now

The exposure depends entirely on what gates the lane, and today nothing does on a public box:

```kotlin
// cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt:2977-2990
private suspend fun RoutingContext.rejectMissingGithubAuth(api: Boolean = false): Boolean {
  val auth = githubAuth ?: return false   // ← no auth configured ⇒ no check
  …
}
```

`preview.coo.ee` currently has GitHub auth unconfigured (`GET /auth/github` → 404). If the playground were enabled there as-is, `/playground`, `POST /api/{v}/compiler/run` and `/pg/<token>` would all serve **anonymously** — a free, unmetered compile-and-execute faucet on an 8 GB box that is also serving 18 catalogs.

## Proposal

1. **Treat GitHub auth as a hard prerequisite for enabling the playground on a public box.** Refuse the lane at startup, with an actionable message, when `--public` is set and no auth is configured — rather than serving it open. (This complements, and does not replace, the containment gate.)
[...]

Issue #3213, opened 2026-08-03
bug(serve): playground Android / Remote Compose modes likely cannot resolve Robolectric's `android-all` inside a bwrap jail


yschimke:
## Suspected bug — needs reproducing before fixing

`bwrapCommand` clears the environment and remaps `HOME` to the snippet's ephemeral work dir:

```kotlin
// cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundSandbox.kt:186-193
add("--clearenv")
add("--setenv"); add("HOME"); add(paths.workDir.absolutePath)
```

On Linux the JVM derives `user.home` from `HOME`, so inside the jail `user.home` is the snippet work dir — not `/root`. Robolectric resolves its `android-all-instrumented` artifact through the local Maven repository under `user.home`, and the prebuilt image bakes that artifact at `/root/.m2/repository/org/robolectric` (`deploy/image/Dockerfile:214-216`).

So a jailed Android snippet JVM should look in `<workDir>/.m2`, find nothing, and have no egress to fetch the ~150–200 MB artifact (`--unshare-net` is the whole point). Expected symptom: the Robolectric sandbox boot fails, `PlaygroundAndroidRenderService` returns no image, and the ANDROID / REMOTE_COMPOSE modes fail with no obvious cause. CMP (Skiko desktop) is unaffected — it has no equivalent runtime download.

`--playground-sandbox-ro /root/.m2` binds the *bytes* into the jail but does not fix the *lookup path*, since `HOME` moved. `docs/design/PLAYGROUND.md` §6.1 mentions prewarming "the Robolectric `android-all` cache" via that knob, which suggests the path problem was not noticed.

## Reproduce

On a host with `bubblewrap` installed:

```bash
compose-preview serve \
  --playground-android-bundle <a wear-m3 bundle> \
  --playground-sandbox bwrap \
  --playground-sandbox-ro "$HOME/.m2" \
  --port 8723
# POST an Android-mode snippet to /api/1/compiler/run and check for a first-frame image
```

Compare against `--playground-sandbox none`, which should work.

## Likely fix

Push an explicit repository location into the jailed daemon rather than relying on `HOME`, alongside the read-only bind — e.g. `-Drobolectric.dependency.dir=<bound path>` (or `maven.repo.local`) in `AndroidBundleLaunch.robolectricSystemProperties()` / the playground daemon opener, set only when a sandbox is active so the unjailed path stays byte-identical.

[...]

Issue #3212, opened 2026-08-03
feat(serve): `--playground-bundle` should accept a served catalog system or a URL, not just a local path


yschimke:
## Problem

`--playground-bundle` / `--playground-android-bundle` resolve a **local filesystem path** only:

```kotlin
// cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt:1203-1214
private fun resolvePlaygroundClasspath(bundlePath: String, workRoot: File, system: String) =
  PlaygroundCatalogClasspath.resolve(
    bundleFile = java.io.File(bundlePath),
    …
  )
```

`deploy/image/entrypoint.sh:203-206` just forwards the env var; nothing fetches anything. So enabling the playground on a box that already serves `compose-m3` means an operator manually downloads that catalog's `liveBundle`, drops it at `/config/playground-cmp.bundle`, and keeps it there.

Two things fall out of that:

1. **It is a manual step that duplicates work the server already does.** `--catalogs` fetched that exact bundle at startup, verified it as `Trusted(Branch)`, and resolved its classpath to run the live lane. The playground then needs a second, hand-placed copy.
2. **The copy goes stale silently.** Catalog auto-refresh (`SERVE_CATALOG_REFRESH`, default 600s) re-fetches the delivery branch and re-points the live lane at the new bundle. The playground's pinned file never moves, so snippets keep compiling against whatever ABI was current the day someone scp'd it. Nothing reports the drift.

## Proposal

Accept three forms, disambiguated the way `--catalogs` entries already are:

| Form | Means |
|---|---|
| `--playground-bundle /config/x.bundle` | today's behaviour, unchanged |
| `--playground-bundle compose-m3` | the classpath of an **already-served catalog**, reusing its fetched + trust-verified liveBundle |
| `--playground-bundle https://raw.githubusercontent.com/…/bundle/x.bundle` | fetch, reusing the `--bundle` fetch + trust path |

The system-id form is the interesting one: it makes the natural deployment `SERVE_PLAYGROUND_BUNDLE=compose-m3` / `SERVE_PLAYGROUND_ANDROID_BUNDLE=wear-m3` with no file to place, and it inherits the trust verdict rather than trusting whatever bytes landed on the config volume.

## Notes

- The seam is already shaped for this: `PlaygroundCompileService.catalogClasspath` is a `(PlaygroundMode) -> Classpath?` function (`PLAYGROUND.md` §8.4), so it can be re-resolved rather than captured once.
[...]

PR #3320, opened 2026-08-05
feat(serve): make the playground deployable on preview.coo.ee

yschimke:
The two changes from #3215's "suggested smallest first step" that live in this repo: the admission gate (#3210) and the bundle source (#3212). Together with GitHub OAuth secrets on the box, this is enough for a working collaborator-only CMP playground on `preview.coo.ee` **without solving container containment at all**.

## 1. Admission: a repo-access-gated lane no longer needs the containment gate (#3210)

`PlaygroundPublicGate.decide` keyed off `--public` alone, so on a public-catalog box the playground was treated as "open to strangers" and refused unless the operator could satisfy the full Phase-4 containment requirement — which the shipped container cannot (#3211: no `bubblewrap`, no systemd for `strict`).

But all three playground surfaces already run `rejectMissingGithubRepoAccess`. On a box with GitHub auth configured the snippet is a *repo collaborator's*, which is the same trust level as the token-gated posture that `decide` admits with no sandbox at all. So `decide` takes `repoAccessGated` as a second, independent basis:

| `--public` | repo-gated | decision |
|---|---|---|
| no | — | Allow — token-gated (unchanged) |
| yes | yes | **Allow — collaborators only; sandbox applied as defence in depth** |
| yes | no | the probe + caps chain, unchanged |
| yes | no, no sandbox | Refuse, now naming **both** remedies (#3214 item 1) |

A configured jail that fails its preflight under the repo-gated posture is warned about rather than fatal — admission never rested on it there. Every `Allow` says which posture admitted it, and the contained-anonymous line says `ANONYMOUS` in as many words, so the two are never confused in a startup log.

`buildGithubAuth()` moves above `openPlaygroundService` so the gate can see it.

## 2. Bundle source: `--playground-bundle compose-m3` (#3212)

The flags resolved a local path only, so enabling the playground on a box already serving that catalog meant hand-placing a copy of its liveBundle on `/config` — duplicating work `--catalogs` does at startup, and going **silently stale** as auto-refresh moved the real catalog underneath it.

Both flags now also take a served system id. The forms are told apart structurally (path separator / existing file / `.bundle`,`.png`,`.zip`,`.jar` suffix ⇒ path; bare token ⇒ system id), and an unknown id is a startup error listing what *is* configured.

Resolution moves to first use, which the system form forces — catalogs are fetched *after* the server is up, so resolving while the lane is wired would find nothing and disable the mode permanently. `PlaygroundClasspathSupplier` memoizes the first success and keeps retrying until then; hence `availableModes` is now computed per read.

Deliberately out of scope (as #3212 allows): following auto-refresh. A resolved classpath's jars are open in live snippet JVMs, so it stays pinned for the process's lifetime — restart to pick up a newer ABI. Asserted by a test rather than left to chance.

## 3. Docs (#3211's minimum fix)

`deploy/image/README.md` recommended `SERVE_PLAYGROUND_SANDBOX=bwrap`, which the containment posture explicitly refuses, on an image that ships neither `bubblewrap` nor `systemd` — following the deploy docs produced a silent refusal and the same 503 page as no configuration at all. It now documents the two postures, states plainly that the contained one is unreachable in this container, and links #3211. `docs/public-preview-server.md`'s "one deliberate exception" callout and `PLAYGROUND.md` §6.2/§8.4 updated to match.

## Test plan

- [x] `./gradlew :cli:test` — full CLI suite green
- [x] `./gradlew ktfmtCheckAll`
- [x] New: `PlaygroundBundleSourceTest` (form parsing) + `PlaygroundClasspathSupplierTest` (deferred resolve, memoization, pinning, retry-not-latch)
- [x] `PlaygroundPublicGateTest` extended: repo-gated with/without sandbox, a failed probe under the repo-gated posture, both-remedies refusal, and the `ANONYMOUS` contained line. Every pre-existing refusal case still passes unchanged.

### Verified against a real running server (not just unit tests)

`serve --catalogs compose-m3 --trust-store trust/producers.json --allow-render-trusted --playground-bundle compose-m3` — **no bundle file placed anywhere**:

```
serve: catalog compose-m3 → 84 preview(s), trust=branch:yschimke/compose-ai-tools@design-artifacts/compose-m3 (live bundle)
serve: playground cmp classpath resolved from served catalog 'compose-m3'
```

`POST /api/1/compiler/run` with a Material3 `@Preview` snippet → compiled in ~7s, `previewId: SnippetKt.Hello`, `previewUrl: /pg/pg_wDR5Z9Q_6wneTFGaGbRJxg`, no diagnostics. Before the catalog finished loading the same request correctly returned `mode CMP is not available on this host` and logged why, then recovered on its own.

The three admission postures, each a real startup:

```
A) --public, no auth, no sandbox
   serve: --playground-bundle … under --public need EITHER GitHub repo-access gating
   (--github-auth-client-id / …) … OR a per-session sandbox: --playground-sandbox …

B) --public + GitHub auth, no sandbox
   serve: playground admitted — public host, repo-access-gated (GitHub sign-in with access to
   --github-auth-repo); no sandbox — a collaborator's snippet runs unconfined on this host …
   serve: playground enabled (POST /api/1/compiler/run) — cmp✓ cmp-render✓

C) --playground-bundle nope-not-a-system
   serve: playground cmp bundle 'nope-not-a-system' is neither a readable file nor a catalog this
   server is configured to serve (configured: compose-m3). That mode is disabled — …
```

And the gating on posture B is real, not just a log line — anonymous callers against all three surfaces:

```
GET  /playground              -> 302 /auth/github/start?return=%2Fplayground
POST /api/1/compiler/run      -> 401 GitHub sign-in required.
GET  /pg/<token>              -> 302 /auth/github/start?return=%2Fpg%2F…
```

No visual evidence embedded: this changes admission gating and classpath resolution only — no `@Preview` output, webview, or overlay is touched, and the playground page markup is unchanged.

## Not in this PR

- **#3211** proper — whether the container can ever provide containment. This makes it non-blocking, not solved.
- **#3213** — Robolectric `android-all` inside a jail. Launching CMP-only sidesteps it; the deploy README now warns off the Android modes on a sandboxed host.
- **#3214 item 2** — per-caller rate limiting. Item 1 (never serve anonymous *and* uncontained) is covered above; the compile lane still has only concurrency slots, which the repo-access gate bounds but does not replace.

### One correction to #3215's measurements

`GET /auth/github → 404` is **not** evidence that GitHub auth is unconfigured — on a box with auth fully configured that path is still 404, because the live route is `/auth/github/start`. The issue's conclusion holds on other evidence; worth knowing before anyone re-probes the box with that URL.

Refs #3210, #3211, #3212, #3214, #3215

---
_Generated by [Claude Code](https://claude.ai/code/session_01RhvrrdcDXZjHeHEpqpgNtr)_


yschimke:
## CI status

Two red checks, both in `serve-lanes-e2e.yml`, and both **failing on `main` already** — not caused by this PR:

- `Serve render lanes honor overrides`
- `Android (Robolectric) serve lane honors overrides`

Evidence: the latest completed `main` run of that workflow ([run 30976700883](https://github.com/yschimke/compose-ai-tools/actions/runs/30976700883), `45e2981`, ~an hour before this branch existed) reports `failed_jobs: 2` — **the same two job names, and no others**. The workflow has been red on `main` since #3302; the two runs in between (`ae29d34`, `21e495c`) were cancelled rather than passing, so this is the most recent signal available.

The failing assertion is in the viewer, which this PR does not touch:

```
preview-harness/serve-lanes.spec.mjs:186
  Received string:  "blob:http://127.0.0.1:8725/74f05823-…"
```

`#cp-img` now carries a `blob:` URL instead of a `.png?knob.label=…` one. That looks like fallout from #3307 (*clarify themed preview loading*), which reworked how `ServeWeb` hands images to the viewer — the spec still asserts the pre-blob contract. Worth its own issue against the harness or #3307; I've left it alone here rather than fold an unrelated viewer fix into an admission-gate change.

Everything this PR does touch is green: **ktfmt**, **Module Unit Tests** (the new `PlaygroundBundleSourceTest` / `PlaygroundClasspathSupplierTest` and the extended `PlaygroundPublicGateTest`), **Build CLI**, **CodeQL**, and **VS Code Extension Tests**. `Playground compiles + redeems a snippet end to end` — the job that most directly exercises this change — was still running at the time of writing; I'll follow up if it goes red.

## One follow-up pushed since the PR opened

`83e0704` — #3313 (*require write access for the playground repo gate*, already in this branch's base) narrowed `hasRepositoryAccess` from "any access" to `admin`/`maintain`/`write`. That is what the repo-access admission posture actually rests on, so the gate's KDoc, its startup log line, `PLAYGROUND.md` §6.2 and the deploy README now say **write** access rather than just "access" — the trust argument in this PR is "someone who can already push to the repo whose CI builds this image", and now the docs match the check that is really made. Docs and one log string only; no behaviour change.

---

PR #3322, opened 2026-08-05
fix(serve): reap idle pooled daemons, budget them, and stop retrying dead previews

yschimke:
Diagnosed from the live box after `preview.coo.ee/pocketcasts/` reported **"Render server: connected · 10 instances"** with zero active streams and a grid of "Theme preview unavailable" cards.

The lease accounting turned out to be correct — `ThemeRenderLeaseManager` bounds concurrent *renders* (tiers `[5,3]` ≤ `maxConcurrentRenders`), never processes. Three separate things were wrong.

## What was broken

**1. Pooled daemons were never reaped.** `ServePerPreviewDaemonPool` only LRU-evicts when *over* `maxOpen`, and `ServeSharedDaemonPool` only closes replicas when the whole host closes. Nothing else reaps them either — a catalog session is registered `pinned`, and `suspendIdle()` skips pinned entries by design, so the one mechanism that sheds daemons excludes exactly the sessions that hold them. Live evidence: `{"instances":10,"pooled":9,"poolCapacity":12,"activeStreams":0}` after 7h uptime. The count was a high-water mark, not a gauge.

**2. Pooled daemons were free.** `LiveSeatLimiter` is acquired at exactly one call site — the WebSocket stream lane — so `/status.json` read `liveSeatsAvailable: 8` of `8` while 10 catalogs held daemons and pocketcasts alone held 10 JVMs. The budget was blind to nearly every process on the box.

**3. A permanently broken preview was advertised as retryable.** `buttons-time-play-button` fails to render live — `NotFoundException: File res/drawable/ic_play.xml … #0x7f080346`, visible through the SVG lane. It therefore never enters the theme cache (pocketcasts' theme optimization: `paused`, **375 of 1267 failed**), so every themed thumbnail went live, hit the shared daemon whose render lock the background optimizer held, backed off at `DAEMON_BUSY_WAIT_MS` (~0.55s measured) → `Busy` → `503 render busy; retry shortly`. The page retried three times and gave up, forever, and each attempt contended the lock that pushed *other* cards into their own back-off. Cards that were already cached returned in ~0.5s, which is why the grid looked patchy rather than uniformly broken.

## What changed

- **Reaping.** `reapIdle` on both pools behind a new `ServeHost.releaseIdleDaemons(idleMillis)`, driven by the existing session reaper. Processes go; the pinned host stays listed and instantly resumable. The monolithic browse daemon is kept — it's what the cold-start design exists to hold on to.
- **Budgeting.** The limiter is built once in `ServeCommand` and shared with `ServeHttpServer` and both pools, which hold permits for as long as a daemon is open, weighted by the catalog's backend. A refused per-preview daemon falls back to baked pixels; a refused replica narrows the burst onto hosts already in circulation rather than failing a render.
- **Terminal failures.** `CatalogThemeCache` latches a key after a run of failures (a successful render clears it). The render lane answers from the latch without touching the daemon, and the HTTP lane returns **409** — which the grid retires on the spot instead of running the retry ladder.
- **Pruner inversion** (the root cause of #3260/#3299 recurring): `AndroidResourcePruner` now takes a *positive* drop-set of resources attributed to a third-party AAR, instead of a retain-set. "We couldn't classify this" and "this is a droppable AAR file" were the same input, and every gap biased toward deleting a resource in use. A name contributed by both an AAR and a project module is an override and stays. Cost of being wrong is now a slightly larger bundle, not a dead render. Keeps @yschimke's schema-recognition guard from #3314.

## Visual evidence

The theme-error card's wording is the user-facing claim, so both states are captured. Left: a retryable failure, unchanged. Right: the new terminal state, reached without the retry ladder.

| Before (what a permanently broken card said) | After (terminal 409) |
| --- | --- |
| retryable wording`` | terminal wording`` |

A `theme-render-terminal` state is registered in `pages-snapshot.spec.mjs`, next to the existing `theme-error` one, so every future change to either message is diffed automatically. That the capture settles without waiting out any backoff is itself the behaviour under test.

## Test plan

- `./gradlew :cli:test :gradle-plugin:test` — green.
- `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot --grep declared-themes` — 2 passed (both themes).
- New tests: idle-reaping and seat-budget behaviour for both daemon pools (including that reaped daemons hand their seats back), failure-latch and un-latch-on-success for `CatalogThemeCache`, terminal-409 assertions in `ServeWebDeferredThemeTest`, and pruner tests rewritten for the inverted set — notably that an empty ownership result now drops nothing.

**Not fixed here:** the deployed pocketcasts catalog still needs a rebuild. It was built with `0.19.30`, which predates the #3299 ownership fix — that is why its `ic_play.xml` is missing. This PR stops that class of bug recurring and stops the server pretending it's retryable, but the existing bundle carries the damage until it is regenerated.

---


yschimke:
Both review findings were correct — fixed in eb87f23.

**Latching retryable optimizer misses.** Confirmed: the optimizer calls `markFailed` whenever it exhausts its attempts, and it does that for a retryable `Busy` (another request or a warm holding the daemon) exactly as it does for a render that threw. Harmless while `failedKeys` only fed the `/status` `failed` count — a real regression now that it feeds `renderFailureLatch()`. `failureReason` now requires both halves: the key is latched **and** a real `RenderOutcome.Failed` supplied a reason. The metric is unchanged, so a contended key still shows as `failed` in `/status` without being reported as terminal to a visitor.

Worth noting the test I'd written asserted the *broken* behaviour (`an optimizer-marked failure reports a reason even when none was captured`) — it's rewritten to pin the correct one.

**Stranded seat on a failed launch.** Confirmed. The ticket is acquired before `openReplica()`, so a throw during temp-dir setup, `openHost`, or the daemon process left permits held by nothing, shrinking the box's budget permanently, one failed launch at a time. Both pools now return the ticket on a throwing open — `ServePerPreviewDaemonPool` had the same shape and is fixed alongside, though only the shared pool was flagged.

Both new tests were checked against the unfixed code and fail there, so neither is vacuous:

```
CatalogThemeCacheTest > an optimizer miss with no captured failure stays retryable() FAILED
ServeSharedDaemonPoolTest > a replica whose launch throws hands its seat back() FAILED
```

`:cli:test` and `:gradle-plugin:test` are green.

---

PR #3328, opened 2026-08-05
chore(samples): add a no-op comment to one preview per sample

yschimke:
**Throwaway PR — do not merge.** This is an instrument, not a change.

## What this is

A control experiment for the preview-diff pipeline. Every sample that has a Kotlin `@Preview` gets exactly one edit: a single `//` comment line inserted above one preview declaration.

Comments never reach bytecode. The compiled classes, the discovered `previews.json`, and every rendered PNG must therefore come back **byte-identical to `main`**. The diff bot should report zero Changed variants.

So the null hypothesis is sharp: **any diff reported on this PR is churn or a flake**, by construction. There is no third explanation.

## Why it triggers a render at all

`.github/preview-scope.json` scopes PR renders to touched modules under `samples/**`. Touching one file in each sample puts every sample module (plus its Gradle dependents) in scope, so this exercises the full sample render surface rather than a corner of it.

## Coverage

20 samples edited. Three are deliberately untouched because they have no Kotlin `@Preview` of their own:

| Sample | Why skipped |
| --- | --- |
| `cmp-android-only` | Deliberately preview-free fixture (mirrors a consumer that applies the plugin to a module with no previews) |
| `cmp-wasm-catalog` | No Kotlin previews — wasm catalog tier |
| `design-catalog-m3-shared` | Shared component library; its previews live downstream in `design-catalog-m3` / `-m3-android`, both of which *are* edited |

## Visual evidence

Intentionally none, and that is the assertion: this PR changes no pixels. The expected evidence is the diff bot reporting **no changed renders**. If it embeds before/after images for anything, that image pair *is* the finding — and it is a bug in the pipeline, not in this diff.

## Test plan

- [x] `./gradlew ktfmtCheckAll` — passes
- [ ] Compose Preview bot reports 0 Changed / 0 New / 0 Removed
- [ ] Any reported diff triaged as churn (nondeterministic render input) or flake (infra/timing)

---
_Generated by [Claude Code](https://claude.ai/code/session_01PNr9JoaGgxCapbFA4oJrjv)_

### DIFF
--- a/samples/android-alpha/src/main/kotlin/com/example/samplealpha/FocusRingPreviews.kt
+++ b/samples/android-alpha/src/main/kotlin/com/example/samplealpha/FocusRingPreviews.kt
@@ -40,8 +40,8 @@ private fun WithRippleConfig(config: RippleThemeConfiguration, content: @Composa
   CompositionLocalProvider(LocalRippleThemeConfiguration provides config, content = content)
 }
 
-// churn probe: comment-only edit, no bytecode or pixel change expected.
-@Preview(name = "Inset Focus Ring — fan-out", widthDp = 480, heightDp = 96, showBackground = true)
+// churn probe phase 2: deliberate visible tweak - this preview MUST render as changed.
+@Preview(name = "Inset Focus Ring — fan-out", widthDp = 480, heightDp = 128, showBackground = true)
 @FocusedPreview(indices = [0, 1, 2, 3])
 @Composable
 fun InsetFocusRingFanOutPreview() {
--- a/samples/android-alpha/src/main/kotlin/com/example/samplealpha/FocusRingPreviews.kt
+++ b/samples/android-alpha/src/main/kotlin/com/example/samplealpha/FocusRingPreviews.kt
@@ -40,6 +40,7 @@ private fun WithRippleConfig(config: RippleThemeConfiguration, content: @Composa
   CompositionLocalProvider(LocalRippleThemeConfiguration provides config, content = content)
 }
 
+// churn probe: comment-only edit, no bytecode or pixel change expected.
 @Preview(name = "Inset Focus Ring — fan-out", widthDp = 480, heightDp = 96, showBackground = true)
 @FocusedPreview(indices = [0, 1, 2, 3])
 @Composable
--- a/samples/android-daemon-bench/src/main/kotlin/com/example/daemonbench/BenchPreviews.kt
+++ b/samples/android-daemon-bench/src/main/kotlin/com/example/daemonbench/BenchPreviews.kt
@@ -32,12 +32,12 @@ import androidx.compose.ui.unit.dp
 // If you grow this set, update the per-render baseline numbers in
 // docs/daemon/baseline-latency.csv at the same time.
 
-// churn probe: comment-only edit, no bytecode or pixel change expected.
+// churn probe phase 2: deliberate visible tweak - this preview MUST render as changed.
 @Preview(name = "RedSquare", showBackground = true)
 @Composable
 fun RedSquarePreview() {
   MaterialTheme {
-    Surface(color = Color(0xFFEF5350)) {
+    Surface(color = Color(0xFF7E57C2)) {
       Box(
         Modifier.size(96.dp).padding(16.dp).clip(RoundedCornerShape(12.dp)).background(Color.White)
       )
--- a/samples/android-daemon-bench/src/main/kotlin/com/example/daemonbench/BenchPreviews.kt
+++ b/samples/android-daemon-bench/src/main/kotlin/com/example/daemonbench/BenchPreviews.kt
@@ -32,6 +32,7 @@ import androidx.compose.ui.unit.dp
 // If you grow this set, update the per-render baseline numbers in
 // docs/daemon/baseline-latency.csv at the same time.
 
+// churn probe: comment-only edit, no bytecode or pixel change expected.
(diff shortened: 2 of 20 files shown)


yschimke:
## Preview-diff probe: results

Two-phase throwaway probe of the preview-diff pipeline. Phase 1 (`c784d7f`) added a single `//` comment above one `@Preview` per sample — comments never reach bytecode, so every render must come back byte-identical. Phase 2 (`cbaeb5b`) made each of those same previews visibly different, scoped so that no shared composable was touched.

### Phase 1 — no over-reporting ✅

Zero pixel diffs. Confirmed from the job log rather than from the bot's silence: `Generate compose PR comment body` ran (it did not hit its `nothing to compare, skipping` guard), and `Post or update compose PR comment` printed `No preview diff and no existing comment — skipping` — a branch reachable only when `compare-previews.py` emits `No visual changes detected.`. Renders are byte-reproducible; no churn, no flakes.

### Phase 2 — no collateral, but two blind spots ⚠️

18 of 20 probed previews reported as **Changed** (31 variants), with exactly the right derived variants (light/dark pairs, the `@FocusedPreview` fan-out, scrolling LONG+GIF, the animated GIF, the `@PreviewParameter` fan-out). **No collateral** — every sibling preview in the same modules stayed unchanged, including `WidgetContainerLargeRemote` and `ImageWidgetSquircleLargePreview`, which confirms the scoping discipline held.

The two that did **not** report:

**1. `SdkMatrixPreview` (`samples:sdk-matrix`) — reported "Unchanged" but never rendered.** This is the real finding. The commit changes its background from `0xFF1B1B1F` to `0xFF102A22`, filling the whole 280×160 frame. The run was `SCOPE_MODULES: full`, so nothing was out of scope, and it finished green (`WORST_RC: 0`, no report artifacts). But the head render tree has **18 module dirs to the baseline's 19** — `renders/samples:sdk-matrix/` is absent — and the comparator listed the preview under *Unchanged (313 functions)*.

A missing render is currently indistinguishable from a clean render, and the comment affirmatively states the preview was verified identical when it was never produced. A baseline entry with no corresponding head render should read as removed/missing, not unchanged. Root cause of the disappearance itself is not established — the module is unconditionally included in `settings.gradle.kts:267` and applies `ee.schimke.composeai.preview`, so it should have rendered.

**2. `samples:sdk21:android-metro-viewmodel` — outside the pipeline by design.** Its `CounterScreenContentPreview` change (`count = 42` → `99`) produced nothing, and the module appears in no render branch at all. Expected: `settings.gradle.kts:688` gates inclusion on `JavaVersion.current() >= 21`, and the preview workflow runs JDK 17. Correct by design, but worth noting the visual-diff bot is structurally blind there and says nothing about it — covered separately by `samples-sdk21.yml`.

### Unrelated pre-existing bug: a11y + notification comments on this PR are false positives

`24 added · 0 changed · 0 removed` on a comment-only diff. `compute-scope.py` emits Gradle paths (`samples:android`), but the a11y/notification manifests store `project.name` — the bare leaf `android` — via `moduleName.set(project.name)` in `ComposePreviewTasks.kt:1464` and `AndroidPreviewSupport.kt:444`. `filter-findings-scope.py` intersects the two spellings, always gets an empty set, and discards the entire baseline, so every render reads as new. Reproduced against the real baseline: **24 entries in, 0 out.**

`test_filter_findings_scope.py` stays green because its fixture uses `{"module": "samples:android"}` — a format the producer never writes. Compose's `baselines.json` does store Gradle paths, which is why the compose comparator is unaffected. Beyond the noise, this silently disables a11y regression detection on change-scoped runs: against an emptied baseline, a newly-broken finding is indistinguishable from the churn.

Throwaway — not for merge.

---

yschimke:
### Follow-up: the a11y/notification false positives are now confirmed by an unintentional A/B on this PR

The two runs of this PR differed in scope, which isolates the bug without touching any code:

| | phase 1 (`c784d7f`) | phase 2 (`cbaeb5b`) |
|---|---|---|
| `SCOPE_MODULES` | 19 explicit `samples:*` paths | `full` |
| scope filter | **runs** | **skipped** |
| Notification previews | `24 added · 0 changed · 0 removed` | `No notification preview changes` |
| A11y | 24 changed / 76 unchanged (100 previews) | 241 unchanged (307 previews) |

Neither commit touches a single notification preview, so the scope filter is the only variable. The guard is [`a11y.sh:167`](.github/actions/apply/pipelines/a11y.sh#L167) (and the mirror at [`notifications.sh:162`](.github/actions/apply/pipelines/notifications.sh#L162)):

```sh
if [ -n "${SCOPE_MODULES:-}" ] && [ "$SCOPE_MODULES" != full ]; then
  python3 "$ACTION_PATH/filter-findings-scope.py" ...
```

So the filter only ever runs on change-scoped PRs — exactly the runs where it empties the baseline because `compute-scope.py` emits `samples:android` while the manifests store `android`. Full-scope runs skip it and compare correctly.

Practical consequence: **the noise is invisible on any full-scope run**, which is why this has survived. It reproduces only on a change-scoped PR — i.e. most sample-only PRs, which are precisely the ones where a11y regression detection matters and is currently disabled.

Note the phase-2 a11y "changed" list is mostly first-time coverage (the `RemoteComposePlayer` `SpeakableTextPresentCheck` errors across ~25 remote previews, etc.), not regressions from this PR — those modules had never been compared before, since phase 1's scoped run only rendered `android`.

---

PR #3335, opened 2026-08-05
feat(remotecompose): serve google: font families to the view player

yschimke:
Third of the per-lane PRs behind the typeface specimens (#3329). This is the **`java`** lane — the AOSP `remote-player-view` player, and the *default* server-side snapshot lane.

It had no downloadable-font path at all. `DefaultTypefaceResolver` resolves a named family by scanning `/system/fonts/` for a filename containing the name, then falls back to `Typeface.create(name, style)`; there is no `FontsContractCompat` call anywhere in `remote-player-view` or `remote-player-core`, and the daemon's Robolectric sandbox has no `/system/fonts/` either. So `google:Orbitron` rendered in the platform default while the same document showed the real face under `cmp-android`, `cmp-jvm` and `js` — the finding the audit doc opened with.

`RcGoogleFontTypefaceResolver` is installed on the player at both view-player call sites (live capture and bundle IR replay) and serves a `google:` name from the shared machine-local Google Fonts cache — the same `(family, weight, italic) -> File` resolution the Robolectric downloadable-font shadow, the figma-svg embed path and the cmp-jvm lane use, so a family resolves to *the same file* in every lane. Everything else delegates to `DefaultTypefaceResolver`: generics, `/system/fonts/` names and inline `FontData` behave exactly as before, and nothing is installed at all when the render was given no font cache directory.

The instance also implements `applyVariationSettings`, rebuilding the face from the cached file with `Typeface.Builder(file).setFontVariationSettings(…)` — what the platform resolver does for a `/system/fonts/` face. Returning the base typeface there (the obvious no-op) is what makes a `wght` ramp draw every line at one weight.

## Visual evidence

The three catalog specimens through this lane, before and after:

java lane typeface before/after

## What this PR does *not* fix

The two axis sheets still draw flat here. They carry their axes as `CoreText` **style** properties (20/21), and the AOSP CoreText renderer doesn't route those into the paint's variation settings — so `applyVariationSettings` is never asked. That is upstream plumbing rather than a resolver gap; a canvas op carrying paint-bundle axes does get a real instance. Recorded in the audit doc rather than papered over.

## Test plan

- [x] `:data-remotecompose-connector:testDebugUnitTest` — new `RcGoogleFontTypefaceResolverTest` covers the two pure rules: which names opt into a fetch (`google:` only — `device:`, bare names and generics never do) and how axis tag/value arrays become the platform settings string, including that an axis missing either half is dropped rather than mis-paired with a neighbour's value.
- [x] `:samples:design-catalog-remote-m3:composePreviewRenderAll` before and after — the images above are those two runs.
- [x] `./gradlew ktfmtCheckAll`.

Docs: `docs/design/RC_PLAYER_TYPEFACES.md` matrix + Java section updated, and finding 1 marked closed.

Remaining lanes: `cmp-android` and `cmp-jvm` (axes), `js` (inline `FontData`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01HtpQQWqQvxXcucxvvurfkd)_


yschimke:
Pushed the Codex review point: a `google:` family the source can't serve is now remembered as a miss, so a text-heavy document makes one fetch attempt per key instead of one per text op. Good catch — `resolve` runs per paint, and only successes were being cached.

On the red **VS Code Extension E2E (cmp-smoke)** check at the previous SHA: not this change. That job runs `:samples:cmp:composePreviewRenderAll` — a CMP *desktop* sample with no Remote Compose in it — and the harness cancelled the render after its 5-minute budget with 0 PNGs written, i.e. a slow/cold-cache run rather than a failure of anything this PR touches. The push above re-runs it.

---

PR #3336, opened 2026-08-05
feat(rc-player-jvm): apply a document's font-variation axes

yschimke:
Fourth of the per-lane PRs behind the typeface specimens (#3329) — the **cmp-jvm** lane's axis half (its typeface half was #3327).

The jvm player read a `CoreText` op's family and ignored the axis arrays beside it, so a document asking for `wght 1000` drew the family's default instance. The layout seam now turns those arrays into a Compose `FontVariation.Settings` and the resolver instances the face at them:

- axis **tags arrive as text ids**, so each is resolved through the document before it is paired with its value;
- an axis counts only when *both* halves are present — pairing a tag with a neighbour's value would draw a real face at silently the wrong instance, which is worse than dropping it;
- families are cached **per axis set**, not per family: a variable file serves many instances and they are different faces, so a cache keyed on `(family, weight, italic)` alone hands a `wght 100` line the `wght 1000` family the previous line built;
- a `wght` axis additionally decides **which file to fetch**, because Google Fonts serves a named family as a static instance per weight and the document's own `fontWeight` stays at its default while the axis carries the intent.

## What this does *not* fix, and why

The catalog's variable specimens still draw flat in this lane, and the reason is upstream of the player. `:data-fonts-google` asks the CSS API with a legacy user-agent to get a `.ttf` at all (Skia and Android can't parse the modern woff2 response), and for a variable-*only* family like Roboto Flex that response is a **static default instance with no `fvar` table** — verified by parsing the downloaded file: `has fvar: False`, and the wght@100 / wght@1000 downloads are byte-identical. There are no axes left in the file to vary.

So this PR is the plumbing being correct, not the specimen turning colourful: a host that supplies a real variable file (the browser lane's manifest does) gets true instancing through this same code, and fetching the variable file from Google means decompressing woff2 — a `:data-fonts-google` change, called out in the audit doc as the next step. I'd rather land the correct plumbing with the blocker named than leave the axes silently dropped.

## Test plan

- [x] `:third-party-rc-embedded-player-jvm:test` — new `RcJvmFontAxisTest`: two `wdth` instances of one file are two different families (and the source is asked once, not per instance); a repeated instance is served from the cache; an empty axis set is the same request as none; a `wght` axis selects which instance is fetched (`GoogleFontKey("Orbitron", 700, …)` from a document whose own weight is 400). Hermetic — the `GoogleFontSource` is faked over the repo's vendored Roboto Flex.
- [x] Full `:third-party-rc-embedded-player-jvm:test` suite and `./gradlew ktfmtCheckAll`.
- [x] End-to-end through `RcJvmRenderHarness` on the three catalog specimens: the four `wght` instances are now fetched as four separate cache entries (`roboto-flex-100/400/700/1000.ttf`) where before only the 400 file was requested — which is how the static-instance limit above was found.

Docs: `docs/design/RC_PLAYER_TYPEFACES.md` matrix + CMP JVM section.

Remaining: `cmp-android` (verify its axis path against the specimens) and `js` (inline `FontData`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01HtpQQWqQvxXcucxvvurfkd)_

### DIFF
--- a/docs/design/RC_PLAYER_TYPEFACES.md
+++ b/docs/design/RC_PLAYER_TYPEFACES.md
@@ -23,6 +23,7 @@ question:
 | **Java** (AOSP `remote-player-view`, server-side) | ✅ framework typefaces | ⚠️ `/system/fonts/` filename scan | ❌ no provider call at all | ✅ `Font.Builder(ByteBuffer)` |
 | **CMP Android** (vendored embedded player, server-side) | ✅ framework typefaces | ⚠️ `Typeface.create(name)` | ✅ `FontsContractCompat` | ❌ ignored |
 | **CMP JVM** (embedded player over Skiko, server-side) | ✅ | ⚠️ host families, else nearest standard | ✅ downloaded via `GoogleFontTypefaceResolver` | ❌ ignored |
+| **CMP JVM** — font-variation axes | ✅ layout ops, applied to the resolved face | — | ⚠️ `wght` picks the instance; Google serves static TTFs | ❌ canvas ops |
 
 Two rows of that table are worth stating as findings, because they make two chips in the *same*
 viewer disagree about the *same* document:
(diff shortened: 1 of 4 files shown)


yschimke:
On the axis-tag encoding: both readings are right, for different paths, so I've made the seam accept either.

The paint bundle's `setTextAxis` op does carry raw four-byte tags — that's the `FONT_AXIS_WEIGHT = 0x77676874` constant the baseline player decodes. But a `CoreText` **style** interns its axis names like any other string, so there the int is a text id. Dumping the catalog's own specimen document (`VariableWeightRemote`, captured through `RemoteText(fontVariationSettings = …)`):

```
[9]  RcTextData(id=44, text=wght)
[10] RcCoreText(textId=42, properties=[…, IntValue(id=8, value=43),
       IntArrayValue(id=20, values=[44]),
       FloatArrayValue(id=21, values=[100.0])])
```

`44`, not `0x77676874` — and the vendored Android player reads the same field the same way (`RcPlayerTextLayout.kt`: `context.getText(id)` → `FontVariation.Setting(name, …)`). So the text-id lookup is correct for this path and the axes weren't being dropped.

Since a writer emitting the raw tag here is plausible, the seam now tries the text table first and unpacks the bytes when there's no entry, with a test pinning both (`0x77676874` → `wght`; a small id like `44` unpacks to control characters and is rejected rather than guessed at). Pushed in f03010d.

---

PR #3347, opened 2026-08-05
fix(serve): let the theme optimizer keep its turn while the server is quiet

yschimke:
Part 2, after #3345 made a cache miss recoverable. With that in, the background pass is a genuine prefetch — this makes it actually keep up.

## The bug

```kotlin
while (cache.get(job.cacheKey) == null && attempts < 3) {
  if (!awaitServerIdle()) return@execute      // ← the FULL 60s window, every render
```

`awaitServerIdle` waits for `themeOptimizationIdleMillis` (60s) of quiet — and it's called **per render attempt**, not once per pass. `serverIdleMillis` is process-wide, so a request to *any* of 21 catalogs resets it. The pass could only advance during a full minute of total silence, which on a public server is close to never.

Measured on preview.coo.ee: **one entry per ~105s** against a sub-second render — roughly 99% waiting. For pocketcasts' 1267 entries that's ~37 hours.

## The fix

Entry is unchanged: the pass still earns the whole quiet window before starting work on a box someone is using. What changes is that it **keeps** that turn while the server stays quiet, and yields as soon as a request actually arrives (`OPTIMIZER_YIELD_MILLIS`, 1.5s).

Both properties still hold:
- a visitor never waits behind more than the render already in flight (the yield is checked between jobs, and the per-render background permit is unchanged);
- an idle box fills the cache at render speed rather than one entry a minute.

## What I found was already right

I'd planned to reorder the job list preview-major so one daemon warm amortises across all themes for a preview. It already is:

```kotlin
catalogIds.flatMap { previewId -> declaredThemes.map { theme -> … } }
```

All 7 themes for a preview render consecutively, so the warm is already shared. No change needed — worth recording so nobody re-proposes it.

## Test plan

- `:cli:test` — green.
- New test models a server that is *quiet but not silent-for-a-minute* (10s idle, 60s entry window): the old gate caches one entry then blocks, the new one caches both themes. **Verified to fail against the unfixed gate.**

## Deliberately not in this PR

Leased bursts still take seats via `acquireBackground`, so they're bounded by the stream reserve rather than competing as foreground work. That's arguably wrong — a leased burst is a user waiting, not background residency — but changing it is exactly the shape of the change that broke the `serve-lanes` E2E on a `--live-seats 1` box earlier today. I'd want to re-run that E2E locally against the change before proposing it, rather than ship it on reasoning alone.

Also unchanged: the single process-wide background render permit. Worth revisiting (contention that matters is per-daemon, and different catalogs' daemons don't block each other), but it's a separate tuning question with its own blast radius.

No visual surfaces changed.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogLiveHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogLiveHost.kt
@@ -183,6 +183,9 @@ class ServeCatalogLiveHost(
   // Whether the optimizer currently holds its turn: set once the full quiet window is met, cleared
   // the moment a request arrives. Without it the pass re-earned the whole window per render.
   private val optimizerHasTurn = AtomicBoolean(false)
+  // When the optimizer last checked for activity. Any activity newer than this happened while it
+  // was rendering, and must cost it the turn even if the server looks quiet again by now.
+  private val optimizerSampledAt = java.util.concurrent.atomic.AtomicLong(0)
   private val warmDaemonIds = ConcurrentHashMap.newKeySet<String>()
   private val warmingInFlight = ConcurrentHashMap.newKeySet<String>()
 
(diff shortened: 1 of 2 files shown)


yschimke:
Codex is right — fixed in 039fe69.

Sampling the instantaneous idle answers the wrong question. A render can outlast `OPTIMIZER_YIELD_MILLIS` several times over, so a request that arrives *during* one has come and gone by the time the pass looks again, and the server reads as quiet. That visitor never cost the optimizer its turn, which is exactly the starvation the gate exists to prevent.

`serverIdleMillis` is the **age** of the last activity, so `now - idle` is when that activity happened. Comparing that against the previous sample answers "did anything happen since we last looked" rather than "is it quiet right now":

```kotlin
val now = clock()
val lastActivityAt = serverIdleMillis()?.let { now - it }
val quiet = lastActivityAt != null && lastActivityAt <= optimizerSampledAt.get()
optimizerSampledAt.set(now)
```

`:cli:test` is green, including the duty-cycle test from the original commit (a genuinely quiet server still keeps its turn — the fix narrows *when* it yields, it doesn't reinstate per-render re-earning).

**One gap I want to flag rather than paper over:** I don't have a regression test for this specific scenario. Reproducing it needs a clock and an idle-clock scripted to advance together across the render boundary, and the honest options were a fiddly scripted-pair fixture or adding a test-only seam to production code. I didn't think a test-only seam was a good trade, and I ran out of room to get the scripted version right. The change is small and the reasoning is in the commit, but it is currently argued rather than pinned — worth knowing if someone touches this gate later.

---

PR #3355, opened 2026-08-05
fix(serve): charge leased burst replicas as foreground seats

yschimke:
Last of the four items from the browsing-performance design. #3345 made a cache miss recoverable, #3347 made the prefetcher keep up — this restores the burst width for the first visitor who picks a theme.

## The bug

A leased burst is a visitor sitting in front of the grid waiting for pixels. Its replicas were charged via `acquireBackground`, so they could only use what the theme optimizer had left behind:

```
budget 8, STREAM_RESERVE 2  →  background ceiling 6, shared across every catalog's pools
```

That caps a visitor's burst at the prefetcher's leftovers, which is backwards — the whole point of the lease tier is that *someone is waiting*.

## The fix

Shared-pool replicas acquire as **foreground** (`LiveSeatLimiter.acquire`), the same class as a stream.

**Scoped to the shared pool only.** The per-preview pool keeps `acquireBackground`, so the stream reserve still does its job: a pooled *render* daemon must never take the last seat from a live *stream*. A shared replica is a different animal — it opens only when leased renders actually overlap, and it's short-lived (burst ends → replica idles → `reapIdle` returns the seat).

## Test plan

I said I wouldn't propose this without running the E2E, because a seat change of exactly this shape broke `serve-lanes` on a `--live-seats 1` box earlier today and I shipped that one on reasoning alone.

Ran against a real daemon-backed `serve --catalogs compose-m3 --allow-render-trusted --live-seats 1` — the same configuration that broke:

```
✓ PNG lane re-renders on knob override (7.5s)
✓ SVG lane bakes the override text into the vector (320ms)
✓ Live WebSocket lane upgrades and pushes a frame (1.3s)
✓ viewer wires the knob into every render lane (7.0s)
✓ Live mode surfaces a visible error when the stream can't activate (461ms)
✓ Wasm iframe re-renders on knob override (3.0s)
6 passed (22.4s)
```

Plus:
- `:cli:test` — green.
- New unit test: with exactly `STREAM_RESERVE` permits free, a background acquire is refused (asserted as a precondition) while an overlapping leased render still widens onto a replica. **Verified to fail with `acquireBackground`.**

## Where this leaves the four requirements

| | Status |
| --- | --- |
| 1. Default theme prebaked | already held; grid opens on baked pixels, never touches the daemon |
| 2. Background fills fast when idle | #3347 (and the job list was already preview-major) |
| 3. First browser gets a burst | **this PR** |
| 4. Single high-priority render on an individual page | #3345 |

Still open, deliberately: the single process-wide background render permit in `ServeBackgroundWork`. Raising it is defensible — the contention that matters is per-daemon, and different catalogs' daemons don't block each other — but it's a separate tuning question and I'd rather it not ride along with a seat change.

No visual surfaces changed.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/LiveSeatLimiter.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/LiveSeatLimiter.kt
@@ -44,7 +44,7 @@ class LiveSeatLimiter(val totalPermits: Int) {
    * coerced down to [totalPermits], so a backend heavier than the whole budget can still run alone
    * rather than being permanently refused.
    */
-  fun acquire(weight: Int, verified: Boolean = true): Ticket? {
+  fun acquire(weight: Int, verified: Boolean = true, countRefusal: Boolean = true): Ticket? {
     val sem = semaphore ?: return Ticket(0)
     if (weight <= 0) return Ticket(0)
     val permits = weight.coerceIn(1, totalPermits)
@@ -54,7 +54,13 @@ class LiveSeatLimiter(val totalPermits: Int) {
     // dropped: on a public box they are mostly noise anyone could generate, but on a `--revisions`
     // box a valid revision is *legitimately* unknown until its first lease builds it, and its
     // refusals are real demand. Two counters keep both readings honest.
-    if (verified) refusals.incrementAndGet() else unverifiedRefusals.incrementAndGet()
+    // Only count callers that actually turn someone away. A caller that degrades instead — the
+    // shared replica pool narrows its burst onto a host already in circulation and still serves the
+    // render — must not inflate this, or the one number that answers "is the budget too small here"
+    // starts reporting throttled batches as refused visitors.
+    if (countRefusal) {
+      if (verified) refusals.incrementAndGet() else unverifiedRefusals.incrementAndGet()
+    }
     return null
   }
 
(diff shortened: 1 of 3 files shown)


yschimke:
Codex is right — fixed in 2c14cf5.

Switching to a foreground `acquire` made a full budget increment `liveSeatRefusals`, but this pool doesn't refuse anything on a miss: it waits for a host already in circulation and serves the render. So any leased burst wider than the remaining budget would have reported visitors as turned away when only replica *widening* was throttled.

That matters more than a cosmetic counter. `refusalCount`'s own kdoc calls it the number that answers *"is the seat budget actually too small here?"* — the evidence any change to the budget should rest on. Inflating it with throttled batches would argue for a bigger budget on a box that never refused anyone, which is exactly the wrong conclusion.

`acquire` now takes `countRefusal` (default `true`, so every existing caller is unchanged) and the pool passes `false`:

```kotlin
// A miss here does NOT refuse the render — the burst narrows onto a host already in circulation.
ticket = liveSeats.acquire(seatWeight(), countRefusal = false)
```

New test: with the budget spent elsewhere, an overlapping leased render still succeeds off the primary and both refusal counters stay at zero. Verified to fail without the flag.

`:cli:test` green on a forced re-run — I'd had an ambiguous result where one invocation reported `BUILD FAILED` while the test task then showed `UP-TO-DATE`, so I re-ran with `--rerun` rather than trust the cache.

---

PR #3363, opened 2026-08-05
perf(serve): prefetch themes in per-preview batches through the replica pool

yschimke:
The prefetcher rendered one entry at a time through the monolithic daemon — while sitting next to a five-wide replica pool it was already wired to. That pool is the lane a visitor gets when they pick a theme; the background pass just never used it.

Measured: **~30s per cache entry** against a sub-second render. pocketcasts' 1267 entries projected to ~10 hours.

## Two changes

**Batch through the leased lane.** Jobs render via `renderLeased`, so the pass borrows the same replicas a visitor's burst does. Batched **by preview** — the unit the daemon warms for, and the unit the job list was already ordered by — so every theme of one preview renders together, one warm is amortised across all of them, and two previews' daemon opens never interleave. Width is the catalog's own burst capacity, which the seat budget narrows on a small box rather than spawning JVMs it can't run.

1267 entries at five wide is **minutes, not hours**.

**Resume on a short window.** The 60s window answers a cold-start question — "may I start work on a box someone might be using?" — worth asking once. After a yield the box has already proved it goes quiet, and all that remains is whether the visitor stopped. Charging a full entry window per interruption is exactly what produced 30s/entry at a ~50% yield rate. Resume waits 2s (above the 1.5s yield threshold, so it can't re-detect the activity it just waited out and livelock). `awaitQuiet` also polls at a fraction of its window rather than a flat second.

## The foreground guarantee is unchanged in substance

The background permit now wraps a **batch** rather than a render — but a batch is bounded by one render, so a visitor arriving mid-batch still waits at most a render. That's what the old per-render permit was expressing. `Busy` stays "ask again" rather than a failure, so a warm still settling doesn't spend the catalog's `failed` count.

## How the design got here

I first tried a fixed-width batch. Two tests failed:

```
idle optimizer renders all themes for a preview before opening the next daemon   FAILED
idle optimizer waits for asynchronous cold warming without spending its retry budget   FAILED
```

Both were correct objections, not stale assertions — a fixed width straddles preview boundaries (breaking warm amortisation) and my first cut had dropped the cold-warm await entirely, deferring those entries to a "next pass" that may never come. Batching on the preview boundary fixes both by construction, and awaiting the warm once per preview is cleaner than the old per-job retry budget. The tests drove the design rather than being adjusted to fit it.

## Test plan

- `:cli:test` — green, including both tests above.
- Not yet measured on the live server: the effect to look for is `themeOptimization.cached` climbing in the tens per minute rather than ~2 per 100s. Worth confirming after release before assuming the target is hit — my last throughput prediction in this area was wrong by 3.5×, in my favour, because I read a cumulative average instead of a steady-state rate.

## Deliberately not included

Raising the background render permit above 1. Three tests encode "one background render at a time" as an invariant; with batching that ceiling now applies per *batch* rather than per *render*, which is the part that mattered. Weakening the invariant itself is a separate decision and I reverted an earlier attempt at it rather than silently rewriting those tests.

No visual surfaces changed.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogLiveHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogLiveHost.kt
@@ -392,6 +392,11 @@ class ServeCatalogLiveHost(
         val byPreview =
           jobs.filter { catalogThemeCache.get(it.cacheKey) == null }.groupBy { it.previewId }
         for ((previewId, previewJobs) in byPreview) {
+          // Gate BEFORE the warm, not just before the renders. `daemonWarmOrScheduling` starts a
+          // cold daemon, which is the single most expensive thing this pass can do to a box that is
+          // still loading catalogs or serving traffic — exactly what the idle gate exists to
+          // prevent. Warming ahead of it let every catalog host kick off a cold start at prewarm.
+          if (!awaitOptimizerTurn()) return@execute
           val previewDaemonId = alias[previewId]
           // Await a cold warm ONCE per preview rather than letting each theme rediscover it. The
           // old per-job loop spent retry budget on this; here it is a precondition of the batch.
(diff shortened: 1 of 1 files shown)


yschimke:
All three findings were correct — fixed in c725854.

**P1, warm ahead of the idle gate.** This one is the serious one and it's a regression I introduced in this PR. The per-preview warm sat *above* `awaitOptimizerTurn()`, so `prewarm()` could start a cold daemon on a server still loading catalogs or serving traffic — the most expensive thing this pass can do, and exactly what the gate exists to prevent. The turn is now taken before the warm, not just before the renders.

**P2, single-daemon batches.** Also correct, and I'd missed the configuration entirely. Under `sharedDaemonRenders=false` every theme of one preview resolves to the *same* per-preview daemon, so a five-wide batch contends on one render lock and all but one returns `Busy` — which I leave unmarked, so those entries would never fill in that configuration. `themeRenderBurstCapacity` is 5 there because different *previews* can run in parallel, which is not what this batch is. Width is now the shared pool's capacity when it actually backs the batch, and 1 otherwise:

```kotlin
if (sharedDaemonRenders && sharedDaemonPool != null)
  sharedDaemonPool.capacity.coerceIn(1, MAX_OPTIMIZER_BATCH)
else 1
```

**P2, latching one transient failure.** Correct, and it interacts badly with my own earlier change: `failureReason` treats a latched key as terminal and answers foreground requests with 409, so one flaky background render would have made that thumbnail permanently unavailable until the catalog generation refreshed. Now uses `recordRenderFailure`, which latches only after a run — that's the retry budget the old per-job loop provided, expressed through the mechanism built for exactly this case.

`:cli:test` green.

Worth noting for anyone reading later: two of these three are cases where the batch rewrite silently changed behaviour in a configuration the tests don't cover (`sharedDaemonRenders=false`), and the third re-broke something an earlier PR in this series had fixed. The batch design is right, but it moved more than it looked like it moved.

---

Issue #3280, opened 2026-08-04
serve: remaining preview.coo.ee performance work (seat accounting, render cache, pocketcasts, badge failure state)


yschimke:
Tracking issue for what's left after the round that landed in #3240, #3243, #3254, #3258, #3264, #3270, #3272, #3274, #3277.

## Done, for context

| change | measured |
|---|---|
| Grid thumbnail prebaking (#3240) | 11.3 MB → 4.4 MB across 7 catalogs |
| Presence keepalive (#3243) | — |
| Viewport-deferred theme renders (#3254) | — |
| Warm on arrival (#3258) | superseded — gated the warm, not the spawn |
| Lazy daemon open (#3270, #3272) | `daemons.running` **17 → 1** at idle |
| Extra-module live lane (#3264) | meshcore-mobile 38 → 44 live components |
| Fetch-then-swap + render-server badge (#3274, #3277) | — |

---

## 1. Daemon residency is not charged to the live-seat budget

**`liveSeatsAvailable` reads 8/8 even with daemons resident.** A prewarm never acquires a seat, so `LiveSeatLimiter` cannot see the residency it nominally bounds. Confirmed repeatedly today, including at `known 18, running 17`.

Two consequences:

- The seat budget (derived in `deploy/image/entrypoint.sh` as `(RAM − 1 GB) / 1.2 GB`, clamped `[2,8]`) protects nothing in practice. Its whole purpose is OOM protection on an Android-heavy box where each daemon is weighted 2.
- `ServeCatalogLiveHost.keepLiveWarm`'s gate — `liveSeats.availablePermits() > 0` — always passes, so presence keepalive cannot be throttled by it either.

Unblocked by the lazy open: there is now a single admission point (`ServeRenderHost.sessionDelegate`) where a seat could be taken and released on `close()`. It was not worth building while every catalog spawned at registration regardless.

Care needed on the release path — `close()` now returns early for a host that never opened, and a seat must not leak when the mid-handshake race in #3272 reaps a session the initializer opened after `close()`.

## 2. The render lane still caches for 5 minutes

`STATIC_RESOURCE_CACHE_CONTROL = "public, max-age=300, stale-while-revalidate=3600"` (`ServeHttpServer.kt`).

The original ask was "cache renders almost indefinitely, until a catalog redeploy". That was answered for grid thumbnails — content-hashed and `immutable` via `ServeHeroImages.gridThumbFor` — but not for the general `/render/<id>.png` lane.

[...]

PR #3373, opened 2026-08-05
feat(serve): report prefetch rate, time split and observed batch width on /status

yschimke:
Answers the question `cached`/`remaining` can't: is the prefetcher keeping up, and if not, is it render-bound or gate-bound?

## Why

Reading progress off those two numbers went wrong **three times** against the live server during this round of work:

| misreading | error |
|---|---|
| timed the SVG lane (render **plus** figma-svg export) and treated it as PNG render cost | ~10× |
| divided by server uptime instead of the pass's active time | 3.5× |
| sampled a nine-minute-old restart and called a working change ineffective | inverted the conclusion |

Every one ran toward the story being told at the time. That's what instrumentation is for.

## What's added to `ThemeOptimizationSnapshot`

- **`entriesPerMinute` / `etaSeconds`** — a *rate*, over the pass's **active** time (render + gate wait) rather than wall-clock. Wall-clock includes stretches where the pass holds no turn at all, which drags the figure toward zero and hides whether it keeps up while running — that was misreading #2.
- **`renderMillis` / `waitingMillis`** — where the wall-clock actually goes. This is the split that separates "nothing to fix without a faster renderer" from "scheduling problem". No existing field distinguishes them, which is why the ~97%-waiting diagnosis took a stack of manual probes to reach.
- **`turnsGranted` / `turnsYielded`** — how often the idle gate grants a turn and how often traffic takes it back.
- **`lastBatchWidth` / `maxBatchWidth`** — the width *actually issued*, not the requested ceiling. A batch collapsing to 1 (single-daemon lane, or a seat budget with no replicas to spare) is otherwise invisible — and that exact collapse was a real bug Codex caught on #3363, found by reading code rather than by any signal the server emitted.

All fields are default-valued, so the JSON stays backward compatible.

## Test plan

- `:cli:test --rerun` — green. `--rerun` deliberately: the plain task resolved `UP-TO-DATE` and skipped execution during this work, exactly as #3280's verification notes warn.
- New tests pin the arithmetic — 5 entries over 60s of active time reports `5.0`/min with a `60`s ETA and a 50/50 render/wait split — and that rate and ETA stay null before there is anything to divide by.

## What this doesn't do

It doesn't render the numbers on the `/status` **HTML** page, only `/status.json`. The JSON is what I'd been probing by hand, so it's the surface that stops the misreadings; the HTML view is a follow-up if you want it visible without a curl.

Related: #3280 item 5 (15 catalogs still pinned below 0.19.35, carrying the pruner bug — `horologist` is failing live now).

No visual surfaces changed.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/CatalogThemeCache.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/CatalogThemeCache.kt
@@ -17,6 +17,34 @@ data class ThemeOptimizationSnapshot(
   val fullyOptimized: Boolean,
   val startedAtEpochMillis: Long? = null,
   val completedAtEpochMillis: Long? = null,
+  /**
+   * Entries cached per minute over the pass's lifetime, and the projected seconds to finish at that
+   * rate. Null before the pass has done enough to divide by.
+   *
+   * The point of publishing a RATE rather than only `cached`/`remaining`: a cumulative count read
+   * twice looks the same whether the pass is keeping up or crawling, and reading progress off a
+   * lifetime average (which includes the server's startup, when the pass is parked) understates the
+   * current rate. Both mistakes were made against this catalog before this existed.
+   */
+  val entriesPerMinute: Double? = null,
+  val etaSeconds: Long? = null,
+  /**
+   * Where the pass's wall-clock actually goes: inside renders vs. waiting for its turn at the idle
+   * gate. This is the split that says whether throughput is render-bound (nothing to fix without a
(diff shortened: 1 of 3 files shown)


yschimke:
All four review findings were correct, and this had already auto-merged by the time they were applied — so they land in #3376 rather than as commits stacked on a merged branch.

Fixed there: rate numerator switched to optimizer-produced entries (foreground renders via `cacheCatalogRender` were inflating it), render timing moved inside `withRenderPermit` with permit queue time charged to `recordWaiting`, the resume path counted as a turn grant, and cold-daemon warm time charged to `renderMillis`.

---

PR #3430, opened 2026-08-07
feat(serve): show a render-history timeline in the viewer

yschimke:
## Summary

The payoff for the pipeline built in #3394 → #3403 → #3415 → #3417 → #3424. `history.json` was being published beside the renders and **nothing read it**. This adds the strip that does, so a preview finally shows how it got to look the way it does.

## What it looks like

Render-history timeline in the viewer

Rendered from the real assets (`serve.css` + `viewer-history.js`) against a real published render and a manifest shaped like the live one. The strip shows the version count over publish count, the `unstable` badge, and one chip per version — date, the source commit it was rendered from, and `×3` where several publishes carried the same bytes.

## How old renders are viewable at all

`raw.githubusercontent.com` serves **any commit**, not just branch names. So a version's `commit` plus the preview's `path` addresses the historical bytes directly — no server round-trip, nothing unpacked. Verified against real published renders before building anything on it:

```
renders/samples:android-alpha/FocusOverlayPreview_Focus_Overlay_FOCUS_0.png
  df4aa9c0 -> http=200  12828 bytes  md5 39d2f859…
  8b9f6f2b -> http=200  12433 bytes  md5 5bc2682c…
```

Two versions of the same preview, genuinely different bytes.

## Decisions worth flagging

**Source commit, not delivery commit.** Chips show `sourceSha` — the commit the render was produced from — because the delivery-branch commit is just a publish marker and means nothing to a human.

**Instability is surfaced, not hidden.** When a render keeps reverting to bytes it already had, the list is trimmed and would otherwise not add up to the publish count sitting next to it. The badge explains the discrepancy rather than leaving it looking like a bug.

**Absent beats broken.** The strip appears only with delivery provenance. An uploaded bundle or local project has no branch to read from, so the attributes are omitted entirely. Half-configured is treated as absent too — a timeline you can't click through to an old render is worse than no timeline.

**Same validation twice.** URL construction refuses a ref where a sha is required, and rejects paths outside the renders tree — in Kotlin and again in the JS that mirrors it, so a malformed manifest can't steer fetches at an attacker-chosen branch.

## Wired for future coverage

Registered as its own page fixture (`serve-viewer-history.html`), which the harness auto-discovers from `fixtures/pages/*.html`. Every future change to this surface gets diffed without anyone remembering to screenshot it — per the repo's rule about not leaving new visual surfaces uncaptured.

The other page fixtures move because the `serve.css` content hash changed; that's cache-busting following the new CSS, not unrelated churn. The `_render-placeholder.png` churn was reverted.

## Test plan

- **9 new tests**: URL construction (delivery-branch manifest URL, commit-addressed render URL, segment encoding) and its rejections (ref-instead-of-sha, traversal, paths outside `renders/`); attribute emission; the both-or-neither rule; omission without provenance.
- **CLI suite: 1593 tests, 0 failures**; `ktfmtCheckAll` clean; fixture golden regenerates stably (re-ran to confirm).

## Remaining

`serve`'s local-git override for project mode — the one piece left, and now the smallest.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UXp51dArWBD4dWia6GbJkV)_


yschimke:
All three review findings are correct, and they share a root cause I got wrong: the timeline mutates the live `` without coordinating with `viewer.js`'s own lifecycle. **This PR should not merge as-is** — I've pushed the CI fix below but not fixes for these.

**Exports lie after selecting an old version (P1, line 125).** The most serious of the three. `select()` changes `cp-img` only, so the panel labelled "The current view as a URL", Copy PNG, the download link, and the issue-report image all still point at the current `/render` URL. Someone looking at a 2026-04 render would copy, or file a bug against, today's pixels. A viewer that misreports what's on screen is worse than one with no timeline.

**Race with the initial render (P1, line 80).** If `history.json` resolves before the first `/render` completes, `img` has no src, `safeSrc` returns null, and the newest version is skipped. A two-version history then renders one chip and is dropped entirely; a longer one silently promotes an *old* version into the `buttons[0]` "current" slot — mislabeled, not just missing. Note the guard I added for the CodeQL alert widens this window slightly, since a non-conforming src now also yields null.

**Override re-renders desync the strip (P2, line 89).** On a provenanced catalog that can re-render, changing a control replaces `img.src` and revokes the previous blob URL. The strip keeps its stale pressed state, so the stage can show the new current render while an old chip reads as selected — and clicking "current" later can load a revoked blob.

Two candidate fixes, and I lean to the second:

1. Keep the swap-in-place design and make it whole: defer building until the snapshot resolves (or derive the current URL from the manifest rather than the DOM), retarget every export affordance on selection, and reset on snapshot refresh. Correct, but it couples the timeline to a fair amount of `viewer.js` internals.
2. Stop mutating the stage — have a chip open that version's render directly. That removes findings 2 and 3 by construction (no export mismatch, no blob interaction, no pressed-state desync) and reduces 1 to cosmetic labelling. Less slick, much less to keep in sync, and the historical bytes are a plain URL anyway.

I don't want to pick that unilaterally at the end of a long session, so I'm leaving it for a fresh pass rather than guessing.

Separately: `b4f1ebb9` fixes a CI failure I introduced — editing `viewer-history.js` changed the hashed asset URL every page embeds, staling the goldens.

---

yschimke:
Correcting a claim in the PR description: **"the harness captures the strip on every future change" is wrong.** I checked the capture the diff bot actually produced rather than trusting that registering the fixture was enough.

[`serve-viewer-history.light.png`](https://raw.githubusercontent.com/yschimke/compose-ai-tools/e5619d665a5d6cdc182556f7704123addc7c398a/renders/serve-viewer-history.light.png) shows the viewer page **with no timeline in it**. `viewer-history.js` fetches the manifest from `raw.githubusercontent.com`, which the harness has no route to, so the strip never renders. The fixture therefore captures a page that is byte-identical whether the timeline works, regresses, or is deleted — zero coverage, not automatic coverage. That is precisely the gap the repo's rule about wiring new visual surfaces exists to close, and I asserted it was closed without checking.

(The capture also renders unstyled, which is worth a look on its own — I have not worked out yet whether that is normal for this fixture path or a second problem.)

Fixing it means giving the fixture a manifest it can read without network — an inline `<script type="application/json">` payload the script prefers when present, falling back to the fetch. That is a small change to `viewer-history.js` plus the fixture, and it has the side benefit of making the timeline testable in the harness at all.

Adding it to the list for the same follow-up pass as the three review findings. The PR already should not merge; this is another reason.

Worth noting the pattern, since it has now happened five times in this feature: every defect here — the missing `formatVersion`, the deleted manifest, the bootstrap deadlock, the whole-tree clobber, and now this — was invisible in review and only appeared by running the thing and looking at the actual output. I caught this one only because I fetched the render instead of trusting the green checkmark.

---

yschimke:
The CodeQL alert survived my fix (now line 101). `safeSrc` rejects the dangerous schemes, but it returns the value unchanged, so the taint tracker still follows `img.getAttribute("src")` → `img.setAttribute("src", …)` and re-reports the same flow. The guard is worth keeping on its own merits — it does block `javascript:`, `data:` and protocol-relative — but it does not, and cannot really, satisfy the analysis while the DOM round-trip exists.

I'm not going to patch it a third time, because it is the same root cause as the three review findings above: **the timeline mutates the live ``.** The redesign I proposed — a chip opens that version's render directly instead of swapping the stage — deletes the `setAttribute("src", …)` sink outright, so this alert, the export mismatch, the blob-revocation desync, and the initial-render race all go away by construction rather than by four separate patches.

That is now four independent problems pointing at the same design decision, which is a strong enough signal that I should change the design rather than keep defending it.

So the follow-up pass on this PR is: swap-in-place → open-directly, plus the inline manifest so the fixture actually captures the strip. Everything below the UI — the extractor, the manifest, the URL layer and their tests — is unaffected and stays.

Still not for merge.

---

yschimke:
CI status, so the next pass does not have to re-derive it.

`Build CLI` and `Module Unit Tests` failed again on the rerun, so this is **not** the repo's usual Maven Central transient. Both come from the same underlying failure:

```
ServeWebFixtureTest > serve web fixtures are in sync with ServeWeb() FAILED
    org.opentest4j.AssertionFailedError at ServeWebFixtureTest.kt:3876
1595 tests completed, 1 failed
Execution failed for task ':cli:test'.
```

The interesting part is that the **same suite is green locally**: `./gradlew :cli:test` gives 1593 tests, 0 failures, on this exact commit, and I regenerated the goldens with `UPDATE_SERVE_WEB_FIXTURES=true` in `b4f1ebb9` before pushing.

**CI runs 1595 tests, my local run 1593.** So CI exercises two tests mine does not, and the fixture comparison disagrees only there. That points at an environment-dependent difference in what gets generated rather than a golden I simply forgot to refresh — regenerating a third time locally would very likely produce the same "green here, red there" result, so I have not done it. The next step is finding which two tests CI has that a plain `:cli:test` does not, and what they add to the fixture set.

Not urgent to fix in isolation: this PR is already blocked on the swap-in-place → open-directly redesign, which changes `viewer-history.js` and therefore the asset hash and every page golden again. Worth folding the fixture investigation into that same pass rather than chasing it separately now.

Recording rather than guessing, because guessing at CI behaviour from local state is exactly what cost three rounds on the publish bug.

---

yschimke:
Redesign pushed in `ee8772d3`. All four problems shared one cause — chip-click replacing `#cp-img`'s src in place — so entries are now **links** that open that version's render, which removes all four by construction rather than by four patches.

| Finding | How linking out resolves it |
|---|---|
| Exports pointed at the wrong render (P1) | The stage is never touched, so nothing can disagree with it |
| Initial-render race (P1) | Every entry, including the newest, is addressed from the manifest, not the DOM |
| Blob-revocation desync (P2) | No pressed state and no captured blob URL to go stale |
| CodeQL DOM-text→`src` (high) | The `setAttribute("src", …)` sink is gone — `grep -c` is now 0 |

**The false coverage claim is fixed too, and this is the part worth checking.** I had said registering the page fixture meant the harness captured the strip. It didn't: the capture the diff bot produced showed a viewer page with *no timeline in it*, because the fetch targets `raw.githubusercontent.com` and the harness has no route there — so the fixture was byte-identical whether the timeline worked or was deleted. The page can now carry an inline manifest, the fixture supplies one, and I rendered the committed fixture offline to confirm the strip actually appears: three chips, `current` marked, the `×3` multi-publish badge, and the `unstable` badge, all styled. That capture is now real coverage.

Inline payloads escape `</` so a manifest can't close the `<script>` element and spill markup into the page — tested in both directions (escaped, and not merely dropped).

CLI suite **1606 green**, `ktfmtCheckAll` clean.

Still outstanding and *not* addressed here: the `ServeWebFixtureTest` local/CI divergence (CI runs 1595 tests to my 1593). This commit regenerates every page golden again, so it's worth seeing what CI says now before digging — the divergence may simply have been a stale golden from the earlier asset-hash change, or it may persist, in which case the next step is finding which two tests CI runs that a plain `:cli:test` does not.

---

yschimke:
CodeQL still reports 1 high on `8e9439c5`. That is three attempts, so I am stopping rather than pushing a fourth guard, and recording where it actually stands.

What I have got wrong so far, in order:

1. `safeSrc` — blocked `javascript:`/`data:`/protocol-relative but returned its input unchanged, so the taint still tracked through it.
2. "Linking out removes the sink" — verified with `grep -c 'setAttribute("src"'`, which only looked for the *old* sink. `link.href = url` is a sink too; the flow moved rather than disappeared.
3. Constraining `repo` — a real guard, but evidently not the only tainted input.

The hypothesis I would test next, and the reason I think guard (3) could not have worked: **the inline manifest I added in the same PR introduced a second DOM-text source into the same sink.** `document.getElementById("cp-history-data").textContent` is DOM text; it is parsed as JSON; and `entry.path` from it is interpolated into `link.href`. `renderUrlAt` does check that path starts with `renders/` and has no `..`, but like `safeSrc` that check returns the value unchanged, which is exactly the shape CodeQL did not accept the first time.

So the fix is probably to *construct* the path rather than pass it through — rebuild it from validated segments, or drop `path` from the manifest entirely and derive it from `module` + `renderBasename` the way `PreviewHistoryManifest.renderPathsToPreviewIds` already does server-side, so nothing DOM-derived reaches the href verbatim.

I have not verified that against the actual alert detail (`/security/code-scanning/392` and its successor), and I am not going to assert it without doing so — the last two times I asserted a fix here I had checked something adjacent to the claim rather than the claim itself. Next pass: read the alert's reported source→sink path first, then fix what it actually names.

Everything else on this PR is in better shape: the three review findings are genuinely resolved by linking out (the stage is never touched), and the CLI suite is 1606 green. This alert and the unverified harness-capture question are what remain.

---

yschimke:
The inline-manifest fix **does not work**, and I now know why. Correcting the coverage claim a second time.

The newest capture is byte-identical to the one from before the fix — `md5 7098b24f2843966ca45bbdf3897f0ac5`, 37375 bytes, on three successive render commits (`e5619d66`, `c3287909`, `1b8a9da7`) spanning both the fix and the redesign.

Root cause: **`fixtures/pages/*.html` are captured with no assets served.** There is no assets handling in `_server.mjs` and no assets directory beside the pages, so the `/assets/serve/<hash>/serve.css` and `.../viewer-history.js` the page references both 404. The unstyled look I noticed earlier and set aside as "possibly normal" was this all along.

So for this fixture type the capture is **server-rendered HTML only — no CSS, no JS**. The inline manifest could never have helped: the script that reads it never loads. That also means:

- My original claim ("registered as a page fixture, so the harness captures the strip") was not merely incomplete, it is unachievable by that route.
- More generally, **no JS-driven viewer surface can be covered by a page fixture**, which is worth knowing beyond this PR.

Three ways to get real coverage, none of which I am doing blind at this point:

1. Extend the harness to serve the `assets/serve/…` route, so page fixtures exercise the real CSS and JS. Highest value — it would retroactively give every existing page fixture visual fidelity, and the current unstyled captures suggest they are all weaker than they look.
2. Inline the CSS and JS into this one fixture. Works, but diverges from the production HTML, so the capture stops being evidence about the real page.
3. Capture this surface through the serve e2e path, which runs a real server, instead of the page-fixture path.

I would pick (1), but it is a harness change with blast radius across 140 captures and should be its own PR with its own review, not a tail-end commit here.

Net: the strip is verified working (I rendered the committed fixture offline with the real assets and the timeline appears correctly), but it is **not** covered by CI, and I am no longer claiming otherwise in the PR description.

---

PR #3439, opened 2026-08-07
feat(design-artifacts): shard the catalog render across parallel jobs

yschimke:
## Summary

The design-artifacts render is one serial `bundle pack`, and it grows close to linearly with the sheet. Measured over four `m3-catalog` runs on `main`:

| Commit | Previews | Render step | Job total |
| --- | --- | --- | --- |
| `d354560` | 287 | 13.6 min | 17.2 min |
| `323e4a2` | 519 | 22.5 min | 24.0 min |
| `5edc70a` | 607 | 26.7 min | 29.6 min |
| `f35c2e5` | 689 | 27.2 min | 28.9 min |

→ `render_minutes ≈ 3.7 + 2.15s × previews`. At ~1500 previews that is ~58 min, past a 40-minute `render-timeout`; at ~2400 it passes the 90-minute job timeout, and **no value of `render-timeout` fixes that**. Only dividing the marginal 2.15 s does.

### `render-shards: N`

New generic input on `design-artifacts-reusable.yml`, **default `1`** — at the default the two new jobs never run and `generate` renders inline exactly as before, byte for byte.

Above 1, each shard runs the same `bundle pack` with `--exclude-preview-id` naming everything that is not its share. Excluded previews stay *listed* in the bundle without a baked PNG (the mechanism render-priority deferral already relies on), so every shard emits a structurally identical bundle — same `previews.json`, same manifest, same re-render classpath — differing only in which `previews/<id>.*` slots are filled.

```
        ┌── shard 1: discover → plan → bundle pack → upload ──┐
 matrix ┼── shard 2: …                                        ┼── merge ── generate ── publish
 [1..N] └── shard N: …                                        ┘
```

### `compose-preview bundle merge` — and why `bundle repack` could not be used

This was the load-bearing open question, and the answer is a **no**:

> `repackRethemedPreviews` swaps re-renders into slots the target **already has**, and only handles `previews/<id>.png` + `previews/<id>.figma.svg`. Against a shard base, every other shard's previews are reported unmatched and dropped — the base's render excluded them, so there is no slot — and the JSON sidecars are explicitly preserved verbatim rather than merged, so any preview that did land would arrive without its `.semantics.json`, which the design-catalog completeness gate fails.

So `bundle merge <base> <shard>… -o <out>` unions the per-preview artifacts: the raster, its `.semantics.json` / `.layout.json` / `.fonts.json` / `.figma.svg` / `.catalog.json` / `.overrides.json` sidecars, the nested `figma-raster/` crops, `ir/<id>.rc` and `extensions/<id>.json`. Base wins on collision, earlier shards over later ones, and it streams each shard's zip so the shared `classes/` + `libs/` payload is read past rather than held.

Manifests, cover and the re-render classpath are **inherited from the base shard, not merged** — every shard packs the same module at the same commit, so they are identical by construction. That is why `publish-live-bundle` + `split-per-preview` need no designated shard.

### Partitioning

Each shard derives its own partition from its own `compose-preview list --json`. A central plan job would have to compile before it could discover, adding its whole ~4.6 min to the critical path instead of ~0.9 min inside each shard's own compile (that compile is shared with the shard's render).

`scripts/design-artifacts/shard-preview-ids.mjs` owns three decisions:

- **by preview id, never by function name** — one function expands to a 30-cell matrix while its neighbour expands to two, and the slowest shard sets wall clock;
- **round-robin over the sorted id list, not contiguous blocks** — ids sort by group, so blocks cluster the template-heavy groups into one shard. Sorting also makes independent shards agree without talking to each other;
- **`modePriority`-deferred ids are removed before partitioning and re-excluded in every shard** — so deferral shrinks the work the shards divide instead of competing with the partition for slots. The two levers compose.

Because the shards plan independently, the merge step cross-checks their uploaded plans (`shard-preview-ids.mjs --verify`: same discovered set, pairwise disjoint, complete cover) before unioning. Without it, a disagreement would surface as a completeness-gate failure naming a component, with nothing pointing at the shards.

### Projection

| Previews | N=1 | N=4 | N=6 | N=8 |
| 689 | 29.4 | 15.8 | 13.7 | 12.7 |
| 1000 | 40.5 | 18.6 | 15.6 | 14.1 |
| 1500 | 58.4 | 23.0 | 18.6 | 16.3 |
| 2500 | 94.3 | 32.0 | 24.5 | 20.8 |

Every shard pays the ~3.7 min compile in full, so **4–6 is the useful range**; 6→8 at 1500 previews saves 2.3 min for two more runners, and 16 buys compile time, not throughput.

Not sharded: the **extra module** (a supplement, small by construction) renders in `generate` as before, and a `@PreviewParameter` provider's **rows** travel whole with their parent id — correct, and the most likely source of a straggler shard. Bin-packing from recorded per-preview times is worth building once one shows up.

### Drive-by

`-o` is now classified in `CliFlags.VALUE_FLAGS`. It has always consumed the next token; the registry test's scanner only matches `--` flags, so it was never caught. Without it a variadic positional list (`bundle merge a.png b.png -o out.png`) reads the output path as another operand.

## Test plan

- `./gradlew :cli:test` — 1616/1617 pass. The one failure, `ServeWebFixtureTest > serve web fixtures are in sync with ServeWeb()`, reproduces on a clean tree at this base and is unrelated to this change.
- New `BundleMergeTest` (9 cases) pins the two things repack cannot do — adding a slot, and carrying the semantics sidecar with it — plus base-wins-on-collision, earlier-shard-wins, overlap reporting, inheritance of manifests/classpath, nested crops + `ir/` + `extensions/`, a deferred preview staying unbaked, and rejection of a non-bundle input.
- New `shard-preview-ids.test.mjs` (20 cases), picked up by the existing `design-artifacts-driver-tests` `node --test` job: disjoint cover, balance within one, determinism under reordering, clamping to the preview count, deferral interaction, and every `verifyShardPlans` failure mode (gap, overlap, divergent discovery, missing shard).
- CLI smoke test of `bundle merge` against two hand-built polyglot bundles: shard 2's preview **and** its `.semantics.json` land in the merged bundle; shard 1's `classes/app.jar` and both manifests survive verbatim.
- `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest` clean; both workflow files parse.

Text-only evidence is correct here — nothing rendered changes. The merged bundle is byte-equivalent in content to the unsharded one; that equivalence is what `BundleMergeTest` and the plan cross-check assert.

## Sequencing

`yschimke/m3-catalog` has the matching one-line `render-shards: 6` on a branch. It must not land before this does — the caller pins `@main`, and GitHub rejects an input the called workflow does not declare, failing the run before any job starts.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KchMGSmbx3Rjp6RZdjMyRa)_

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -283,11 +283,13 @@ jobs:
     name: ${{ inputs.system }} · plan
     if: ${{ inputs.render-shards > 1 }}
     runs-on: ${{ inputs.runs-on }}
-    timeout-minutes: 5
+    timeout-minutes: 10
+    permissions:
+      contents: read
     outputs:
       matrix: ${{ steps.plan.outputs.matrix }}
     steps:
-      # No checkout, no toolchain: this is `[1, 2, …, N]` and nothing else. The real partition is
+      # No checkout, no Gradle: this is `[1, 2, …, N]` and nothing else. The real partition is
       # computed inside each shard from its own discovery, which is what keeps the fan-out free of a
       # serial "discover once, then start the shards" prefix — that prefix would cost a full Gradle
       # configure + compile (~3.7 min) before any shard began, which is the same fixed cost each
@@ -299,6 +301,54 @@ jobs:
         run: |
           set -euo pipefail
           echo "matrix=[$(seq -s, 1 "$SHARDS")]" >> "$GITHUB_OUTPUT"
+
+      # The merge runs LAST, after every shard has spent its ~20 minutes — so a CLI that predates
+      # `bundle merge` would fail with "Unknown bundle subcommand" having burned the whole fan-out.
+      # That is not a hypothetical: the documented default is `cli-source: released`, and
+      # `cli-version: catalog` pins the CLI to the applied plugin version, which for any catalog that
+      # has not bumped its pin is a release older than this feature. So probe it here, once, before
+      # anything expensive starts.
+      #
+      # Only for the released path. `cli-source: build` builds the CLI from the `driver-ref`
+      # checkout, which is the same checkout the merge step builds from, so a probe would mean
+      # building it twice; a `driver-ref` old enough to lack the subcommand is a compose-ai-tools
+      # internal misconfiguration, not a consumer-facing pin.
+      - name: Set up JDK 21
+        if: ${{ inputs.cli-source == 'released' }}
+        uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
+        with:
+          distribution: temurin
+          java-version: '21'
+
+      # `cli-version: catalog` reads the pin out of the caller's version catalog, so the checkout is
+      # needed for the install action to resolve it.
+      - name: Check out the calling repo
+        if: ${{ inputs.cli-source == 'released' }}
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.ref }}
+          persist-credentials: false
+
+      - name: Install compose-preview CLI
+        if: ${{ inputs.cli-source == 'released' }}
+        uses: yschimke/compose-ai-tools/.github/actions/install@main
+        with:
+          version: ${{ inputs.cli-version }}
+          catalog-key: ${{ inputs.catalog-key }}
+          catalog-path: ${{ inputs.catalog-path }}
+          github-token: ${{ github.token }}
+
+      - name: Check the CLI can merge shard renders
+        if: ${{ inputs.cli-source == 'released' }}
+        run: |
+          set -euo pipefail
+          version="$(compose-preview --version | head -1)"
+          if compose-preview bundle help 2>&1 | grep -q 'bundle merge'; then
+            echo "✓ $version supports 'bundle merge'."
+            exit 0
+          fi
+          echo "::error title=CLI too old for render-shards::${version} has no 'compose-preview bundle merge', which the sharded render needs to reassemble the shard bundles. Raise cli-version (or, with cli-version: catalog, the ${{ inputs.catalog-key }} pin in ${{ inputs.catalog-path }}) to a release that carries it, or set render-shards: 1."
+          exit 1
 
   render-shard:
     name: ${{ inputs.system }} · shard ${{ matrix.shard }}/${{ inputs.render-shards }}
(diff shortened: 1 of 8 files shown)


yschimke:
All three review findings were real. Fixed in `a207a26`, with tests for each (the driver suite is now 29 cases).

**P1 — partition only ids the name filter keeps.** Correct: the plan was partitioning the unfiltered discovery output while the render applies `renderFilterPatterns`, so a shard whose share happened to be all deferred-function ids would report work to do and then exclude every id the filter kept — which `composePreviewRender` refuses. `shardRenderPlan` now takes the filter and drops those ids before the partition is drawn (and does not re-exclude them; the name filter already handles them). `previewNameMatches` is a faithful port of `PreviewNameFilter.matches` — glob-vs-equality-or-substring, simple *and* package-qualified candidates — because both directions are wrong here: too permissive re-opens the empty-render case, too strict drops a sticker from every shard.

**P1 — decouple merge from the caller's pinned CLI.** Also correct, and it is the *default* configuration that hits it: `cli-source: released` + `cli-version: catalog` pins to the applied plugin version, so any catalog that has not bumped its pin gets a CLI with no `bundle merge` — discovered only after every shard has spent its twenty minutes. Rather than build a second CLI in the merge job, `shard-matrix` now probes `compose-preview bundle help` for the subcommand *before* the fan-out and fails with the pin to raise. Skipped for `cli-source: build`, where the CLI comes from the same `driver-ref` checkout the merge builds from — an old ref there is an internal misconfiguration, not a consumer-facing pin.

**P2 — compare id sets, not sizes.** Right, and the example is exact: `["a"]` and `["d"]` at `renderable: 2` are disjoint with a union of the right size, so the cover check called it agreement. Each plan now carries a digest of its sorted renderable id set, `--verify` compares those, and plans with no digest are rejected rather than waved through.

---

**On the two red checks — `Build CLI` and `Module Unit Tests`:** both fail on `ServeWebFixtureTest > serve web fixtures are in sync with ServeWeb()`, and both are red on `main` itself, including at `29e5ecb` (the tip this branch is rebased onto). Reproduced on a clean tree at that base here too. The drift is a content-hashed asset URL in a committed golden — `serve-viewer-history.html`'s `/assets/serve/<hash>/viewer.js` — plus a `_render-placeholder.png` byte difference; `UPDATE_SERVE_WEB_FIXTURES=true` regenerates both. Deliberately not doing that in this PR: it is unrelated to this diff, it is live territory (#3430 / #3437), and blessing a golden without knowing whether `ServeWeb` or the fixture is the stale side can paper over a real regression. Auto-merge will pick this up once `main` is green.

Nothing else in this PR touches the render output — the merged bundle is content-equivalent to the unsharded one, which is what `BundleMergeTest` and the plan cross-check assert.


---

PR #3460, opened 2026-08-07
fix(serve): refuse a render whose validated overrides can't be applied

### LINKED ISSUE #3449: Valid render overrides are silently ignored — 200 with byte-identical baked pixels
When the live daemon can't serve a re-render, some overrides fail loudly and others quietly return the baked PNG with a `200`. The quiet ones are the problem: the response is indistinguishable from a successful override.

## Reproduction

Against `https://preview.coo.ee/m3-catalog/render/button-filled__ideal__default__light.png`, while the live daemon is down (#3447):

| request | status | body |
|---|---|---|
| *(no params)* | 200 | baked PNG, 3303 bytes |
| `?fontScale=2.0` | 200 | **byte-identical to baked** |
| `?darkMode=true` | 200 | **byte-identical to baked** |
| `?themeProvider=<valid>` | 503 | `render busy; retry shortly` |
| `?themeProvider=<bogus>` | 400 | `unknown themeProvider '…'; this catalog declares …` |
| `?fontScale=abc` | 400 | *(validation error)* |
| `?zzzBogusParam=1` | 200 | baked — unknown params ignored, which is fine |

```
$ curl -s -o a.png URL
$ curl -s -o b.png 'URL?fontScale=2.0'
$ cmp -s a.png b.png && echo identical
identical
```

## The bug

`fontScale` is a **recognised, validated** parameter — `fontScale=abc` returns 400, so the server parsed it, accepted `2.0` as well-formed, and then served pixels that do not reflect it. Same for `darkMode=true`. A caller gets `200 image/png` and no way to tell the override was dropped.

`themeProvider` on the same request, in the same state, returns 503. So the handling is inconsistent between override kinds, and the more common ones picked the silent path.

## Why it matters beyond this outage

The silent fallback isn't specific to #3447 — any condition where the live lane is unavailable (`livebundle-unavailable`, an untrusted catalog, a missing classpath, no free seat) can serve a baked PNG under a 200 for an override that was never applied. Anything comparing renders across override values — a diff bot, a parity check, an agent iterating on a theme — reads "no visual difference" and concludes the override had no effect on the UI, when in fact it was never rendered. That is a wrong answer delivered confidently, which is worse than an error.

## Suggested direction

Pick one and make every override kind agree:

- **Preferred:** if a validated override cannot be honoured, don't return `200 image/png`. Fail with a status and a body naming the reason.
- If a baked fallback is deliberate, mark it unmissably — a response header (e.g. `x-compose-preview-render: baked-fallback` plus the dropped params) at minimum, since the pixels themselves carry no signal.

Either way `fontScale` / `darkMode` / `themeProvider` should behave the same way in the same state, and the viewer should surface it rather than showing a snapshot as though the override applied.

Observed on preview.coo.ee, compose-preview serve **0.19.43**.

yschimke:
Fixes #3449.

## The bug

When the live render lane can't serve — daemon down, `livebundle-unavailable`, an untrusted catalog, an unmapped Android-only variant, no free live seat — `GET /render/<id>.png` answered an override-bearing request with the preview's **baked snapshot** under a `200 image/png`. Those pixels are byte-identical to the un-overridden render, so a diff bot, a parity check, or an agent iterating on a theme reads "no visual difference" and concludes the override had no effect on the UI, when in fact it was never rendered. `themeProvider` already returned `503` in that same state while the commoner `fontScale` / `uiMode` took the silent path, so the handling also disagreed between override kinds.

## The fix

Baked pixels answering an override-bearing request are, by definition, pixels that don't reflect the override (`Generation.BAKED` means no renderer was involved). Every override kind now agrees:

| state | response |
|---|---|
| no override, or one the baked variant already satisfies (`uiMode=light` on a `…__light` id) | `200` + `X-Compose-Preview-Generation: baked` |
| the override was applied | `200` + `X-Compose-Preview-Generation: daemon` (or `daemon-cache` / `catalog-cache`) |
| the preview **has** a live lane, just not right now | `503` + `Retry-After: 2` — same as a pure `themeProvider` request already did |
| the preview has **no** live lane at all | `409` — retrying can't help, and the viewer already treats 409 as terminal |
| `&fallback=baked` | `200` + `X-Compose-Preview-Render: baked-fallback` |

All four override-carrying responses carry **`X-Compose-Preview-Dropped-Overrides`** — exactly the params the returned pixels don't reflect (`fontScale,uiMode`, `knob.label`, `rc.stopColor`, …) — so a `curl -I` can tell. An unrecognised param is still ignored silently (a cache-buster must not 409 a page) and a malformed value still `400`s at parse time.

`CatalogLiveRouting.overridesAffectRender` stays the authority on *whether* anything was dropped (it compares against a defaults instance, so a newly added override field is covered without touching the naming code); the new `droppedOverrideNames` adds the human detail on top, falling back to a catch-all `overrides` for a field it doesn't spell.

## Visual evidence

Viewer at `/p/button-filled__ideal__default__light?fontScale=2.0` on a baked-only session — the state the issue was reported in. Before, the stage shows the snapshot as though the font scale applied; after, it names what wasn't rendered.

| before | after |
|---|---|
| <img width="440" src="``https://raw.githubusercontent.com/yschimke/compose-ai-tools/c5e02e9034c40221e40bffceb089a1d912af7c8e/docs/design/evidence/serve-override-honesty/viewer-fontscale-before.png"&gt;`` | <img width="440" src="``https://raw.githubusercontent.com/yschimke/compose-ai-tools/c5e02e9034c40221e40bffceb089a1d912af7c8e/docs/design/evidence/serve-override-honesty/viewer-fontscale-after.png"&gt;`` |

## Test plan

- `./gradlew :cli:test` green (1626 tests), including new coverage:
  - `CatalogLiveRoutingTest` — dropped-override naming, the `uiMode`-matches-baked no-op, and the invariant that anything `overridesAffectRender` routes to a daemon is reportable.
  - `ServeHttpRoutingTest` — 409 for a static session, 503 + `Retry-After` for a live lane that's merely down, `fallback=baked` opting back in with both headers, and unrecognised params still serving the snapshot.
  - `variant render remains non cacheable` updated to the new contract (the opted-in fallback is still `no-store`).
- Wired into the e2e harness so future changes are covered automatically: the `bundle-upload` spec (baked-only serve, real uploaded bundle) asserts the 409 + `fallback=baked` pair, and `serve-lanes` asserts the honest side — a real daemon render carries no dropped-override header.
- Reproduced end-to-end against a locally built CLI serving a static bundle, replaying the issue's own table:

```
(no params)                      200 | Generation: baked
?fontScale=2.0                   409 | Dropped-Overrides: fontScale
?uiMode=dark                     409 | Dropped-Overrides: uiMode
?uiMode=light (matches variant)  200 | Generation: baked
?knob.label=Hi                   409 | Dropped-Overrides: knob.label
?zzzBogusParam=1                 200 | Generation: baked
?fontScale=abc                   400 | fontScale must be a positive number, got 'abc'
?fontScale=2.0&fallback=baked    200 | Render: baked-fallback | Dropped-Overrides: fontScale
```

  On the pre-fix build the same `?fontScale=2.0` request returned `200` with bytes `cmp`-identical to the plain render.
- Docs: new *"When an override can't be applied, the render is refused"* section in `docs/public-preview-server.md`, next to the fallback conditions that cause it.

https://claude.ai/code/session_013SzM3BTqbHwsNaPa1Nd1Fb

---
_Generated by [Claude Code](https://claude.ai/code/session_013SzM3BTqbHwsNaPa1Nd1Fb)_


yschimke:
Both Codex review points were real; fixed in b709663.

1. **`background=default` was being refused.** `clearBackground = false` means "preserve the preview's authored background" — what the baked render drew — so the snapshot honours it. It now shares one `withoutBakedNoOps` normalization with the uiMode-matches-the-variant no-op, so the routing predicate and the dropped-override names can't disagree: neither wakes a daemon for it, and neither names it. `background=clear` still needs a real re-render. Covered by a new `CatalogLiveRoutingTest` case.

2. **Bookmarked interactive modes stalled after a refusal.** The bootstrap waited on the stage ``'s load/error, which a failed render never fires (no `src` is assigned), so `?mode=wasm&fontScale=2.0` sat 8s before entering the lane that can apply the override. A snapshot request now reports settled whichever way it went, and the bootstrap enters the bookmarked mode on that — which also fixes the stall for every pre-existing failing-render case, not just the new refusal.

`:cli:test` green.

---

Issue #2930, opened 2026-07-29
rc-player: `WatchScreenRemote` renders a completely empty canvas (75.97% mismatch)


yschimke:
The vendored TypeScript player produces **literally nothing** for the `remote-m3` catalog's
`WatchScreenRemote` preview — not a partial or mis-scaled render, an entirely blank canvas.

Measured on the three-way `rc-compare` lane added in #2929, all three images at 454×454:

| | distinct RGB values | mid-pixel |
|---|---|---|
| baked PNG | 770 | `51,46,60,255` |
| **JS player** | **1** | `128,128,128` (the flatten background — nothing was drawn) |
| embedded player | 732 | `51,46,60,255` |

`pixelmatch` scores it **75.97%** against the baked PNG. The embedded player renders the same
document at **2.94%**, so the document itself is fine and decodes correctly — this is specific to the
TypeScript player.

![WatchScreenRemote across both players](``https://raw.githubusercontent.com/yschimke/compose-ai-tools/a7bbc4375eb343a1389e7dcc06f2fb58d05b8069/docs/design/evidence/rc-embedded-player/rc-compare-three-way.png)``

_(second row of the page — baked, blank JS render, all-red diff, then the embedded render which is
nearly identical to baked)_

## Why this is not one of the already-tracked issues

Checked against the three open rc-player issues before filing:

- **#2770** (DP density audit) — that's mis-*scaling*; content still draws. Here nothing draws at
  all. Its "largest single residual" is `IconRemote` at 5.83%; this is 75.97%.
- **#2741** (`CanvasOperations` / fill-path) — that's an empty fill *shape* inside an otherwise
  rendering document. Here the whole canvas is empty.
- **#2735** (`CoreSemantics` opcode 250) — that truncation mode is already fixed, and it would
  present differently anyway: `rc-compare` flags unknown-opcode documents as *"not decodable"* from
  the `Unknown operation opcode` console warning. **This document was flagged as rendered**, so the
  player parsed it, ran, and drew nothing.

That last point is the useful diagnostic: it is not a parse failure. Something in the draw path
silently produces no output for this document.

## Suspicion worth checking first

`WatchScreenRemote` is the catalog's only **Wear** preview (227dp round watch face, and the only
454×454 document). Whether the player has a gap around Wear-specific layout or the round-clip path
would be the first thing to rule out.

## Reproducing

```sh
node scripts/design-artifacts/rc-compare.mjs \
[...]

PR #3465, opened 2026-08-07
fix(rc-player): refresh the vendored player to 53e19e93 and keep weighted children measurable

yschimke:
Refreshes `third_party/remote-compose-player/` from `d8b07da2` to `53e19e93` (five upstream commits, mostly `c3a08e1` "typescript: fix six layout and variable-resolution defects") and rebuilds `cli/src/main/resources/rc-player/bundle.js` from the result.

## Two of our deltas stop being ours

Upstream implemented both in the meantime, so they come from upstream now:

- `CoreSemantics.ts` is deleted in favour of upstream's `AccessibilitySemantics` — identical wire reads, and both were registered for opcode 250, so the map held two entries for it.
- Our `TextLayout` is replaced by upstream's, which additionally decodes the dynamic-colour flag. The one line of ours kept on top is `readNanIdBits()` for the NaN-boxed `fontSize`/`fontWeight`: same four bytes as upstream's `readInt()`, but only that one is the id-remapping hook `LoomWireBuffer` overrides under macro expansion.

## One fix upstream's new code needed

Upstream's new `WEIGHT` branch sizes a weighted child from its modifier-defined size (`mPadBeforeWidth`, normally 0) rather than the full `maxWidth` — right for a weight that never gets distributed, on the cross axis or in a wrapping parent. But `RowLayout`/`ColumnLayout` hand the distributed share down *as the incoming constraint*, re-measuring each weighted child with `minWidth == maxWidth == childWidth`; taking the modifier-defined size unconditionally discards that and re-measures the child at ~0. It now clamps into the incoming `[min, max]`, so a tight constraint wins and a loose one keeps upstream's behaviour.

The child is then laid out **and painted** at that zero width, so its text re-wraps one word per line and each line is centred about a zero-width box — landing at a negative x outside the component, where the clip removes it.

| before the refresh | refresh without the fix | this branch |
|---|---|---|
| <img width="260" src="``https://raw.githubusercontent.com/yschimke/compose-ai-tools/cfdcf4017bd155629daafa591aa968aaa7904429/docs/design/evidence/rc-weighted-child/watchscreen-before-refresh.png"&gt;`` | <img width="260" src="``https://raw.githubusercontent.com/yschimke/compose-ai-tools/cfdcf4017bd155629daafa591aa968aaa7904429/docs/design/evidence/rc-weighted-child/watchscreen-refresh-without-fix.png"&gt;`` | <img width="260" src="``https://raw.githubusercontent.com/yschimke/compose-ai-tools/cfdcf4017bd155629daafa591aa968aaa7904429/docs/design/evidence/rc-weighted-child/watchscreen-after.png"&gt;`` |
| 78.8454% covered, 1046 colours | 78.8464% covered, 859 colours | 78.8454% covered, 1046 colours |

Both card titles are gone from the middle render. The values survive because they sit in a different, unweighted component. Full write-up in [`docs/design/evidence/rc-weighted-child/`](docs/design/evidence/rc-weighted-child/README.md).

## The test that should have caught it

`rc-round-clip.test.mjs` asserted canvas coverage above 40% — the right shape of assertion for the blank-canvas failure it was written for (#2930), and blind to this one. Two titles are ~0.001% of a 454×454 canvas, so coverage moved from 78.8454% to 78.8464%: *up*, four decimal places below the margin that guards against font drift. The document parsed cleanly and warned about nothing, so nothing else reported it either.

It now also asserts on the text draws themselves — one `fillText` per expected string (a re-wrap arrives as two), and no negative offsets — both properties of a correct layout rather than of a font's metrics, so they stay stable where a glyph-level pixel comparison would not.

## Test plan

- `npm test` in `scripts/design-artifacts` with Chromium available: **528 pass, 0 fail, 1 skip** (the CMP/Wasm frame-pacing test, which needs a built wasm player dist). Both browser suites that drive the built bundle — `rc-round-clip` and `rc-webfonts` — run and assert.
- The new assertion verified **failing** against a bundle built without the `WEIGHT` fix, and passing with it.
- `tsc --noEmit` over the vendored source: 30 errors, all pre-existing, two fewer than before — upstream's `CoreText.OP_CODE` widening removes the `TextLayout` static-side clash, and `nearestFontStretch` is now typed as the `fontStretch` enum rather than `string`.
- Round-clip fixture rendered before/after: identical coverage (78.8454%) and colour count (1046) to the pre-refresh bundle.

The same player-side fixes are staged for upstream on `claude/remotecompose-player-sync-h4ditg` in `yschimke/remotecompose-experiments`, where `base.rc`/`canvas.rc`/`pie_chart.rc` render byte-identical through `render.mjs`, `sweep.mjs` reports no docfaults, and `tsc` reports the same error set as before.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nk3pKb4dr6weM1GHmon57g)_


yschimke:
Two of the three review findings are fixed in c5e1816; the third I'm leaving.

**Refresh pre-size padding after resolving modifiers — fixed.** Confirmed by reading `refreshPadding()`, whose own doc comment states the reason: `inflate()` snapshots padding before `PaddingModifier.updateVariables` has resolved variables or applied DP density scaling. That helper was written to undo exactly this for the padding *totals*, and the new ordered accumulators were simply never added to it — so they stayed on the load-time snapshot while feeding the measured size. `.padding(30).width(200)` measured 260 instead of 320 at density 2, and variable padding dropped out of the size entirely (an unresolved modifier reads 0 at load time). `inflate()` now records which padding modifiers fall before each size modifier and `refreshPadding()` re-sums their resolved values. The partition itself is fixed by the modifier list, so only the values need re-reading.

**Clamp inset child constraints to zero — fixed.** Reachable as described: padding can exceed the component's own size and the size modifier is not widened to fit. There were four such subtractions, not one (the wrap inset, the scroll host size, the scroll re-measure inset, and the plain `computeSize` inset), all now going through a single `contentExtent()` helper that floors at zero.

**Register the impulse timing variables as listeners — not taking this one.** The reasoning looks right and `ImpulseOperation` does lack `registerListening()`, but it's untouched upstream code that this branch only vendors, and I have no document that drives an impulse's `duration`/`startAt` from a variable — so I'd be making an unverified behaviour change to timing code on a vendoring refresh. It belongs upstream in `camaelon/remotecompose-experiments` with a document that exercises it. The player-side fixes from this branch are staged there on `claude/remotecompose-player-sync-h4ditg` in `yschimke/remotecompose-experiments`; this is worth filing alongside them rather than folding in here.

Suite after the fixes: 528 pass, 0 fail, 1 skip. Round-clip fixture unchanged at 78.8454% coverage / 1046 colours, and `tsc` reports the same 30 pre-existing errors.

---

Issue #3447, opened 2026-08-07
m3-catalog live renders all fail with Skiko UnsatisfiedLinkError (PathBuilder_nMakeFromPath)


yschimke:
Every live render on the `m3-catalog` daemon at preview.coo.ee is failing. Baked PNGs still serve fine, so the catalog looks healthy until you request an override.

## Symptom

```
$ curl -i 'https://preview.coo.ee/m3-catalog/render/button-filled__ideal__default__light.png?themeProvider=ee.schimke.m3catalog.LightHighContrastTheme'
HTTP/2 503
retry-after: 2

render busy; retry shortly
```

Persistent — not a blip. Reproduces across ~40s of retries and on every valid theme (`LightMediumContrastTheme` too). The request itself is well-formed; the server validates the parameter and a bogus theme gets a correct 400 listing `LightHighContrastTheme` among the declared ones.

## Actual failure

From `/status.json`, the daemon is failing **3794 of 3990 renders (95%)**, all with one reason:

```
render failed: UnsatisfiedLinkError:
  'long org.jetbrains.skia.PathBuilderKt.PathBuilder_nMakeFromPath(long)'
```

```json
"id": "m3-catalog", "backend": "desktop", "uptimeSeconds": 865,
"renderStats": { "renders": 3990, "ok": 196, "failed": 3794, "timedOut": 0, "busy": 0 }
```

It is the only failing daemon on the box, and the only **desktop/Skiko** one — `jetlagged`, `jetsnack` and `pocketcasts` are Android backends at zero failures. Server-wide `failed=3794` is entirely this catalog.

## Likely cause

The signature is a **Skiko JVM-vs-native version skew**: the loaded `libskiko` doesn't export a symbol the `org.jetbrains.skia` classes call, so the bundle's skiko classes are newer than the native library actually loaded in-process. `PathBuilder` is a recent Skia API, which fits. Skiko loads its native once per JVM, so a live bundle carrying newer skiko classes than the serve process's already-loaded native would fail exactly this way, on every render, forever.

Inference from the error shape rather than something I've confirmed against the bundle — the next step is comparing the skiko version on the m3-catalog live bundle's classpath against what the serve box loads.

## Timing

- Catalog regenerated **2026-08-07T15:45:42Z** (818 previews)
- Daemon uptime at time of report: **865s** (≈15:51), failing since
- Catalog generated by compose-ai-tools **0.19.41**; server running **0.19.43**

[...]

PR #3485, opened 2026-08-08
fix(ci): restore green main — iOS 26 SDK for the player, Rule 3 only where it applies

yschimke:
Three jobs are red on `main` ([CI run 31228260935](https://github.com/yschimke/compose-ai-tools/actions/runs/31228260935), [Integration run 31228260860](https://github.com/yschimke/compose-ai-tools/actions/runs/31228260860)). All three are fallout from the Compose Multiplatform 1.11.1 bump (#3462) — one from the SDK it now needs, two from Rule 3, the render-graph exclusion that PR added to contain it.

## CI / CMP Remote Compose Player

`:rc-player-compose:linkDebugTestIosSimulatorArm64` dies ~17 min in:

```
ld: warning: Could not find or use auto-linked framework 'UIUtilities'
Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_UIViewLayoutRegion", referenced from:
       in liborg.jetbrains.compose.ui:ui-uikit-cache.a[20](CMPLayoutRegion.o)
```

`UIViewLayoutRegion` is an iOS 26 API. The `macos-15` image has Xcode 26.0.1 / 26.1.1 / 26.2 / 26.3 installed but still **defaults** to 16.4, whose iOS 18.5 SDK has no such class. Added a step that selects the newest installed `Xcode_26*.app` before the Gradle invocation — newest-installed rather than a pin, so an image refresh retiring a point release doesn't take the job with it, and a hard failure with the available-Xcode listing if none is present.

## Integration / wear-os-samples (ComposeStarter)

Every preview fails in the renderer's own code, before user code runs:

```
RuntimeError: renderFailed: {'kind': 'classpathSkew',
  'message': 'NoClassDefFoundError: androidx/compose/material3/ColorScheme'}
    at ee.schimke.composeai.daemon.RenderEngine…evaluate$lambda$1$0$2(RenderEngine.kt:442)
```

A Wear app themes with `androidx.wear.compose.material3` and has **no** `androidx.compose.material3` anywhere on its graph. Material 3 is `compileOnly` on `:daemon:android` precisely so the consumer's copy wins at runtime, and Rule 3 stopped ours being smuggled in behind its back — so `CaptureMaterialTheme` had nothing to load.

The daemon now probes the classpath and skips both the theme capture and the `compose/theme` payload (which reads Material 3 objects out of the slot tables) when there is none. Everything downstream is already nullable, so such a render degrades to "no Material 3 theme reported" rather than failing. This is the shape `renderers/android` has always had — `WearMaterialTheme` reads Wear's theme reflectively, and its Material 3 reads sit behind theme-catalog render modes, which is why the Gradle render path survived this and the daemon path didn't.

## Integration / wear-os-samples (WearTilesKotlin)

`composePreviewRender` succeeds but every one of the 188 renders writes an `.error.json` sidecar, so `Verify PNGs were produced` fails.

Both of the original hypotheses were wrong, and the diagnostics commit is what settled it. Neither the Rule 3 gate as first written nor `compileSdk = 37` exceeding Robolectric's max SDK was the cause — the `compileSdk 37` warning is real and appears on every run of this leg, but it is a clamp notice, not the failure. The sidecars named the actual cause, twice, and the two exceptions turned out to be two faces of one skew.

**First: the classes were too old.**

```
NoSuchMethodError: 'kotlin.jvm.functions.Function1 androidx.compose.ui.node
  .ComposeUiNode$Companion.getApplyOnDeactivatedNodeAssertion()'
    at …renderer.RobolectricRenderTestBase.MeasuredWrapBox(RobolectricRenderTest.kt:2998)
```

The gate was reading **the plugin's own injection** as the consumer's Compose. For a tile-only consumer the plugin injects `androidx.compose.ui:ui` and `androidx.compose.foundation:foundation` at `RENDERER_COMPOSE_FLOOR_VERSION` (1.9.5) onto the main variant precisely *because* it has none — the merged unit-test resource APK needs those R classes. The probe saw a declared `androidx.compose.ui`, concluded the consumer had Compose, and left Rule 3 on. Rule 3 then dropped the Compose Multiplatform transitives carrying compose-ui 1.11.2, leaving the render classpath on the 1.9.5 floor, which the renderer's own bytecode does not link against. Circular: we injected the Compose that convinced us not to supply Compose.

Fixed by tracking the dependencies the plugin contributes (`addPluginDependency`, an identity set on the project's extra properties) and skipping them in the probe. Identity rather than coordinate, so a consumer's own `androidx.compose.ui:ui` still registers next to our floor pin — `ComposeStarter` still matches, and the R$id skew of #3447 that Rule 3 exists for stays fixed.

One deliberate limit, tested and documented rather than worked around: Gradle's `DependencySet` collapses *equal* dependencies, so a consumer declaring the identical coordinate **and** version we inject reads as Compose-less. That is the safe direction — its only Compose would be the floor, and Rule 3 ON would strip our transitives and leave the renderer unlinkable against it.

**Then: the resources were too old.** With Rule 3 correctly off, the same 188 previews failed on a *different* exception:

```
NoSuchFieldError: Class androidx.compose.ui.R$id does not have member field
  'int androidx_compose_ui_view_compose_view_context'
    at ComposeView_androidKt.getComposeViewContext(ComposeView.android.kt:762)
```

The render classpath's compose-ui was now ours (1.11.2), but the merged unit-test resource APK is built from the consumer's **main variant**, still pinned at the 1.9.5 floor. Newer classes, older R class.

So the main-variant `androidx.compose.ui` / `foundation` pins are now versioned by *which Compose actually runs*, rather than always by the renderer's compile floor:

* **Compose consumer** — Rule 3 on, its own Compose runs and its BOM wins over our floor anyway. Unchanged; raising the floor would push it off its own Compose line for no reason.
* **Compose-less consumer** — Rule 3 off, OUR compose-ui runs, so its resources have to be on that same line.

Both version facts were read off the published artifacts rather than inferred, which is the habit this leg has repeatedly punished skipping: `ui-android` 1.9.5's `R.txt` carries no `androidx_compose_ui_view_compose_view_context` and 1.11.2's does, and the CMP→androidx mapping comes from `org/jetbrains/compose/ui/ui/1.11.1/ui-1.11.1.pom` — it does **not** track the CMP version (1.11.1 maps to 1.11.2), so the constant documents that trap for the next bump.

Two things this exposed that are **not** fixed here, flagged for a decision:

* The `wear-os-samples` cells track `ref: main`, so upstream drift lands in our CI unannounced. WearTilesKotlin has since moved to AGP 9.2.0 / `compileSdk = 37`, which trips our own "exceeds Robolectric's max supported SDK (36); clamping" warning on every run of this leg. Harmless today — it was the runner-up hypothesis and the evidence cleared it — but it is a live source of unrelated red.
* `WearTilesKotlin` is not `pr_blocking`, so on a normal PR every step in it is skipped and the leg reports green regardless. That is what let a broken leg look fixed twice; the runs proving these fixes are manual `workflow_dispatch` runs, where `event_name != 'pull_request'` makes the matrix do real work.

## Diagnostics

`Verify PNGs were produced` used to print the sidecar *paths* and nothing else — 188 filenames and not one word about what threw, with the only route to the exception being a download of the renders artifact (which the network policy blocks from some environments). It now groups sidecars by `(exception, message)` and prints the dominant cause's stack trace, riding in the bundle like `android-all-cache-key.py` since these legs have no checkout. That is what turned this diagnosis from inference into evidence.

## Test plan

- `./gradlew -p gradle-plugin :test --tests '*RenderGraphComposeExclusionTest*'` — ✅ 11 tests. New: our injected floor doesn't make a tile-only consumer look like a Compose consumer (asserted on both the probe and the resulting exclude rules); a real Compose consumer still counts with our floor beside it; the equal-coordinate collapse above; and the main-variant pin following the CMP runtime version on a tile-only consumer while staying on the floor for a Compose consumer.
- `./gradlew -p gradle-plugin :test` — ✅ full plugin unit suite.
- `./gradlew :daemon:android:testDebugUnitTest --tests '*Material3ClasspathProbeTest*'` — ✅ pins both answers of the probe, with a hiding `ClassLoader` standing in for a Wear consumer's render classpath.
- `./gradlew ktfmtFormatAll` — ✅ clean.
- Summarizer exercised over hand-built sidecars: several previews sharing one cause, a second distinct cause, a sidecar with no `topAppFrame`, and an unparseable file.
- End-to-end gate is a manually dispatched Integration run, since the tiles leg does no work on a `pull_request` event.

## Note on evidence

No visual evidence: nothing here changes a rendered surface. The two Integration legs render the same previews they always did — the fix is that they render *at all* again, which the job's own `Verify PNGs were produced` and daemon round-trip steps assert.

### DIFF
--- a/.github/ci/summarize-render-errors.py
+++ b/.github/ci/summarize-render-errors.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python3
+"""Summarize `*.png.error.json` render sidecars for a failing CI leg.
+
+When `composePreviewRender` produces no PNGs, the reason is in the per-preview
+sidecars (`PreviewRenderError`, schema `compose-preview-error/v1`) the renderer
+drops next to where each PNG would have gone. The verify steps used to print
+only the sidecar *paths* — on a leg like `wear-os-samples (WearTilesKotlin)`
+that means 188 filenames and not one word about what actually threw, so
+diagnosing it costs a download of the renders artifact (and that artifact is
+not reachable from every environment).
+
+Print the exceptions instead. Failures of this kind are overwhelmingly one root
+cause replicated across every preview, so group by (exception, message) and show
+the full stack trace for the largest group, with a one-line head for the rest.
+
+Usage: summarize-render-errors.py <sidecar> [<sidecar> ...]
+Always exits 0 — this is a diagnostic, the caller owns the failure exit.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from collections import defaultdict
+
+# Enough to identify the throw site without burying the log in 188 copies.
+STACK_LINES = 40
+
+
+def main(paths: list[str]) -> int:
+    groups: dict[tuple[str, str], list[dict]] = defaultdict(list)
+    unreadable: list[tuple[str, str]] = []
+
+    for path in paths:
+        try:
+            with open(path) as handle:
+                payload = json.load(handle)
+        except (OSError, ValueError) as error:
+            unreadable.append((path, str(error)))
+            continue
+        payload["_path"] = path
+        groups[(payload.get("exception", "?"), payload.get("message", ""))].append(payload)
+
+    if not groups and not unreadable:
+        return 0
+
+    total = sum(len(items) for items in groups.values())
+    print(f"Render error sidecars: {total} preview(s) failed, {len(groups)} distinct cause(s).")
+
+    # Largest group first: the dominant cause is nearly always the real one.
+    ranked = sorted(groups.items(), key=lambda item: len(item[1]), reverse=True)
+
+    for index, ((exception, message), items) in enumerate(ranked):
+        sample = items[0]
+        print()
+        print(f"[{index + 1}/{len(ranked)}] {len(items)} preview(s): {exception}: {message}")
+        frame = sample.get("topAppFrame")
+        if frame:
+            print(f"  at {frame.get('file', '?')}:{frame.get('line', 0)} ({frame.get('function', '?')})")
+        print(f"  e.g. {sample.get('_path', '?')}")
+
+        # Full trace for the dominant cause only; the rest are identified by
+        # their exception + message + top frame above.
+        if index == 0:
+            trace = (sample.get("stackTrace") or "").splitlines()
+            if trace:
+                print("  stack trace:")
+                for line in trace[:STACK_LINES]:
+                    print(f"    {line}")
+                if len(trace) > STACK_LINES:
+                    print(f"    … {len(trace) - STACK_LINES} more frame(s)")
+
+    if unreadable:
+        print()
+        print(f"{len(unreadable)} sidecar(s) could not be parsed:")
+        for path, error in unreadable[:10]:
+            print(f"  {path}: {error}")
+
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv[1:]))
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,6 +315,26 @@ jobs:
         with:
           persist-credentials: false
       - uses: ./.github/actions/setup
+      # Compose Multiplatform 1.11's `ui-uikit` links against iOS 26 SDK symbols — `CMPLayoutRegion`
+      # references `UIViewLayoutRegion`, which does not exist before iOS 26. The `macos-15` image
+      # ships several Xcode 26.x alongside 16.4 but still *defaults* to 16.4 (iOS 18.5 SDK), so
+      # `:rc-player-compose:linkDebugTestIosSimulatorArm64` dies with
+      # `Undefined symbols for architecture arm64: "_OBJC_CLASS_$_UIViewLayoutRegion"` — the whole
+      # job, ~17 min in, on every push since the CMP 1.11.1 bump. Select the newest installed
+      # Xcode 26 rather than pinning one, so an image refresh that retires a point release doesn't
+      # take the job with it.
+      - name: Select an Xcode with the iOS 26 SDK
+        run: |
+          set -euo pipefail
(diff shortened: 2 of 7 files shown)


yschimke:
Thanks — one of these is real, one isn't. Both addressed in e2f14128d.

**P2 (`recordInjectedDependency` version) — correct, fixed.** The injections had moved to `mainVariantComposeVersion(...)` while both doctor entries still said `RENDERER_COMPOSE_FLOOR_VERSION`. For a Compose consumer those agree; for a Compose-less one they diverge 1.9.5 vs 1.11.2, so `doctor.json` misreported exactly in the scenario the doctor exists to explain. Now resolved once into `mainComposeVersion` and used for both `addPluginDependency` calls and both records, so they can't drift again, with each `reason` stating what the version follows.

**P1 (compiler plugin treated as Compose) — the premise doesn't hold, no behaviour change.** It rests on WearTilesKotlin applying `org.jetbrains.kotlin.plugin.compose`. It doesn't, on two counts:

- `WearTilesKotlin/build.gradle.kts` has `alias(libs.plugins.compose.compiler) apply false` — declared at the root, never applied. `hasPlugin` is per-project, so this reads **false** in the app module.
- `WearTilesKotlin/app/build.gradle.kts` applies only `com.android.application` and `libs.plugins.kotlin.android`.

The empirical side agrees: on `12bcab77` the tiles leg rendered its PNGs and passed `Verify at least one TILE preview was rendered` ([job](https://github.com/yschimke/compose-ai-tools/actions/runs/31234774751/job/93045751848)). Had the shortcut fired, Rule 3 would have stayed on and the `NoSuchMethodError` would have come straight back.

The citation was fair, though — the comment at lines 423–424 *did* claim that shape, and it's stale. Upstream moved the app module to `kotlin.android` since it was written (these cells track `ref: main`). Corrected in the same commit, and I've documented the shortcut's direction on `consumerBringsOwnCompose`: it's a positive signal only, and sound in the direction used, since a module compiling `@Composable` code cannot do so without a Compose runtime on its graph.

---

PR #3488, opened 2026-08-08
feat(serve): offer the imported design spec as a viewer lane beside the players

yschimke:
## What

When a served catalog publishes a design reference for a preview — what design-parity imports from Figma (or a PNG bundle / HTML export) into `references/index.json` — the viewer's renderer row now grows a **`SPEC:`** group beside the Remote Compose player chips. Pressing it swaps the imported spec onto the same stage in place of the render.

So the strip that chooses *which player draws the code* now also chooses *what the design says*, and flipping between them is a comparison on the same pixels rather than a navigation.

Previously the only Figma affordance was the `figma spec` provenance link, which sends you out to figma.com; the spec itself was reachable only via the separate compare page.

## Visual evidence

meshcore-mobile's viewer (`serve-viewer-path`), captured by the preview-harness — re-shot on top of the Material 3 chrome from #3490:

| Before — no spec lane | After — the `SPEC:` group beside the renderers |
| --- | --- |
| Renderer row with no spec lane | SPEC group with a Figma chip |

The chip pressed — the imported spec on the stage, the design side's purple accent, `◇ Figma` badge, and the lane labelled "imported design spec — not a render":

The imported spec on the stage

The same lane in dark

## How

- **`ServeWeb.viewerPage`** takes a new `designReference: DesignReference?`. Non-null renders the `SPEC:` chip group, a `` on the stage, a hidden `cp-mode` radio, and a `view diff →` link into the focused Reference/Diff/Actual page for that exact mapping. Null (every catalog that has not adopted design-parity) renders byte-identically to before.
- **`ServeHttpServer`** passes the first reference for the preview — the same precedence `ServeFigmaSpec` already uses.
- **`viewer.js`** gains `openSpec`/`closeSpec` and a `spec` branch in `enterMode`, so the lane joins the existing machinery: `?mode=spec` is bookmarkable, Back/Forward works, every other lane closes it, the zoom limit applies to the spec image, and the override controls are disabled while it owns the stage (no device size, locale or theme re-points a fixed imported raster).
- **`backend-badge.js`** shows `◇ <provider>` — an outline diamond, deliberately not the renderers' `▪`/`▶`, because what is on the stage was not rendered by this server.
- **Palette:** pressed, the chip wears `.cp-parity-lane--figma`'s purple container rather than the renderers' green, so "green is what we rendered, purple is what the design says" holds here as it does on the parity dashboard.
- **Provider-neutral**: a reference from a committed PNG bundle / HTML export / Stitch gets the same lane, labelled `Spec` rather than `Figma`.

**Nothing is fetched from Figma**, here or anywhere else in `serve`: the raster is the catalog's own inert PNG, downloaded and hash-checked at catalog-refresh time and served from this box's `/reference/<id>.png` — the same bytes the comparison page scores against — and it is only requested once the chip is pressed. Its URL is origin-checked before it reaches the stage, the same way the Wasm iframe's src is.

## Test plan

- `./gradlew :cli:test --tests '*ServeWeb*' --tests '*ServeThemeCss*' --tests '*ServeHttpRoutingTest*' --tests '*ServeFigmaSpec*'` — green. Three new `ServeWebTest` cases cover the lane's markup + reference URL, its absence without a reference, and the non-Figma provider wording.
- `ServeWebFixtureTest` golden refreshed (`serve-viewer-path.html`).
- `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot` — 81 passed, after merging #3490; the shots above are from that run.
- Harness wiring so this is diffed from now on: `serve-viewer-path` added to `STYLED_FIXTURES` (the chip's claim is how it is painted), plus a new `spec-lane` fixture state that presses the chip and shoots the spec on the stage.
- `docs/public-preview-server.md` documents the lane next to the existing figma-spec section.


yschimke:
All three review findings fixed in a4c6406:

- **CodeQL / DOM text reinterpreted as HTML** — the raster is now assigned with `specImg.src = …` instead of `setAttribute("src", …)`, matching every other image/frame lane in `viewer.js`. The `setAttribute` form is indistinguishable from setting a script/iframe source on an element whose type isn't known statically, which is what the rule flagged; the value itself was already server-built (literal prefix + url-encoded, validated reference id) and same-origin.
- **Zoom** — `applyZoom()` now also caps `#cp-spec-img`, so a phone-shaped imported reference obeys the same 72vh Fit-screen limit as the render it is being compared against.
- **SVG toggle** — pressing SVG now leaves the spec lane via `setMode("png")` instead of taking the direct-refresh branch. Extended the same routing to the RC canvas and RC/Wasm frame, which had the identical pre-existing bug (badge flips to SVG, hidden snapshot reloads under a lane still on screen with its chip pressed).

Re-verified: `:cli:test --tests '*ServeWeb*'` green, and the full 81-capture `pages-snapshot` harness run passes.

---

yschimke:
Is this from a cached source of the figma?

yschimke:
Yes — cached, three hops away from Figma, with no live call anywhere in `serve`:

1. **Publish time (design-parity).** The Figma adapter exports the mapped node to a canonical PNG and publishes it into `references/index.json` on the catalog's delivery branch, alongside the `figma:<fileKey>/<nodeId>` handle.
2. **Catalog refresh (`ServeCatalogStore.writeDesignReferences`).** The box downloads those raster bytes from the delivery branch into its own store, validating the declared dimensions and the optional `sha256` before accepting them, and rewrites each reference's path to a local `references/<id>.png`.
3. **View time.** The Spec lane points at this box's `/reference/<id>.png` — the local bytes from step 2. No Figma credential exists in `serve`, and no request leaves for figma.com.

So the freshness is "as of the last catalog publish + refresh", not "as of now". That is deliberate rather than a limitation: these are exactly the bytes the parity scorer and the Reference/Diff/Actual page already compare against, so the lane shows *what the verdict was computed from*. A lane that fetched something fresher could disagree with the diff sitting next to it.

The live escape hatch is the `figma spec` link in the provenance row above — that opens the node itself, which is where you'd go to see anything newer than the last publish.

---

yschimke:
Test if any changes needed after the material design change that just landed in main

yschimke:
Merged #3490 in and yes, one change was needed.

The M3 chrome makes `.cp-live-toggle[aria-pressed="true"]` a green container. The Spec chip only recoloured its dot (amber), so on top of M3 a pressed spec chip read as a *green renderer chip with an odd amber light* — exactly the confusion the lane exists to avoid.

Fixed in 236fb38 by taking the colour the palette already assigns to the design side: `.cp-parity-lane--figma`'s purple container, light and dark, overriding the inherited green outright rather than tinting one dot. "Green is what we rendered, purple is what the design says" now holds in the viewer as it does on the parity dashboard.

Everything else survived the merge untouched — the chip already picks up M3's pill shape, min-height and hover state from `.cp-live-toggle`, and the `SPEC:` group label shares `.cp-rc-backends-label` with the RC selector.

Also re-shot the four doc screenshots under the new chrome (they still showed the old one) and refreshed the PR body's evidence to the current commit. `:cli:test --tests '*ServeWeb*' --tests '*ServeThemeCss*'` and all 81 harness captures pass on the merge.

| Before — no spec lane (M3) | After — pressed, purple (M3) |
| --- | --- |
| Renderer row with no spec lane | Spec on the stage |

---

PR #3489, opened 2026-08-08
feat(serve): give the viewer the theme bar + Background/Transparent, as flow-row buttons

yschimke:
The catalog grid carried a Theme picker and a Background/Transparent pair; the single-preview viewer carried neither. Its Theme axis was a `<select>` buried in the ⚙ Overrides drawer, and the page honoured a stored `cp-bg` choice it gave no way to change.

Rebuilt on top of the Material 3 chrome (#3490), which landed while this was open.

## What changed

**Both controls are now common to the grid and the viewer.**

- The viewer bar carries the same theme chips the grid does — Day/Night (Night alone on a dark-first system) plus one chip per app-declared `@ThemeCatalog` theme, valued exactly like the select's options. `#cp-theme` stays in the DOM as the axis's single state holder (viewer.js, the sticky script and URL hydration all read and write it) but is visually removed like the render-mode radios, so the page never shows two controls for one value. A chip click assigns the select and fires its `change`, so every existing lane — daemon re-render, Wasm `?uiMode`, URL state, the catalog-scoped sticky key — applies unchanged, and `syncThemeBar()` mirrors whatever `syncServerControls()` has just decided about the select and each option rather than re-deriving it.
- Transparent moves into a shared `bg-toggle.js` loaded by both pages (the landing script no longer carries its own copy). The CSS already backed `.cp-stage` off `cp-bg-transparent` and the pre-paint script already restored the choice on every page, so the viewer was missing the control, not the behaviour.

**Filter chips in a flow row, not segmented buttons.** M3 specifies a segmented button for a small, connected, *fixed* set of options. The Theme picker is none of those: it grows a chip per declared theme and wraps, and a connected track that wraps draws one keyline around two or three ragged rows. So the picker and the two toggles are M3 **filter chips** instead — the component M3 specifies for a wrapping, variable-length set — sharing `.cp-state-btn`'s shape and the same `secondary-container` selected fill the segmented track used. Disabled chips keep M3's 38% on-surface treatment: a theme this lane cannot render is information, not something to hide.

**Background/Transparent and Fit screen/Fit width are one toggle each** (review ask). Both are two-state axes with a default, which is exactly what `aria-pressed` on a single button expresses — the label names the non-default state (`Transparent`, `Fit width`) and pressed-ness says whether it is on. The pairs spent twice the toolbar width to say the same thing, and each always showed one segment that did nothing when clicked. A dropdown would have been a second interaction for a boolean.

**Harness coverage:** `serve-viewer-themes` joins `STYLED_FIXTURES` — it is the only viewer fixture with app-declared themes, and neither the chip treatment nor the runtime enabled-state sync is visible in a bare capture. Its contract test now drives the *chip* rather than the hidden select, so a bar wired to nothing would fail.

## Visual evidence

Captured with `preview-harness` `pages-snapshot`; **before = `main` at 9eea184** (M3 already applied), so these show only this PR's change. Light theme shown; dark captured too.

**Viewer with declared themes** — the whole axis arrives on the bar: Day/Night, three declared themes, then the two single toggles.

| Before (main) | After |
| --- | --- |
| viewer with declared themes, before | viewer with declared themes, after |

**Viewer on a catalog palette (JetNews)** — the chips take the served design system's M3 palette, and are honestly greyed on a static snapshot with no live lane.

| Before (main) | After |
| --- | --- |
| viewer on the JetNews palette, before | viewer on the JetNews palette, after |

**Catalog grid** — Light/Dark become chips and the Background/Transparent segment becomes one Transparent toggle.

| Before (main) | After |
| --- | --- |
| catalog grid, before | catalog grid, after |

## Test plan

- [x] `./gradlew :cli:test` — green, including the regenerated `ServeWebFixtureTest` goldens.
- [x] New `ServeViewerThemeBarTest`: chips exist and carry option values the select actually offers; the select stays present, `aria-hidden`-wrapped and `tabindex="-1"`; a dark-first system offers Night alone; a theme specimen withholds declared chips from bar and select alike; the grid and the viewer emit the *identical* Transparent toggle, and neither page emits a `data-bg-choice` / `data-zoom-mode` pair any more.
- [x] `ServeUrlStateTest` follows the bg wiring into `bg-toggle.js` and asserts both pages load it.
- [x] `preview-harness` `pages-snapshot` run locally for the three fixtures above (light + dark) plus the theme/knob contract tests — the declared-theme contract clicks the chip and still sees exactly one bounded `?themeProvider=` render.
- [x] `./gradlew ktfmtCheckAll`, `npm --prefix vscode-extension run format`.

Branch renamed from the session's `claude/…` to `agent/…` per the repo's branch convention.


yschimke:
Check if any changes needed, post the material redesign that landed

yschimke:
Can you change Background/Transparent and Fit Screen/Width, to be toggle buttons. So a single button that toggles mode, or a combo / drop down?

yschimke:
Merged the Material 3 chrome (#3490) in and rebuilt on top of it, then took the toggle ask.

**What the M3 merge needed.** Only `serve.css` conflicted, in the two blocks this PR touches. #3490 made the Theme picker and Background/Transparent M3 *segmented buttons*; I kept the M3 tokens but not that component. A segmented button is specified for a small, connected, fixed set of options, and the Theme picker is none of those — it grows a chip per declared `@ThemeCatalog` theme and wraps, so the connected track drew one keyline around two or three ragged rows exactly when a catalog has something to pick from. They're M3 **filter chips** now, sharing `.cp-state-btn`'s shape and the same `secondary-container` selected fill the track used, so nothing about the M3 language is given up. Disabled chips keep M3's 38% on-surface treatment rather than vanishing — a theme this lane can't render is information.

**Toggles.** Background/Transparent and Fit screen/Fit width are one button each, not a dropdown: both are booleans with a default, which is what `aria-pressed` on a single toggle says. The label names the non-default state (`Transparent`, `Fit width`) and pressed-ness says whether it's on — a combo box would make a boolean cost two interactions, and the pairs were each spending twice the toolbar width while showing one segment that did nothing when clicked.

Before/after (against `main` at 9eea184, so M3 is already in the "before"):

| Before (main) | After |
| --- | --- |
| viewer with declared themes, before | viewer with declared themes, after |

The full set (viewer on a catalog palette, catalog grid) is embedded in the PR description. `:cli:test` is green with regenerated goldens; the evidence images moved to `docs/images/` to match #3490's convention.

---

PR #3496, opened 2026-08-08
feat(serve): one chip + one combo for the viewer's renderer, and a single-line export bar

yschimke:
The viewer's lane row had grown a chip per lane — `Live preview · In-browser (Wasm) · RC: JS CMP Wasm Java CMP Android CMP JVM · Spec: Figma · SVG · static snapshot` — so answering *"what is drawing this?"* meant reading up to eight independent pressed-states, and every new lane added another chip. See [`/remote-m3/p/appcard__ideal__default__compact`](https://preview.coo.ee/remote-m3/p/appcard__ideal__default__compact).

## The renderer row

**Before** — every lane its own chip:

Renderer row before

**After** — the chip *names* the current renderer and its status dot says whether that render is interactive (clicking it still toggles live); the combo holds the alternatives; a subtle link steps out to every player side by side; SVG applies to whatever the chip is showing:

Renderer row after

Switching player renames the chip, rather than lighting one of six:

Renderer row after switching to CMP Android

Details:

- **`#cp-live-toggle`** is the chip. It names the current lane (`Java`, `CMP Android`, `Figma spec`, `Live`) and its dot is the live indicator — a hollow grey ring when static, a filled green disc in a halo when interactive, so the two states differ in *shape* as well as colour. Clicking toggles into the best live lane and back out, and its tooltip is re-derived on every transition from the same state (a fixed one would promise "click for live" on a chip whose click now exits to the snapshot).
- **`#cp-lane-select`** is the combo. It is a *command* menu (`Switch renderer…`), not a state field — the chip beside it already holds the state, and a combo repeating that name read as two controls arguing about one fact. It covers every lane: the five Remote Compose players (unavailable ones listed as such), the in-browser Wasm app, and the imported design spec. The per-lane chips `#cp-wasm-btn`, `#cp-rc-backends` and `#cp-spec-btn` are gone; the wiring behind them (`rcPlayer=<wire>`, `setMode("rc"/"rc-wasm"/"wasm"/"spec")`) is unchanged.
- **`compare players →`** opens `/compare?format=rc&preview=<id>` — the existing player wall, focused on this screen. Shown only when there are ≥2 players to compare.
- The spec lane keeps its `spec diff →` step; `#cp-spec-lane` survives as the carrier for `data-spec-src`.

## The export bar

**Before** — the primary hand-off behind a disclosure, and a full-width read-only URL field per format whose only affordance was a `title`:

Export before

**After** — always visible, one line, three plainly-named actions per format:

Export after

The URL itself still exists as a `#cp-url-<ext>` field (`refreshLinks` writes it, both copy buttons and the lane e2e read it) but is taken out of the flow — nobody reads a 200-character absolute `/render` URL, and "Copy link" says what `title="Click to copy the URL"` never managed to.

## Also

Fixes a latent gap this refactor surfaced: `rcPlayer=` rode the page URL (it's in `URL_STATE_PARAMS`) but nothing ever read it back, so a shared `?rcPlayer=cmp-android` link opened on the default player under a control naming it. It now hydrates from the offered options, and falls back to the default for an unknown or unavailable wire.

## Test plan

- `./gradlew ktfmtCheckAll :cli:test --tests '*Serve*'` — green.
- `ServeWebTest` / `ServeHttpRoutingTest` / `ServeWebFixtureTest` updated for the new markup; the fixture goldens are regenerated and back in sync.
- preview-harness `pages-snapshot` — **85/85 passed**, including the new fixture and state below.
- **New coverage:** `serve-viewer-rc-players` is added as a page fixture (styled, with a `player-cmp-android` runtime state), so the CI visual-diff bot renders and diffs this row on every future PR. No existing viewer fixture carried more than two lanes, which is why the row above had never been diffed at full width.
- The images in this description are those harness captures, cropped; they are committed under `docs/design/evidence/renderer-picker/` with a README.

https://claude.ai/code/session_01NGDrV4BJamKdyoV8ugEwvs


yschimke:
resolve conflicts

PR #3516, opened 2026-08-08
feat(design-catalog-remote-m3): declare Roboto Flex and Google Sans Flex typeface themes

yschimke:
The Remote Compose catalog declared no themes at all, so the preview server's **Theme** select was empty for it while its Wear sibling offered four. Follow-up to #3515, which fixed the Wear catalog's themes not changing their typeface.

## Scope: built-in family id 0, and nothing else

A Remote Compose document reaches a typeface in one of three ways (`docs/design/RC_PLAYER_TYPEFACES.md`): a **built-in family id** — `0 = default`, `1 = sans-serif`, `2 = serif`, `3 = monospace` — a **named family**, or an **embedded face**. These themes move the built-in default and stop there.

The named-family stickers are deliberately untouched: `BrandedTextRemote`, `TypefaceSpecimenRemote`, `VariableWeightRemote` and `VariableWidthRemote` are the only coverage the `Named(…)` and font-variation-axis paths have across the five player lanes, and a theme that reached them would delete it. Generic ids 1–3 are out of scope for the same reason — a family is chosen at the `RemoteText` call site, so a sticker that asked for `Monospace` asked for monospace. No sticker here does, which leaves id 0 as the whole surface a theme has to move.

## How the choice reaches the document

`RemoteMaterialTheme` is `@RemoteComposable`, so it can only be installed *inside* the document — but a `PreviewWrapperProvider` wraps the whole `@Preview`, which is outside `RemoteSticker`'s `RemoteOverridablePreview`. So the provider publishes its choice as a plain `LocalRemoteCatalogFont` and `RemoteSticker` installs the matching `RemoteTypography` once it is in remote scope. The three widget-container stickers bypass `RemoteSticker` entirely (the Glance Wear preview path owns its own capture), so `CenteredWidgetContent` installs it the same way — otherwise the Theme select would silently skip them.

## Vendoring, and the wasm size ratchet

The CMP Wasm lane is **manifest-only and never fetches**, so a named family it doesn't carry fails `RcComposeSupport.fontFamilyIssue`'s availability check outright rather than substituting a face. Roboto Flex was already vendored as the `role: "default"` family; Google Sans Flex was not, which would have left the themed documents unrenderable in that one lane while the other four resolved them. Both weights are now vendored.

That tripped `:rc-player-wasm:wasmPlayerDist`, which copies the catalog's fonts directory wholesale into the distribution and sums raw bytes: +256,316 took a 23,256,891-byte dist to 23,513,207 against a 23,500,000 ratchet. The ratchet's own comment says it moves when the growth is understood, and this growth is a deliberate payload, so it is raised to 23,780,000 — the same ~257 KB slack the previous value carried, so an unintended jump still fails. Measured locally: `23513207 / 23780000`, BUILD SUCCESSFUL.

## Verification

**All 27 existing sticker PNGs are byte-identical** to an `origin/main` worktree render, verified with `cmp`. The only new files are the two synthesised theme sheets.

**One `.rc` sidecar does change, with no pixel change**, and it is worth stating plainly rather than rounding to "byte-identical": `CircularProgressRemote`'s document differs from `main` in a handful of id bytes. Adding the theme branch to `RemoteSticker` moves id allocation in the emitted document, and that sticker's named value (`rememberOverridableRemoteFloat("progress")`) lands in the shifted range. Rendering `origin/main` in the *same* tree reproduces the baseline bytes exactly, so this is the change and not capture nondeterminism; two alternative shapes (hoisting the branch outside `RemoteOverridablePreview`, and a static CompositionLocal) were tried and neither removes it. The document is semantically unchanged — same ops, same named value, same raster — so it is recorded rather than worked around.

No before/after images: the diff adds no sticker and changes no pixel of the 27 that exist. The two new files are the *synthesised* `@WearThemeCatalog` specimen sheets, and they are not evidence of anything for this catalog — the renderer builds them by reflecting over `androidx.wear.compose.material3.MaterialTheme`, a phone/watch Compose theme these providers never install, because a Remote Compose theme isn't one. (It could not show a typeface regardless: that sheet lays its type rows out past the bottom of its fixed canvas — see `docs/design/evidence/wear-m3-theme-fonts/` in #3515.) The themes' real effect shows in the preview server, on the stickers themselves, via `?theme=theme:com.example.designcatalogremotem3.RemoteGoogleSansFlexThemeCatalog`.

Worth flagging as a follow-up: a Remote-aware theme specimen — one that reflects `RemoteMaterialTheme` the way the Wear strategy reflects the Wear one — would make this catalog's themes diffable in CI like every other surface.

## Test plan

- `:samples:design-catalog-remote-m3:testDebugUnitTest` — passes, including the new `RemoteCatalogTypographyTest` (pins that a theme re-points the built-in default at a `google:`-namespaced family, and that the stock typography leaves it alone)
- `:samples:design-catalog-remote-m3:composePreviewRenderAll` — BUILD SUCCESSFUL; both themes discovered and rendered
- `:rc-player-wasm:wasmPlayerDist` — BUILD SUCCESSFUL at the new ratchet
- byte-comparison against `origin/main`, plus a same-tree `origin/main` control render to attribute the one sidecar delta


yschimke:
Both review findings were correct and are fixed in 1d83fd1.

**P2 — widget-container previews.** Confirmed: all three bypass `RemoteSticker` (the Glance Wear preview path owns its own capture), so the Theme select skipped them, most visibly on `WidgetContainerLargeRemote` which draws through `RemoteMaterialTheme.typography`. `CenteredWidgetContent` — the shared helper all three go through — now installs a selected typeface the same way the component frame does. Re-verified that the un-themed contract still holds: 29/29 renders byte-identical by `cmp`, 0 differing.

**P1 — Google Sans Flex on the CMP Wasm lane.** Confirmed: that lane is manifest-only and never fetches, and the vendored `fonts.json` carried Roboto Flex (as `role: "default"`) but not Google Sans Flex, so the themed documents would have failed `fontFamilyIssue`'s availability check there while resolving in the other four lanes.

This one needed an owner decision rather than a code fix, because the face is in **no license directory** of the google/fonts corpus — `deploy/image/README.md` records that it is baked into the runtime image only under explicit redistribution clearance, and vendoring it here is a second redistribution. Confirmed with the repo owner, so both weights are now vendored as a `role: "named"` family, and `fonts/README.md` states the clearance and that a fork does not inherit it (with drop-it instructions if a fork can't re-clear).

Worth noting for anyone reading later: the manifest regeneration preserves `role: "named"` families whose files are still vendored, so the new entry survives the design-artifacts export rather than being dropped by the recorder.

---

PR #3517, opened 2026-08-08
feat(daemon-desktop): honour @ScrollingPreview(END) in the served render

yschimke:
## Summary

Completes the `END` story started in #3499 (Android daemon) and #3513 (CMP Desktop renderer). The desktop daemon — the `compose-preview serve` path — was the last backend still serving the resting top for a preview annotated to capture its settled bottom, so the frame the panel showed disagreed with the PNG the Gradle render shipped beside it.

It was skipped in both earlier PRs for a real reason: it captures through `ImageComposeScene`, which has neither a test main-clock nor `onNode` interactions — the two things `renderScrollPreview` leans on. But it does expose both halves needed to drive a scroll directly:

- `scene.semanticsOwners` reaches the same `SemanticsActions.ScrollBy` the test harness invokes, and
- `scene.render(nanoTime)` *is* the frame clock an animated scroll advances on.

So the drive ticks the scene by hand instead of a `MainTestClock`. No move onto `runSkikoComposeUiTest` was needed after all — which also means the semantics / layout / figma-svg products keep coming off the same scene as before.

## Two details worth a reviewer's eye

**Where it runs.** In `render()`, ahead of `renderOnce` — which is also the interactive session's per-input entry point, and must never re-drive under a user who is scrolling the preview themselves.

**The frame clock.** The one-shot path pins `render()`'s `nanoTime` to zero for determinism, and an animated scroll needs time to actually pass. The drive ticks a cursor forward and the capture frames continue from where it stopped, rather than snapping back to zero and handing every in-flight animation a large negative delta on the very frame being captured. Previews without an END drive never touch the cursor, so their frame clock — and their pixels — are exactly what they were.

Because the drive happens before capture, the semantics, layout and figma-svg read off the same scene follow the scroll too. The test asserts on the emitted `compose-figma.svg` for exactly that reason: it proves the whole derived-artifact family moved, not just the raster.

`reduceMotion` is deliberately not honoured here, unlike the Android daemon — it gates Wear's `TransformingLazyColumn` edge transforms, and Wear Compose is Android-only.

## A defect this surfaced (not fixed here)

While testing this I probed what these scrollables actually publish, and the answer explains a lot:

```
[LazyColumnListPreview]  value=0.0  maxValue=100.0   <- item-space, not pixels
[DateRangePicker]        value=0.0  maxValue=0.0     <- not measured yet on the probe frame
```

`ScrollAxisRange` is **not** a pixel budget. The existing `captureLong` driver treats it as one — it sizes each step as `min(stepPx, maxValue - value)` and feeds the result to `ScrollBy`, which *is* in pixels. On a `LazyColumn` that turns a 336 px stride into a ~2 px one, so a `scroll-long` capture of the 30-row fixture stitches 13 slices into a 453 px image instead of a full-page one. The same wrong instrument is why a `DateRangePicker` gets reported as having no scrollable at all.

This PR's drive converges regardless (it measures progress and iterates until nothing moves), so `END` is correct — but `LONG`/`GIF` need the instrument replaced with a geometric one. That's the next change, and it's what "long-scroll screenshots in the preview server" actually depends on.

## Test plan

- [x] New `RenderEngineScrollEndTest`, end-to-end through `DesktopHost`:
  - an `END` preview is captured at the bottom of its list (SVG carries `Row 30`, not `Row 1`)
  - a preview with no scroll intent keeps its resting top
  - a `TOP` preview is not driven
  - `maxScrollPx` bounds the drive
- [x] `:daemon:desktop:test --tests "*RenderEngineTest*" --tests "*RenderEngineWrapContent*" --tests "*RenderEngineFigmaSvgScroll*"` — the neighbouring render-path suites, unchanged
- [x] `ktfmtFormat`

---
_Generated by [Claude Code](https://claude.ai/code/session_01E98YxF6aPxrecEMqV2gWMu)_

### DIFF
--- a/daemon/desktop/src/main/kotlin/ee/schimke/composeai/daemon/RenderEngine.kt
+++ b/daemon/desktop/src/main/kotlin/ee/schimke/composeai/daemon/RenderEngine.kt
@@ -224,9 +224,17 @@ class RenderEngine(
         // so the PNG *and* every tree derived from the same scene (semantics, layout, figma-svg)
         // describe the frame the Gradle render ships. Ahead of `renderOnce`, which is also the
         // interactive session's per-input entry point and must never re-drive under the user.
+        val driveStartNs = System.nanoTime()
         trace.section("render:scrollToEnd") { driveStaticScrollToEnd(state) }
+        val driveNs = System.nanoTime() - driveStartNs
         trace.section("render:once") {
-          renderOnce(state, requestId, sandboxStats = sandboxStats, trace = trace)
+          renderOnce(
+            state,
+            requestId,
+            sandboxStats = sandboxStats,
+            trace = trace,
+            elapsedBeforeNs = driveNs,
+          )
         }
       } finally {
         trace.section("compose:tearDown") { tearDown(state) }
@@ -634,8 +642,16 @@ class RenderEngine(
     trace: PerfettoTraceDataProducer.Recorder =
       PerfettoTraceDataProducer.recorder(state.spec.outputBaseName, backend = "desktop"),
     useWallClockFrameTime: Boolean = false,
+    /**
+     * Wall-clock nanoseconds already spent on this render *before* this call — currently the
+     * `@ScrollingPreview(END)` drive, which runs ahead of `renderOnce` (see [render]) but is part
+     * of the render body by any honest accounting: it can cost tens of scene renders. Folded into
+     * the reported `tookMs` so the daemon's latency and cost-model numbers cover the real work. `0`
+     * for every other caller, including the interactive session.
+     */
+    elapsedBeforeNs: Long = 0L,
   ): RenderResult {
-    val startNs = System.nanoTime()
+    val startNs = System.nanoTime() - elapsedBeforeNs
 
     // Flush snapshot state written out-of-composition since the last render so the held scene
     // observes it before painting. The held-session scrub path mutates `lottieProgressState` from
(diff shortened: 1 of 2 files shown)


yschimke:
**CMP/Wasm Parity is red here and it isn't this change.** Not merge-blocking (`continue-on-error: true` on the job), but leaving it unexplained would be worse than a note.

Why this diff can't reach that row: it touches one file, `daemon/desktop/.../RenderEngine.kt`, and everything it adds is behind `staticScrollFor(previewId) ?: return`. `VariableWeightRemote` carries no `@ScrollingPreview`, so the drive returns before doing anything. The parity job also renders through the CMP/Wasm player, not the desktop daemon — there's no path from this code to those pixels.

The broader pattern, across four unrelated PRs with identical inputs to this job:

| PR | row that moved | delta |
| --- | --- | --- |
| #3496 (serve/viewer) | `NamedLabelRemoteButton` | 4.12% → 4.92% |
| #3499 (scroll daemon) | `FilledRemoteButton` | 3.67% → 4.99% |
| #3496 / #3499 | `CustomShapeRemoteButton` | "improved" on both, to *different* values (4.22% vs 3.37%), moving in opposite directions |
| #3517 (this one) | `VariableWeightRemote` | 1.38% → 1.82% |

A different row each time, none related to the diff, one row both improving and regressing across runs. Every input to the job — published corpus bundle, `rc-player` sources, `rc-compare.mjs` — is unchanged between those branches and `main`, so the ±0.25 pp threshold appears to sit below the headless render's own run-to-run variance.

Not touching it in this PR; flagging it as worth either raising the threshold or stabilising the render, since as it stands it will cry wolf on essentially every PR.

---

PR #3531, opened 2026-08-08
fix(serve): grey out the controls an IR replay can't honour, and stop refusing two it can

yschimke:
Follow-up to #3512, which stopped the server reporting an override an IR-replayed preview cannot apply as applied. Two gaps left over — one of them a bug that PR introduced.

## `fontScale` and `uiMode` were wrongly on the dropped list

They look inert against `remote-m3` — every render returns the baked bytes — but that is a property of *those documents*, not of replay. A document can defer both to the host and resolve them at paint time, with no recomposition:

```
RemoteComposeView.getDefaultTextSize() = 14f * density * Configuration.fontScale
onDraw(): paint theme from Configuration.isNightModeActive()
          when the player's theme is THEME_UNSPECIFIED
```

Both read the live Android `Configuration`, which `RenderEngine` already sets per render spec — so the wiring is end-to-end today, and a document that reads the host values genuinely responds. `remote-m3` simply baked absolute text sizes and concrete colours at capture.

Naming them meant 409ing an override the replay can honour, which is exactly the false-refusal failure the list's narrowness exists to prevent. Their silence on a constant-folded document is authored behaviour, not a server lie.

The bar for entering the list is now **no representation in the document at all**, not "observed inert against one catalog". That leaves:

| named | why |
|---|---|
| `localeTag` | `stringResource()` resolved to a literal at capture, and `RemoteContext` carries no locale among its system variables — a document cannot defer it |
| `themeProvider`, `knob.*` | seeded into a composition that never runs; no document-side counterpart |
| string `rc.*` | reaches the player but is swallowed below `RemoteComposeState.overrideData` (unchanged from #3512) |

## The viewer still offered those controls

A 409 is more legible than silent unchanged pixels, but inviting someone to drag a slider into an error isn't the finished job. The page now carries `data-ir-replay`, read from the same host question the refusal derives from — deliberately its own flag rather than the RC canvas lane's `data-has-rc-doc`, so a future host that serves a document for a class-backed preview can't grey that preview's live controls. `viewer.js` disables exactly the refused set, with a hover explanation rather than a control that is merely dead.

Two nuances in the gating:
- **Day/Night and Font scale stay live**, per the above.
- **String `rc.` knobs are gated on the lane, not the preview** — the browser RC lanes drive a different player and apply them client-side, so they stay editable there and are disabled only on the server lane.

## Visual evidence

Viewer controls column, all drawers expanded, same preview with and without the flag.

| Before (`irReplay = false`) | After (`irReplay = true`) |
|---|---|
| controls before | controls after |

In the "after": **Locale** greyed, **Font scale** still live (the correction above), **Overrides** `label` + `count` greyed, and under **Remote Compose** the string `label` greyed while `progress` and `shaderColor` stay editable.

Corroborated from the DOM rather than eyeballed:

```
before {"irReplay":null,"localeTag":false,"fontScale":false,"themeProviderOpts":[false],
        "knobs":[["label",false],["count",false]],
        "rc":[["label","string",false],["progress","float",false],["shaderColor","color",false]]}
after  {"irReplay":"1","localeTag":true,"fontScale":false,"themeProviderOpts":[true],
        "knobs":[["label",true],["count",true]],
        "rc":[["label","string",true],["progress","float",false],["shaderColor","color",false]]}
```

## Test plan

- `IrReplayDroppedOverridesTest` — now asserts `fontScale` / `uiMode` are **never** named; the negative assertions are the load-bearing ones, since a wrong positive here refuses a working override.
- `ServeViewerIrReplayControlsTest` — pins the flag and its independence from the RC-doc gate.
- `:cli:test --tests 'ee.schimke.composeai.cli.serve.*'` green; `:cli:ktfmtFormat` clean.

## Follow-up not in this PR

The screenshots above were captured by driving the served viewer HTML through headless Chromium by hand. That surface has no standing fixture in `vscode-extension/preview-harness`, so the next change to these control states won't be diffed automatically. Wiring an `ir-replay` scenario into `serve-lanes.spec.mjs` is the gap to close — worth a fast-follow rather than bundling here.


yschimke:
Codex's `.cp-knob` finding was correct — fixed in 812b9ba.

`gateForIrReplay` only ever *adds* the disable, on the documented assumption that every call site assigns `el.disabled` from its own lane logic first. The locale and RC-knob paths do; the `.cp-knob` path didn't, so a knob switched off for the snapshot lane stayed off after entering Wasm — and `wasmOverridePatch()` skips disabled knobs, so the edit never reached the app.

Author knobs are the one control here with no lane-derived enabled state to restore: nothing re-synced them across transitions before this change, so their base is whatever the server rendered. The fix records that once in `data-base-disabled` and restores it on each pass before gating.

Same commit re-attributes the authorship that tripped `Reject agent attribution`.

---

PR #3532, opened 2026-08-08
feat(catalog): record WHY a component has no design reference

yschimke:
`@CatalogComponent(reference = "")` currently means four different things, and a consumer cannot tell them apart:

- nobody has looked yet
- the kit retired the pattern and publishes no replacement
- the kit never published this component
- a node exists, but pointing at it would report something that is not a defect

Only the first is a gap. The rest are findings about the design system, and today they read as neglect, because absence has no voice.

`noReference` gives the absence a reason that travels with the component, the way `reference` carries the handle when there is one:

```kotlin
@CatalogComponent(
  id = "NavigationDrawer/Modal",
  noReference = "the kit files its only drawer under Deprecated",
)
```

## Where this came from

Mapping the M3 catalog against the Material 3 Design Kit. Five of its 77 components have no live counterpart for reasons worth recording rather than rediscovering:

| component | why |
|---|---|
| the three navigation drawers | the kit's only drawer node sits under `Navigation / Deprecated`, and a search of the whole file finds no replacement |
| `NavigationBar/Standard` | Compose's 80dp `NavigationBar`; the kit publishes only the 64dp bar |
| `Search/TopBar` | the kit's `Search bar` is the same component minus the window insets `TopSearchBar` carries, so mapping it would report the insets as a height defect |

Each of those is a fact about the kit that a reader should get from the catalog, not from an archaeology session.

## Scope

- `noReference` on the annotation, defaulted to empty so nothing existing changes.
- Threaded through discovery to `previews.json`.
- Mirrored on the `preview-data-api` wire type, so `compose-preview show --json`, contrib scripting and MCP readers see it instead of having `ignoreUnknownKeys` drop it silently.

Nothing is inferred. A component that says nothing still reports both fields null, which stays distinguishable from one that states its reason.

## Rebuilt on current `main`

The earlier version of this branch shared **no merge base** with `main` — its parent (`b962e2f0`, Aug 1) is not on `main` at all, so the two histories were disjoint and GitHub reported the PR unmergeable rather than conflicted. There was nothing to resolve; the change is re-applied on top of `main` instead, same five files.

Two things changed in the re-apply, both because `referenceSet` (#3530, #3534) landed in the same declaration meanwhile:

- **`noReference` is appended** to the parameter list rather than slotted in after `reference`, for the reason #3534 established: a parameter inserted ahead of an existing one silently re-points positional calls.
- **The "nobody has looked yet" assertion moved** from `FilledButton` to `PlainSticker`. `FilledButton` now names both kit handles, so it can no longer stand for the silent case; `PlainSticker` carries no arguments at all, which is exactly that case.

## Test plan

- `:gradle-plugin:functionalTest --tests "*attaches @CatalogComponent*"` — **passes** on the rebuilt branch.
  - A new `RetiredButton` component declares `noReference` and the test asserts the reason lands on the manifest with `reference` still null.
  - `PlainSticker` asserts `reference`, `referenceSet` and `noReference` are **all** null — the silent case, kept distinguishable from a stated absence.
- `ktfmtCheck` on all four touched modules — clean.

Non-visual: a metadata field travelling through JSON. No rendered surface changes.


yschimke:
fix merge conflicts

Issue #3559, opened 2026-08-08
perf: re-tune catalog render sharding now that captures run on a warm renderer


yschimke:
m3-catalog's shard count was chosen against a cost model that #3548 has invalidated, and it is currently pinned to `render-shards: 1` for a separate correctness reason. Both need revisiting, in that order.

## The old arithmetic

`design-artifacts.yml` records the reasoning:

> growing linearly with the sheet — `render_minutes ~= 3.7 + 2.15s x previews` […] Six, not sixteen: every shard pays the ~3.7 min of Gradle configure + compile in full, so past about six the fixed cost dominates and extra runners buy compile time, not throughput.

That model has two terms. #3548 collapsed the **marginal** one: captures now draw on a warm renderer instead of forking a JVM each time, so the per-preview cost is no longer ~2.15 s. The **fixed** ~3.7 min per shard is untouched.

Changing only the marginal term moves the optimum *down*, not up — when marginal work is cheap, paying 3.7 min of fixed cost to split it buys less. So the honest expectation is that the right shard count is now **lower than six**, possibly 1–2, and that the previous "six is the ceiling" conclusion no longer follows from anything measured.

## What this needs

Fresh numbers, not a re-derivation. Specifically:

- the new per-preview marginal cost on real CI hardware (the ~36 ms figure in #3540 is an in-process measurement on a container, not an end-to-end CI one);
- whether the ~3.7 min fixed cost per shard still holds;
- the resulting optimum, and whether sharding earns its complexity at all for a 1095-preview catalog.

`design-artifacts.yml` and the reusable workflow in design-parity are where the number lives; the comment block explaining the choice should be updated with whatever the new measurements say rather than left describing a superseded model.

## Blocked on a correctness fix first

Re-enabling sharding at any count is currently unsafe, and not for performance reasons. From the same comment:

> BACK TO ONE, until the exclusion matching is anchored. Sharding cost the catalog three quarters of its renders — silently, on a green run.
>
> `--exclude-preview-id` matches by equality OR substring […] and a base id is a substring of its own variant ids — `…SwitchOn_Light` matches `…SwitchOn_Light_VARIANT_off`. So a shard that excludes some other shard's base preview also deletes every variant of it that it was supposed to render.

267 captured of 1095 assigned, 235 merged, on a green run.

[...]

Issue #3558, opened 2026-08-08
CMP/Wasm Parity gives different verdicts for identical rendering input


yschimke:
The `CMP/Wasm Parity` check produced three different verdicts across six consecutive CI runs of #3557 — a PR that changed **only** `.github/` files (workflows, `reap-pr-runs.mjs`, its test). Nothing in that diff can reach a renderer, sample, catalog, or any Kotlin source, so every one of those runs rendered identical input.

## Evidence

| Run | Head | Verdict |
|---|---|---|
| [31275822205](https://github.com/yschimke/compose-ai-tools/actions/runs/31275822205) | earlier commit | 0 regressions, 27 unchanged |
| [31276165454](https://github.com/yschimke/compose-ai-tools/actions/runs/31276165454) | earlier commit | 0 regressions, 27 unchanged |
| [31276795638](https://github.com/yschimke/compose-ai-tools/actions/runs/31276795638) | `fb681f39` | 0 regressions, 27 unchanged |
| [31277419169](https://github.com/yschimke/compose-ai-tools/actions/runs/31277419169) | `db24b285` | **2 improvements**, 25 unchanged |
| [31278058863](https://github.com/yschimke/compose-ai-tools/actions/runs/31278058863) | `67da6ffa` | 2 improvements, 25 unchanged |
| [31278742208](https://github.com/yschimke/compose-ai-tools/actions/runs/31278742208) | `36f0dbc4` | **1 regression** — check red |

The specific movements:

```
VariableWidthRemote_width_320dp_height_240dp_dpi_320:   2.45% → 0.19%   (improvement)
VariableWeightRemote_width_320dp_height_240dp_dpi_320:  1.98% → 0.12%   (improvement)
TypographyRemote_width_320dp_height_240dp_dpi_320:      0.46% → 0.79%   (regression, +0.33 pp vs the 0.25 pp gate)
```

All three are `com.example.designcatalogremotem3.CatalogPreviewsKt` text-rendering fixtures — the backend-sensitive category `docs/design/RC_CMP_WASM_PLAYER.md` already flags as tolerance-dependent (L415, L427–429, L528–529).

## Two distinct causes, and the second is the real one

[...]

PR #3570, opened 2026-08-08
fix(design-artifacts): anchor every generated render exclusion

yschimke:
Second half of the #3559 blocker. #3561 taught the three matchers to understand `=<id>`; this teaches the pipeline to *emit* it, which is what actually makes sharding safe to re-enable.

## The bug, on the producing side

`--exclude-preview-id` matches a plain pattern by equality **or substring**, and preview ids are hierarchical, so a base id is always a substring of its own fan-out. The sharder expresses each shard's share as "every id that isn't mine", so a shard excluding another shard's `SwitchOn_Light` also deleted `SwitchOn_Light_VARIANT_off` — which it was itself assigned.

m3-catalog: **267 captured of 1095 assigned, 235 merged, on a green run**, hence `render-shards: 1`.

The pathological case is not rare, it is the common one:

```
shard 2 renders  SwitchOff_Light_VARIANT_on, SwitchOn_Light_VARIANT_off
shard 2 excludes SwitchOff_Light,            SwitchOn_Light
```

Every id it excludes is a substring of an id it renders, so unanchored it renders **nothing at all** — and reports success. That case is now a test.

## The fix

Anchor at the one boundary where an id becomes a CLI pattern:

- `shard-preview-ids.mjs` — the `--out` exclusion list (partition + deferred).
- `deferred-preview-ids.mjs` — a new `--anchored-out`, for the unsharded path that passes deferred ids straight to `--exclude-preview-id`.

Ids stay **plain** everywhere else, which is the reason this is two outputs rather than one: the sharder removes deferred ids from the partition by `Set.has`, and `verifyShardPlans` compares `previews` against discovered ids. Anchoring either would silently break the thing it was meant to protect — the deferral would stop competing-for-slots correctly, and the merge-side cross-check would match nothing.

## The unsharded path had it too

Not a sharding-only bug. Mode deferral is also a generated exclusion list, so deferring `Switch_Dark` also deferred `Switch_Dark_VARIANT_off` — a preview the spec never deferred, absent from the sheet, surfacing later as a completeness-gate failure naming a component with nothing pointing back at the mode filter. Fixed in both callers (`design-artifacts.yml` and the reusable workflow).

## Old CLIs

`cli-version: catalog` can pin a release that predates `=`. The two paths differ in what that costs, so they differ in how loud they are:

| path | old CLI does | response |
|---|---|---|
| sharded | anchored pattern matches nothing → each shard renders everything → merge picks winners, catalog is fine but the fan-out bought nothing… **or**, without the fix, silently publishes a quarter of the catalog | **hard error** up front, beside the existing `bundle merge` probe |
| unsharded | deferred modes render too — every sticker still baked, run just slower | `::warning` |

A capability probe beats a version comparison here: `cli-source: build` and `cli-version: catalog` can both land on a CLI whose version string says little about what it supports.

## Also

Corrected the stale cost model in the sharder's header — it still presented `render_minutes ≈ 3.7 + 2.15s × previews` as current guidance, which #3548 invalidated. It now says the marginal term collapsed, the fixed term didn't, the optimum therefore moved *down*, and #3559 is re-measuring. No count is changed here.

## Test plan

```
cd scripts/design-artifacts && node --test
```

**578 tests, 554 pass, 21 skip, 3 fail — all 3 pre-existing** (`package-version`, `rc-compare-pixels`, `rc-size-modifier` need `npm ci`; `Cannot find package 'pngjs'`). Verified identical on a clean `git stash` of this branch.

New coverage:

- *every emitted exclusion is anchored, so a shard cannot delete its own variants* — asserts the property as the matcher sees it (no anchored pattern names an id this shard renders), then pins the concrete shard-2 case above and checks that the **unanchored** form really would have matched all of its own work, so the test fails if the anchoring is removed.
- *the plan's previews stay PLAIN ids while its exclusions are patterns* — plus `verifyShardPlans` still passing on that plan, pinning the two-shapes decision.
- *anchorExclusions is idempotent and drops nothing*.
- deferred: *the CLI form of a deferred id is anchored*, *the plain ids stay plain, because the sharder matches them by set membership*.

Every `run:` block in both workflows bash-syntax-checked with `${{ }}` stubbed out; both files parse as YAML.

No visual surface changes — this decides which previews a filter selects, not how any of them are drawn.

## Not in this PR

Re-enabling sharding or changing the count. That is the rest of #3559 and it wants measurements, not a guess; `render-shards: 1` stays where it is until there are numbers. What changes here is that turning it back up is no longer unsafe.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uk2xtVbDfuqQP2Bkzdndz9)_

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -333,17 +333,32 @@ jobs:
           catalog-path: ${{ inputs.catalog-path }}
           github-token: ${{ github.token }}
 
-      - name: Check the CLI can merge shard renders
+      # Two capabilities, one probe, because a CLI can have the first without the second and each
+      # failure mode is silent in its own way:
+      #
+      #  - `bundle merge` reassembles the shard bundles. Without it the fan-out runs to completion
+      #    and dies at the last step, having burned every shard's ~20 minutes.
+      #  - `=<id>` anchored exclusions (#3561) are what make a generated exclusion list safe. Without
+      #    them `--exclude-preview-id` falls back to substring matching, and since ids are
+      #    hierarchical each shard also deletes the variants of every id it excluded — m3-catalog
+      #    captured 267 of 1095 assigned previews that way, on a GREEN run. An old CLI here does not
+      #    fail; it silently publishes a three-quarters-empty catalog, which is strictly worse than
+      #    refusing to start, hence the hard error rather than a warning.
+      - name: Check the CLI can shard and merge
         if: ${{ inputs.cli-source == 'released' }}
         run: |
           set -euo pipefail
           version="$(compose-preview --version | head -1)"
-          if compose-preview bundle help 2>&1 | grep -q 'bundle merge'; then
-            echo "✓ $version supports 'bundle merge'."
-            exit 0
+          help="$(compose-preview bundle help 2>&1 || true)"
+          if ! grep -q 'bundle merge' <<<"$help"; then
+            echo "::error title=CLI too old for render-shards::${version} has no 'compose-preview bundle merge', which the sharded render needs to reassemble the shard bundles. Raise cli-version (or, with cli-version: catalog, the ${{ inputs.catalog-key }} pin in ${{ inputs.catalog-path }}) to a release that carries it, or set render-shards: 1."
+            exit 1
           fi
-          echo "::error title=CLI too old for render-shards::${version} has no 'compose-preview bundle merge', which the sharded render needs to reassemble the shard bundles. Raise cli-version (or, with cli-version: catalog, the ${{ inputs.catalog-key }} pin in ${{ inputs.catalog-path }}) to a release that carries it, or set render-shards: 1."
-          exit 1
+          if ! grep -q '=<id>' <<<"$help"; then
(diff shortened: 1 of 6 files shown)


yschimke:
`CMP/Wasm Parity` is red here and I am not fixing it in this PR — it cannot be caused by this diff, and it is #3558.

The job's own log shows what it compares. The corpus is fetched from the delivery branch, not built from this PR:

```
git fetch --depth=1 origin design-artifacts/remote-m3
git show FETCH_HEAD:bundle/bundle.png > $RUNNER_TEMP/corpus/bundle.png
node rc-compare.mjs --bundle .../corpus/bundle.png --player cli/src/main/resources/rc-player/bundle.js --cmp-wasm rc-player/wasm/build/wasmDist
```

So the inputs are the delivery-branch bundle, the RC player and the wasm player. This PR touches `scripts/design-artifacts/{shard,deferred}-preview-ids.mjs` and two workflow files — nothing on that list, and nothing that renders a pixel.

The two rows that moved are both variable-font text:

| preview | baseline | here |
|---|---|---|
| `VariableWeightRemote` | 0.12% | 1.98% |
| `TypographyRemote` | 0.46% | 0.79% |

A 1.86 pp swing on a variable-weight specimen between two runs of identical input is the signature in #3558. Supporting evidence in the same log: `VariableWidthRemote` sits at **2.45%** and `TypefaceSpecimenRemote` at **1.08%** without being flagged, because they were already above the line in the baseline — the whole variable-font family is unstable, and which members cross the 0.25 pp delta on a given run is the coin-flip #3558 describes.

The check is advisory (`This check does not block merging`), so the correct action is to read it and decide, which is this comment. Fixing it means fixing the non-determinism, which belongs in #3558.

---

PR #3583, opened 2026-08-09
fix(serve): stop offering declared themes a replayed preview can never apply

yschimke:
Picking a Typeface theme on <https://preview.coo.ee/remote-m3/> turns the whole grid into "This preview can't render live".

## What's happening

The render lane is answering honestly. Every `remote-m3` preview emits a Remote Compose document, so `bundle pack` carries the captured `ir/<id>.rc` and drops the module bytecode that authored it (`BundlePreviewTask`: *"an IR preview contributes no module bytecode"*). The bundle daemon therefore redraws by **replaying that document**, never by re-running the composable — and a declared theme is a `PreviewWrapperProvider` wrapped around a *composition*. There is nothing to wrap, so the server refuses:

```
$ curl -sID '``https://preview.coo.ee/remote-m3/render/template-watchscreen__ideal__default__compact.png?themeProvider=com.example.designcatalogremotem3.RemoteGoogleSansFlexThemeCatalog'``
HTTP/2 409
x-compose-preview-dropped-overrides: themeProvider

override not applied: themeProvider — this preview is replayed from its captured
document, which cannot be recomposed, so the override can never apply
```

409 is terminal by design (#3449, #3531) — serving the unchanged snapshot under a 200 would tell a diff bot the theme made no difference to an artifact that was never rendered.

The bug is that the **landing grid offers the chips anyway**. `themeOverridable` gated on renderability alone — *does this preview have a daemon twin?* — which `remote-m3` passes: it is fully live, it just can't recompose. #3531 taught the **viewer** this exact distinction (`data-ir-replay`, greying the provider options) and stopped at the viewer; the grid was never given the same fact, so one click fires a `?themeProvider=` render per card and every one comes back 409.

## The change

Feed the landing the same predicate the refusal is derived from (`droppedOverridesAreTerminal`), per preview, as `irReplayFor`. Read at the call site rather than re-derived, so grid, viewer and refusal can't drift.

Per-preview, deliberately: a mixed session withholds the chips on exactly its replayed cards. A catalog whose previews all replay keeps its baked **Light/Dark** axis — a document *can* defer that to the host and resolve it at paint time — and drops only the composition-only choice.

## Visual evidence

**Before** — chips offered, every card 409s (this is the existing `theme-render-terminal` harness state, i.e. what `remote-m3` does in production today):

Landing grid with declared-theme chips; every card shows "This preview can't render live"

**After** — same catalog, IR-replayed: Light/Dark stays, the Typeface chips are withheld, cards keep their pixels:

Landing grid with only the Light/Dark pair and no error cards

## What this does not do

It does not make the published `remote-m3` themable. That needs a document captured **per declared theme** at pack time — nothing in the replay path can substitute a typeface after the fact. Recorded in `RemoteThemeCatalogs.kt` beside the providers, which now say where the Theme select does and doesn't offer them (a live session against the module still does). The two synthesised specimen sheets remain the published catalog's evidence of both faces.

## Test plan

- `ServeLandingIrReplayThemeTest` (new) — two fake hosts over real HTTP, both fully live, differing only in whether they carry a `.rc` document. Asserted through the server because the defect was the wiring, not the page builder: reverting `irReplayFor` to `{ false }` fails it, and the page-level tests still pass.
- `ServeWebFixtureTest` — new committed fixture `serve-landing-ir-replay-themes.html` and its assertions (baked pair kept, `theme:` chips and `themeBase` gone).
- Registered in `STYLED_FIXTURES` so the visual-diff bot shoots the withheld control on every future PR — an absent chip is a purely visual claim, and captured unstyled the toolbar isn't even laid out.
- `./gradlew :cli:test ktfmtCheckAll` green; harness captures above produced with `npx playwright test -c preview-harness/playwright.config.mjs snapshot`.

---


yschimke:
The CMP/Wasm parity regression is pre-existing on `main`, not from this branch — and this branch is one row *better*, not worse.

`main` @ [`7677a3f`](https://github.com/yschimke/compose-ai-tools/actions/runs/31304700014/job/93223141775) (the tip this branch forked from), same job, same baseline:

```
CMP/Wasm parity vs baseline: 2 regression(s), 0 improvement(s), 25 unchanged
✗ VariableWidthRemote:  mismatch grew 2.26 pp — 0.19% → 2.45%
✗ VariableWeightRemote: mismatch grew 1.86 pp — 0.12% → 1.98%
```

This PR reports one of those two (`VariableWidthRemote`, identical 0.19% → 2.45%); `VariableWeightRemote` landed under the 0.25 pp gate this time, so the difference between the runs is variance in the variable-font axis rendering, not the diff.

Why it surfaced here at all: the only file this PR touches under `samples/design-catalog-remote-m3/` is a **KDoc block** in `RemoteThemeCatalogs.kt` — no bytecode change — but that path is what pulls the RC parity job into CI scope. On the recent `main` merges it was simply out of scope. Nothing in the diff reaches a render path: the code changes are `ServeWeb` / `ServeHttpServer` HTML generation, tests, a fixture, and docs images.

The drift itself is the delivery branch going stale. The baseline is `rc-compare-summary.json` on `design-artifacts/remote-m3`, and renderer/plugin merges deliberately don't regenerate the delivery branches — the Monday cron and the release chain pick that up. Both `wght` and `wdth` axis stickers moving together points at the variable-font path having changed since that branch was last published, which is a separate question from this PR and the reason the check is marked non-blocking.

No action taken here. Worth a look on its own if the axis rendering genuinely moved rather than the baseline aging out.

---

yschimke:
**Do not merge as-is — the rule this encodes is wrong.** Correcting my own premise.

A replayed Remote Compose document *can* be re-themed. `remote-m3` carries its scheme as named remote state, and the player's override family reaches it. Strings from the published `.rc` documents:

| Sticker | Named colour state |
|---|---|
| `template-watchscreen` | `USER:WearM3.background`, `.surfaceContainer`, `.onSurface`, `.tertiary` |
| `theme-colorscheme` | `USER:WearM3.primary`, `.surfaceContainer`, `.onBackground` |
| `card` | `USER:WearM3.surfaceContainer`, `.onSurfaceVariant` |
| `button-filled` | `USER:WearM3.primary`, `.onPrimary` |

Demonstrated against production, on the replay lane, with no recomposition:

```
GET /remote-m3/render/template-watchscreen__ideal__default__compact.png
      ?rcPlayer=cmp-jvm&rc.WearM3.background=color:%23FF0000   → 200, pixels changed
```

The watch face comes back with a red background. `setNamedColorOverride` is already wired end to end here — `StoreBackedRemoteContext`, `RcPlayer.namedColorOverrides`, `RcJvmRenderer`'s `RcSeed.ColorValue`, and serve's own `rc.<name>=` seeds.

So `CatalogLiveRouting.irReplayDroppedOverrideNames` dropping `themeProvider` for *any* IR-replayed preview is too wide — the exact failure its own doc block warns about for `fontScale`/`uiMode` ("an axis earns its place only by having no representation in the document at all"). `themeProvider` has a document-side counterpart for colours. This PR's `irReplayFor` gate inherits that error and would withhold a colour theme that would have worked.

What is genuinely inexpressible on replay is the **typeface**: the player's named-override family is string / float / int / bool / colour, with no named font override, and the family is resolved into the text ops at capture.

Two further findings while checking:

1. The published per-theme token sets are **byte-identical** for both providers, and both report `roboto-flex` — including `RemoteGoogleSansFlexThemeCatalog`'s. #3573 exports them by reflecting `androidx.wear.compose.material3.MaterialTheme`, which these Remote providers never install (the catalog KDoc already flags that caveat for the specimen sheets). The colour half being identical is correct for these two — they are typeface-only — but the typography half is wrong, and the export can't be the colour source for a themed replay until it reflects the theme actually installed.

2. Because both `remote-m3` themes are typeface-only, themed replay will correctly apply nothing for these two chips. The capability is worth having; it does not by itself make this catalog's Theme control move pixels.

Reworking toward: seed a theme's colour tokens onto the document's `USER:<prefix>.<token>` names on the replay lane, and narrow the drop to what the document genuinely cannot express, rather than gating on the lane.

---

yschimke:
`VS Code Extension E2E (interactive-e-g)` is failing on `main`, not because of this branch.

| Where | Head | Result | Shape |
|---|---|---|---|
| `main` | `b7840f8` ([run 31306811909](https://github.com/yschimke/compose-ai-tools/actions/runs/31306811909/job/93228386333)) | failure | ~3.5 min, fail-fast cancels the sibling shards |
| this PR | `0709804`, attempts 1 and 2 | failure | scenario E hangs after `composePreviewDiscover`, 30-min mocha timeout |

Same shard, same scenario (`E. rename a @Preview function; old id + PNG drop, new id surfaces, index pruned`). Both attempts here stalled at the identical point — 09:11:52 → 09:37:46 and 09:44:22 → 10:10:00, with `F` and `G` passing in ~7s each either side — so it is deterministic, not a flake. I re-ran once to check; I haven't re-run again, since a second identical stall is evidence rather than noise.

Nothing in this diff reaches it: the scenario drives `:samples:cmp` through the Gradle plugin and the extension runtime, and the changes here are `ServeWeb`/`ServeHttpServer` HTML generation, CLI tests, a committed serve fixture, a Playwright spec this Electron/mocha suite doesn't run, docs images, and a KDoc block.

One observation for whoever picks the `main` failure up: #3568 added `assertRefreshRendered` and relaxed the per-task Gradle cap. On `main` that produces the intended fast failure. On a cold-cache branch the same underlying "render never completes" lands inside `composePreviewDiscover` with nothing to cut it short, so it consumes the full 30-minute per-test budget instead of reporting. The hang looks like the real defect; the fast failure is the part that is already visible.

---

Issue #3579, opened 2026-08-09
figma-svg fidelity sweep: four differences found across the published catalogs


yschimke:
Swept `design-artifacts/{compose-m3,wear-m3,remote-m3}` (regenerated 2026-08-08 from `e6eebcf9`) by serving each branch and harvesting `compare.html`'s in-browser SSIM headlessly, then re-diffing the worst rows against a 1:1 rasterisation of each SVG.

Four differences, plus one problem with the report itself.

## Status

| # | finding | state |
| --- | --- | --- |
| 0 | `compare.html` under-reports raster-bearing SVGs | open, diagnosed |
| 1 | Wear `CompactButton` paints the touch target | fixed (#3580) |
| 2 | `TextField/Outlined` draws its value twice | fixed (#3580) |
| 3 | Wear `TransformingLazyColumn` morph not applied | fixed (#3585) |
| 4 | Text metrics drift in both catalogs | open — the obvious fix is wrong, see below |
| — | variant states export identically to their base | open, unrelated |

## 0. `compare.html` under-reports raster-bearing SVGs

The page scores each SVG by drawing it into an `<img>`-backed canvas. That blocks external `<image href>` children, and the blob URL it builds gives them no base to resolve against either — so any SVG using the raster fallback scores low regardless of fidelity. `remote-m3`'s `WidgetContainer/Small` reads 95.2% on the page and 0.4% pixel-difference when rasterised through `<object>`.

`<object>` is not a drop-in fix: it can't be read back from canvas. Scoring these correctly means inlining the referenced rasters as `data:` URIs into the SVG text before making the blob.

## 1. Wear `CompactButton` paints the touch target — fixed in #3580

SVG `rect y=16 h=96` against a render whose pill is `y=32..95`, i.e. 48dp instead of 32dp. Two independent misses: `hasMinimumInteractiveSize()` keyed on a modifier named `minimumInteractiveComponentSize`, which Wear does not use, and Wear pads through `Modifier.padding(PaddingValues)` whose `PaddingValuesElement` exposes no per-edge `Dp`, so `paddingInsets` resolved null and no `paintInset` was captured either. 24.6% → 5.1%.

## 2. `TextField/Outlined` draws its value twice — fixed in #3580

[...]

PR #3585, opened 2026-08-09
fix(figma-svg): fade a paint fill with its content when the alpha trails it

yschimke:
Addresses finding 3 of #3579 — the Wear `TransformingLazyColumn` morph.

A `TransformingLazyColumn` item fills through `Modifier.paint` (Wear's `surface()` helper) and carries the morph's alpha *behind* it in the chain:

```
graphicsLayer(1.0) · fillMaxWidth · TLCCompositeParentData · defaultMinSize ·
fillMaxWidth · paint · graphicsLayer(0.5551) · combinedClickable · padding
```

The export classified any alpha sitting after the first drawing modifier as content-only, so each card kept a full-strength background while its label faded — which is what the scaling-list stickers were mostly differing by.

Content-only is the *right* answer for `Modifier.background`, which draws its rect and only then delegates, and `graphicsLayerAfterBackgroundNestsOnlyTheInnerContent` pins that — so I didn't loosen the rule for every trailing alpha. Wear's `SurfaceTransformation` is the exception: it paints the surface already faded. That is measured, not inferred — at the item whose alpha is 0.5551 the render draws its `#332E3C` card at `(28,25,33)`, i.e. the base colour times the alpha:

| item | alpha | render fill | base × alpha |
| --- | --- | --- | --- |
| 1 | 1.0 | (51,46,60) | (51,46,60) |
| 2 | 0.9577 | (48,44,57) | (49,44,57) |
| 3 | 0.5551 | (28,25,33) | (28,26,33) |

So the rule is keyed on the fill coming from `paint` rather than from `background`.

## Visual evidence

Render PNG (left) vs exported SVG (right), 1:1. Before — the second card's background at full strength while its label fades:

scaling list before

After — background and label fade together:

scaling list after

## Test plan

Catalog-level, from a local `compose-preview bundle pack --module :samples:design-catalog-wear-m3 --with-semantics`, every preview's `compose-figma.svg` diffed against its own render:

| preview | before | after |
| --- | --- | --- |
| `ScalingListSticker_Extra Large Round` | 9.97% | **3.87%** |
| `ScalingListSticker_Large Round` | 7.46% | **4.11%** |
| `TimeTextScaffoldTemplate_Large Round` | 7.46% | **4.11%** |
| `EdgeButtonSticker_Extra Large Round` | 3.05% | **1.95%** |
| `TitleCardSticker` (control, text-only) | 5.36% | 5.36% |
| `CompactButtonSticker` (control) | 5.12% | 5.12% |

The two controls are unchanged, so the change moved only the items carrying a morph alpha. What's left on the moved rows is the text-metric gap (finding 4), untouched here.

Unit coverage:

- `FigmaLayeredSvgTest.graphicsLayerAfterAPaintFillFadesTheFillWithTheContent` — new; fails on `main`.
- `graphicsLayerAfterBackgroundNestsOnlyTheInnerContent` and `finalFrameRasterDoesNotReapplyLocalOrAncestorOpacity` unchanged and green — the `background` case and the frame-crop case keep their existing behaviour.
- `./gradlew :renderer-android:testDebugUnitTest :data-layoutinspector-core:test :data-layoutinspector-connector:test ktfmtCheckAll` — green.

## Note on the other open findings

**Finding 4 (text metrics) is diagnosed but not fixed, and the obvious fix is wrong.** `googleFontsWeight` snaps Compose's arbitrary weights to the nearest 100-step instance, so Wear's 550 and 450 runs embed 600 and 500 faces. Requesting the exact weight does work — CSS2 serves `wght@550` — but for a non-100-step weight Google returns the **variable** font, and `FontSubsetter` (FontBox) has no variable support, so it flattens to the default instance: the embedded face came back with no `fvar` at all, and 450 and 550 resolved to the same 43 KB source. That trades one wrong weight for another, so I reverted it. A real fix needs the font instanced at the requested axis values before subsetting. Recorded on #3579.

Also on #3579 and not addressed here: `compare.html` scores each SVG by drawing it into an ``-backed canvas, which both blocks external `<image href>` children and gives a blob URL no base to resolve them against — so raster-bearing SVGs are under-reported. Fixing it means inlining those rasters as data URIs before scoring, since an `<object>` can't be read back from canvas.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XrscSvCccWovuNg44XxmRm)_


yschimke:
Both review comments are right, and together they sink this approach. Closing rather than patching it — the discriminator is wrong, not just too broad.

I probed both cases through real Compose rather than reasoning about them:

**A trailing alpha does not fade a plain `paint` fill.** `Box(Modifier.size(40.dp).paint(ColorPainter(Color.Red)).graphicsLayer { alpha = 0.5f })` renders its centre pixel at `(255,0,0)` — fully opaque red. So `paint` behaves exactly like `background` here, and promoting the alpha to the group is wrong for every ordinary painter fill, as the first comment says.

**The painter's own alpha would be applied twice.** `Modifier.paint(painter, alpha = 0.5f)` captures as `bg=#80FF0000` with `alphas=[(2, paint=0.5)]` — `painterColorHex` has already folded the alpha into the fill colour, *and* the paint modifier's own index lands it in `contentOpacity` (its index equals `firstDraw`, and the filter is `>=`). Promoting that to the group gives an effectively 25%-opaque fill.

So what actually fades the Wear card? Not the trailing layer. The item carries two lambda-form graphicsLayers:

```
idx=0  graphicsLayer  alpha=1.0     keys=[alpha, block]
idx=6  graphicsLayer  alpha=0.9577  keys=[alpha, block]
```

The measured morph alpha is attributed to the trailing one, but a trailing layer demonstrably cannot fade the `paint` above it — so either the attribution or the ordering model is wrong for this chain. `SurfaceTransformation` drives these blocks per frame, so the static read and the live value need not agree, and the leading layer (which *would* fade the fill, being outer) reports 1.0.

The render-side facts still stand and are worth keeping: the fill genuinely fades with the alpha — at the item whose morph alpha is 0.5551 the render draws its `#332E3C` card at `(28,25,33)` — and the export leaves those backgrounds opaque. So finding 3 is real; my mechanism for it was not. Re-opened on #3579 with this diagnosis so the next attempt starts from which layer owns the alpha rather than from which modifier paints.

---

PR #3593, opened 2026-08-09
test(vscode): pair the preview filter with missingRenders=warn, and bound every wait

yschimke:
Scenario E ("rename a `@Preview` function") was the residual flake in the `interactive-e-g` shard, and the cause of both post-fix failures on `main` ([31306997859](https://github.com/yschimke/compose-ai-tools/actions/runs/31306997859), [31305131855](https://github.com/yschimke/compose-ai-tools/actions/runs/31305131855)).

This PR does two things, and **does not make scenario E pass** — the remaining failure turned out to be a poisoned build-cache entry, which no change to this file can fix. See "What this does not fix" below.

## Fix 1 — the filter needs `missingRenders=warn`

The suite hands every `gradlew` invocation `-PcomposePreview.filter=…`, narrowing the render to the 6 previews the scenarios assert on. But `composePreviewRenderAll` finishes with a completeness check across the **whole** manifest, and `composePreview.missingRenders` defaults to `fail`:

```
composePreviewRenderAll: render produced no output file for 59 of 66 preview(s)
```

So the render task was failing on **every refresh this suite has ever made**, including the pre-rename baseline. Every scenario only ever exercised `renderWithDiskFallback` — the degraded path that paints the on-disk manifest — never the happy path.

`ci.yml`'s bundle job already pairs a narrowed render with `-PcomposePreview.missingRenders=warn` for exactly this reason. The suite didn't. It does now, and `FIXTURE_PREVIEW_FILTER`'s doc comment states the pairing is mandatory so it can't be dropped again.

Effect, measured: `interactive-e-g` went from a 30-minute bare Mocha timeout to a **2-minute** run, with F and G passing and the baseline refresh reaching the real render path (`Rendered 6 preview(s)`) for the first time.

## Fix 2 — bounded, self-describing waits

A bare `Timeout of 1800000ms exceeded` is not an acceptable failure mode: three quite different bugs collapse into one indistinguishable symptom (`setPreviews` never posted / renamed id never appeared / old id survived), at 30 minutes of runner time each.

- `waitFor` gains an optional `describeState`, rendered into the timeout message: the latest cmp `setPreviews` and its ids, old/new id presence, `getLastWarmDaemonError()`, and 40 lines of the Compose Preview output channel — which carries the `> <task> completed` / `FAILED` / `cancelled` Gradle lines the in-repo e2e otherwise routes to the extension log and discards.
- `refreshWithinBudget` bounds the awaited `triggerRefresh` itself. `PANEL_UPDATE_BUDGET_MS` only bounds the *poll after* the refresh resolves, which does nothing for a refresh that never resolves — and that is precisely what run 31306997859 was.
- The panel's two terminal empty states are matched explicitly and fail immediately with the message text, rather than waiting out a budget for a republish that is never coming.
- Budgets are clamped against the test's remaining Mocha time (`budgetWithin`). A fixed budget isn't enough on its own: E runs four bounded refreshes in one 30-minute test and 4 × `REFRESH_BUDGET_MS` is 48 minutes, so a wedge in a *late* refresh would blow the ceiling before its own timer fired.
- Scenarios F and G are bounded the same way, sharing a generic `describeModulePanelState`.

**Deliberately no retry.** A second forced refresh would very likely make E pass, and "edits not updating" is precisely the failure this suite exists to catch.

## What this does not fix

Scenario E (and scenario B, in `interactive-a-d`, which is red on `main` too) fail on a **poisoned remote build-cache entry**, not on anything in this file. Reproduced with plain Gradle in a clean container — no VS Code, no extension, no daemon:

| same renamed source | `compileKotlin` | `.class` files | `previews.json` |
|---|---|---|---|
| build cache on | `FROM-CACHE` | **0** | **3** |
| `--no-build-cache` | executed | 46 | **66** |

The cache serves an empty `:samples:cmp:compileKotlin` output. Restoring it empties the class directory; `DiscoverPreviewsTask` scans `classDirs`, finds nothing, and writes an asset-only manifest from `resourceDirs`; every filter pattern then matches nothing and `composePreviewRender` fails. That explains the 1-second CI build, why it only bites after a source edit, and why `interactive-a-d` is red on `main` with no PR involved.

Remediation (purging the entries, making the tasks non-cacheable, or narrowing the cache-push policy) is a separate decision and is not attempted here.

## Test plan

- `npm run compile:host` — clean.
- `npm test` — 1674 passing, unchanged (this file is e2e-gated, so the unit suite only proves it still compiles).
- `npm run format:check` — clean.
- `interactive-e-g` on this PR: F and G pass; E fails in ~2 minutes with the manifest state printed, which is how the cache entry was found.

Scenarios A–D still pass `this.timeout()` as a wait budget. Left alone — they run *before* E, and converting latent slow-but-passing waits into failures in the same change would muddy the signal.

No visual surface changes — test harness only, so text-only evidence is the right bar.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VmbwB3s3xBmk4Q4RWBjpjM)_

### DIFF
--- a/vscode-extension/src/test/electron/realGradleApi.ts
+++ b/vscode-extension/src/test/electron/realGradleApi.ts
@@ -10,6 +10,11 @@ import type { GradleApi } from "../../gradleService";
  */
 const KILL_GRACE_MS = 5_000;
 
+/** One enqueued-but-not-yet-spawned `gradlew` invocation. */
+interface QueuedInvocation {
+    cancelled: boolean;
+}
+
 /**
  * Real {@link GradleApi} that shells out to the repo's `./gradlew` wrapper.
  * Used by the daily/manual e2e suite to drive the *actual* Gradle plugin
@@ -69,7 +74,95 @@ export class RealGradleApi implements GradleApi {
         return this.cancelledTaskNames;
     }
 
+    /**
+     * Tail of the invocation chain. Every {@link runTask} links onto it, so at
+     * most one `gradlew` process is alive at a time.
+     *
+     * Production reaches Gradle through `vscjava.vscode-gradle`, which talks to
+     * one long-lived daemon, and `gradleService.ts` leans on that: several of
+     * its comments say a second invocation simply "queues behind it on the
+     * daemon's serialised queue". Shelling out to `./gradlew` breaks that
+     * assumption — a client that finds the daemon busy starts *another* one, so
+     * the invocations genuinely run in parallel.
+     *
+     * That is not theoretical. On run 31310881459 the suite's forced
+     * `composePreviewRenderAll` and a save-loop `composePreviewCompile` landed
+     * 0.6s apart on `:samples:cmp`, two daemons started, and the compile
+     * rewrote the module's class output while the other build's
+     * `composePreviewDiscover` was scanning it. Discovery scans `classDirs`, so
+     * the manifest collapsed from 66 previews to the 3 asset-only ones that
+     * come from `resourceDirs`, and every `-PcomposePreview.filter` pattern
+     * then matched nothing:
+     *
+     *     composePreviewRender --preview matched no previews … Available previews:
+     *       lottie/spin.json
+     *       svg/badge.svg
+     *       svg/star.svg
+     *
+     * Scenarios B and E both edit a `.kt` fixture on disk and so both hit it.
+     *
+     * Serialising here restores the one-daemon invariant the extension is
+     * written against. It does **not** fix the underlying product race — two
+     * concurrent Gradle clients against one project can corrupt each other's
+     * task outputs whatever drives them — that belongs in `GradleService`.
+     */
+    private queueTail: Promise<void> = Promise.resolve();
+
+    /**
+     * Invocations that have been enqueued but not yet spawned, by
+     * `cancellationKey`. {@link cancelRunTask} can only kill a live child;
+     * without this, a supersession that arrives while a task is still waiting
+     * its turn would be dropped and the stale build would run anyway — exactly
+     * the "leftover previews from the cancelled refresh" that scenario D pins.
+     */
+    private readonly queued = new Map<string, QueuedInvocation[]>();
+
     runTask(opts: {
+        projectFolder: string;
+        taskName: string;
+        args?: ReadonlyArray<string>;
+        showOutputColors: boolean;
+        onOutput?: (output: {
+            getOutputBytes(): Uint8Array;
+            getOutputType(): number;
+        }) => void;
+        cancellationKey?: string;
+    }): Promise<void> {
+        const key = opts.cancellationKey;
+        const entry: QueuedInvocation = { cancelled: false };
+        if (key) {
+            const waiting = this.queued.get(key);
+            if (waiting) waiting.push(entry);
+            else this.queued.set(key, [entry]);
+        }
+        const run = this.queueTail.then(() => {
+            if (key) {
+                const waiting = this.queued.get(key);
+                if (waiting) {
+                    const at = waiting.indexOf(entry);
+                    if (at >= 0) waiting.splice(at, 1);
+                    if (waiting.length === 0) this.queued.delete(key);
+                }
+            }
+            if (entry.cancelled) {
+                // Worded so `gradleService`'s CANCELLED_RE maps it to a
+                // TaskCancelledError, same as a SIGTERM'd child.
+                throw new Error(
+                    `gradlew ${opts.taskName} cancelled before it started`,
+                );
+            }
+            return this.spawnTask(opts);
+        });
+        // Never let a rejection break the chain — the next invocation still has
+        // to run, and `run` already carries the failure to its own caller.
+        this.queueTail = run.then(
+            () => undefined,
+            () => undefined,
+        );
+        return run;
+    }
+
+    private spawnTask(opts: {
         projectFolder: string;
         taskName: string;
         args?: ReadonlyArray<string>;
@@ -189,6 +282,18 @@ export class RealGradleApi implements GradleApi {
         // keyed by `cancellationKey`. Without a key there's nothing
         // specific to cancel (gradleService always supplies one).
         const key = opts.cancellationKey;
+        // Tasks still waiting their turn have no child to signal — mark them so
+        // they abort when they reach the head of the queue instead of running a
+        // build the caller has already superseded. Not recorded in
+        // `cancelledTaskNames`: nothing was truncated, so a suite reading that
+        // list for partial renders would be misled.
+        const waiting = key ? this.queued.get(key) : undefined;
+        if (waiting?.length) {
+            for (const entry of waiting) entry.cancelled = true;
+            this.onLog(
+                `[realGradleApi] cancel ${opts.taskName} (key=${key}) — ${waiting.length} queued invocation(s) dropped before spawn`,
+            );
+        }
         const child = key ? this.liveChildren.get(key) : undefined;
         if (!child || child.pid === undefined || child.exitCode !== null) {
             return;
--- a/vscode-extension/src/test/electron/realGradleApi.ts
+++ b/vscode-extension/src/test/electron/realGradleApi.ts
@@ -10,11 +10,6 @@ import type { GradleApi } from "../../gradleService";
  */
 const KILL_GRACE_MS = 5_000;
 
-/** One enqueued-but-not-yet-spawned `gradlew` invocation. */
-interface QueuedInvocation {
-    cancelled: boolean;
-}
-
 /**
  * Real {@link GradleApi} that shells out to the repo's `./gradlew` wrapper.
  * Used by the daily/manual e2e suite to drive the *actual* Gradle plugin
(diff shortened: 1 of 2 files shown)


yschimke:
`interactive-e-g` is red on this PR, and that is expected — flagging it explicitly rather than leaving it to be read as a regression.

**The shard was already red on `main` before this PR.** Both prior failures ([31306997859](https://github.com/yschimke/compose-ai-tools/actions/runs/31306997859), [31305131855](https://github.com/yschimke/compose-ai-tools/actions/runs/31305131855)) are the same scenario, and it failed the same way in the original 2026-08-08 breakage. This PR does not make it redder; it makes the identical failure fast and legible.

**What the change bought, measured on this PR's own runs:**

| | before | after |
|---|---|---|
| shard wall time | ~30 min | **2 min** |
| failure message | `Timeout of 1800000ms exceeded` | the actual cause, below |

```
Error: after the rename the panel resolved to its empty state instead of republishing:
  "No @Preview functions in this file (3 in other files in this module)."
  latest cmp setPreviews: <none posted>
  > Task :samples:cmp:composePreviewRender FAILED
  > composePreviewRender --preview matched no previews for
    'com.example.samplecmp.RedBoxPreview*', 'com.example.samplecmp.BlueBoxPreview', …
```

**What that tells us.** After the rename, `Previews.kt` yields **zero** discoverable previews and the module manifest comes back with only 3 — so `composePreviewRenderAll` hard-fails (a filter matching nothing is fatal for that task, as `FIXTURE_PREVIEW_FILTER`'s own comment warns), `refresh` settles into the panel's empty state, posts no `setPreviews`, and returns `'completed'` — which is why `assertRefreshRendered` never caught it.

**Why I am not fixing the underlying defect in this PR.** I ruled out the obvious cause: nothing else in the module references `RedBoxPreview` (`grep` finds only the declaration), so the rename does not break compilation. Why discovery collapses to 3 module-wide needs either a local repro — which needs the Android SDK, xvfb and a VS Code host I do not have here — or a targeted CI experiment, and it is a different question from the harness problem this PR set out to fix.

There are two candidate fixes and they are not equivalent:

- **Test-side** — force a compile before the post-rename render so the filter always has something to match. Narrow, but it blunts the scenario's ability to catch exactly the "edits not updating" race it was written for.
- **Product-side** — `composePreviewRender` treating "filter matched no previews" as fatal is arguably wrong mid-edit. A user renaming a preview today gets "No @Preview functions in this file" instead of their renamed preview, which is a real bug this scenario has been quietly catching all along.

I lean product-side, but it changes plugin behaviour other paths depend on, so it is worth a deliberate decision rather than my guess.

---

yschimke:
The `missingRenders=warn` fix worked, and the new diagnostics immediately paid for themselves — but they surfaced a deeper bug that this PR does not yet fix. Recording it here as the blocker.

## What improved

`interactive-e-g` went from a 30-minute bare Mocha timeout to a **2-minute** run with a named failure. F and G both pass now, and the baseline refresh reaches the real render path for the first time:

```
composePreviewRender: 6 capture(s) drawn on a warm renderer (no per-capture JVM fork).
Rendered 6 preview(s)
composePreviewRenderAll: missing-renders policy=warn — … 59 of 66 preview(s)
[refresh] rendered 6 preview(s) for Previews.kt
```

## What it exposed

Both remaining failures are the same bug, and both happen **immediately after the test writes to `Previews.kt`**:

* `interactive-e-g` — scenario E, post-rename refresh:
  ```
  > composePreviewRender --preview matched no previews for '…RedBoxPreview*', … Available previews:
      lottie/spin.json
      svg/badge.svg
      svg/star.svg
  [refresh] done — 0 visible previews in Previews.kt, module has 3
  ```
* `interactive-a-d` — scenario B, post-edit refresh: `:samples:cmp render: refresh reported 'failed'`.

`DiscoverPreviewsTask` scans compiled `classDirs`; the three survivors are asset previews from `resourceDirs`. So the module's **class output was empty when discovery ran** — the manifest collapsed from 66 previews to 3, and every filter then matches nothing.

Why it's empty: two `gradlew` processes are running against `:samples:cmp` at once.

```
11:31:38.7  gradlew :samples:cmp:composePreviewRenderAll   ← the test's forced refresh
11:31:39.3  gradlew :samples:cmp:composePreviewCompile     ← the daemon save loop, from the file watcher
…
Starting a Gradle Daemon (subsequent builds will be faster)
Starting a Gradle Daemon (subsequent builds will be faster)   ← two of them
```

`gradleService.ts` assumes concurrent invocations queue up behind each other ("the daemon serialises clients"). That holds for one daemon. Here the second invocation can't reuse the busy daemon and **spawns its own**, so the compile rewrites `samples/cmp`'s class directory while the other build's `composePreviewDiscover` is scanning it.

Note the same signature — `module has 3` — is present in the original 2026-08-08 failure, so this race is the root cause the timeout was hiding all along, not a regression from this PR. What this PR changed is that the render no longer fails on *every* refresh, so the collapse now surfaces as a hard failure instead of being absorbed by `renderWithDiskFallback`.

Next step is a design call between serialising at the harness level and serialising in `GradleService`; picking that before pushing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VmbwB3s3xBmk4Q4RWBjpjM)_

---

yschimke:
Pushed `b96b49e` — the harness now serialises its `gradlew` invocations, so the render and the save-loop compile can no longer overlap and clobber `:samples:cmp`'s class output mid-discovery.

Two details worth flagging for review:

* **Queued cancellation.** `cancelRunTask` could only kill a live child. With a queue in front of it, a supersession that arrives while a task is still waiting would have been dropped and the stale build would have run late — which is exactly what scenario D pins. Queued invocations are now marked cancelled and abort when they reach the head, worded so `gradleService`'s `CANCELLED_RE` maps them to a `TaskCancelledError` like a SIGTERM'd child. They are deliberately *not* added to `cancelledTasks`: nothing was truncated, and `e2e.test.ts` reads that list to warn about partial renders.
* **The continuous-compile worker bypasses this.** It spawns `gradle --continuous` directly rather than through `GradleApi`, so it can't wedge the queue — and equally isn't serialised by it. Not exercised by these scenarios.

The product race is untouched and is the more interesting half: `GradleService` assumes concurrent invocations serialise on a single Gradle daemon, and a `./gradlew` client that finds the daemon busy just starts another one. Two clients against one project can corrupt each other's task outputs whatever drives them. Worth its own issue rather than smuggling a `GradleService` change into a test-harness PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VmbwB3s3xBmk4Q4RWBjpjM)_

---

yschimke:
**My concurrency diagnosis was wrong.** Reverted the serialisation in `0c0b326`.

The run on `b96b49e` shows the invocations cleanly serialised — and scenario E fails identically:

```
12:14:34.9  gradlew :samples:cmp:composePreviewRenderAll   ← baseline: renders 6, manifest has 66
12:15:00.2  gradlew :samples:cmp:composePreviewRenderAll   ← post-rename: BUILD FAILED in 1s, manifest has 3
12:15:02.4  gradlew :samples:cmp:composePreviewCompile     ← after, not overlapping
```

No overlap, no second daemon, same three asset previews. So the compile was never clobbering the class directory; the earlier 0.6s overlap was a coincidence of timing, not the cause.

What the serialised run does show more sharply: the post-rename build takes **1 second**. Nothing recompiles and nothing rediscovers — the render task reads an already-on-disk manifest that by then contains only the three `resourceDirs` assets. So the manifest is being reduced to 3 somewhere between the successful baseline render and the next `composePreviewRenderAll`, and the only thing that happens in that window is the test's `fs.writeFileSync` on `Previews.kt`. Finding what writes it is the next step, and I don't want to keep guessing at it through CI.

Reverting rather than keeping-and-fixing, for two reasons: it buys nothing against the actual bug, and @chatgpt-codex-connector's note on it is correct — `GradleService.runTask` starts its 10-minute cap *before* calling into the API, so queued time would eat the execution budget and could manufacture exactly the truncated renders the change was meant to prevent. Not worth carrying a real defect for a fix that doesn't fix anything.

What stands on this PR: the `missingRenders=warn` pairing (which is what took this shard from a 30-minute bare timeout to ~2 minutes and got F and G passing) and the bounded, self-describing waits. Scenario E still fails, but now in 2 minutes with the manifest state printed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VmbwB3s3xBmk4Q4RWBjpjM)_

---

yschimke:
## Root cause found, and it is not in this repo's test code at all

Reproduced with plain Gradle in a clean container — no VS Code, no extension, no daemon, no concurrency:

```
$ ./gradlew :samples:cmp:composePreviewDiscover        # pristine source
previews: 66

$ # rename RedBoxPreview -> RedBoxPreviewRenamed, change nothing else
$ ./gradlew :samples:cmp:composePreviewDiscover
BUILD SUCCESSFUL in 4s
> Task :samples:cmp:compileKotlin FROM-CACHE
> Task :samples:cmp:composePreviewDiscover
previews: 3      ['lottie/spin.json', 'svg/badge.svg', 'svg/star.svg']

$ find samples/cmp/build -name '*.class' | wc -l
0
```

Control, same renamed source:

| | `compileKotlin` | `.class` files | `previews.json` |
|---|---|---|---|
| build cache on | `FROM-CACHE` | **0** | **3** |
| `--no-build-cache` | executed | 46 | **66** |

**The remote cache is serving an empty `:samples:cmp:compileKotlin` output.** Restoring it empties the class directory; `DiscoverPreviewsTask` scans `classDirs`, finds nothing, and writes an asset-only manifest from `resourceDirs`; every `-PcomposePreview.filter` pattern then matches nothing and `composePreviewRender` fails.

This accounts for every observation that didn't fit the earlier theories:

* the CI build takes **1 second** — nothing compiles, it's a cache hit;
* serialising the harness changed nothing — there was never a race;
* it only bites *after* a source edit — the pristine state hits a good entry, the edited state hits the poisoned one;
* scenario B fails the same way — it also edits `Previews.kt`;
* `interactive-a-d` is red on `main` too, with no PR involved.

The entry cannot have come from this container: the first local build here used pristine source. It came from the remote cache.

## Consequences

1. **Scenario E cannot pass while the cache serves an empty compile output.** The `missingRenders=warn` pairing and the bounded diagnostics on this PR are correct on their own merits — that combination is what took this shard from a 30-minute unattributed timeout to a 2-minute failure that prints the manifest state, and it is how this got diagnosed. But E stays red until the cache entry is dealt with.

2. **This is a correctness problem on `main`, not a test problem.** Any consumer resolving that key gets an empty compile. `#3588` widened cache *pushing* to many workflows, so a poisoned entry now propagates to every reader — and the e2e suite, which renames the function and whose compiles are routinely SIGTERM'd, is the obvious producer.

Deliberately not acting on (2) unilaterally: purging the poisoned entries, marking `compileKotlin` / `composePreviewDiscover` non-cacheable, and revisiting the `#3588` push policy are materially different calls, and `#3588` is already merged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VmbwB3s3xBmk4Q4RWBjpjM)_

---

PR #3596, opened 2026-08-09
fix(plugin): floor the render classpath's compose-ui line so the renderer links

### LINKED ISSUE #3590: fix(renderer): every authored @Preview fails with NoSuchMethodError getApplyOnDeactivatedNodeAssertion on 0.19.49
## Summary

On `yschimke/home-assistant-android`, **every** authored `@Preview` in both catalogued modules fails to render on compose-ai-tools **0.19.49**, with:

```
java.lang.NoSuchMethodError: 'kotlin.jvm.functions.Function1 androidx.compose.ui.node.ComposeUiNode$Companion.getApplyOnDeactivatedNodeAssertion()'
```

This looks like a Compose UI ABI skew between the Compose runtime on the render classpath and the Compose the app module was compiled against — `getApplyOnDeactivatedNodeAssertion` is a newer `ComposeUiNode.Companion` accessor, so an older `compose-ui` is being loaded under a newer compiled app (or the reverse).

The failure is total, not partial: 13/13 catalogued phone components and 22/22 catalogued Wear components produced no render.

## Impact

The run still went **green** and force-pushed an empty catalog over a good one, so both published delivery branches are now unusable:

- `design-artifacts/home-assistant-android` — `catalog.json` went from **13 components → 0**
- `design-artifacts/home-assistant-wear` — **22 components → 0**

`preview.coo.ee` consequently reports `catalog had no usable images` for both systems (`ServeCatalogStore.kt:431`), and the published `index.html` pages render empty. The only images that survived are synthetic manifest-activity captures, which the specs deliberately exclude from catalog entries.

Separate follow-ups filed for the two reasons this shipped silently rather than failing: #3591 (`--allow-incomplete` should fail on a total render failure) and #3592 (surface failed renders in the UI).

## Reproduction

Failing run: https://github.com/yschimke/home-assistant-android/actions/runs/31270371437

Both jobs `success`, both with `Render catalog bundle` reported as success.

The trigger commit is a **one-line version bump** with no source, spec, or workflow change:

```
f04727b4  Merge pull request #4 — chore(deps): update compose-ai-tools to 0.19.49
 gradle/libs.versions.toml | 2 +-
```

| Renderer | Date | Phone components | Wear components |
| --- | --- | ---: | ---: |
| 0.19.38 | 2026-08-05 | 13 | 22 |
| 0.19.49 | 2026-08-08 | **0** | **0** |

So the regression is somewhere in `v0.19.38..v0.19.49`.

## Suspect

`a73a88e1b fix(plugin): scope Rule 3's Compose exclusion to our own render dependencies` (#3483, released in 0.19.46) is the commit in that range that changes which Compose artifacts reach the render classpath. Not confirmed — it is the strongest candidate by inspection, and bisecting 0.19.45 / 0.19.46 against this repo would settle it.

Worth noting the affected repo uses `embed-deps: true` and `split-mode: view-only`, which may be why this shows up here and not in the sample catalogs on the same renderer.

## Evidence from the log

Stack frames land in our own render engine:

```
at Apply(ComposeDataExtensionHooks.kt:366)
at Around(PreviewOverridesDataProduct.kt:83)
at Around(ComposeDataExtensionHooks.kt:457)
at <lambda>(RenderEngine.kt:466)
[daemon semantics] render failed for '…ComponentPreviewsKt.PreviewMdcAlertDialog': NoSuchMethodError: …
```

And the driver's own summary of the damage:

```
##[warning][home-assistant-android] missing renders for: Onboarding/Welcome, Onboarding/Notification permission,
Settings/Wear companion, Onboarding/Matter commissioning, Settings/Quick settings tiles, Settings/App shortcuts,
Settings/Android Auto favorites, Settings/Sensor list, Settings/Widgets, Settings/Wi-Fi SSIDs, Settings/External URL,
Settings/Notification detail, Components/Icon dialog

##[warning][home-assistant-wear] missing renders for: Home/Landing, Home/Entity list, Home/Sensors,
Entities/Detail panel, Tiles/Shortcut selection, Tiles/Shortcut order, Tiles/Camera selection, Tiles/Template settings,
Tiles/Refresh interval, Tiles/Camera settings, Sensors/Manager, Settings/Set favorites, Settings/Favorites,
Settings/Complication, Onboarding/Install phone app, Conversation/Speech bubbles, Entities/Rows,
Entities/Selection chip, Controls/Toggle states, Sensors/Rows, Components/List header, Components/Time text
```

## Expected

An authored `@Preview` that compiled against the module's own Compose version renders on the renderer that module is pinned to. A Compose ABI mismatch on the render classpath should be detected and reported as a classpath problem, not surface as a per-preview `NoSuchMethodError` on every entry.

yschimke:
Fixes #3590.

## Summary

Rule 3 hands the Android render classpath to the consumer's Compose, but nothing ever checked that Compose is new enough for **our own bytecode** to link against it.

`RENDERER_COMPOSE_FLOOR_VERSION` (1.9.5) reads like it covers this, but it is only ever stamped on plugin-injected coordinates with no version source of their own — `ui-test-manifest`, `ui-test-junit4`, `runtime-tracing`. Nothing pins `ui` / `foundation` / `runtime` / `animation`, which are the artifacts that actually carry the symbols the renderer calls. A consumer below the floor therefore handed the renderer an unlinkable classpath, and every preview died before user code ran:

```
java.lang.NoSuchMethodError: 'kotlin.jvm.functions.Function1
 androidx.compose.ui.node.ComposeUiNode$Companion.getApplyOnDeactivatedNodeAssertion()'
   at ee.schimke.composeai.daemon.RenderEngine…(RenderEngine.kt:2900)
```

`yschimke/home-assistant-android` is on `compose-bom = "2025.01.00"` (ui line 1.7.6) and lost **13/13** phone and **22/22** Wear catalog entries, republishing both design-artifacts branches with `"components": []`. That is the whole catalog, on a one-line renderer bump, reported green.

## What this does

Adds one `eachDependency` rule to `applyRenderGraphResolutionRules` — the same shape as the hamcrest-core rule already there, and for the same class of reason (our bytecode needs a method the resolved artifact lacks). It raises **only** the compose-ui version line to `RENDERER_COMPOSE_LINK_FLOOR_VERSION`.

The floor is set to `RENDERER_COMPOSE_CMP_RUNTIME_VERSION` (**1.11.2**) deliberately: that is what the Rule-3-off path already renders against, so it is the one Compose line the renderer is known-good on. It is also what home-assistant-android rendered against on 0.19.38–0.19.45, when our CMP transitives still supplied the render Compose.

Scoped tightly:

- **Only the ui line** (`animation`, `foundation`, `runtime`, `ui`). `material` / `material3` version independently — there is no `androidx.compose.material3:material3:1.11.2` — so a bare `androidx.compose.` prefix would resolve to versions that do not exist.
- **Only upward.** A consumer at or above the floor keeps its own version, so Rule 3's documented symmetry still holds for every consumer that is not already broken.
- **Only when comparable.** `+`, `1.9.+`, ranges and blanks are left alone: forcing a version we cannot order risks dragging a consumer *backwards*, which is worse than a classpath we merely suspect is old.

## Reviewer attention please — the resource-APK coupling

The `RenderGraphComposeExclusionTest` header documents #3447: the merged unit-test resource APK is built from the *consumer's* graph, so an upgraded Compose can mean bytecode looking up `R.id` fields the resources never declared (`NoSuchFieldError: androidx.compose.ui.R$id`).

This change raises the ui line on the render configuration, so that hazard is in scope for consumers below the floor. Two reasons I still think this is the right move, but it wants a second opinion:

1. `mainVariantComposeVersion` decides the resource APK's Compose and returns `RENDERER_COMPOSE_FLOOR_VERSION` for a Compose-bringing consumer. Aligning it with the link floor is the obvious follow-up, **but** it cannot see a BOM-driven consumer's version at configuration time (home-assistant-android declares `platform(compose-bom)` plus an unversioned `ui`, so the declared version string is null). The resolution-time rule here is the only place the real version is visible.
2. The affected population is exactly the consumers who currently render **nothing**, so the realistic downside is trading a total failure for a different failure — while consumers above the floor are untouched by construction.

The e2e matrix (`ComposeStarter`, `WearTilesKotlin`, `cadence`, `meshcore-mobile`) is the real gate here, since those cover both the #3447 and #3483 shapes that Rule 3 exists to protect.

## Test plan

- New `ComposeLineFloorTest` covers the decision function: below-floor raised, at/above untouched, material/material3 and non-Compose groups skipped, alpha-of-floor treated as below, uncomparable versions never touched.
- `RenderGraphComposeExclusionTest` still passes unchanged — the exclusion behaviour is not touched, only the version that survives it.
- `./gradlew :gradle-plugin:test --tests '*ComposeLineFloorTest*' --tests '*RenderGraphComposeExclusionTest*'` — **BUILD SUCCESSFUL** locally.
- Real verification is a `home-assistant-android` render on a build with this in: pass condition is 13 phone / 22 Wear components, not zero. That repo is pinned to 0.19.45 as a stopgap and now has `workflow_dispatch` on its design-artifacts workflow, so unpinning and re-rendering is a one-click check once this ships.

No visual surface changes — this is build wiring, so there is no before/after render to embed. The renders it *restores* are the 35 catalog entries above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01141C3L3ahTgzWxEvjWBF7U)_

### DIFF
--- a/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
+++ b/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
@@ -369,6 +369,85 @@ internal object AndroidPreviewSupport {
   internal const val RENDERER_COMPOSE_CMP_RUNTIME_VERSION: String = "1.11.2"
 
   /**
+   * Lowest `androidx.compose` version on the **compose-ui version line** that the renderer's own
+   * bytecode links against. Rule 3 defers the render classpath to the consumer's Compose (see
+   * [applyRenderGraphResolutionRules]); this is the floor under which deferring stops being safe.
+   *
+   * Distinct from [RENDERER_COMPOSE_FLOOR_VERSION], which is only ever stamped on plugin-injected
+   * coordinates that have no version source of their own (`ui-test-manifest`, `ui-test-junit4`,
+   * `runtime-tracing`). Nothing pinned `ui` / `foundation` / `runtime` / `animation`, so a consumer
+   * below this floor handed the renderer a classpath its bytecode cannot link against and every
+   * preview died identically:
+   * ```
+   * NoSuchMethodError: 'kotlin.jvm.functions.Function1
+   *   androidx.compose.ui.node.ComposeUiNode$Companion.getApplyOnDeactivatedNodeAssertion()'
+   * ```
+   * (issue #3590 — `yschimke/home-assistant-android` on `compose-bom` 2025.01.00 lost all 13 phone
+   * and 22 Wear catalog entries, republishing both design-artifacts branches empty.)
+   *
+   * Set to [RENDERER_COMPOSE_CMP_RUNTIME_VERSION] deliberately: that is the version the Rule-3-off
+   * path already renders against, so it is the one Compose line the renderer is known-good on.
+   * Keep the two in lockstep.
+   */
+  internal const val RENDERER_COMPOSE_LINK_FLOOR_VERSION: String =
+    RENDERER_COMPOSE_CMP_RUNTIME_VERSION
+
+  /**
+   * The `androidx.compose.*` groups that share the compose-ui version line (1.7.6 / 1.9.5 / 1.11.2
+   * move together). Deliberately NOT `androidx.compose.material` / `material3`, which version
+   * independently — material3 has no 1.11.2 — nor a bare `androidx.compose.` prefix, which would
+   * sweep them in.
+   */
+  private val COMPOSE_UI_LINE_GROUPS =
+    setOf(
+      "androidx.compose.animation",
+      "androidx.compose.foundation",
+      "androidx.compose.runtime",
+      "androidx.compose.ui",
+    )
+
+  /**
+   * [RENDERER_COMPOSE_LINK_FLOOR_VERSION] if [group] is on the compose-ui line and [version] is
+   * below the floor, else `null` (leave the dependency alone).
+   *
+   * Kept pure and separate from the `eachDependency` wiring so the decision is unit-testable without
+   * resolving a configuration. An unparseable or dynamic version (`+`, `1.9.+`, a range, empty)
+   * returns `null`: we only ever raise a version we can prove is too low, since forcing one we
+   * cannot compare risks dragging a consumer *backwards*.
+   */
+  internal fun composeLineFloorUpgrade(group: String, version: String?): String? {
+    if (group !in COMPOSE_UI_LINE_GROUPS) return null
+    val current = version?.takeIf { it.isNotBlank() } ?: return null
+    return if (isBelowVersion(current, RENDERER_COMPOSE_LINK_FLOOR_VERSION)) {
+      RENDERER_COMPOSE_LINK_FLOOR_VERSION
+    } else {
+      null
+    }
+  }
+
+  /**
+   * Whether [candidate] orders below [floor], comparing dotted numeric components left to right and
+   * treating a pre-release suffix as lower than the same numbers without one (`1.11.2-alpha01` <
+   * `1.11.2`), which matches how AndroidX ships alphas ahead of a stable.
+   *
+   * Returns `false` for anything non-numeric or dynamic — see [composeLineFloorUpgrade] for why
+   * "cannot compare" must mean "do not touch".
+   */
+  private fun isBelowVersion(candidate: String, floor: String): Boolean {
+    val candidateBase = candidate.substringBefore('-')
+    val floorBase = floor.substringBefore('-')
+    val candidateParts = candidateBase.split('.').map { it.toIntOrNull() ?: return false }
+    val floorParts = floorBase.split('.').map { it.toIntOrNull() ?: return false }
+    for (index in 0 until maxOf(candidateParts.size, floorParts.size)) {
+      val left = candidateParts.getOrElse(index) { 0 }
+      val right = floorParts.getOrElse(index) { 0 }
+      if (left != right) return left < right
+    }
+    // Numerically equal: a pre-release of the floor is still below it.
+    return candidate.contains('-') && !floor.contains('-')
+  }
+
+  /**
    * Version for the main-variant `androidx.compose.ui` / `foundation` pins, which decide the R
    * class in the merged unit-test resource APK. See the call site for why the two cases differ: a
    * Compose consumer's own BOM wins over our floor anyway, while a Compose-less consumer runs
@@ -521,6 +600,21 @@ internal object AndroidPreviewSupport {
    * (`alignDesktopToolWithConsumerGraph`).
    */
   internal fun applyRenderGraphResolutionRules(configuration: Configuration) {
+    configuration.resolutionStrategy.eachDependency {
+      val target = composeLineFloorUpgrade(requested.group, requested.version)
+      if (target != null) {
+        useVersion(target)
+        because(
+          "Rule 3 hands the render classpath to the consumer's Compose, but the renderer's own " +
+            "bytecode still has to link against it. Below " +
+            "$RENDERER_COMPOSE_LINK_FLOOR_VERSION the compose-ui line lacks entry points the " +
+            "renderer calls and every preview dies before user code runs: " +
+            "NoSuchMethodError androidx.compose.ui.node.ComposeUiNode\$Companion" +
+            ".getApplyOnDeactivatedNodeAssertion(). Raise to the floor; consumers already at or " +
+            "above it keep their own version (issue #3590)."
+        )
+      }
+    }
     configuration.resolutionStrategy.eachDependency {
       val req = requested
       val targetName = kmpAndroidSiblingName(req.group, req.name)
--- a/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
+++ b/gradle-plugin/src/main/kotlin/ee/schimke/composeai/plugin/AndroidPreviewSupport.kt
@@ -1924,9 +1924,18 @@ internal object AndroidPreviewSupport {
         // `floorComposeLine` follows `manageDependencies`: the floor only holds if the matching
         // main-variant `ui`/`foundation` pins move with it, and the opt-out branch deliberately
         // skips those ("consumer must ensure androidx.compose.ui:ui is on the main variant").
-        // Raising the render graph there would put 1.11.2 classes over the consumer's own
-        // resources — the #3484 `R$id` NoSuchFieldError — so opt-out keeps the consumer's line and
-        // the required-dependency check reports it instead.
+        // Raising the render graph there would put floor-version classes over the consumer's
+        // own resources — the #3484 `R$id` NoSuchFieldError — so opt-out keeps the consumer's
+        // line, whatever it is.
+        //
+        // KNOWN GAP: an opt-out consumer whose Compose is BELOW the floor still hits #3590's
+        // NoSuchMethodError, and nothing reports it — `validateExternallyManagedDependencies`
+        // checks that the required coordinates are DECLARED, never what they resolve to. Failing
+        // that validation on a below-floor compose-ui line is the fix, and it wants the resolved
+        // version rather than the declared one (a BOM consumer declares no version at all). Same
+        // root as the `enforcedPlatform` case: whenever the resource graph cannot be moved to the
+        // floor, the honest outcome is a configuration-time failure naming the required version,
+        // not a silently-raised render graph.
         applyRenderGraphResolutionRules(this, floorComposeLine = manageDependencies)
       }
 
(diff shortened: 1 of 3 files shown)


yschimke:
## Correction to the PR body, and a datapoint that strengthens the floor choice

The body claims the floor version is "what home-assistant-android rendered against on **0.19.38–0.19.45**". That range is wrong — only **0.19.38** is proven good. I pinned that repo to 0.19.45 as a stopgap and the render still published 0 components, with a completely different failure:

```
java.lang.NoClassDefFoundError: androidx/compose/material3/ColorScheme
   at ee.schimke.composeai.daemon.RenderEngine$render$statement$1.evaluate$lambda$1$0$2(RenderEngine.kt:442)
```

That is #3483 — the config-wide Rule 3 exclusion dropping our injected `androidx.compose.material3`. home-assistant-android is a Material2 app with no `material3` of its own, so it had nothing to fall back on.

The real picture for that consumer:

| Renderer | Result |
| --- | --- |
| 0.19.38 | 13 / 22 components — last known-good |
| ~0.19.39 – 0.19.45 | `NoClassDefFoundError: material3/ColorScheme` (#3483) |
| 0.19.46 – 0.19.51 | `NoSuchMethodError: getApplyOnDeactivatedNodeAssertion` (#3590) |

**This does not change the fix, and I think it argues for it.** `a73a88e1b` was a genuine fix — scoping the exclusion to our own subtree is what keeps material3 on the classpath for a Material2 consumer. Reverting it would trade #3590 for #3483. The only way out is forward, which is what this PR does: keep the scoped exclusion and add the version floor it was missing.

The floor value itself is unaffected — 1.11.2 is `RENDERER_COMPOSE_CMP_RUNTIME_VERSION`, justified by being the version the Rule-3-off path renders against today, independently of what any one consumer historically resolved.

One extra note for reviewers on the #3447 resource-APK question: home-assistant-android carries **no `androidx.compose.material3`** and takes Material2 at 1.7.6 from its BOM. This PR raises only `ui`/`foundation`/`runtime`/`animation` to 1.11.2, leaving `compose-material` at 1.7.6 — Material2 compiled against an older ui running on a newer ui, which is the forward-compatible direction, but it is the specific combination the e2e matrix should be read against.

---

yschimke:
## Addressed the P1 — main variant now moves on the same floor

Codex is right, and this was the open question I flagged in the PR body. Fixed in `150a13d5e`.

Flooring only the render graph left the merged unit-test resource APK on `RENDERER_COMPOSE_FLOOR_VERSION` (1.9.5) while 1.11.2 classes ran against it — trading #3590's `NoSuchMethodError` for #3484's `NoSuchFieldError` on `androidx.compose.ui.R$id`. This repo's own test already documents the exact receipt: *"`ui-android` 1.9.5's `R.txt` has no such id, 1.11.2's does"*.

`mainVariantComposeVersion` now returns `RENDERER_COMPOSE_LINK_FLOOR_VERSION` in both branches, so resources and bytecode move together or not at all.

### On the concern I had about it being unresolvable

I'd written that the main-variant side "cannot see a BOM-driven consumer's version at configuration time", which is true but turned out not to matter — it doesn't need to. The pin already relies on Gradle's max-version conflict resolution, so returning the floor unconditionally is raise-only by construction:

- consumer **above** the floor → its own Compose line wins, untouched;
- consumer **below** → raised to the floor, which is exactly when the render graph is raised too.

That also makes the two branches coincide, so the `consumerBringsOwnCompose` probe is no longer load-bearing there and the function collapses to the constant. The probe still does its real job in `addRenderGraphDependency`.

### One behaviour change worth naming explicitly

The old code pinned a Compose consumer's main variant at 1.9.5; it is now 1.11.2. So a consumer resolving **between 1.9.5 and 1.11.2** — e.g. the 1.10.6 case in `main-variant compose pin follows whichever compose actually runs` — now gets raised where it previously would not have been.

That is required rather than incidental: the render-graph rule floors *that* consumer to 1.11.2 as well, so its resource APK has to follow. I updated that test's second assertion and its rationale rather than working around it, since the old rationale ("its BOM wins over our floor anyway") only holds while the consumer sits above the pin.

Tests: `ComposeLineFloorTest` 6/6, `RenderGraphComposeExclusionTest` 11/11, both green on a forced re-run. The e2e matrix is still the real gate for the #3447 / #3483 shapes.

---

yschimke:
## Both P1s addressed in `3aa0f750d`

Both were real, and both were ways the floor could be silently undone rather than disagreements about the design.

### 1. Sibling substitution discarded the floor

Correct, and this one was genuinely subtle. Gradle hands every `eachDependency` action the **original** `requested` selector, so the sibling rule's `useTarget(… "version" to req.version)` overwrote the `useVersion` my earlier block had applied. `androidx.compose.ui:ui-jvmstubs:1.9.5` — the exact coordinate the existing sibling tests use — would be floored to 1.11.2 and then re-pinned to `ui-android:1.9.5`, landing on precisely the artifact the floor exists to keep off the graph. Silent, and invisible to a unit test of either rule alone.

Both halves now resolve in a single rule. I pulled the decision out into a pure `renderGraphTarget(group, name, version, floorComposeLine)` so the interaction is testable without resolving a configuration, which is what let me cover it properly rather than just reordering and hoping.

### 2. `manageDependencies=false` raised the graph with no matching resource pin

Also correct. The opt-out branch deliberately skips the main-variant `ui`/`foundation` pins (*"consumer must ensure `androidx.compose.ui:ui` is on the main variant"*), while `applyRenderGraphResolutionRules` runs unconditionally at both call sites — so opt-out consumers below the floor would get 1.11.2 classes over their own older resources, i.e. the #3484 `R$id` `NoSuchFieldError`.

`floorComposeLine` now follows `manageDependencies`. I chose "don't floor in opt-out" over "require the consumer's line to meet the floor" because it matches the opt-out contract everywhere else in that branch — we report what's missing rather than silently managing it. The sibling substitution is unrelated to dependency management and still applies either way.

### Tests

`ComposeLineFloorTest` 6 → **10**, adding: sibling carries the floor; sibling above the floor keeps its own version; opt-out leaves the compose line alone but still substitutes siblings; non-Compose module with no sibling untouched. With `RenderGraphComposeExclusionTest` that's **21 green** on a forced re-run.

Worth noting for anyone reading later: three rounds of review have each found a different path by which the floor gets dropped or de-synced from the resource APK. That is a good argument for the e2e matrix being the merge gate here rather than the unit tests, which is still where I'd want a green before this lands.

---

yschimke:
## Both correct. One fixed, one recorded — and I think the pattern is now the story

### The false claim is fixed (`9c42c7670`)

The opt-out comment said the required-dependency check reports a below-floor consumer. It does not — `validateExternallyManagedDependencies` checks that coordinates are **declared**, never what they resolve to. So `manageDependencies=false` + old Compose still hits #3590 with no diagnostic at all, and my comment asserted otherwise. That was my error and the comment now says what actually happens.

### The `enforcedPlatform` case is real too

`useVersion` on the render configuration overrides an enforced BOM's constraints, while the ordinary `${variantName}Implementation` pin cannot — so the render graph goes to the floor and the resource APK stays on the enforced older line. Same `R$id` `NoSuchFieldError`, reached by a third route.

### The pattern

Four review rounds have found four distinct ways the render graph and the resource APK can end up on different Compose lines:

1. the main-variant pin not moving with the floor,
2. the KMP sibling substitution discarding the floored version,
3. `manageDependencies=false` skipping the main pins entirely,
4. `enforcedPlatform` making the main pin unable to win.

1 and 2 are fixed. 3 and 4 share one shape: **the render graph can be floored, and the resource graph cannot follow.** Patching each route individually is how this keeps sprouting — the mechanism-specific detection is exactly what keeps missing a case.

The invariant is what matters, and it is one line: *whatever Compose the renderer links against, the merged resource APK must be on the same line.* Where we can move both, align them silently (rounds 1–2). Where we provably cannot, the honest outcome is a **configuration-time failure naming the required version** — not a silently-raised render graph that fails later with a stack trace pointing at Compose internals.

That check wants the **resolved** version, not the declared one — a BOM consumer declares no version at all, which is what defeated my earlier attempt to answer this from `mainVariantComposeVersion` — and it should cover both remaining routes at once rather than detecting `enforcedPlatform` specifically.

I've recorded both as a single known gap in the code rather than half-implementing it, because a resolution-time validation hook is a design decision about this plugin's contract with opt-out consumers, and it deserves a deliberate answer rather than a fifth patch from me on a review round. Happy to implement it if you want it in this PR; my read is it is cleaner as a follow-up with its own tests, since what is here now is strictly an improvement for the managed-dependencies path that covers `home-assistant-android` and every other BOM consumer.

---

yschimke:
## `VS Code Extension E2E (interactive-e-g)` is red on `main`, not from this PR

Not fixing it here — it reproduces on the base branch and predates this branch.

**Evidence.** Every recent `main` commit whose run wasn't cancelled fails the same shard:

| `main` commit | interactive-e-g |
| --- | --- |
| `4354d4017` | **failure** |
| `bdc176dc1` | **failure** |
| `b7840f87e` | **failure** |
| `09c74cb12`, `d9df173e4`, `dba138464`, `76e52b00f`, `b0fe7cac3` | cancelled (superseded) |

`4354d4017` fails in the same step (`Run e2e tests under xvfb`) with the same ~33-minute duration, i.e. the same 30-minute mocha timeout.

**Failure shape here:** a timeout, not an assertion —

```
1) Compose Preview interactive scenarios (real Gradle)
     E. rename a @Preview function; old id + PNG drop, new id surfaces, index pruned:
   Error: Timeout of 1800000ms exceeded.
```

F and G then pass in 154s and 10s. The log shows `composePreviewDaemonStart` at 12:28:02 followed by a ~27-minute gap with no further Gradle invocation, so test E is absorbing a cold build and blowing the 30-minute budget; the later tests are fast because the build is warm by then. This run also had the BuildFetch remote cache **read-only** (`ON_CI: false`), which makes the cold leg materially slower.

**Why it isn't this change:** the failing test drives `:samples:cmp` — Compose Multiplatform, on the desktop renderer. This PR only touches `applyRenderGraphResolutionRules`, which is applied to the **Android** render configurations; the desktop side is aligned by `alignDesktopToolWithConsumerGraph` and is explicitly out of Rule 3's scope. Nothing here can add 27 minutes to a CMP daemon start.

The other four E2E shards — `interactive-a-d`, `cmp-smoke`, `cached-preload`, `wear` — all pass on this head.

Worth noting separately: a 30-minute per-test timeout that the first test in a shard can exhaust on a cold build is order-dependent by construction. Whichever test runs first wears the build cost, so which one fails is a function of shard ordering and cache warmth rather than of the code under test. That looks like a CI-shape issue worth its own issue rather than something to chase per-PR.

---

PR #3604, opened 2026-08-09
feat(remote-m3): port the Wear M3 palettes and apply themes at replay, not capture

yschimke:
Brings `remote-m3`'s declared themes up to the Wear sibling's five, and changes *when* a theme is applied: previews are recorded once under the default theme, and a selected theme is seeded onto the recorded document as named values.

## Why the seam moved

`RemoteMaterialTheme`'s scheme is **not** constant-folded into the document. Every role it draws through is emitted as named state — `USER:WearM3.primary`, `USER:WearM3.surfaceContainer`, `USER:WearM3.onSurface` — and the player's `setNamedColorOverride` reaches exactly those. So a theme is expressible on an already-recorded document, with no recomposition.

Recording a theme *in* would bake its colours in as constants: one capture per theme, and a published catalog could only ever show the one it was packed with. One theme-independent document plus a small override map replaces that.

## Evidence — the published catalog, re-themed

These are rendered from the **published** `design-artifacts/remote-m3` documents, whose module bytecode `bundle pack` dropped, by seeding `remoteCatalogThemeColors(<theme>)` onto them:

| | M3 (stock) | Coral | Teal | KotlinConf |
|---|---|---|---|---|
| `Button/Filled` |  |  |  |  |
| `Progress/Circular` |  |  |  |  |

Reproducible against production as it stands today:

```
curl "https://preview.coo.ee/remote-m3/render/button-filled__ideal__default__compact.png\
?rcPlayer=cmp-jvm&rc.WearM3.primary=color:%23FF6F61&rc.WearM3.secondary=color:%23FFB4A9"
```

## What's in it

- **Five themes**, name- and group-matched to `:samples:design-catalog-wear-m3`: M3, Coral, Teal, Google Sans Flex, KotlinConf. Palettes ported role-for-role from `wearColorScheme`.
- **`RemoteSticker` and the widget frame install nothing.** Providers `Wrap` nothing — they exist so `@WearThemeCatalog` has something to annotate, which is what populates the Theme select.
- **Inter vendored** into `:samples:cmp-wasm-catalog`'s `fonts.json` (400/700, SIL OFL-1.1 from the google/fonts corpus). JetBrains Mono was already there. That lane is manifest-only and never fetches, so an unlisted family fails the availability check rather than substituting.

## Limits, stated plainly

- **The typeface half does not apply yet.** A named value carries a colour, float, int or bool — not a face. The typeface is host-side in its own way: stickers emit the built-in default family id and the player picks the face at draw time. `remoteCatalogFont` / `remoteCatalogDisplayFont` are published as data for a lane to configure its resolver from. Until one does, **Google Sans Flex and KotlinConf apply their colours and not their faces.**
- **`?themeProvider=` doesn't expand into these seeds yet** — that's the server half, and it's what makes the Theme select work rather than requiring hand-written `rc.` URLs.
- **Only components drawing `primary`/`secondary` move**, faithful to the sibling's `wearColorScheme`. The watch-screen template paints from `background`/`surfaceContainer`/`onSurface`/`tertiary` and is untouched — verified, not assumed.
- The specimen sheets remain byte-identical across all five themes; they read the Wear `MaterialTheme` reflectively, which these providers never install. Not evidence, and the header says so.

## Test plan

- `:samples:design-catalog-remote-m3:testDebugUnitTest` — `RemoteCatalogThemeTest` pins the declared set against the sibling's, that every override key is a `WearM3.`-prefixed role (a typo silently no-ops at render time, so this is the guard), that M3 and Google Sans Flex override nothing, and the font data.
- `composePreviewRenderAll` + `ktfmtCheckAll` green.
- **Un-themed pixels unchanged: 27/27 byte-identical**, verified by rendering with and without the change in one tree. The published delivery branch is *not* a usable control — it lags renderer drift, so every sticker differs from it for unrelated reasons.
- `CircularProgressRemote`'s `.rc` shifts back to its pre-theme bytes: the id-allocation artefact the old capture-time branch introduced, undone with it. PNG unaffected.

---


yschimke:
Both Codex comments are correct about the no-op wrappers. Fixed one, and the other is real but pre-dates this PR — details below, since the difference matters for who should fix it.

## Live theming (line 162) — fixed in `c3439b9`

A regression I introduced. Moving theming to replay left a recomposing session — `?themeProvider=` on a module-backed daemon — selecting a theme and getting the default one, because nothing installed it.

The wrapper installs the scheme again, derived from the same `remoteCatalogThemeColors` map the replay path seeds. One definition feeding both lanes, deliberately: a theme that recomposed to one palette and replayed to another would be worse than either, and nothing would catch it, because both renders succeed.

The recorded documents are unaffected and stay theme-independent, which is what the replay path needs. That falls out of how they're recorded rather than needing enforcement — `composePreviewRenderAll` renders each `@Preview` with no provider, so the local is null and the un-themed branch is taken. Re-verified: 27/27 stickers byte-identical against a build of this branch without the change.

On the second half of that comment — packed documents can't use the map because `irReplayDroppedOverrideNames` still drops every `themeProvider` — that's right, and it's the next PR. Worth noting the mechanism already works end to end without it; against the published catalog, whose bytecode was dropped at pack time:

```
curl "https://preview.coo.ee/remote-m3/render/button-filled__ideal__default__compact.png\
?rcPlayer=cmp-jvm&rc.WearM3.primary=color:%23FF6F61&rc.WearM3.secondary=color:%23FFB4A9"
```

returns a coral button today. What's missing is `?themeProvider=` expanding into those seeds, plus narrowing the drop. The PR body lists both as known gaps.

## Token sidecars (line 186) — real, but not from this PR

`WearThemeCatalogStrategy` reads `androidx.wear.compose.material3.MaterialTheme` reflectively. A Remote provider installs `RemoteMaterialTheme`, which is a different type — so the sidecar reads the stock Wear theme no matter what the wrapper does. Restoring installation in `c3439b9` does **not** fix this, and capture-time theming didn't either.

Checked against what's actually published on `design-artifacts/remote-m3` today, before any of this:

```
$ diff themes/…remotegooglesansflexthemecatalog.dtcg.json \
       themes/…remoterobotoflexthemecatalog.dtcg.json && echo IDENTICAL
IDENTICAL
```

Both also report `fontFamily: roboto-flex` — including the Google Sans Flex theme's, whose entire purpose is a different face. So the exported metadata already claims the themes are identical, with two of them.

This PR does make it more visible by taking that from two themes to five, and I'd rather not leave that unsaid. But the fix is in the export strategy — it needs a Remote-aware reader, since there is no Wear `MaterialTheme` to reflect on — not in these declarations. I'd rather not fold an export-driver change into a sample PR; happy to take it as the next one after the server wiring if you'd prefer it ordered differently.

---

PR #3608, opened 2026-08-09
feat(serve): apply a declared theme to a replayed preview as named colour seeds

yschimke:
A theme could not reach a preview whose composable was dropped at pack time. `themeProvider` substitutes a `PreviewWrapperProvider` around a composition; a replayed document has none, so the render lane refused it — correctly, but that left a published catalog's Theme select with nothing it could do.

The colours it stands for *can* reach it, so the server expands the request instead of refusing it.

## Why this works

`RemoteMaterialTheme`'s scheme is **not** constant-folded into the document. Every role it draws through is emitted as named state — `USER:WearM3.primary`, `USER:WearM3.surfaceContainer`, `USER:WearM3.onSurface` — and the player's `setNamedColorOverride` rewrites exactly those, on a replayed document, with no recomposition.

So a `themeProvider` on a replayed preview becomes the `rc.<role>=color:…` seeds that mean the same thing, carried by machinery that already exists. Both replay lanes — the daemon and the isolated cmp-jvm player — go through one expansion, so they can't disagree about what a theme is.

Already demonstrable against production, which is what this generalises:

```
curl "https://preview.coo.ee/remote-m3/render/button-filled__ideal__default__compact.png\
?rcPlayer=cmp-jvm&rc.WearM3.primary=color:%23FF6F61&rc.WearM3.secondary=color:%23FFB4A9"
```

returns a coral button from bytes whose module bytecode was dropped at pack time. This PR is what lets `?theme=` do that for you.

## Three properties, each pinned by a test

- **The provider key is consumed, not forwarded.** Leaving it would have `droppedOverridesFor` report an un-applied override on the very render that applied it — refusing a `200` it had earned (#3449).
- **A preview whose lane can recompose keeps its provider untouched.** Re-running the composable applies the whole theme, typeface included, which no named value can carry. Seeding colours there would silently narrow what selecting a theme means.
- **A catalog that declares themes but publishes no mapping keeps refusing.** Rendering default pixels under a theme's name is precisely the failure #3449 exists to prevent; an empty map must not become a silent success.

## Scope, and what it does not do yet

`ServeHost.themeReplayColors` defaults to empty, so **every host behaves exactly as it does today** until one publishes a mapping. Nothing in this PR changes a live deployment's behaviour.

Populating it from the delivery branch is the next change, and it's the interesting one. The obvious route — teach the theme-catalog render strategy to read the Remote theme — turns out to be impossible, which is worth recording because I told Codex the opposite on #3604:

```
public final RemoteColorScheme getColorScheme(Composer, int);
  RuntimeVisibleAnnotations:
    androidx.compose.remote.creation.compose.layout.RemoteComposable
```

`RemoteMaterialTheme`'s accessors are all `@RemoteComposable` — readable only inside remote scope. `WearThemeSpecimen` is a plain `@Composable` sheet outside any document, so no reflection reaches the value; it doesn't exist there. The per-theme token set has to come from the **document** instead: render a sticker under the provider, read the named colour state, diff against the unthemed capture. That same artifact is what feeds `themeReplayColors`, so the token-set fix and this data source are one piece of work rather than two.

Also still open, and unchanged here: the typeface half. It's resolved host-side from the built-in default family id, so it needs the resolver override rather than a named value.

## Test plan

- `ServeThemeReplaySeedTest` (new) — three fake hosts over real HTTP, differing only in whether the preview replays and whether a mapping is published. Asserts the seeds that reach the render lane, that the provider is consumed, that a recomposing lane is untouched, and that an unmapped catalog still gets its terminal 409.
- Full `:cli:test` and `ktfmtCheckAll` green — including the existing IR-replay and landing-gate tests, which this had to keep honest.

Text-only PR: no rendered surface changes, since no host publishes a mapping yet.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHost.kt
@@ -137,8 +137,16 @@ interface ServeHost : AutoCloseable {
    */
   fun themeReplayColors(providerFqn: String): Map<String, String> = emptyMap()
 
-  /** Whether any declared theme can be applied by replay ([themeReplayColors]). */
-  fun canThemeByReplay(): Boolean = declaredThemes.any {
+  /**
+   * The declared themes a **replayed** preview can actually be rendered under — those this host
+   * publishes a [themeReplayColors] mapping for.
+   *
+   * Per theme, not per host: a catalog may publish mappings for some of its themes and not others,
+   * and a theme that moves only typography legitimately has no colours to seed at all. Offering the
+   * whole declared set because *one* of them is mapped puts the unmapped ones back on the terminal
+   * 409 the gate exists to prevent.
+   */
+  fun replayableThemes(): List<ServeTheme> = declaredThemes.filter {
     themeReplayColors(it.providerFqn).isNotEmpty()
   }
 
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHost.kt
@@ -116,6 +116,33 @@ interface ServeHost : AutoCloseable {
   fun canRenderOverridesFor(previewId: String): Boolean = canRenderOverrides
 
   /**
+   * The **named-value overrides that apply a declared theme to an already-recorded document** —
+   * `<name>` to a colour literal (`#RRGGBB`), keyed by the document's own state names.
+   *
+   * This is what lets a theme reach a preview whose composable is gone. A Remote Compose document
+   * emits the roles it draws through as named state (`USER:WearM3.primary` and friends) rather than
+   * folding them into constants, so the player's `setNamedColorOverride` can re-theme a *replayed*
+   * document with no recomposition. Seeding these is that operation, and it is the only route a
+   * theme has on a session that cannot recompose — a published catalog whose module bytecode was
+   * dropped at pack time.
+   *
+   * Empty (the default) means this host publishes no such mapping for [providerFqn], and a
+   * `themeProvider` render of a replayed preview stays the terminal refusal it is today. That is
+   * deliberate: applying nothing and answering `200` would be the #3449 failure — a render that
+   * claims a theme it never applied, indistinguishable from one where the theme changed nothing.
+   *
+   * Only consulted for previews that replay. A session that can recompose applies the provider by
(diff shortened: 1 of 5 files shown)


yschimke:
All four findings were correct — fixed in `2f82935`. Taking them in order of how quietly each would have failed.

## Theme lease admission (P1)

The worst of the four, because it fails silently. `pureThemeProvider` classified the request from the **parsed** overrides, and expansion removes `themeProvider` from them — so an expanded render read as an ordinary seeded one: no `themeRenderLeases.admit`, no rejection of an expired or saturated lease, no serial permit. Nothing errors. Every page sharing a catalog just runs the full advertised concurrency straight at the global render semaphore, and a theme selection turns into a burst of 503s under load.

Classification now comes from the request rather than the overrides: the expansion carries out the provider it consumed, and a request whose only override param was `themeProvider` is still a pure theme render after becoming seeds.

## The viewer (P1)

Right, and I'd left the two surfaces disagreeing — the grid offered the chips while `/p/<id>` greyed them, for a render route that could serve them.

Added `replayThemes` as its own flag rather than softening `irReplay`, per your suggestion. That distinction matters: everything else `irReplay` disables — locale, author knobs, string `rc.` seeds — genuinely still cannot be honoured by a replay, and widening the existing flag would have re-enabled all of it.

## Per-theme availability (P2)

`canThemeByReplay()` returned true on the first mapped provider, so a partially-mapped catalog offered every chip and the unmapped ones hit the terminal 409 the gate exists to prevent. Your typography-only case is the sharper version of it — such a theme legitimately has an empty colour map and would be advertised as replayable forever.

Replaced with `replayableThemes()`, and both pages now offer only what the session can apply.

## Blank explicit seeds (P2)

Correct, and it lands in the worst category: `rc.WearM3.primary=` won precedence via `putIfAbsent`, but `ServeOverrides.parse` skips blank `rc.` values outright, so the role fell back to the document's authored colour while the response reported a themed render — a theme applied in part, claimed in full. Blank now means absent for the merge.

## Tests

Two added to `ServeThemeReplaySeedTest`: the blank-seed precedence over real HTTP, and per-theme narrowing.

The second is worth a note for anyone extending these fakes — I first wrote it as `object : ServeHost by unmapped { … }`, which passes for the wrong reason. `replayableThemes()` is an interface default, so delegation forwards it to the delegate and evaluates it against *that* host's themes, not the overrides in the object expression. It needs a real fake.

`:cli:test` and `ktfmtCheckAll` green.

---

PR #3628, opened 2026-08-10
fix(rc-player): honor document density behavior

yschimke:
## Summary

- decode the Remote Compose document density-behavior header
- apply AndroidX alpha16 density semantics exhaustively to row, column, flow, collapsible layouts, padding, offset, marquee, rounded clipping, borders, and dimension constraints
- preserve the distinct `DimensionIn` rule and the legacy pre-1.1 border-width behavior
- preserve component modifier wire order (the complete change from merged PR #3630)
- add codec, operation-matrix, and integrated high-density layout coverage

## Merge audit

PR #3630 was merged into this feature branch rather than `main`. Its modifier-order commit is now intentionally included here, rebased with the density work onto current `main`. Merging this PR lands both changes; nothing from #3630 remains stranded.

## Home Assistant validation

Rerendered all 122 captured `homeassistant-remotecompose` documents with the updated Wasm player:

- mean pixel mismatch: **19.14% → 7.23%**
- mean coverage delta: **12.85% → 2.94%**
- Sensor: about **33% → 1%** mismatch
- VerticalStack: **34–36% → 0.8–3.5%**
- Markdown, Calendar, Logbook, Light, Statistics Graph, Area, and responsive Tiles now land around **0.5–2%**

Grid improves substantially but retains a separate content/color issue. Toggle is unchanged and is being investigated separately.

## Validation

- `./gradlew :rc-player-protocol:desktopTest :rc-player-compose:desktopTest --tests '*RcDocumentCodecTest*' --tests '*RcComposeSupportTest*'`
- `./gradlew :rc-player-compose:desktopTest --tests '*RcLayoutRenderTest.legacyFlowSpacingRemainsInPhysicalPixelsAtHighDensity*'`
- `./gradlew :rc-player-compose:ktfmtCheck :rc-player-protocol:ktfmtCheck`
- complete 122-document `rc-compare` rerender
- after rebasing onto current `main`: `./gradlew :rc-player-protocol:desktopTest :rc-player-runtime:desktopTest :rc-player-compose:desktopTest :rc-player-wasm:wasmPlayerDist`

PR #3633, opened 2026-08-11
fix(rc-player): preserve fractional pixel padding totals

yschimke:
## Summary

- retain Remote Compose padding as physical pixels through measurement
- round the combined horizontal and vertical inset once, matching the float-based AndroidX layout model
- add a high-density Flow regression with four exact-fit cards and 31.5px edges

## Why

Home Assistant's Grid is 819px wide and emits four 189px cards separated by three 21px gaps: exactly 819px. Each card is a 126px body with 31.5px padding on both sides. Compose `Dp` padding rounds each edge to 32px, inflating every card to 190px and wrapping the fourth card. AndroidX accumulates the float padding before layout, so the pair remains 63px.

## Validation

- `./gradlew :rc-player-compose:desktopTest :rc-player-compose:ktfmtCheck`
- 108 Compose player tests pass
- focused regression verifies the fourth card stays on the first Flow row at density 2.625

## Stack

Merge #3628 first. PR #3630 was merged into #3628, so #3628 now supplies both the density semantics and modifier wire ordering required here. This branch has been rebased onto the repaired #3628 head and passes Compose desktop tests plus Wasm compilation.


yschimke:
Superseded by #3636, which contains this change plus the remaining Remote Compose operation fixes on a clean base of .

PR #3635, opened 2026-08-11
fix(rc-player): preserve AndroidX weighted row spacing

yschimke:
## Summary

- reproduce AndroidX `RowLayout` weight measurement when `spacedBy` is present
- distribute all remaining width to weighted children before adding visual gaps
- keep ordinary unweighted rows and Flow layouts unchanged
- add a pixel regression proving the final weighted child overflows and is clipped like AndroidX

## Why

AndroidX alpha16 measures weighted row children from the full remaining row width, then adds `spacedBy` during positioning. Compose `Row` normally reserves arrangement spacing before distributing weights, making each weighted child narrower. This is visible in the Home Assistant weather card's four equal-width extra-info panels.

## Test plan

- `./gradlew :rc-player-compose:desktopTest`
- focused Home Assistant Weather rerender: dark **1.98% → 0.95%**, light **6.51% → 1.13%**

## Stack

Merge #3628 first. PR #3630 was merged into #3628, so #3628 now supplies the modifier-order and density behavior this patch builds on. This branch has been rebased onto the repaired #3628 head and passes Compose desktop tests plus Wasm compilation.

### DIFF
--- a/rc-player/compose/src/commonMain/kotlin/ee/schimke/composeai/rcplayer/compose/RcComposePlayer.kt
+++ b/rc-player/compose/src/commonMain/kotlin/ee/schimke/composeai/rcplayer/compose/RcComposePlayer.kt
@@ -639,10 +639,19 @@ private fun RenderLayoutNode(
           theme = theme,
         )
       } else {
+        val hasWeightedChildren =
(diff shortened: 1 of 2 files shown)


yschimke:
Superseded by #3636, which contains this change plus the remaining Remote Compose operation fixes on a clean base of .

Issue #3680, opened 2026-08-11
Epic: Add an iterative component parity issue and acceptance workflow


yschimke:
## Summary

Build an iterative workflow for reviewing design specs, component implementations, preview fixtures, and rendered comparisons.

A reviewer should be able to:

1. See differences between a component&#39;s design reference and rendered preview.
2. Report a difference against the exact component, preview variant, and UI element.
3. See existing GitHub issues directly on the component page.
4. Classify whether the fix belongs to the spec, component, preview, renderer, or comparison engine.
5. Optionally accept one known, issue-linked difference without accepting the whole component.
6. Re-render and verify the fix.
7. Detect when an acceptance is resolved, stale, or no longer matches the recorded difference.

Example: [m3-catalog#40](https://github.com/yschimke/m3-catalog/issues/40) tracks a tonal IconButton whose container matches the Material 3 kit but whose glyph uses a different color token. We should be able to accept that exact glyph-color delta while continuing to detect changes to the container, geometry, glyph shape, and other variants.

## Motivation

The preview server already provides component pages, Reference/Diff/Actual comparisons, catalog-wide parity scoring, annotations, and a prefilled GitHub issue link. However, reported issues are not first-class catalog data:

- A component page does not show its open issues.
- Issues are not reliably joined to component and preview identities.
- Comparisons cannot distinguish a known issue from an untriaged regression.
- There is no committed, issue-linked acceptance model.
- Global thresholds or ignored rectangles can hide unrelated regressions.
- Closing an issue does not prompt verification or removal of its acceptance.

This makes parity reporting a one-way report rather than an iterative workflow.

## Proposed model

    Spec/reference ─┐
    Component ──────┼─&gt; preview variant ─&gt; comparison ─&gt; GitHub issue
    Preview fixture ┘                                  └─&gt; scoped acceptance

The stable locator should contain:

    repository + system + componentId + previewId + referenceId + variant

The preview-server URL is a reproduction link, not the identity. Hosts, tokens, branches, and URL structures may change.

## GitHub issue metadata

Use labels for low-cardinality workflow categories:

    area:spec
    area:component
    area:preview
    area:renderer
    area:comparison
    parity:regression
    parity:known-difference
    parity:verification-needed

Do not create a label for every component. Component identity should live in a machine-readable issue-body locator, with GitHub issue or Project fields as optional enrichment.

Every generated issue should contain a versioned locator such as:

    

The visible issue body should present the same information as readable Markdown.

[...]

PR #3689, opened 2026-08-11
docs(parity): give acceptance its own portable comparison path

yschimke:
Follow-up to #3687, which auto-merged before its last review findings could be applied. Docs only, one file.

Forty-two review rounds landed here (rounds 5–46; #3681 and #3687 carried the first four). Every finding was checked against the code before the doc was touched. Organised by what the document now says, since this text becomes the squashed commit message.

## The central conclusion, and a change of approach

**Acceptance cannot be a mask bolted onto the browser scorer.** The boundary rule #3687 landed wasn't conservative — `drawImage` averages **signed** luma, so on a footprint straddling a mask edge an accepted difference can cancel an opposite unaccepted regression before `scorePlanes` sees the pixel. And the canonical resample is undefined across engines, so identical candidate bytes can produce different canonical pixels and falsely invalidate.

**So §4 no longer specifies a pixel algorithm.** Earlier revisions carried a numbered pipeline; successive rounds found a real defect in every version of it — a gate placed before the data it reads existed, a resample that mixed what a previous step had separated, a delta computed in the wrong direction, and a scorer description that turned out not to match `scorePlanes` at all (four corrections, mostly from partial reads of one function; the last of them was that the walkthrough described *mismatch* and never stated the returned value is a clamped `(1 − mismatch) × 100`, so a port would have published every number inverted). A planning document cannot validate an algorithm. What replaced it: **ten invariants** with the reason each is not negotiable, and an **open-problems list** for Phase 3 (the portable pixel path; mask participation in `edgeMask`; the masked pass's denominator; what "accepted contribution" means, given it can legitimately go negative; sub-pixel rounding; the match metric shared by the candidate gate and the resolution test).

**The portable kernel moves the raw score once, and the doc says so.** I10 preserves the *geometry* — one resample, source → score plane, at the candidate box's dimensions — not the value. Any kernel that isn't `drawImage` produces different pixels, so the number shifts once at a **versioned rebaseline**: bump the score schema version, regenerate committed baselines in the same change, note it in the release notes.

## Identity and scope

**Overrides are the whole render map, not a named list of families.** An earlier revision scoped four — `uiMode`, `fontScale`, device, locale — but `viewer.js`'s `overrides()` also sends the size fields, every checked overlay toggle, and every author-declared knob as `knob.<key>`, with the Remote Compose lane adding `rc.<name>`. A knob *is* the component's state, so an enumerated scope lets a mask authored at `knob.label=Send` suppress pixels at `knob.label=Delete`. The rule is structural instead: every key the render lane accepts is in scope, because a key that didn't affect the render wouldn't be in the map.

**The locator carries overrides, as canonical JSON.** `variant` is axes only, so `overrides` is its own field in the identity and in the `compose-parity-locator/v1` block — without it an issue filed at `fontScale=1.5` serialised identically to one filed at the default. The first form tried, `key=value` joined by `;`, cannot round-trip its own inputs: knob values are free text, so a label of `a;knob.color=red` reads back as two overrides. The value is now a one-line JSON object with keys sorted by code point — the same encoding §4 already stores in the record — always present, `{}` meaning the default render.

**Scope matching uses every recorded field** — `system`, `component`, `previewId`, `referenceId`, `variant` and `overrides`. Served ids are unique only *within* a system, so matching on the comparison page's `(previewId, referenceId)` key would let a `wear-m3` acceptance suppress pixels in `m3`. Matching is exact over the whole override set, because overrides change layout and a mask is geometry.

**The orphan walk checks every scope field the catalog can resolve**, not just the two ids. Asking only "does this preview exist? does this reference exist?" misses a component rename that leaves both ids untouched — the acceptance matches nothing, reports nothing, and outlives the thing it was about invisibly. So the walk resolves `previewId` within its `system`, requires the resolved preview's component and axes to match, and requires `referenceId` to hang off *that* preview. `overrides` is named as the one field it cannot check: an unopened override combination is *unused*, not orphaned.

## Statuses and gates

Strict precedence: `refused` → any non-`candidate-changed` invalidation → `resolved` → `candidate-changed` → `valid`.

- **`resolved` outranks `candidate-changed` only**, and is guarded on that gate having *fired* — otherwise a tolerant metric marks an unchanged candidate resolved at authoring time, and a changed reference that happens to agree closes an acceptance against a spec nobody reviewed.
- **Causes and reasons are arrays**, deterministically ordered. **Ambiguity short-circuits the bounds check only** — a vanished tag is still `element-moved`. **`plane-changed` short-circuits the element gates**, which is what makes a single tag index sufficient.
- **Only `valid` acceptances contribute a mask to the scoring union** — keeping a `resolved` mask would remove its pixels as neighbourhood candidates and could hide a regression beside the thing just fixed.
- **`stale` and `unknown` are a second axis, not two more statuses.** They come from the issue index, which one engine may have and the other may not; folding them into `status` would make the conformance fixtures depend on a file outside the contract. The join happens at the dashboard and the gate, and only `closed` + not-`resolved` is called stale.
- **`refused`** covers both artifact hashes, a reference with no `sha256`, decode failure, degenerate/mismatched dimensions, a non-binary, animated or empty mask, oversized artifacts, out-of-range tolerances, unportable or uncontained paths, unsafe ids, schema violations, orphaned targets, bad headers, and no-op acceptances. **Duplicate ids, unkeyable records and over-budget documents fail the whole document.**
- **`validationFailures` carries one entry per `(record, reason)` pair**, and `duplicate-id` one entry per duplicated *value*, ordered by first occurrence. Both were previously open enough that two engines could emit different arrays for the same rejected file.

## Safety and schema

- **Both artifacts hashed and verified before decode**; `acceptedCandidateSha256` was recorded but never checked. **Hashes compare normalised** — `ServeDesignReferenceStore` lowercases a reference hash to validate it and then serves the original spelling, so raw string inequality would report `reference-changed` for an unchanged reference.
- **Element identity is tag uniqueness, not ref path shape.** `SemanticsRefs` indexes siblings sharing an anchor, so `r/role:Button[0]` silently retargets. A server-computed **tag index** (`tag → {count, bounds}` in canonical coordinates, from the *same* render transaction as the scored PNG) is a hard Phase 2 prerequisite — and selection must be drivable from it, since tag-only nodes produce no annotation to click.
- **Every author-controlled number has a named ceiling.** `candidateTolerance` is `[0, 8]` in 8-bit channel distance and `element.tolerance` is `[0.0, 0.25]`, both fixed by `v1` rather than deferred — a bound each engine picks for itself is not a bound. The budget is 256 acceptances, 128 megapixels, **8192 px per axis**, and **8 MiB encoded per artifact**: the area cap does not imply the axis cap (a `1 × 128,000,000` raster is inside the budget and undecodable in every browser), and neither implies the byte cap (`ServeCatalogStore` refuses any catalog asset over 25 MB, so an artifact inside every pixel limit could be evaluated offline and unfetchable when served).
- **Encoding is checked in the header, not inferred from the samples.** The browser normalises every decode to 8-bit RGBA, so an indexed or RGBA mask with strictly binary samples passes a value check while violating the declared encoding — and an **APNG** passes every check in the contract while a decoder reads `IDAT` and an `` advances the animation. The preflight validates bit depth `8` and colour type `0` for the mask, and rejects an `acTL` chunk in either raster, walking chunk headers only.
- **Paths are contained *and* portable.** `isSafeRelativePath` rewrites `\` to `/` before splitting, so `a\b.png` is checked as two segments and opened as one filename on POSIX; `#` and `?` become URL syntax the moment the host fetches the artifact instead of reading it. Restricted to `[A-Za-z0-9._-]` segments joined by `/`, which removes the encoding question rather than answering it.
- **Ids are constrained three ways** — as path segments (the artifact directory derives from them), as map keys (`__proto__` is a fine path segment and a catastrophic object key), and against **both** dot names: `.` passes a `..`-only check and normalises to the `known-differences` root, letting one acceptance address its siblings' artifacts.
- **An unkeyable `id` has a token.** `schema-invalid` presupposes an id to key the status by and `document-unreadable` describes a file that would not parse; absent, blank, or wrong-typed ids are `id-missing`, positional, document-fatal — with `id-not-safe` keeping the present-but-unsafe case as a per-acceptance refusal.
- **Wire shapes are written out** — the selector payload, the result object, and `validationFailures`' three entry shapes (named record, positional record, document-wide).
- **`kind: producer` and the structured `finding` matcher are cut from v1** — neither had an authoring path or defined semantics.

## Fixtures, delivery, reporting

Inputs are `(reference, candidate, acceptances[], semanticsPayload)`; the acceptance **set** is where engines diverge. Stages pin content boxes, the tag-index projection, selector resolution, per-acceptance separation, canonical, the surviving-union rebuild, and both score-plane sets. **Three runners, not two** — the JS scorer, design-parity's suite, and the server projector's own Kotlin tests, since production JS consumes a prebuilt index and neither of the first two exercises the code that builds it.

Resolution detection sits in **Phase 3**, since Phase 3's own fixtures require the `resolved` case; Phase 4 surfaces statuses rather than computing them, and **stale requires positive closure evidence** — the index publishes closed rows, absence reports *unknown*. **Closure is scoped to one owning document**: "every acceptance linked to this issue" is not evaluable from a run that reads one `known-differences.json`, so `v1` gives an issue a single owning document and a run that cannot establish ownership deletes its resolved records but omits the closing keyword. Reporting moves to the focused comparison; **interactive lanes keep reporting disabled outright**, since overrides are query parameters and interaction is not.

## Test plan

Docs only — no build, test, or rendered surface is affected, so there is no visual evidence to embed. Claims were verified by reading `format-compare.js` (`scorePlanes` including `edgeMask`, the full `directed`, and its clamped return; `scoreImages`, `normalisedBoxes`, `contentBox`, `boxCanvas`, the constants block), `viewer.js` (`overrides` / `liveOverrides` / `refreshLinks` / `refreshReportLink` / `specActualUrl`), `ServeWeb.kt`'s `overrideKnobsHtml`, `ServeCatalogStore.kt`'s `fetchCatalogAsset` and `MAX_FETCH_BYTES`, `SemanticsRefs.kt` and `SemanticsRefsTest`, `ServeDesignAnnotations.kt`, `ComposeSemanticsModels.kt`, `ServeAnnotations.kt`, `ServeDesignReferences.kt` (`isSafeRelativePath`, `hasValidRaster`), `ServeRenderHost.kt`'s `renderAnnotations`, `ServeHttpServer.kt`'s `handleReferenceComparison`, `.github/actions/apply/lib/push-branch.sh`, and the publish step in `design-artifacts-reusable.yml`.

**CI note:** `Renderer Android Unit Tests` was red on `main` independently of this PR — the same five `FigmaSvg*` assertions failing at `5278cd0a`. `main` has since taken #3693 (`fix(figma-svg): stop rastering content the vector model can represent`), and this branch has merged `main` in so CI re-runs against the fixed base.

Refs #3680

---


yschimke:
Note on the two red checks — neither is from this PR, which changes one markdown file.

**`Renderer Android Unit Tests` — failing on `main` too.** Five `FigmaSvg*` assertions:

```
FigmaSvgSegmentedButtonTest > segmented button labels export as editable text, not raster crops   FigmaSvgSegmentedButtonTest.kt:90
FigmaSvgVectorIconRenderTest > a rendered ImageVector Icon exports as an editable path            FigmaSvgVectorIconRenderTest.kt:183
FigmaSvgVectorIconRenderTest > a tinted Icon exports its paths in the tint colour                 FigmaSvgVectorIconRenderTest.kt:203
FigmaSvgVectorIconRenderTest > an intrinsically tinted vector exports in its tint colour          FigmaSvgVectorIconRenderTest.kt:226
FigmaSvgWearPickerTest > picker options and cleared-semantics separator survive as visible SVG    FigmaSvgWearPickerTest.kt:107
219 tests completed, 5 failed, 1 skipped
```

The same five, at the same line numbers, with the same totals, fail on `main` at `6178d986` ([run 31543931950](https://github.com/yschimke/compose-ai-tools/actions/runs/31543931950)) — which predates this branch's current head. That run is also red on `Design Artifacts Driver Tests`, `Actions Script Tests` and `Module Unit Tests`. I'm not touching them: a docs-only change can't have caused them, and fixing the SVG export lane is out of scope here. I'll re-check when the base branch recovers.

**`Reject agent attribution` — transient runner fault, already re-run.** It never reached the scan; the first step died fetching the PR title and body:

```
GET /repos/yschimke/compose-ai-tools/pulls/3689 - 500
HttpError: self-signed certificate ... DEPTH_ZERO_SELF_SIGNED_CERT
```

That's the runner's TLS handshake to `api.github.com`. The identical workflow passed on every earlier commit of this branch (`c811ee5`, `f6b70e1`, `a12824c`, `16a43af`) and on `main` pushes in the same window, so it's a one-off rather than anything about the commits or the body. Re-running the failed jobs was the only applicable action.

---

Issue #3676, opened 2026-08-11
samples(android): make the granted permission preview actually capture a granted environment


yschimke:
## Problem

The permission component correctly calls the public Android API:

```kotlin
val granted =
  ContextCompat.checkSelfPermission(
    context,
    Manifest.permission.CAMERA,
  ) == PackageManager.PERMISSION_GRANTED
```

However, both committed preview functions call the same component with the same environment:

```kotlin
@Preview(name = "Camera permission — denied")
fun CameraPermissionDeniedPreview() = PermissionGatedCameraScreen()

@Preview(name = "Camera permission — granted")
fun CameraPermissionGrantedPreview() = PermissionGatedCameraScreen()
```

The source comment acknowledges that the “granted” preview renders the denied branch during a normal static build unless a live agent manually pushes an override.

Source: [`PermissionGatedPreview.kt`](https://github.com/yschimke/compose-ai-tools/blob/main/samples/android/src/main/kotlin/com/example/sampleandroid/PermissionGatedPreview.kt).

Related permissions UI/e2e work: #1400.

## Why this is misleading

The component architecture is good, but the preview name promises a state the capture does not establish. CI can therefore publish two denied screenshots while reporting one as granted.

Passing a `granted: Boolean` only in the preview would make the pixels look right while bypassing `checkSelfPermission()`, which would repeat the original gesture-sample mistake. The environment must be overridden behind the platform API.

## What real composable code looks like

The production composable should remain unchanged:

```kotlin
@Composable
fun PermissionGatedCameraScreen() {
  val context = LocalContext.current
  val granted =
    ContextCompat.checkSelfPermission(
      context,
      Manifest.permission.CAMERA,
    ) == PackageManager.PERMISSION_GRANTED

  if (granted) {
    CameraContent()
  } else {
    CameraPermissionRequest()
  }
}
```

The preview should declare environment state, not pass fake UI state:

```kotlin
@Preview(name = "Camera permission — granted")
@PermissionPreview(
  grants = ["android.permission.CAMERA=granted"],
)
@Composable
fun CameraPermissionGrantedPreview() {
  PermissionGatedCameraScreen()
}
```

The exact annotation/API is open, but it should route through the existing permissions connector and Robolectric shadow.

## Acceptance criteria

- [ ] Static CI captures distinct granted and denied pixels.
- [ ] Both variants call the same unmodified composable.
- [ ] The granted state flows through `checkSelfPermission()`.
[...]

Issue #3675, opened 2026-08-11
samples(wear): keep placeholder preview overrides out of the component API


yschimke:
## Problem

`PlaceholderCard` reads a compose-ai-tools override as its default application state, and its loading preview installs the override CompositionLocal manually:

```kotlin
@Composable
fun PlaceholderCard(
  loading: Boolean = placeholderActive(default = false),
) {
  val placeholderState = rememberPlaceholderState(loading)
  // real Material placeholder modifiers
}

@Preview
@Composable
fun PlaceholderCardLoading() {
  CompositionLocalProvider(LocalPlaceholderActive provides true) {
    PlaceholderCard()
  }
}
```

Source: [`PlaceholderPreviews.kt`](https://github.com/yschimke/compose-ai-tools/blob/main/samples/wear/src/main/kotlin/com/example/samplewear/PlaceholderPreviews.kt).

Related implementation work: #2646.

## Why this is fake

The rendered placeholder itself is real Wear Material 3 UI, but the component is coupled to `ee.schimke.composeai.overrides`. A consumer copying the sample learns to source production loading state from preview infrastructure.

The static loading preview also knows how the daemon installs its override, which violates the intended boundary: application code owns loading state; preview plumbing supplies or overrides that state outside the component.

## What real composable code looks like

```kotlin
@Composable
fun SessionCard(
  loading: Boolean,
  title: String,
  subtitle: String,
) {
  val placeholderState = rememberPlaceholderState(loading)
  TitleCard(
    onClick = {},
    title = {
      Text(title, Modifier.placeholder(placeholderState))
    },
    subtitle = {
      Text(subtitle, Modifier.placeholder(placeholderState))
    },
    modifier = Modifier.placeholderShimmer(placeholderState),
  )
}

@Preview
@Composable
fun SessionCardLoadedPreview() {
  SessionCard(
    loading = false,
    title = "Morning run",
    subtitle = "5.2 km · 27 min",
  )
}
```

A loading variant may pass an honest state fixture, while daemon overrides should intercept/seed placeholder state through the connector without requiring the component to import compose-ai-tools.

## Proposed preview shape

Keep the sample's public component state explicit. Move `placeholderActive` and `LocalPlaceholderActive` usage into renderer-side wrapping or a preview-only adapter owned by the tooling module. Align this with #2646's first-class placeholder state capture.

## Acceptance criteria

- [ ] Sample application components do not import compose-ai-tools override APIs.
[...]

Issue #3674, opened 2026-08-11
samples(catalogs): remove LocalInspectionMode component swaps from preview content


yschimke:
## Problem

The Compose M3 and Wear M3 catalogs change behavior and sometimes component implementations based on `LocalInspectionMode`:

```kotlin
fun Sticker(id: String) = CatalogSticker {
  CatalogComponent(id, interactive = !LocalInspectionMode.current)
}
```

```kotlin
if (interactive) StatefulSwitch(on)
else Switch(checked = on, onCheckedChange = {})
```

Sources:

- [`CatalogStickers.kt`](https://github.com/yschimke/compose-ai-tools/blob/main/samples/design-catalog-m3/src/main/kotlin/com/example/designcatalogm3/CatalogStickers.kt)
- [`CatalogComponents.kt`](https://github.com/yschimke/compose-ai-tools/blob/main/samples/design-catalog-m3-shared/src/commonMain/kotlin/com/example/designcatalogm3/shared/CatalogComponents.kt)
- [`CatalogInteractive.kt`](https://github.com/yschimke/compose-ai-tools/blob/main/samples/design-catalog-wear-m3/src/main/kotlin/com/example/designcatalogwearm3/CatalogInteractive.kt)

## Why this is fake

The baked preview is not always the same composable that runs in the live lane. It swaps stateful controls for inert controls and returns no-op callbacks specifically to preserve old pixels.

A static render does not click a correctly authored stateful component, so ordinary remembered state is already deterministic at its initial value. Where time-based animation exists, paused-clock capture should control it. `LocalInspectionMode` should not be an application behavior switch.

This split also makes interactive-session failures harder to diagnose because the same preview can silently fall onto the inert branch.

## What real composable code looks like

```kotlin
@Composable
fun WifiSwitch(initiallyChecked: Boolean) {
  var checked by remember(initiallyChecked) {
    mutableStateOf(initiallyChecked)
  }
  Switch(
    checked = checked,
    onCheckedChange = { checked = it },
  )
}

@Preview
@Composable
fun WifiSwitchPreview() {
  WifiSwitch(initiallyChecked = true)
}
```

The initial state can come from ordinary component arguments or named preview overrides. The component implementation and callback remain the same in baked and live renders.

## Proposed preview shape

Make catalog components unconditionally stateful/interactive. Use override values only to seed initial state and content. Use paused clocks for progress/animation, and use real input dispatch for interaction captures.

## Acceptance criteria

[...]

Issue #3673, opened 2026-08-11
samples(android): type into the soft-keyboard preview through public input APIs


yschimke:
## Problem

The animated keyboard sample drives a renderer-internal controller directly:

```kotlin
LaunchedEffect(index) {
  KeyboardController.notifyKeyDown(slot.pressed)
}
DisposableEffect(Unit) {
  onDispose { KeyboardController.notifyKeyUp() }
}
```

Source: [`SoftKeyboardAnimatedPreview.kt`](https://github.com/yschimke/compose-ai-tools/blob/main/samples/android/src/main/kotlin/com/example/sampleandroid/SoftKeyboardAnimatedPreview.kt).

## Why this is fake

Normal applications do not import `ee.schimke.composeai.daemon.KeyboardController`. The preview manually updates the keyboard overlay and separately constructs the displayed text, so it can appear to type even if focus, IME visibility, key dispatch, and text input are broken.

The idle preview already has the right architecture: it calls Compose's public `LocalSoftwareKeyboardController.show()` and lets the connector observe that platform-facing call.

## What real composable code looks like

```kotlin
@Composable
fun MessageEditor() {
  var value by remember { mutableStateOf("") }
  val focusRequester = remember { FocusRequester() }

  TextField(
    value = value,
    onValueChange = { value = it },
    modifier = Modifier.focusRequester(focusRequester),
  )

  LaunchedEffect(Unit) {
    focusRequester.requestFocus()
  }
}

@Preview(widthDp = 360, heightDp = 640)
@Composable
fun MessageEditorPreview() = MessageEditor()
```

The capture harness should focus the field and replay key/text events through the same `interactive/input` dispatch path used by a held session.

## Proposed preview shape

Add a recording/capture declaration for a deterministic input script, or let an existing recording fixture target the preview by semantics. The keyboard connector may render the band and pressed-key overlay, but sample code should only contain the real text field.

## Acceptance criteria

- [ ] No sample imports `KeyboardController`.
- [ ] Text shown in the field is produced by real text-input dispatch.
- [ ] IME visibility follows public focus/keyboard-controller behavior.
- [ ] Pressed-key overlay is derived from the dispatched input.
- [ ] Static and animated captures are deterministic.

PR #3691, opened 2026-08-11
fix(samples): make previews capture the state they claim

### LINKED ISSUE #3671: samples(android): replace the preview-only Glance widget with shared production content
## Problem

`WeatherGlanceAppWidget` implements only the preview entry point and leaves its production entry point empty:

```kotlin
private class WeatherGlanceAppWidget : GlanceAppWidget() {
  override suspend fun providePreview(context: Context, widgetCategory: Int) {
    provideContent { /* weather UI */ }
  }

  override suspend fun provideGlance(context: Context, id: GlanceId) {
    // Production runtime not exercised by this preview-only sample.
  }
}
```

The same file now also contains native `androidx.glance.preview.Preview` coverage, so this older Phase-A helper sample is no longer a good consumer example.

Source: [`AppWidgetPreviews.kt`](https://github.com/yschimke/compose-ai-tools/blob/main/samples/android/src/main/kotlin/com/example/sampleandroid/AppWidgetPreviews.kt).

The legacy RemoteViews widget has a similar problem: its `RemoteViews` construction is duplicated between the real `WeatherAppWidgetReceiver` and every preview variant.

## Why this is fake

A user copying this sample gets a widget that renders in previews but renders nothing when installed. The preview is exercising a separate implementation rather than the component shipped to the launcher.

Duplicated RemoteViews setup can drift in exactly the same way: the preview can stay green while production `onUpdate()` changes.

## What real composable code looks like

```kotlin
@Composable
private fun WeatherWidgetContent() {
  Column(
    modifier = GlanceModifier.fillMaxSize().padding(12.dp),
  ) {
    Text("San Francisco")
    Text("67°")
    Text("Partly cloudy · H 70° / L 55°")
  }
}

class WeatherGlanceAppWidget : GlanceAppWidget() {
  override suspend fun provideGlance(context: Context, id: GlanceId) {
    provideContent { WeatherWidgetContent() }
  }

  override suspend fun providePreview(context: Context, widgetCategory: Int) {
    provideContent { WeatherWidgetContent() }
  }
}

@androidx.glance.preview.Preview(widthDp = 312, heightDp = 152)
@Composable
fun WeatherWidgetPreview() = WeatherWidgetContent()
```

For RemoteViews, extract one `weatherRemoteViews(context, state)` factory and call it from both `onUpdate()` and `AppWidgetContent`.

## Acceptance criteria

- [ ] `provideGlance()` renders the same content as the preview.
- [ ] Native Glance preview discovery is the canonical sample.
- [ ] Remove or clearly quarantine the obsolete Phase-A helper demonstration.
- [ ] Production and preview RemoteViews use one shared factory.
- [ ] Launcher size/resize variants wrap the real widget content without duplicating it.

yschimke:
Closes #3671, #3672, #3674, #3675, #3676.

Four sample-fidelity defects where a preview's label promised a state the capture did not establish, or where the baked capture was not a photograph of the composable that actually runs.

All "before" images below were rendered from `origin/main` in a pristine worktree; all "after" images from this branch, through the same `composePreviewRenderAll` pipeline on the same container. Sources are committed under [`renders/preview-fidelity/`](https://github.com/yschimke/compose-ai-tools/tree/c8046006671767f707e9b006664def432b84ad26/renders/preview-fidelity).

---

## #3676 — the "granted" permission preview rendered the denied branch

`CameraPermissionGrantedPreview` and `CameraPermissionDeniedPreview` called the same composable in the same environment, so CI published two identical PNGs and labelled one of them granted. The before-image is **byte-identical** to the denied capture — same 19,633 bytes.

| Before — labelled "granted", shows denied | After — actually granted |
|---|---|
| granted before | granted after |

The denied variant, unchanged, for comparison:

denied

New `@PermissionPreview(grants = ["android.permission.CAMERA=granted"])` seeds Robolectric's grant state through the already-shipped permissions connector (#1400) before first composition, following the existing `@AmbientPreview` / `@GestureHintPreview` groove through discovery and the renderer.

**The composable is untouched** and still reads `ContextCompat.checkSelfPermission(...)`. Passing a `granted: Boolean` would have made the pixels look right while bypassing the platform call — the environment is overridden *behind* the API, not the UI state in front of it.

## #3672 — pressed/focused states were forged, not driven

The catalogs emitted `PressInteraction` / `FocusInteraction` into hand-made sources from a `LaunchedEffect`. The Android M3 focus ring shows why that mattered — the forged interaction produced a *visibly weaker* result than real focus:

| Before — forged `FocusInteraction` | After — real focus traversal |
|---|---|
| focus ring before | focus ring after |

The catalog was publishing a hairline where the real M3 inset focus ring is a thick stroke over a gap. Wear's focused sticker moves the same way (`#E2D5F8` → `#D4C8EC`):

| Wear focused — before | Wear focused — after |
|---|---|
| wear focused before | wear focused after |

### Two exceptions, both deliberate and marked in-source

**The Wear `pressed` sticker was NOT converted.** `@FocusedPreview(pressed = true)` was tried on it and measured: the capture came out pixel-identical to the focus-only one across the entire button container (both `#D4C8EC`), differing **only** in the label glyphs. The indirect-pointer Press does not reach Wear M3's `Button` interaction source, so the capture documented *focus*, not press. Shipping that under `state = "pressed"` would have been the same defect #3676 exists to remove. It keeps its held interaction source as a marked stopgap and stays byte-identical to `main`:

wear pressed unchanged

**The CMP/desktop catalog was not converted** either: `renderers/desktop/` has no focus handling at all, so `@FocusedPreview` is Robolectric-only. Its helpers are commented as a stopgap. **This needs a follow-up issue** — desktop `ImageComposeScene` focus traversal + indirect-pointer press dispatch.

## #3671 — the Glance widget rendered nothing on a launcher

`provideGlance()` was an empty no-op while `providePreview()` carried the UI, so a consumer copying the sample got a widget that previewed fine and rendered nothing when installed. Both entry points now serve one composable.

The duplicated RemoteViews construction **had already drifted**: the receiver set the condition line to `"Partly cloudy · H 70° / L 55°"` while three of seven previews had decayed to `"Partly cloudy"`. One shared `weatherRemoteViews(context, ...)` factory now feeds production and previews, with genuinely-varying fields passed as arguments.

| Before — drifted short string | After — production string |
|---|---|
| widget before | widget after |

Only this one of the three drifted previews actually moved pixels — the resize walks clip the condition text mid-word (`"Partly"`), so the longer string is invisible there.

## #3675 — the Wear placeholder card sourced production state from tooling

`PlaceholderCard(loading: Boolean = placeholderActive(default = false))` took its default from compose-ai-tools' override runtime, teaching consumers to source production loading state from preview infrastructure. `loading` is now required and the signature carries no tooling import; both existing captures are pixel-unchanged.

A marked preview-only wrapper keeps the live override lane exercised (new capture, unforced so identical to the loaded state):

placeholder override driven

Worth noting: neither *existing* preview actually reached the live lane before — `Loading`'s inner `CompositionLocalProvider` shadowed any daemon-provided value — so this is a net gain in coverage, not just preservation.

## #3674 — catalogs swapped components based on `LocalInspectionMode`

The M3 and Wear catalogs branched on `LocalInspectionMode` to swap stateful controls for inert look-alikes, so the published capture was not always the composable that runs live.

Every branch was audited individually. In all but three cases the inert side produced an **identical first frame** (a counter at 0 renders the bare label; `remember(initial)` starts at `initial`), so the branch is gone with **zero pixel movement** across the desktop catalog. The three real divergences:

- **Progress** — the lanes composed genuinely different overloads (indeterminate vs determinate). Kept determinate everywhere, so published pixels are unchanged; the wasm tier loses its animated ring. Showing it properly wants its own catalog id, which is an inventory change.
- **Cards** — the lanes picked plain vs clickable overloads. Kept **plain** everywhere: the clickable overload adds a clickable node to the semantics tree, and that tree is itself published via `a11y/touchTargets` greenlines and the `compose/semantics-wireframe` variant. A card click-tally isn't worth invalidating two published data products.
- **Slider** — found a latent bug: `StatefulSlider()` hard-coded `0.5f` and ignored its `catalogOverrideFloat("value", …)` knob entirely, so a live seed did nothing. Now honoured.

---

## Test plan

- `:gradle-plugin:preview-discovery:test` — 17 suites green, incl. new grant-parsing coverage (10 cases: parsing, case-insensitivity, malformed entries, duplicates, manifest round-trip)
- `:gradle-plugin:functionalTest --tests "*DiscoveryFunctionalTest*"` — new TestKit coverage that `@PermissionPreview` is discovered by FQN and stamped on every capture
- `:samples:android:testDebugUnitTest` — new `PermissionPreviewPixelTest` asserts the granted and denied renders differ; **passes**
- `:samples:design-catalog-m3-shared:desktopTest` + `:samples:design-catalog-wear-m3:test` — `CatalogInteractivityTest` reworked to assert the property that replaced the lane split: same control, same initial frame under either `LocalInspectionMode`, still responsive
- `composePreviewRenderAll` green across all five affected modules; full before/after PNG diff run to confirm exactly which captures move

**Pre-existing failure, not from this branch:** `ScrollPreviewPixelTest` has two failing GIF-capture cases (`animates red to blue`, `following END resets scroll`). Verified failing identically on `origin/main` in a clean worktree.

## Notes for review

- `GlanceWeatherWidgetPreview`'s `@Preview(name = ...)` changed to mark the helper API as non-canonical, which **does** rename its output PNG — the sanitised preview name is part of the render filename. Nothing in the repo references the old name, so this is a one-time delete+add in the visual diff rather than a broken reference.
- Docs updated: `docs/design/DESIGN_CATALOGS.md`'s "Two lanes" section documented the now-deleted `interactive` flag throughout.
- Container setup gap found: this environment had no Android SDK, and `scripts/install.sh --android-sdk` fails because it tries to apt-install JDK 17 (404 on a stale Ubuntu pin) even though a working JDK 17 is present at `/opt/jdk17`. Android renders are unreachable in a fresh web session until that precondition short-circuits — worth its own issue.

## Not in this PR

Per triage of the eight-issue sweep: **#3670** (Glimmer environment compositing) should stay open as the only tracker for `:data-glimmer-environment-connector`, rescoped from "the sample is fake" to "build the connector". **#3673** (keyboard) is a real smell but needs static-lane input-script infrastructure. **#3669** (splash) is recommended for closure — `SplashScreenSurface` is a deliberately published preview runtime, sister to `:notification-preview-runtime`, and the platform `SplashScreenView` it proposes capturing is system-constructed and out of reach; its own build file records that `core-splashscreen` exposes no reusable Compose surface for tooling.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HQ1RBBig7q5aAHJiBWRpku)_


yschimke:
Addressed all three review findings in `2ebf191`. All three were correct; the P1 in particular was a genuine defect and I've added a regression test that proves it.

**P1 — stale grants leaking between captures. Fixed.** Confirmed exactly as described: `syncRobolectricGrants` denied only the names in the *incoming* map, and `set(null)` passed an empty array so it revoked nothing at all. The KDoc claimed "absent from the override are explicitly denied so a re-render with a shrunk grant map doesn't leak" — the implementation never did that.

Each sync now denies the **union** of every previously-applied name and the incoming keys before granting, tracked in a new `appliedPermissionNames` set. Renders that never seed an override still touch nothing, so an un-annotated capture keeps its manifest baseline.

The worst case is worse than a shrunk map: a preview carrying no `@PermissionPreview` never constructs the extension at all, so nothing runs to clear anything — it would simply inherit whatever the previous capture granted. Exactly the "capture shows a state it never established" defect this PR exists to remove.

New `PermissionsControllerRevocationTest` pins it at the platform boundary (`Context.checkSelfPermission`) rather than the controller's snapshot map, since the snapshot was never what leaked — it has always been replaced wholesale. Verified it bites: reverting the fix fails 3 of its 4 cases (the explicit-`DENIED` case passes either way, since denied keys were already in the deny array).

**P2 — permissions dropped from artifact-only capture modes. Partially fixed.** Correct about `@AnimatedPreview`: both animation-capture constructions now carry `permissions`, so an `@AnimatedPreview` owning a function outright no longer leaves the renderer's first-non-null scan empty.

The `@ScrollingPreview` half is **not** fixed, deliberately. Normal scroll captures (including `GIF`) already go through the standard `Capture` path and carry `permissions`. But `LONG`/`GIF` also emit `dataProducts` entries, and that manifest type has no permission field at all — adding one means changing the data-product wire shape plus renderer plumbing, which is disproportionate here and better as its own change. Flagging rather than silently leaving it.

**P2 — docs contradiction on Wear pressed. Fixed, and this one was my error.** I rewrote that section before making the Wear revert and didn't re-read it afterwards. It now splits focused (real input on both Android sheets, with the M3 inset ring as visible proof) from pressed (works on `design-catalog-m3-android`, does **not** reach Wear M3's `Button`), and lists both stopgaps — Wear's `ButtonPressed` and the whole CMP/desktop sheet — with the condition for deleting each.

---

**On the CI failures:** the four red checks on `c804600` were all pre-existing on `main`, verified by running each in a clean worktree at `origin/main` — `ScrollPreviewPixelTest` (2 GIF cases), `FigmaSvg*` (5 cases in Renderer Android Unit Tests), the design-artifacts `catalog-export` version pin (`0.1.46` vs expected `0.1.45`), and `test-sandbox-tooling.sh` (bubblewrap missing from the Dockerfile, #3211). This branch touches nothing under `deploy/`. This push is rebased onto current `main`, which now includes #3688 greening those up, so the next run should be clean on that front.

---

yschimke:
Following up on the one CI failure I hadn't diagnosed when I commented above — **Serve render lanes honor overrides**. It is *not* from this branch, but my previous comment was too quick to lump it in with the others, so here is the actual evidence.

Two live-lane tests fail:

- `Live lane types the visitor's keystrokes into the field` — frame diff `0.35` / `0.28` against a `> 0.4` threshold
- `Live lane selects text on a mouse drag` — frame diff `0.0` / `0.07` against the same threshold

Both are input-delivery assertions (keystrokes reaching the field, a mouse drag starting a selection), and the run is against `SERVE_SYSTEM: compose-m3` — a catalog this PR does touch, which is why I did not want to wave it off without checking.

`serve-lanes-e2e.yml` has failed on `main` for the **last six consecutive runs**, starting `2026-08-11T19:00:40Z` (#3661) and continuing through #3668, #3677, #3678 and #3683 at `22:38Z`. This branch was cut from `main` after all of those. The last green run on `main` was `18:27Z` (#3653), immediately before #3661 merged.

The prime suspect is therefore **#3661, "fix(daemon): preserve natural pixel input coordinates"** — the first failing run, and a change to input coordinate handling, which is exactly what these two assertions measure (a drag that produces a `0.0` diff reads like pointer input not landing at all). That is circumstantial from run ordering rather than a bisect, so treat it as a lead, not a conclusion.

Worth noting #3688 greened the other four checks but did not cover this one; `main` was still red here at `22:38Z`. This branch fails the same two tests and no others, so it neither introduces nor worsens the regression. Happy to look into #3661 separately if that would help — it is unrelated to this PR's scope, so I have not touched it here.

---

yschimke:
Correcting one thing I got wrong earlier: I said the rebase onto #3688 should green all four of the originally-red checks. It greened **two** — Actions Script Tests ✅ and Design Artifacts Driver Tests ✅. The other two are still red on `main` itself, and I've now verified each directly rather than inferring it:

- **Renderer Android Unit Tests** — same 5 `FigmaSvg*` failures (`FigmaSvgSegmentedButtonTest`, 3× `FigmaSvgVectorIconRenderTest`, `FigmaSvgWearPickerTest`). I ran `:renderer-android:testDebugUnitTest` on this branch and on a clean checkout of **current** `main` (`5278cd0`): `219 tests completed, 5 failed, 1 skipped` on both, identical test names. #3688 did not cover these.
- **Serve render lanes honor overrides** — pre-existing, detailed in the comment above.

Everything else on this run is green, including the checks that actually exercise this change: Plugin Unit Tests, Plugin Functional Tests (the new `@PermissionPreview` discovery coverage), Build Samples, all three Daemon Harness jobs, CMP/Wasm Parity (0 regressions), and CMP Remote Compose Player. Module Unit Tests — which runs the new `PermissionPreviewPixelTest` and `PermissionsControllerRevocationTest` — is still in flight; both pass locally.

So the two remaining reds are inherited from `main`, not introduced here.

---

yschimke:
One clarification on the preview-diff comment above, because it is misleading for this PR and a reviewer would reasonably draw the wrong conclusion from it.

The bot lists Wear's **`ButtonFocused` under "Unchanged"**. That is not a measurement — the scoped CI run never re-rendered `samples:design-catalog-wear-m3`, so there was nothing to compare against.

Checked against the bot's own render commits:

- baseline `bbc64331` → `renders/samples:design-catalog-wear-m3/ButtonFocused.png` exists, `md5 831d0119…`
- head `04291e7f` → same path **404s**; the module has no render in that commit at all

So the "unchanged" entry is carried over from the baseline. Consistent with the Changed list containing only `samples:android` and `samples:design-catalog-m3-android` — the two modules the run actually rendered — and with `Build Samples (full render suite)` being **skipped** on this run.

For what it is worth, the baseline PNG the bot holds is **byte-identical** to the `before` image committed in this PR (`renders/preview-fidelity/wear-focused-before.png`, same `md5 831d0119…`), which is a nice independent check that the baseline I rendered locally from `origin/main` is the same artifact CI has. The `after` differs (`md5 3ba335b0…`): container `#E2D5F8` → `#D4C8EC`, the real focus state layer.

Net: `ButtonFocused` **does** change, the committed before/after in `renders/preview-fidelity/` is the evidence for it, and the bot's Unchanged listing for that module should be read as "not rendered", not "no change". `ButtonPressed` genuinely is unchanged — that one was reverted to its stopgap deliberately.

---

yschimke:
**Module Unit Tests** went red. It is pre-existing, and I checked it carefully because this is the job that runs both new tests from this PR.

Four Gradle tasks failed. I ran all of them locally on this branch and on a clean checkout of current `main` (`5278cd0`), with `--continue` so nothing short-circuits:

| Task | Branch | current `main` |
|---|---|---|
| `:data-layoutinspector-core:test` | 277 completed, **8 failed** | 277 completed, **8 failed** |
| `:daemon:desktop:test` | 265 completed, **3 failed** | 265 completed, **3 failed** |
| `:daemon:android:testDebugUnitTest` | **280** completed, **5 failed** | **276** completed, **5 failed** |
| `:renderer-android:testDebugUnitTest` | 219 completed, 5 failed | 219 completed, 5 failed |

Same failing test names on both sides in every task — all in the same family (`FigmaSvg*`, `MaterialIconRef*`, `RenderEngineTest > figmaSvg*`, `DesktopLayoutInspectorTest`), i.e. vector/SVG export, which this PR does not touch.

The one column that differs is the useful one: `:daemon:android` runs **280** tests on this branch against **276** on `main`. That +4 is `PermissionsControllerRevocationTest`, and the failure count is 5 on both sides — so the new tests ran and passed rather than being skipped. Confirmed in the result XML:

- `PermissionsControllerRevocationTest` — `tests="4" failures="0" errors="0"`
- `PermissionPreviewPixelTest` — `tests="1" failures="0" errors="0"`

So all three remaining red checks on this PR (Module Unit Tests, Renderer Android Unit Tests, Serve render lanes) are inherited from `main`. Nothing in this branch fails that passes on `main`, and the two tests this PR adds both pass in CI's own job.

Worth flagging separately: `main` currently has **21 failing tests** across four tasks in this one job, plus the serve-lanes regression. That is a lot of red for a branch that is presumably meant to be green, and #3688 only addressed part of it.

---

Issue #3694, opened 2026-08-12
renderer(wear): indirect-pointer Press doesn't reach Wear M3 Button, so `pressed = true` captures focus instead


yschimke:
Follow-up to #3672, narrowed: **CMP/desktop renderer focus + press dispatch is being handled separately** and is no longer in scope here. This issue is now only the Wear half.

## The problem

On Wear the focus walk works and the press does not.

`@FocusedPreview(indices = [0], pressed = true)` was applied to the Wear `ButtonPressed` sticker and measured. The resulting capture was pixel-identical to the focus-only capture across the whole button container — both `#D4C8EC` — differing **only** in the label glyphs. The Press never reaches the component's interaction source, so the capture documents *focus*, not press.

That is why `ButtonPressed` in `samples/design-catalog-wear-m3/.../CatalogPreviews.kt` was reverted to a held `MutableInteractionSource` rather than shipped in #3691: a sticker published under `state = "pressed"` that actually shows the focused visual is the same mislabelled-artifact defect #3676 existed to remove. Better an honest forged state layer than a confidently wrong label.

The helper carries an in-source comment naming this blocker and the condition for deleting it.

## Worth knowing before fixing

`FocusedPreview`'s KDoc claims `Modifier.clickable` registers a fallback indirect-pointer handler, so the pressed visual "still arrives" on plain touch surfaces. That is not what is observed on Wear M3's `Button`. Establish which of these is true before changing anything:

- the fallback handler does not apply to Wear M3's `Button` (it may route interactions differently from `Modifier.clickable`), or
- the dispatch is not landing on the focus-walk target on Wear, or
- the press lands but Wear M3's pressed state layer is not visually distinct from its focused one at rest

The third would be a documentation fix, not a renderer fix — and would mean the Wear catalog should stop advertising a separate `pressed` variant at all.

## Acceptance

- [ ] Root cause identified from the three above
- [ ] Either `pressed = true` yields a Wear capture visually distinct from the focus-only capture, or `FocusedPreview`'s KDoc states plainly that it cannot on Wear and the catalog drops the variant
- [ ] If fixed: the `pressedSource()` helper in `samples/design-catalog-wear-m3/.../CatalogPreviews.kt` is deleted and `ButtonPressed` moves to `@FocusedPreview(indices = [0], pressed = true)`
- [ ] `docs/design/DESIGN_CATALOGS.md`'s "Pressed stickers are the unfinished half" section updated

## Verifying a fix

[...]

PR #3699, opened 2026-08-12
feat(renderer-desktop): drive @FocusedPreview with real focus and press on CMP desktop

### LINKED ISSUE #3672: samples(catalogs): drive pressed and focused states through real input overrides
## Problem

Three catalog surfaces forge pressed/focused visuals by emitting interactions from `LaunchedEffect`:

```kotlin
@Composable
fun pressedSource(): MutableInteractionSource {
  val source = remember { MutableInteractionSource() }
  LaunchedEffect(source) {
    source.emit(PressInteraction.Press(Offset.Zero))
  }
  return source
}
```

Occurrences:

- [Compose M3 shared catalog](https://github.com/yschimke/compose-ai-tools/blob/main/samples/design-catalog-m3-shared/src/commonMain/kotlin/com/example/designcatalogm3/shared/CatalogComponents.kt)
- [Android M3 focus-ring catalog](https://github.com/yschimke/compose-ai-tools/blob/main/samples/design-catalog-m3-android/src/main/kotlin/com/example/designcatalogm3android/FocusRing.kt)
- [Wear M3 catalog](https://github.com/yschimke/compose-ai-tools/blob/main/samples/design-catalog-wear-m3/src/main/kotlin/com/example/designcatalogwearm3/CatalogPreviews.kt)

## Why this is fake

No user focus traversal or pointer press occurred. The sample injects an incomplete interaction directly into a Material component and never pairs it with `Release`, `Cancel`, or `Unfocus`.

This bypasses the behavior we need to validate: focusability, input mode, traversal order, hit testing, and the component's own interaction wiring. It can display a state layer even when the real component cannot receive that state.

The repository already demonstrates the correct approach in `samples/android-alpha/FocusRingPreviews.kt`: plain buttons plus `@FocusedPreview`.

## What real composable code looks like

```kotlin
@Composable
fun FocusableButtons() {
  Row {
    Button(onClick = {}) { Text("Save") }
    Button(onClick = {}) { Text("Edit") }
  }
}

@Preview
@FocusedPreview(indices = [0, 1])
@Composable
fun FocusedButtonsPreview() {
  MaterialTheme { FocusableButtons() }
}

@Preview
@FocusedPreview(indices = [0], pressed = true)
@Composable
fun PressedButtonPreview() {
  MaterialTheme { FocusableButtons() }
}
```

## Proposed preview shape

Extend/reuse the focus and input connectors for the catalog modules, including Wear. Preview annotations should move actual focus and dispatch a real press through the host. Catalog components should not accept preview-only interaction sources.

## Acceptance criteria

- [ ] Remove manually emitted `PressInteraction` and `FocusInteraction` from samples.
- [ ] Focused captures use real focus traversal.
- [ ] Pressed captures dispatch through the input host and have correct press lifecycle.
- [ ] Wear focus runs under the appropriate rotary/D-pad input mode.
- [ ] Captures prove that the real component, not merely its state layer, receives the interaction.

yschimke:
Finishes the CMP/desktop half of #3672 — the piece #3691 deliberately left out because `renderers/desktop` had no focus or press dispatch at all.

## The defect, measured

The CMP catalog forged its two interaction stickers with a held `MutableInteractionSource` (`pressedSource()` / `focusedSource()`). Rendered from `main`, both came out **pixel-identical to a resting button** — the forged interaction did not produce a weak state layer, it produced *none*, and the sheet was publishing two labelled state stickers that showed no state.

| | `Button/Filled` resting (control) | `keyboard-focus` | `pressed` |
|---|---|---|---|
| **before** (`main`) | resting | focused before | pressed before |
| container colour | `#6750A4` | `#6750A4` | `#6750A4` |
| **after** (this branch) | unchanged | focused after | pressed after |
| container colour | `#6750A4` | `#7661AD` | `#8471B5` |

Both after-captures keep the exact framing of their before-captures (351×210 / 342×210) — a state, not a reflow.

## What changed

**`renderers/desktop` — new `renderFocusPreview`.** Walks focus with a real `FocusManager.moveFocus(...)` traversal under a synthetic keyboard input mode, then captures. No focus logic is reimplemented in the renderer: it wraps content with `:data-focus-connector-desktop`'s `FocusOverrideExtension` — the same seam the daemon's `renderNow.overrides.focus` path uses and the same shape `RobolectricRenderTest` consumes on Android — and only decides when to flip `FocusController`.

Uses `runSkikoComposeUiTest` rather than `ImageComposeScene` for the reason the scroll path does: the walk needs a paused clock (Material's state-layer crossfade must land *before* the capture) and `pressed` needs an input host. The AS-parity wrap-measure box and round clip are shared with `renderPreview`, so a focused capture is framed identically to the resting one.

**Press dispatch differs from Android on purpose.** Android sends an *indirect* pointer event, which routes to the focused composable because XR Glasses have no touchscreen. Desktop has no indirect channel, so this dispatches an ordinary pointer down onto the focused element — hit-tested like a real click, which is stronger evidence, not weaker: the pressed state can only appear if the real component consumed the event. As on Android the press is held, not released, so `onClick` never fires.

**Traversal replays its prefix.** This backend composes one scene per capture; the Android renderer keeps one composition alive across a preview's captures and inherits where focus was left. So step 3 of `traverse = [Next, Next, Previous]` replays all three from `Enter` — `focusTraversalPrefix` in `RenderPreviewsTask` computes that list. Without it every step of a traversal sheet would render step 1's frame under step N's filename, which is the same mislabelled-artifact defect this issue family exists to remove.

**Plugin.** `RenderPreviewsTask` forwards the per-capture focus state as renderer args 33–38 (four `"0"` placeholders keep the existing size-bound slots 29–32 positionally stable). `-1` / empty means "no focus intent", so every preview without the annotation stays byte-identical on the undriven path, and the Android lane — which reads the manifest, not argv — is untouched.

**Connector.** `:data-focus-connector-desktop` gains `enterPlacesFocus` parity with the Android walk, plus `FocusOverlayDesktop` (`overlay = true`): the same marker the Android overlay draws, over bounds read from the semantics tree instead of reflected out of `AndroidComposeView.focusOwner`.

**Catalog.** `button-filled-pressed` / `button-filled-focused` are now plain `Button`s; `pressedSource()` / `focusedSource()` are gone, and the previews carry `@FocusedPreview(indices = [0])` / `@FocusedPreview(indices = [0], pressed = true)`. The live wasm tier composes the same button and now shows the state when a visitor actually focuses or presses it.

## Not in scope

`gif = true` remains Android-only — a desktop `gif = true` preview writes an undriven single frame into its `.gif` output, as it did before. Now documented on the annotation and in `DESIGN_CATALOGS.md` rather than left to be discovered. Wear's `ButtonPressed` stopgap is unaffected and stays tracked by #3694.

## Test plan

- `:renderer-desktop:test` — new `DesktopFocusRendererTest`, the desktop twin of `:samples:android-alpha`'s `FocusedPreviewPixelTest`, over a fixture of plain `Button`s with no interaction source of their own:
  - focused capture differs from the undriven one, at identical dimensions
  - indexed fan-out `0/1/2` produces three distinct captures
  - traversal `Next` / `Next,Next` / `Next,Next,Previous` walks 0 → 1 → 0, and step 1 equals the `indices = [0]` capture
  - pressed capture differs from the focus-only one
  - `overlay = true` paints the marker and preserves the `.raw.png` companion
  - a preview with nothing focusable declines, so the caller falls back to the undriven render
- `:gradle-plugin:test` — new `FocusTraversalPrefixTest` pins the prefix ordering.
- `:cli:test --tests "*BundleRendererSizeBoundsArgsTest*"` — arg indices 28–31 still land where the bundle path expects.
- `:samples:design-catalog-m3:composePreviewRenderAll` (filtered) on this branch and in a pristine `origin/main` worktree produced the before/after captures above; colours sampled from the PNGs.
- `ktfmtCheckAll`, `:samples:design-catalog-m3-shared:compileKotlinWasmJs`, `:samples:design-catalog-m3:compileKotlin` all green.

Closes #3672.

---
_Generated by [Claude Code](https://claude.ai/code/session_019QX56Q7v2sC7zxT8pU9Gmq)_


yschimke:
Pushed `1759662`. All four Codex findings were real; one of them was hiding behind a race the diff bot caught.

**The race — worth reading even if you skip the rest.** The preview-diff bot reported `FilledButtonPressed` changed and `FilledButtonFocused` **unchanged**, from the same run. That should be impossible: both stickers take the same path, and the pressed one only presses *after* the focus walk lands. The explanation is that the walk lives in a `LaunchedEffect`, and the first version assumed a fixed `advanceTimeBy` window contained its dispatch. When it didn't, the node ended up focused — so the `landed` probe was happy — but its indication had never been given a frame to fade in, and the published sticker was pixel-identical to an untouched button. The pressed capture won the race only because the press path adds two more settle windows behind it.

That is the same defect class this issue exists to remove (a capture that doesn't show the state its filename claims), arrived at from the other direction, and it would have shipped green. The walk now waits for idle before each advance, waits for focus to actually land, then gives the indication its own settle window. Nothing about the capture depends on dispatcher timing any more.

The bot's own output on the next run is the check on this: `FilledButtonFocused` should move from "Unchanged" to "Changed".

**The other three:**
- **Traversal prefix deduplicated per step.** `captures` is the cross product of the scroll / timing / focus fan-outs, so a traversal crossed with another dimension carried each step once per row — step 2 of a two-timing traversal would have dispatched four `Next` moves and landed past the element its filename names. Covered by a new case in `FocusTraversalPrefixTest`.
- **Overload and wrapper resolution now shared with `renderPreview`** (`findComposableMethodWithArgs` / `resolveWrapper`, both widened to `internal`) instead of near-copies, so a `@PreviewParameter` overload with matching arity but wrong types can't win and a substituted `@PreviewWrapper` (Remote Compose) is still substituted. The unused `classLoader` parameter went with them.
- **END + focus now compose** rather than the focus branch silently winning: `driveScrollToEnd` is extracted from `captureEnd` and runs first, in the same scene, so a capture carrying both intents documents both. LONG / GIF stay with the scroll branch — their stitched product isn't something a single focused frame can stand in for. New test `endScrollAndFocusComposeInOneCapture` renders a 20-item list and asserts the END+focus capture differs from the focus-only one (it now shows items 17–20 with focus on the first visible, instead of items 1–3).

**CI, for the record — neither red check is from this branch:**
- `Serve render lanes honor overrides` is red on `main` and has been for the last ~7 merges, including the `78138e1` this branch forked from. Same spec, same assertion family (`serve-lanes.spec.mjs` live-lane pixel-diff thresholds). On this PR the same "mouse drag" test was flaky-then-passed and its sibling typing test failed instead. I'll re-run when the base recovers.
- `Plugin Functional Tests` failed with `java.lang.OutOfMemoryError: Java heap space` inside ClassGraph, in the TestKit build for `BundleFunctionalTest > module runtime resources are packed into the app jar for a live re-render` — a runner memory failure, not an assertion. The full `:gradle-plugin:functionalTest` suite passes locally on this branch (92 tests, 7m37s).

Local verification of the new commit: `:renderer-desktop:test` (7 tests), `:gradle-plugin:test`, `:cli:test --tests "*BundleRendererSizeBoundsArgsTest*"`, `ktfmtCheckAll` — all green. The committed evidence PNGs re-render byte-identical after the settle change.

---

PR #3700, opened 2026-08-12
feat(serve): exploded 3D view of a screen, one sheet per composable

yschimke:
Adds `?exploded=1` to `/render/<id>.svg` and a **3D** chip beside SVG in the viewer: the layered `compose/figma-svg` export tilted back and pulled apart into one floating sheet per level of composable nesting, each labelled with the composables that drew it.

A flat render answers *what does it look like*. This answers *what is it made of* — and unlike a tree view it keeps every layer's real pixels.

## The picture

A phone screen exploded into one sheet per level of composable nesting: Frame, Scaffold, TopAppBar/LazyColumn/FAB, Card, Text

The same transform over a **real** export — `renders/material-icon-refs/compose-figma.svg`, captured from a Robolectric render of `MaterialIconRowPreview`. The hoisted Material-icon `<defs>` still resolve from the sheet that `<use>`s them, and the labels see through `ReusableComposeNode` to the icon each layer actually draws:

The committed MaterialIconRowPreview export exploded, its icons still resolving from the shared defs

## In the viewer

`?exploded=1` presses the 3D chip, switches the stage to the vector lane and fetches the exploded SVG — a shareable deep link, not tab-local state:

The viewer opened on ?exploded=1, with the 3D and SVG chips pressed and the exploded diagram on the stage

The camera axes live in the overrides drawer, beside the other controls that shape an export:

The Exploded 3D group in the overrides drawer: Lean, Spin, Separation and Layers sliders with live readouts

## Why an SVG rewrite rather than a WebGL scene

The layered export already emits **every composable as a `<g id="…">`, nested exactly as the composables nest**, so the structure this needs is sitting in bytes the SVG lane already serves. `ExplodedSvg` is a pure SVG→SVG transform — no second render, no new data product, no daemon capability to negotiate. Two consequences that a canvas wouldn't have:

- **Any host that can answer `.svg` can answer this**, including a fully static catalog serving a baked `figma/<slug>.svg` off its delivery branch. No daemon in the path.
- **The result is still one static SVG.** It embeds in a PR body (as above), opens in a browser, `curl`s from an agent, rasterizes in the visual-diff bot, and pastes into Figma with its layers intact. "Copy link" / "Download" for SVG hand over the exploded view whenever the chip is pressed, because it is the same URL.

Each drawing element is assigned to the plane matching its named-group depth; each plane is re-emitted as a *structural copy* of the original tree with only that plane's elements kept, so the enclosing `transform` / `clip-path` chain still places everything exactly where it was, and `<defs>` / `<clipPath>` / `<style>` are carried over once at the top.

A plane is flat, so the camera reduces to one affine `matrix(…)` per sheet. Two calls worth reviewing:

- **Lean is measured from face-on** (`0` leaves a portrait preview reading as a portrait preview), and **the separation is a screen-space offset rather than a distance along the plane normal**. Physically the sheets separate along the normal, which projects to `sin(lean)` — vanishing exactly as the view approaches the angle a UI is meant to be read at, so a physically-honest camera can only explode a screen by first laying it flat. Every printed exploded diagram cheats the same way, and decoupling them makes both sliders mean what their labels say. Pinned by a test, since it is the property a future "make the projection more correct" change would quietly delete.
- **Fixed light chrome, not `prefers-color-scheme`.** The media query answers for the reader's OS while the background the SVG lands on is chosen by the host, so the two disagreed in both directions during development (light-grey labels invisible on the viewer's white stage; then a dark card floating in that same white stage). One fixed pairing can be wrong about its surroundings but never about itself.

Every axis rides the URL (`&explodeTilt=` / `&explodeSpin=` / `&explodeGap=` / `&explodeDepth=`), untouched axes are omitted so the common link stays `?exploded=1`, and an out-of-range or unparseable value falls back to the default rather than 400ing a stale bookmark — this is a view axis on an export URL, not a render override, and it is deliberately never reported as a "dropped override". It composes with `?mode=web` and `?scroll=long`, because all three are post-processing steps over one render.

## Automatic future coverage

Per the repo rule, this doesn't stop at a one-off screenshot:

- `ExplodedSvgFixtureTest` re-renders both committed goldens through the **production** renderer and fails on drift, so a change to the camera, the sheet split, the plates or the labels is a reviewable text diff. It also pins the drawer sliders' defaults to `ExplodedSvg.Options`' own, across module boundaries.
- `pages-snapshot.spec.mjs` serves one of those goldens as the `?exploded=1` render lane's stub, so the new `serve-viewer-exploded` fixture's screenshot contains the **real** projection rather than a hand-drawn stand-in — and `vscode-preview-diff` diffs it, plus its drawer-open `controls` state, on every PR with nobody remembering to capture anything.

## Test plan

- `./gradlew :data-layoutinspector-core:test` — 14 new `ExplodedSvgTest` cases (plane assignment, transform/clip preservation, resources carried once, layer ids not duplicated across sheets, depth folding, face-on separation, canvas growth, XXE refused, degradation to the flat SVG). The 8 pre-existing `FigmaSvg*` failures in that module reproduce unchanged on `main`.
- `./gradlew :cli:test` — green. New `ExplodedSvgFixtureTest` (4 cases) and 2 new `ServeHttpRoutingTest` cases covering the lane on a **static** catalog: serves exploded, knobs reshape without a re-render, bogus values fall back, `?exploded=0` is byte-identical to the flat export, and nothing is reported as a dropped override.
- `npx playwright test … pages-snapshot` — 96 passed, including the new `serve-viewer-exploded` fixture in both themes and its `controls` state.
- `npm test` (vscode-extension) — 1674 passing.

## Notes

- The session handed me a `claude/…` branch; renamed to `agent/…` per the repo's branch-prefix rule before pushing.
- `docs/public-preview-server.md` gains a section for the lane, with the same images.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vq9v33ZPideWib3VmahzHG)_


yschimke:
**`Serve render lanes honor overrides` is red on `main` too — not caused by this PR.**

Two live-lane tests failed here: `Live lane selects text on a mouse drag` and `Live lane types the visitor's keystrokes into the field`. Both report the same symptom — the streamed frame never visibly reacted (`frameDiffPct` of `0` and `0.067%` against a `0.4%` floor, after the full 30s poll).

The drag test is **already failing on `main`**, on a commit *ahead* of this branch's base:

- [main @ `78138e1` — run 31564855663, job "Serve render lanes honor overrides"](https://github.com/yschimke/compose-ai-tools/actions/runs/31564855663/job/94014514663) → `1 failed / 11 passed`, `Live lane selects text on a mouse drag`, received `0.0666%` then `0` on retry.
- This PR branched from `93c9e1c`; `78138e1` is a descendant of that, so the failure postdates my branch point and cannot be inherited from it.

**Why the keystroke test isn't mine either.** The only production file this PR changes on that path is `viewer.js`, and its diff is inert for the live lane:

- `dropVectorModes()` is a drop-in for the `svgToggle.setAttribute("aria-pressed", "false")` line it replaced in all five `enterMode` branches. Its extra work sits behind `if (explodeToggle && explodeOn())`, and these tests never press the 3D chip — so on the live path it executes exactly the statement it replaced, then returns.
- `withSnapshotFormat` → `withExplode` returns its input unchanged unless `ext === ".svg"`. The live lane pins `snapshotExt = ".png"`, so no query parameter changes and no extra request is made.
- Nothing in the diff touches pointer/keystroke forwarding, the stream transport, or the canvas.

Both failures are the same lane failing the same way — "the live daemon's text field didn't repaint" — under Xvfb + Mesa software GL, and `main` drew one sample from that distribution while this run drew two. Everything else in the job passed (10/12 here, 11/12 on main).

I'm not fixing the underlying live-lane instability in this PR — it's unrelated to the exploded view and it's already red on the base branch, so a fix belongs on its own change where it can be judged on its own evidence. I'll re-run this job and re-check once the base branch is green again.

---

Issue #798, opened 2026-05-05
v1.0 cleanup checklist


yschimke:
Punch list of cleanups before tagging 1.0.0. Items are observed at branch tip (post-0.9.3); each row links to a file:line and a one-line action. Group decisions ("ship in 1.0" vs "explicitly deferred to 1.x") are the main thing missing — once those are made, most rows collapse to either a small PR or a doc edit.

## A. Doc inconsistencies (single source of truth: `gradle/libs.versions.toml`)

- [ ] **Kotlin / AGP versions disagree across docs.** Catalog has Kotlin `2.3.21` + AGP `9.2.0`; docs say otherwise.
  - `README.md:62` — "AGP 9.1+ … Kotlin 2.2.21"
  - `docs/AGENTS.md:106` — "Kotlin 2.3.20, … AGP 9.1.1"
  - `docs/HOW_IT_WORKS.md:112-113` — "AGP 9.1.0 … Kotlin 2.2.21"
  - Action: sync all three to `2.3.21` / `9.2.0`, or generate from the catalog at build time.
- [ ] **`.github/workflows/daemon-harness.yml:4,77,108`** references `docs/daemon/TODO.md`, which doesn't exist (replaced by `docs/daemon/ROADMAP.md`). Update the comments.

## B. Reserved-but-unimplemented protocol surface

The protocol document advertises fields and notifications that the daemon never actually emits. Either implement, drop, or label clearly as "1.x" in `PROTOCOL.md`.

- [ ] `leakDetection: ("light" | "heavy")[]` — `docs/daemon/PROTOCOL.md:128` ("always [] today").
- [ ] `renderFailed.kind` discriminants `compile|runtime|capture|timeout` — daemon hard-codes `"internal"` (see `JsonRpcServer` in `daemon/core/`). Either populate or document.
- [ ] `sandboxRecycle`, `daemonWarming`, `daemonReady` notifications — `PROTOCOL.md:504/523/533`, never emitted; `docs/daemon/ROADMAP.md` "Sandbox lifecycle | Planned". Decide 1.0 scope.
- [ ] `HistoryPixelNotImplemented` (`-32012`) error and pixel-mode `history/diff` (`PROTOCOL.md:55,347,362`) — phase H5, not implemented; flag as 1.x in protocol doc.

## C. Partially-implemented features

[...]

PR #3710, opened 2026-08-12
chore(release): cut the next release as 1.0.0

yschimke:
## Summary

Pins the root package in `release-please-config.json` with `"release-as": "1.0.0"`, so the next release cuts as **1.0.0** instead of the `0.19.x` patch the conventional-commit history would produce.

The open release PR #3692 (`chore(main): release 0.19.62`) is not stale work to redo — the `update-release-pr` job re-runs release-please on the next push to `main`, re-reads this config, and rewrites that same PR to `chore(main): release 1.0.0`. Merging it then tags, drafts, builds, and finalizes exactly as before; nothing else in the release chain changes.

### The removal step, documented

`release-as` is **sticky** — it isn't consumed by the release it forces. Left in place, every later release PR proposes 1.0.0 again, the tag already exists, and the release wedges. This repo has already been bitten by that once (an override pinning 0.7.0 that outlived its release). So `docs/RELEASING.md` grows a section saying, in order: delete the override once `v1.0.0` publishes, delete `bump-minor-pre-major` / `bump-patch-for-minor-pre-major` with it (they only apply while the major is `0`, so they go inert at 1.0.0), land it as a `chore:` PR.

That section also spells out what changes for PR titles afterwards. Those two pre-major keys are why `feat:` has been landing as a *patch* bump for the whole `0.x` line; once they're gone the usual mapping applies — `fix:` → patch, `feat:` → **minor**, `feat!:` / `BREAKING CHANGE` → **major**. A `feat!:` title after 1.0.0 cuts 2.0.0, not 1.0.1.

I left the two pre-major keys in place here rather than removing them in the same commit: they're inert at 1.0.0 either way, and keeping them means this PR can't quietly change `0.x` bumping behaviour if 1.0.0 doesn't cut for some reason.

### Docs that said "pre-1.0"

Five places used "pre-1.0" to mean "the version number is below 1.0" — statements that go false the moment this lands. Each of them was really an unstable-API caveat, so they now say that directly (`docs/RELEASING.md` daemon artifacts ×4, `docs/NON_GRADLE_INTEGRATION.md` ×2, `docs/daemon/STREAMING.md` status line). No stability guarantee is added or withdrawn — flag it if you'd rather 1.0.0 mean something stronger for the `daemon-*` surface, since that's a call for you, not the version bump.

## Test plan

- `release-please-config.json` parses, and `release-as` is a valid package-level key in the [release-please config schema](https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json) the file already pins via `$schema`.
- Confirmed against release-please's [`src/strategies/base.ts`](https://github.com/googleapis/release-please/blob/main/src/strategies/base.ts) that `buildNewVersion` short-circuits to `Version.parse(this.releaseAs)` before any bump computation, and that the `changelogEmpty` skip is not a concern here — there are releasable `fix:` / `feat:` commits on `main` since `v0.19.61`, which is why release PR #3692 already exists.
- Docs-and-config only: no Kotlin touched, and `.github/ci/format-paths.json` ignores `docs/**`, `**/*.md`, and `**/*.json`, so the `format` gate is scoped out.
- Non-visual change — no rendered surfaces affected, so no before/after images.

---
_Generated by [Claude Code](https://claude.ai/code/session_015iYCc36RqdAi8y9MLfvKRo)_

### DIFF
--- a/docs/NON_GRADLE_INTEGRATION.md
+++ b/docs/NON_GRADLE_INTEGRATION.md
@@ -239,7 +239,7 @@ fun main() {
 }
 ```
 
-Maven coordinates (pre-1.0):
+Maven coordinates (unstable API — see the caveat below):
 
 ```kotlin
 implementation("ee.schimke.composeai:render-session-api:<version>")
@@ -321,8 +321,9 @@ Pass `--help` to any of them for the full flag list.
 - **`previews.json` schema versioning.** The wire field is `"schema":
   "compose-previews/v1"`. The daemon tolerates unknown fields; clients
   that produce manifests should round-trip schema-stable fields.
-- **Maven Central artifacts are pre-1.0.** Pin to a specific version;
-  expect API changes across minor versions until 1.0.
+- **These Maven Central artifacts do not carry a stability guarantee
+  yet.** Pin to a specific version; expect API changes across releases
+  until the surface settles.
 
 ## Reference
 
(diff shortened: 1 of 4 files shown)


yschimke:
Codex's P1 landed while this was merging, so the reconciliation is in #3716 rather than here.

The two facts it cites check out — `protocolVersion` is a single `Int` on both `InitializeParams` and `InitializeResult`, with `JsonRpcServer` failing the handshake on mismatch, and there is no `--json-schema` anywhere under `cli/`. Auditing the rest of the policy docs turned up two more in the same family: the `// API: stable` source tags and the `@Stable` / `@Incubating` DSL tiers, neither of which exists in the tree.

Where I disagree is that this blocks 1.0. All four are documented as *post-1.0 evolution* rather than 1.0 gates — `API_STABILITY.md` literally reads "post-1.0; today a single `int`" — and the 1.0 readiness punch list (#798) is closed as completed. Three are ways to *discover* the contract rather than the contract itself, and the missing DSL tier makes the surface stricter, not looser: with no incubating escape hatch, a property is committed the moment it ships. The guarantees that 1.0 actually activates rest on the deprecation policy, the tolerant-decode rules, and the three test layers, all live today.

What was genuinely wrong is that the docs let 1.0 *imply* those four. #3716 names them in a table in `VERSIONING.md § 10`, says at each description site that it isn't in 1.0.0, and rewrites the pre-1.0 disclaimer into what 1.0 does turn on.

---

PR #3714, opened 2026-08-12
feat(serve): show whole design screens with the code laid over them

yschimke:
design-parity's page importer has been able to read a key screen out of a design file for a while — the frame, the rectangle of every component instance on it, and the code handle each maps to. Nothing consumed it on the server, so the only way to look at one was to dispatch a workflow and download a zip from a blob host some networks block. The manifest's own schema names *"a preview server"* as a consumer and carries a preview id on every placement specifically so one could render the component. This is that consumer, plus the publishing path that gets a screen onto a delivery branch.

## The surface

A catalog that publishes backdrops gets an `N screens` action chip beside its comparison actions; one that publishes none is unchanged and the route 404s rather than serving an empty stage.

The page backdrop: a design screen with a rectangle over every component instance, coloured by how it was linked

Green Code Connect, blue design-map, amber name match, dashed red for **unlinked** — a part of the screen with no code behind it, which is the finding the surface exists for. Clicking a rectangle opens that component's viewer; one with no code behind it deep-links into Figma instead.

**The overlay is why this belongs on the server.** The offline viewer design-parity ships takes pre-baked render PNGs as inputs. The server already renders every preview on demand, so laying code over design is just an image element pointed at the existing render lane, and it inherits the live daemon, the visitor's theme, and any refresh — the design half is fixed at import, the code half stays live.

The same screen with the catalog's renders laid over it at the difference blend

A render is pinned top-left and scaled to the placement's **width**, never stretched to the box: a component that renders taller than its slot is a real finding and stretching would hide it. "Only what we don't implement" mutes everything mapped, which is the fastest read of a screen's coverage:

Only the unlinked rectangles, with everything this catalog implements muted

And the screens index, plus the dark rendering:

The screens index — one card per published screen with its coverage

The same screen on a dark surface — the design's own pixels are never tinted

> The backdrop and the renders in these shots are the harness's stand-ins — there is no design file or daemon in a headless capture. The geometry, colouring, controls and blend are the production drawing code.

## What's in it

- **`ServePageBackdrops.kt`** — the serve host's view of the manifest, reading the shapes from `:preview-data-api`'s existing contract mirror rather than re-declaring them, and adding only what is serve-specific: which pages this host will draw, and how to read their bytes without escaping the manifest's directory. Fail-soft throughout — a missing file, an unsupported version, an unroutable id, a traversing image path or a backdrop that isn't a readable PNG drops the record, never the catalog's grid.
- **Routes** — `/{system}/pages`, `/{system}/pages/{id}`, and the backdrop image off the same route with a `.png` suffix, the convention `/reference/{name}` already uses. Both the session-query and canonical path forms.
- **Geometry converted once, server-side.** The manifest is in frame-local design units; the page emits percentages, so `page-backdrop.js` does no geometry at all and the stage survives being responsive, zoomed or printed. Pinned by a test that renders under `Locale.GERMANY`, where a locale-sensitive format call would emit a comma decimal that is not CSS — and no English test run would ever notice.
- **The overlay images ride an inert `<template>`** and are cloned into the stage on first toggle. Template content is parsed but not loaded until adopted, so a visitor who never opens the overlay costs the server nothing — a screen holds a couple of dozen placements and on a live catalog each is a daemon render.
- **`emit-design-pages.mjs` + `design-pages.mjs`** — re-keys each placement from the repo's discovery preview id onto the catalog's serve preview id (the same id problem `design-references.mjs` exists to solve; without it the server gets ids that render nothing), then writes the manifest and one PNG per screen into the bundle. A linked placement the catalog publishes no sticker for keeps its mapping and warns, rather than being dropped and silently understating coverage.
- **Reusable workflow step** — runs the import (it already holds the Figma token) then the emitter, gated on the repo committing a `design-pages.json` with the opt-in set. A repo that has never heard of the feature pays one probe and publishes exactly what it did before.

The server still holds no Figma credential: a rectangle's deep link is reassembled from a validated file key and node id against a literal origin, like the per-preview spec link.

## Future coverage

Wired into the preview harness as `serve-page-backdrop` and `serve-page-backdrop-index`, with the overlay and unlinked-only **states** captured — both are pure runtime CSS states the committed HTML can't hold, so without them a regression in the positioning, opacity or blend would move no baseline. The page-backdrop image lane gets a screen-shaped stand-in rather than the component placeholder, since the hotspots are the subject. The public-landing golden gained a `screenCount` so the new action chip is diffed too.

## Review rounds

Three rounds of automated review, and one of the findings pointed at something bigger: `api/preview-data-api` already carried a mirror of this wire contract, landed on `main` mid-flight. This branch had re-declared the same shapes, which that module's own header rules out — so the store now consumes it, which subsumed two findings on its own (typed `PlacementLink`, and a `ref` an older producer never wrote).

Everything else fixed: the `.png` suffix and `.`/`..` are reserved from page ids; a declared-but-unresolved preview id no longer falls through to the function-name lookup, and an ambiguous fallback across component ids is declined; unsupported `confidence` values are dropped rather than republished (with the strict enum, one bad string would hide every screen); the emitter is fail-soft on malformed JSON *and* malformed collections; backdrop sources are contained by real path plus regular-file and PNG-signature checks; staged backdrops are capped before fetching; the page-backdrop package is pinned and the opt-in probe no longer sees the Figma token; hidden and muted hotspots leave the tab order.

CodeQL flagged the lazy-overlay `src` assignment three times, through a prefix test and then a URL parse. The template above is the resolution: no URL ever becomes a JavaScript string, so there is no sink to sanitise.

**One finding declined**, with reasoning in `acdf31b`: the overlay renders the catalog's default theme rather than the visitor's declared-theme selection. The producer deliberately pins each placement to the light sticker because a design export is a light-mode image, so following a dark selection would make the comparison less faithful, not more. Carrying the selection is a follow-up worth having alongside a light/dark axis on this page, not a correction to this.

## Test plan

- `./gradlew :cli:test` — **1862 tests, 0 failures**, including 13 store, 9 routing and 29 fixture tests. Covers fail-soft loading, path containment, ratio positioning, suffix and dot-segment ids, both URL forms, the chip appearing only when screens exist, and the locale regression.
- `node --test design-pages.test.mjs` — 15 pass, dependency-free like its sibling.
- `npm run harness:snapshot -- --grep page-backdrop` — 2 fixtures × 2 themes plus two states; the overlay shot now asserts a render actually loaded, where `every()` over an empty list had made it vacuously green.
- `ktfmtFormatAll` clean.

Consumer side: yschimke/m3-catalog#52, merged — it turns this on for the Material 3 kit's Upcoming-Mobile screen. Its delivery branch will grow a `pages/` directory on the first publish after this lands.

https://claude.ai/code/session_01PNKmA16L2b1Ngw2gRoz83Z


yschimke:
Fix merge conflicts

yschimke:
Merged `main` (the exploded-viewer conflict in `pages-snapshot.spec.mjs` was two additions to the same image-lane list — both kept) and addressed the review in `acdf31b`. Six of the seven findings were real:

| Finding | Fix |
| --- | --- |
| `.png` not reserved from page ids | Refused in both the server's sanitiser and the producer, so such a page never reaches a delivery branch to be silently dropped. Tests both sides. |
| Declared preview id falling through to the function-name lookup | That emptiness is a *refusal* from the bundle-id collision guard, not a miss — falling back could overlay a sibling variant. The fallback now applies only when no id was declared. |
| Malformed `pages.json` failing the publish | Fail-soft: warns and publishes the catalog without screens, matching the posture the rest of the lane documents. |
| Symlink escaping the containment check | Real-path comparison, plus regular-file and PNG-signature checks, so the published bytes are what the server will accept. |
| `latest` + the Figma token in scope | Pinned to `0.1.47`; the opt-in probe now runs under `env -u FIGMA_TOKEN` since it only reads a committed JSON file. |
| Hidden hotspots still focusable | Hidden *and* muted hotspots leave the tab order and the a11y tree. |

**Declined: carrying the visitor's declared-theme selection into the overlay render URLs.** The producer deliberately pins each placement to the *light* sticker, because a design export is a light-mode image — so following a dark selection would make the comparison less faithful rather than more. The current behaviour is the correct default for the input, not an oversight. Carrying the selection is worth having as a follow-up alongside a light/dark axis on this page, where the design side could be re-exported to match; it isn't a correction to this change.

---

yschimke:
Second round addressed in `ed6bf03`, and one of the findings turned out to point at something bigger.

**`api/preview-data-api` already mirrors this contract.** `PageBackdrop.kt` landed on `main` while this branch was in flight, hand-written and pinned by `PageBackdropParseTest` against the producer's own fixture. This branch had re-declared the same shapes in `serve` — exactly what that module's header rules out, since one foreign contract cannot have two mirrors in a repo without them drifting. `ServePageBackdrops.kt` now consumes those types and keeps only what is serve-specific: which pages this host will draw, and how to read their bytes without escaping the manifest's directory.

The shared mirror is better than what it replaced, and that subsumes two findings outright:

- **Refs omitted by older manifests** — the deep link is now built from the placement's own `nodeId`, which `ref` is *defined* as containing. So an unlinked hotspot, where that link is the only one it has, no longer depends on a field newer than the contract. (`refFor` is there too, and tested.)
- **Unknown link methods** — `PlacementLink` is a typed enum, so an unrecognised value fails the parse rather than degrading per-placement. Harsher, and right: the alternative is guessing what an unknown method means while colouring it as coverage. Additive producer changes still parse, since the shared decoder ignores unknown keys.

Plus the rest of the round:

- **Preview mapping on a degraded link** — `renderablePreviewId` gates on the link as well as the id, so a manifest carrying `unlinked` beside a stale preview id can't put a render *and* a viewer link on a rectangle the same page paints dashed-red.
- **Malformed collections** — `{"pages": {}}` is valid JSON and threw `object is not iterable` into the workflow's `set -e`. Both the collection guards and the `planPageBackdrops` call are inside the fail-soft path now, with tests.
- **Ambiguous function-name fallback** — declined rather than guessed. Two components whose previews share a member name share a bucket, and taking the first would overlay component A inside component B's rectangle. Same posture as the collision guard it sits next to.

**CodeQL** (high, `page-backdrop.js`): only a same-origin relative path is promoted from `data-src` to a live `src` now, and the hover pairing compares attribute values instead of interpolating a design-file node id into a selector. DOM text isn't a trust boundary even when the server wrote it. I re-ran the harness to confirm the overlay still actually loads rather than being silently dropped by the new guard — the `overlay` state shot still composites the renders.

`main` merged twice more along the way. 1836 CLI tests and 13 producer tests green.

---

yschimke:
Fix merge conflicts

yschimke:
Conflicts fixed and the latest review round addressed.

**The conflict was substantive, not textual.** #3709 replaced the catalog landing's run of muted text links with a row of M3 action chips, and my "N screens" link was written in the old idiom. I took that design wholesale and added a screens chip to its list rather than preserving my version — same for the `designToolLabel` parameter, which had collided with `screenCount` in the same argument list. Also caught a second conflicted file (`ServeHttpServer.kt`) that the merge output didn't headline.

The chip renders where it should, between the download and playground actions:

| | |
| --- | --- |
| `download all (.zip)` · **`2 screens`** · `try in playground` | captured in `serve-landing-public` |

I added `screenCount` to that golden deliberately: no landing fixture carried a screens chip, so the new element could have been restyled or lost without moving any baseline.

**CodeQL, third time on the same sink.** After a prefix test and then a URL-parsing guard both failed to register as barriers, I stopped trying to phrase a sanitiser the analysis recognises and removed the sink instead. The overlay images now ride an inert `<template>` with real `src` attributes and get cloned into the stage on first toggle. Template content is parsed but not loaded until adopted, so it's the same laziness with no client-side URL handling at all — no `getAttribute`, no assignment, nothing to validate. The URLs stay where every other URL on the page is built and escaped. The clone lands before the first hotspot so the rectangles keep painting over the renders.

While there I fixed a hole in my own test: the overlay capture waited on `every()` over the render list, which is vacuously true for an empty list — so a regression that armed nothing would have kept it green. It now requires at least one render with a non-zero natural width.

**Three findings, all real:**

- `.` and `..` pass the id alphabet but are *path segments* — a browser normalises `/pages/..` to `/` before the request is sent. Refused in producer and server.
- An unsupported `confidence` is dropped rather than republished. Moving to the shared contract's strict enum made this load-bearing: an unrecognised value fails the parse for the **whole** manifest, so one bad string would hide every screen a catalog publishes.
- Staged backdrops are capped before anything is fetched. Chunking bounded concurrency but not the total.

1862 CLI tests and 15 producer tests green. Separately: the earlier `Actions Script Tests` failure was a checkout dying with git exit 128 — infrastructure, same shape as one that hit `main` — and the `interactive-a-d` e2e failure is on a test path this PR doesn't touch (no `vscode-extension/src` changes at all); I'm waiting on its re-run rather than assuming.

---

yschimke:
**`VS Code Extension E2E (interactive-e-g)` is failing, and I don't believe it's from this change.** Recording the evidence rather than just re-running.

The failure isn't an assertion — it's Gradle discovery coming back empty mid-test:

```
composePreviewRender --preview matched no previews for 'com.example.samplecmp.RedBoxPreview*', …
Manifest contains 0 code preview(s) and 3 asset preview(s) at samples/cmp/build/compose-previews/previews.json.
Available previews:
    lottie/spin.json
    svg/badge.svg
    svg/star.svg
[refresh] done — 0 visible previews in Previews.kt, module has 3
```

Three things point away from this PR:

1. **The diff cannot reach that code path.** `git diff --name-only origin/main...HEAD` touches only `cli/**/serve`, `scripts/design-artifacts`, `vscode-extension/preview-harness` (the Playwright page-capture harness, not the electron e2e), `docs`, and workflows. Nothing in `gradle-plugin/`, `renderers/`, `samples/` or `daemon/` — the entire discovery-and-render path this failed in.
2. **It failed *after* succeeding in the same job.** Earlier in that run: `composePreviewRender: 6 capture(s) drawn on a warm renderer` / `[refresh] rendered 6 preview(s) for Previews.kt`. The module compiled and discovered fine, then the post-edit rediscovery found zero. That's the test's own edit/recompile cycle racing, not a static property of the branch.
3. **It moves between shards.** The previous run failed `interactive-a-d` (a temp preview's PNG missing after edit+refresh — the same edit-then-render family) while `e-g` passed; this run failed `e-g` while `cmp-smoke` and `cached-preload` passed. A deterministic break from a code change doesn't wander.

What I could *not* establish: whether `main` currently fails the same way. Its recent `VS Code Extension E2E` runs are almost all `cancelled` — merges landing fast enough to cancel each other — so there's little completed signal to compare against. The last completed one I found (`63af4978`) had `interactive-a-d` passing and a different job failing at `actions/checkout` with git exit 128, which is the same infrastructure flake that hit `Actions Script Tests` here and has since gone green.

I tried `rerun-failed-jobs` and got `403 This workflow is already running` — `wear` and `interactive-a-d` are still in flight on this run. I'll re-run the failed shard once the run completes and report what it does.

---

yschimke:
Latest round: three fixed in `e3f2aad`, one declined.

**Placements are now capped per page, not just pages per catalog.** Fair catch — last round's `MAX_PAGE_BACKDROPS` bounds how many backdrop PNGs get staged but says nothing about one page's manifest entry, and every placement expands into a hotspot, an overlay image and a list row. Capped at 500; the densest real screen in the Material 3 kit carries eleven.

**An out-of-range `depth` is normalised to 0.** `Number.isInteger(2147483648)` is true, so it was republished and then failed the consumer's Kotlin `Int` decode — taking the whole manifest with it. Exactly the failure shape as the `confidence` finding: one bad field in one placement hiding every screen a catalog publishes.

**The `figma_token` secret description now names both consumers.** It promised the token was used "only" for design-map references, which stopped being true when this branch added the page import. A caller decides whether to hand over a credential from that text, so it now discloses the second consumer, that it's a separately distributed npm package, and the opt-in it's gated on.

**Declined: reserving a session named `pages`.** The shadowing is real, but it isn't something this adds — it's the documented behaviour of the entire route table. From the routing block in `ServeHttpServer.kt`, unchanged by this PR:

> Constant first segments (`/healthz`, `/version`, `/bundle.zip`, `/wasm/…`, …) score higher than `/{system}` in Ktor routing, so they still win — only genuinely unknown single segments fall through to a session lookup.

`/compare`, `/parity`, `/reference` and `/rc-compare` have had precisely this property since before this branch; a session named `compare` is shadowed today. Reserving session names would change behaviour for all of them and belongs in its own change, not smuggled in here.

16 producer tests and the backdrop suites green; `main` merged again.

---

PR #3717, opened 2026-08-12
fix(serve): make link unfurling work in Slack and Google Chat

yschimke:
## Why

Links to `preview.coo.ee` were not unfurling in Slack or Google Chat. The Open Graph markup was never the problem — it is correct and is served to bots, and the images are reachable and fast. I audited the live server against how those services actually fetch:

| Check | Result |
| --- | --- |
| `og:*` block present in initial HTML, bot UAs | ✅ correct on `/`, `/{system}/`, `/{system}/p/{id}` |
| `og:image` reachable by `Slack-ImgProxy` | ✅ 200, `image/png`, ~0.6 s |
| gzip single-encoded, TLS chain, rate limits | ✅ fine |
| **`HEAD` on any route** | ❌ **405 on `/` and `/status`, 404 on everything else** |
| **`og:image:width` / `height`** | ❌ absent; images 300×210 and 216×480 |
| **`robots.txt` / `sitemap.xml`** | ❌ both 404 (styled HTML error page) |
| **`Cache-Control` on every HTML page** | ⚠️ `private, no-store`, `Vary: Cookie` |

The HEAD result is the confirmed defect. `ServeHttpServer` registers 60 `get`, 14 `post`, 3 `delete` and **zero `head`** routes, and `AutoHeadResponse` was never installed — so the probe an unfurler sends before committing to a download got a 4xx across the whole site. Link checkers and `curl -I` were being told the same thing.

## What changed

**HEAD answers GET on the cheap lanes.** `install(AutoHeadResponse)` — the plugin replays the GET pipeline and drops the body, so content type, length, `Cache-Control`, ETag and the generation marker all match the GET, which is what the probe is asking about.

**…and is refused on the lanes that do work.** Replaying the GET pipeline is right for a page or a baked PNG and wrong for anything that *makes* something: `HEAD /bundle.zip` would render every preview in the catalog and pack a zip only to discard it, so an anonymous `curl -I` — or a link checker, or an uptime monitor — could cold-start a catalog's daemons and burn its render capacity while downloading nothing. `/bundle.zip`, `/bundle/{name}`, `/history/render`, any `/render` carrying override params, and the non-PNG render products (`.svg`, `.slots`, `.a11y`, `.annotations`, `.rc` — all produced on demand, query or not) answer `405` + `Allow: GET`. Only `<id>.png` or no suffix is a baked replay, and that is exactly what an unfurler probes. The token gate runs **before** this refusal, so a gated host still answers its concealment 404 rather than advertising that a route exists.

**The unfurl image declares its size.** New `ServeHost.bakedRenderSize` reads the PNG's IHDR (8 bytes, no decode, no fetch — reusing the existing `readPngSize`) so pages emit `og:image:width`/`height`. That also sizes the card honestly: a 300×210 component render now asks for `summary` instead of the `summary_large_image` it could never fill. The front door advertises the **full render** rather than the downscaled `/hero/` thumbnail it lays out — that one is card-sized (216×480), under every unfurler's floor and under Google's 512² guidance. The size is remembered in `CatalogMeta`, not read live, because `peekHost` returns null for a suspended catalog and a live lookup would silently fall back to advertising the thumbnail whenever the featured catalog happened to be idle. Dimensions are omitted when the URL carries overrides, since the image URL inherits the page's query and the baked size would then describe a different render.

**Real `robots.txt` and `sitemap.xml`** (new `ServeSiteIndex`, pure and unit-tested). robots.txt splits on cheap published bytes vs. work: landings, viewers and baked PNGs stay open; the playground, `bundle.zip`/wasm, `history/render`, the compare/parity/reference and machine lanes are closed — as is **a render URL carrying a query string**, which is the rule that actually protects the daemons, since `/…/render/<id>.png` serves a baked file but the same path with `?theme=` re-renders and the grid links the override forms. That rule is scoped to the render lane rather than to every query string on the server: a blanket form would also close `…/compose-m3/?tab=components&theme=…`, which is precisely the link someone pastes into a chat after picking a theme, and Googlebot obeys this group. Link unfurlers get a wildcard-free group with no crawl delay (their parsers are simple-prefix, and a delayed unfurl is a missing unfurl); `Googlebot` is deliberately **not** in it — robots groups are winner-takes-all, so naming it there would exempt Google's whole fleet from every rule above, and it doesn't need the exemption. sitemap.xml lists the front door, each listed catalog landing and one URL per preview viewer, stamped with `<lastmod>` from the catalog's own `generatedAt` (not the refresh clock, which moves on every poll whether or not bytes changed), built off remembered metadata via `peekHost` so it never wakes a daemon. A token-gated host disallows everything and serves no sitemap.

**Anonymous pages are storable again.** `pageCacheControl` gains a request-scoped `signedIn`: an anonymous request on a `--github-auth-*` server gets the signed-out rendering everyone gets, so it is now `public, max-age=0, s-maxage=300, must-revalidate` with `Vary: Cookie` instead of `private, no-store`. `max-age=0` keeps the browser exactly where `no-store` had it — every visit revalidates, so the sign-in chip can never be replayed stale, and there is deliberately no `stale-while-revalidate`. Signed-in requests and token-gated hosts are unchanged. The one thing lost is `no-store`'s bfcache block: a visitor who signs in and presses Back may see signed-out chrome until they reload.

**A HEAD no longer counts as a view** — `AutoHeadResponse` replays the GET pipeline, which would otherwise tally the probe *and* the fetch that follows it, double-counting every shared link.

## Generated output

<details>
<summary><code>/robots.txt</code> (public host)</summary>

```
# compose-preview — https://github.com/yschimke/compose-ai-tools
# Catalog landings, preview viewers and their baked PNGs are open to crawl.
# Render-on-demand, code-running and bulk-download lanes are not.

User-agent: *
Disallow: /admin/
Disallow: /auth/
Disallow: /docs
Disallow: /d/
Disallow: /playground
Disallow: /api/
Disallow: /*/api/
Disallow: /status
Disallow: /healthz
Disallow: /readyz
Disallow: /version
Disallow: /parity
Disallow: /*/parity
Disallow: /compare
Disallow: /*/compare
Disallow: /reference/
Disallow: /*/reference/
Disallow: /rc-compare/
Disallow: /*/rc-compare/
Disallow: /history/render/
Disallow: /*/history/render/
Disallow: /bundle.zip
Disallow: /*/bundle.zip
Disallow: /bundle/
Disallow: /*/bundle/
Disallow: /wasm/
Disallow: /rc-player-wasm/
Disallow: /rc-player/
Disallow: /doc-player/
Disallow: /index.json
Disallow: /*/index.json
Disallow: /iframe.html
Disallow: /*/iframe.html
Disallow: /*.svg$
Disallow: /*.rc$
# Override params re-render a preview; the bare path serves the bake.
Disallow: /render/*?
Disallow: /*/render/*?
Crawl-delay: 10

# Link unfurlers: one fetch per shared URL, and they need the og:image too.
User-agent: Slackbot-LinkExpanding
User-agent: Slackbot
User-agent: Slack-ImgProxy
User-agent: Twitterbot
User-agent: facebookexternalhit
User-agent: LinkedInBot
User-agent: Discordbot
User-agent: TelegramBot
User-agent: WhatsApp
User-agent: SkypeUriPreview
User-agent: Iframely
Disallow: /admin/
Disallow: /auth/
Disallow: /docs
Disallow: /d/
Disallow: /playground

Sitemap: https://preview.coo.ee/sitemap.xml
```

</details>

<details>
<summary><code>/sitemap.xml</code></summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <url>
    <loc>https://preview.coo.ee/</loc>
    <lastmod>2026-07-17T12:34:56Z</lastmod>
  </url>
  <url>
    <loc>https://preview.coo.ee/compose-m3/</loc>
    <lastmod>2026-07-17T12:34:56Z</lastmod>
  </url>
  <url>
    <loc>https://preview.coo.ee/compose-m3/p/button-filled__ideal__default__dark</loc>
    <lastmod>2026-07-17T12:34:56Z</lastmod>
  </url>
  ...
</urlset>
```

</details>

No visual evidence: nothing rendered changes. The diff is HTTP-layer plumbing, `<meta>` tags, two new text/XML endpoints and response headers — no `@Preview` output, page chrome, overlay or theme is touched.

## Test plan

- `./gradlew :cli:test` — green (1300+ tests), plus `ktfmtCheckAll`.
- New `ServeSiteIndexTest`: token-gated host disallows everything and advertises no sitemap; query strings closed on the render lane only, with the blanket `Disallow: /*?` asserted absent so a stateful page URL stays fetchable; the link-preview group carries no wildcard or end-anchor rule and can never contain `Disallow: /`; general-purpose indexers stay in the restricted group; `<lastmod>` stamping; an unparseable `generatedAt` drops only that timestamp rather than invalidating the whole document; a well-shaped but impossible date (`2026-13-40T25:99:99+99:99`, `2026-02-30`) is rejected too, since a date a crawler refuses invalidates the file; ids needing percent-encoding survive both the URL and the XML.
- New in `ServeHttpRoutingTest`: HEAD answers with the same status, content type and cache-control as GET across eight representative routes (front door, status, version, catalog landing, viewer, render PNG, robots, sitemap); HEAD is refused with `405` + `Allow: GET` on the work lanes and on all five non-PNG render suffixes while the baked PNG still answers `200`; an unauthenticated HEAD on a gated host is concealed with `404` and no `Allow` header, and answers `405` once the token is supplied; a HEAD probe does not move the view tally; an overridden viewer link declares no image dimensions; robots.txt opens the browse surface and closes the render/code lanes; the sitemap lists landings and viewers but not images, closed lanes, or unlisted catalogs.
- New in `ServeWebTest`: a known image size is declared and picks `summary_large_image`; a 300×210 render drops to `summary`; unknown size omits the dimensions and keeps the large card; a half-known size is omitted entirely.
- New in `ServeAuthTest`: the anonymous-on-auth-server header, no `stale-while-revalidate`, and token-gated hosts staying on `no-store`.
- Updated one existing routing assertion: "the front door does not put a render request on the server" is now scoped to `src`/`href` attributes rather than the whole body, because `og:image` deliberately names the render — a meta tag is inert until an unfurler reads it, and one baked PNG per shared link is not the per-visitor render load that assertion guards.

https://claude.ai/code/session_01CaAPNqjKC8rqonbraRjid3

---
_Generated by [Claude Code](https://claude.ai/code/session_01CaAPNqjKC8rqonbraRjid3)_


yschimke:
All five review findings were real; fixed in 6fb0dee.

**P1 — bulk renders on a HEAD probe.** Correct, and I introduced it. `AutoHeadResponse` reruns the whole GET handler, so `HEAD /bundle.zip` rendered every preview and packed a zip only to discard it — anonymously repeatable on a public host. The work lanes (`/bundle.zip`, `/bundle/{name}`, `/history/render`) now answer `405` + `Allow: GET`, as does an override-bearing `/render`. The bare render path an unfurler probes for `og:image` still answers, which is the only part the plugin was installed for.

**Query rule blocking stateful shared links.** Correct, and it contradicted my own rationale in the file. Narrowed from `/*?` to `/render/*?` + `/*/render/*?`, so `…/compose-m3/?tab=components&theme=…` — the link someone actually pastes after picking a theme — stays fetchable by the group Googlebot obeys, while the override render lane stays closed.

**Dimensions vs override render URLs.** Correct. `imageUrl` inherits the page's query suffix, so the baked size could describe a different image. Dimensions are now omitted whenever the query names any override param (`ServeOverrides.isOverrideParam` — a cheap key-level test that fails safe, since omitting just means the fetcher measures). Same guard on the catalog landing's hero.

**Homepage unfurl image after suspension.** Correct, and worse than an edge case — which catalog is featured depends only on publisher grouping, so a suspended featured catalog is the usual state on a quiet server. The hero's full-render size is now remembered in `CatalogMeta` alongside `heroImage`, and the homepage reads that instead of `peekHost`.

**Sitemap timestamp validation.** Correct. The regex passed `2026-13-40T25:99:99+99:99`. Now shape-checked by the regex (the sitemap profile is narrower than what `java.time` accepts — a local date-time with no offset is rejected) and then parsed with `LocalDate`/`OffsetDateTime`, emitting only on success.

Each has a test: `HEAD is refused on the lanes whose GET does real work`, `query strings are closed on the render lane only`, `an overridden viewer link declares no image dimensions`, `a well-shaped but impossible date is rejected too`. Full `:cli:test` and `ktfmtCheckAll` green.

One assertion needed loosening rather than fixing: the routing test's "must not close the browse surface" check compared substrings, and the legitimate `Disallow: /*/render/*?` rule has the open form as a prefix — it now compares lines exactly.

---

yschimke:
Both second-round findings were real too; fixed in 3f641ee.

**Token gate ordering.** Correct — my guard ran ahead of `rejectBadToken`, which answers 404 specifically to conceal that a route exists. Refusing first with `405` + `Allow: GET` said the opposite, and since `/history/render` is only registered when the repository-backed surface is enabled, the difference was a probe for optional configuration. The gate now runs first on all four handlers.

**HEAD on non-PNG render products.** Correct, and the more useful catch of the two: I keyed the render guard on override params alone, so an override-free `HEAD …/render/{id}.svg` (or `.slots`, `.a11y`, `.annotations`, `.rc`) still ran the whole handler and took the render semaphore to have its body discarded. Only `<id>.png` or no suffix is a baked replay, so those suffixes now refuse on the suffix alone — which is the more robust half of the check, since it doesn't depend on the caller spelling a query.

Tests: `HEAD is refused on the lanes whose GET does real work` covers all five suffixes, and a new `an unauthenticated HEAD on a gated lane is concealed, not refused` stands up a token-gated server and asserts 404 with no `Allow` header, plus 405 once the token is supplied. Full `:cli:test` and `ktfmtCheckAll` green.

---

PR #3728, opened 2026-08-12
docs(versioning): stop overselling API stability

yschimke:
## Summary

`VERSIONING.md` and `API_STABILITY.md` describe enforcement machinery that mostly isn't built, and #3716 made it worse — its § 10 asserted the guarantees were "live today". They aren't. This walks that back and marks each site as intent rather than fact.

Every claim below was checked against the tree, not inferred from the docs:

| Claim in the docs | What's actually there |
|---|---|
| Kotlin BCV gates `:gradle-plugin` / `:preview-annotations` (named in **three** places) | No BCV plugin applied, no `.api` golden files, no `apiCheck` in any workflow. A breaking API change ships silently |
| Wire enums decode tolerantly to `UNKNOWN` | Most `Messages.kt` enums are plain enums — `FileKind`, `ChangeType` among them. `ignoreUnknownKeys` covers unknown *fields*, not unknown enum *values* |
| `apply()` fails an out-of-matrix consumer with "supports AGP A.B–C.D; found E.F" | `apply()` checks the Gradle version only. `doctor` reports skew, which is a report, not a gate |
| `previews.json` + `HistoryEntry` carry `schemaVersion` | Neither writes one — not in the daemon's `PreviewManifest`, the viewer's, or `HistoryEntry` |
| A wire message without a fixture fails CI | `fixtureInventoryMatchesExpected` compares against a hand-maintained set, so a new message lands green. History, interactive, stream, XR, recording and extension methods are already uncovered |
| VS Code supports daemon `protocolVersion` N..N-1 | `daemonProtocol.ts` exports one hard-coded `PROTOCOL_VERSION = 2`; `handleInitialize` rejects anything else. The staggered-release case the window exists for fails the handshake |
| "Anything not in this list is internal" | `bundle.json` (CLI + bundle-viewer) and `daemon-launch.json` (VS Code's `daemonProcess.ts`, plus the documented non-Gradle integrations) are versioned formats read by detached consumers and appear in no list |

**What the docs now say.** § 10 becomes "What is and is not enforced": `1.0.0` is a version number that ended the `0.x` free-for-all and states an intent, and it explicitly did not switch on enforcement. What genuinely holds is named — review, the deprecation cycle, and the real coverage the fixture and integration suites do have — followed by a table of what's described but missing. Each description site says the same thing locally, so a reader who lands mid-document isn't misled.

Nothing here changes behaviour or weakens a contract we were actually keeping; it stops claiming ones we weren't.

## Test plan

- Each row verified directly: `grep` for `binary-compatibility-validator` / `apiValidation` / `*.api` / `apiCheck` → nothing; `UNKNOWN` appears 4× against 31 `enum class` in `Messages.kt`; `GradleVersionCheck.kt` is the only `apply()`-time gate; `PreviewManifest` (both copies) and `HistoryEntry` have no `schemaVersion`; `MessagesTest.fixtureInventoryMatchesExpected` is a literal `setOf(...)`; `PROTOCOL_VERSION = 2` in `daemonProtocol.ts` vs the `!=` rejection at `JsonRpcServer.kt:747`.
- Anchors updated where § 10's heading changed — no `#10-status-at-10` references remain anywhere in the repo.
- Docs only; `.github/ci/format-paths.json` ignores `docs/**`, so the `format` gate is scoped out.

## Follow-ups this implies

Not in this PR, and each is additive whenever it's wanted: wire up BCV, retrofit `UNKNOWN` onto the wire enums, add the two `schemaVersion` carriers, derive the fixture inventory from the protocol surface, and either implement the N-1 window or drop it from the docs.

---
_Generated by [Claude Code](https://claude.ai/code/session_015iYCc36RqdAi8y9MLfvKRo)_

### DIFF
--- a/docs/API_STABILITY.md
+++ b/docs/API_STABILITY.md
@@ -4,7 +4,9 @@ Cross-cutting design doc for what counts as a public contract in this repo, how 
 
 ## 1. The contracts
 
-Nine externally-observable surfaces, each with its own evolution story. Anything **not** in this list is internal and may move without notice.
+Nine externally-observable surfaces, each with its own evolution story. Anything **not** in this list is internal and may move without notice — with the caveat below, and read alongside [VERSIONING.md § 10](VERSIONING.md#10-what-is-and-is-not-enforced), which records how much of the machinery described here is actually implemented.
+
+> **The list is not exhaustive.** At least two versioned formats are exchanged with detached consumers without appearing above: `bundle.json` inside a packed preview bundle (read by the CLI and the bundle viewer, carrying `BUNDLE_SCHEMA_VERSION`) and `daemon-launch.json` from the published `daemon-launch-builder` (read by VS Code's `daemonProcess.ts`, which gates on its `schemaVersion`, and by the non-Gradle integrations in [NON_GRADLE_INTEGRATION.md](NON_GRADLE_INTEGRATION.md)). Both carry their own schema version and both would strand existing artifacts or producers if changed carelessly. Treat "not in this list" as "not yet classified", not as licence to break them.
 
 | # | Surface | Lives in | Consumed by | Stability tier |
 |---|---|---|---|---|
@@ -24,7 +26,7 @@ The annotation library and per-data-product schemas are intentionally narrow. Th
 
 ### 2.1 Daemon JSON-RPC (surface 1)
 
-**Negotiation:** `initialize` request carries a single `protocolVersion: Int`; the daemon answers with its own plus a `ServerCapabilities` bag, and any mismatch is `InvalidRequest` and the daemon exits. Range negotiation — client and daemon each sending `{min, max}` and operating at `min(client.max, server.max)` — is **not in 1.0.0**; it lands in a later `1.x` (see [VERSIONING.md § 10](VERSIONING.md#10-status-at-10)).
+**Negotiation:** `initialize` request carries a single `protocolVersion: Int`; the daemon answers with its own plus a `ServerCapabilities` bag, and any mismatch is `InvalidRequest` and the daemon exits. Range negotiation — client and daemon each sending `{min, max}` and operating at `min(client.max, server.max)` — is **not in 1.0.0**; it lands in a later `1.x` (see [VERSIONING.md § 10](VERSIONING.md#10-what-is-and-is-not-enforced)).
 
 **Feature detection:** capability bag, never daemon `semver`. The bag already covers `supportedOverrides`, `dataProducts`, `dataExtensions`, `previewExtensions`, `knownDevices`, `backend`, `androidSdk`, `recordingFormats`, `interactive`, `recording`. New features add a capability entry, not a behavior change under an existing field.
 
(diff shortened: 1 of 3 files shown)


yschimke:
All four P2s were right, and they're the same complaint this PR makes — aimed at its own replacement text. Verified each and pushed corrections in `1dd428a`:

- **Fixture round-trips** — `daemonProtocol.test.ts` reads via `JSON.parse(...) as T` and asserts selected properties on 16 of the 28 fixtures. That's a parse-and-spot-check, not a round-trip; Kotlin is the only side that round-trips. Fixed in both `VERSIONING.md § 9` and `API_STABILITY.md § 2.1`.
- **Integration coverage** — it's two different things and the table conflated them. A PR runs one cell (`wear-os-samples (ComposeStarter)`), `agp8-min` is skipped there, and a safe-paths diff skips the matrix entirely with the required legs re-emitted green; the full matrix runs on `main` and the nightly cron. Now stated as PR-scope vs main/nightly-scope separately.
- **`RENDERER_COMPATIBILITY.md`** — titled "Renderer compatibility notes", and that's what it is: skew failure modes, mitigations, desktop Compose/Skiko floors. No supported-combination table. "Collects the known-good combinations" pointed readers at a matrix that doesn't exist, so § 2.5 now says the matrix is documented nowhere.
- **`daemon-launch.json`** — `daemonProcess.ts` requires `schemaVersion` to equal its own exactly and discards the descriptor otherwise. Fail-closed-and-rebuild, not a tolerant read.

The one thing I did not act on is the implied ask to make `NON_GRADLE_INTEGRATION.md`'s v1 guidance agree with the v2 gate. That's a real inconsistency, but it's a docs-correctness bug in its own right rather than part of scaling back the stability claims, so it wants its own change.

---

yschimke:
Three more, all confirmed. Pushed in `8379fb8`.

Two were self-contradictions the earlier corrections created — rules left standing that the corrections themselves had made unfollowable:

- **§ 6** required one release of warning from `apply()` before dropping a toolchain corner, while the paragraph directly below established `apply()` never sees an AGP/Kotlin/Compose version. A maintainer could follow the documented process and consumers would get no notice whatsoever. The warning now goes to the CHANGELOG and release notes, with the missing in-build channel stated.
- **§ 4.5** required daemon support for the previous protocol version for one minor cycle, which the same section then said is impossible. Removed from the requirement list, with why it's held back and a note to restore it in the change that implements range negotiation. A requirement nobody can satisfy is worse than an acknowledged gap.

The third is the more useful catch: **the DSL surface pointed at the wrong artifact.** `PreviewExtension` / `DaemonExtension` are in `:gradle-plugin-config`, published as `compose-preview-config`, which the runtime plugin `api`-depends on — and which Gradle conflict-resolves to a single version when a consumer-pinned config plugin meets a CLI-injected runtime on the buildscript classpath. `AGENTS.md` already calls it "the one artifact that must stay binary-stable across versions"; the stability docs were pointing at `:gradle-plugin` and leaving it unclassified. Surface 4 now names it, § 2.4 leads with where the DSL lives and why that artifact is load-bearing, and the BCV row notes the named modules are the wrong target as well as ungated.

---

yschimke:
Right, and it's the last sentence of § 10 that was wrong — "each is additive and can land in any `1.x`". Fixed in `bc5765f`.

`protocolVersion` is a number on both sides (`Int` in `Messages.kt`, `number` in the TypeScript interface and PROTOCOL.md § 3), so putting `{min, max}` in that field makes either peer fail to decode `initialize` before it can negotiate anything — the one message where a decode failure is guaranteed fatal. § 10 now names the two real paths: a coordinated `protocolVersion` bump, or a staged rollout that adds the range as a new optional field existing peers ignore, retiring the numeric one a cycle later.

Worth stating plainly since it makes the deferral less casual than "additive, land it whenever": only the staged path avoids the coordinated release that range negotiation exists to make unnecessary in the first place.

---

PR #3736, opened 2026-08-12
feat(apply): add fork-safe render/publish phase split

yschimke:
## Why

`apply` can't post a preview comment on a PR raised from a **fork**. On a `pull_request` event from a fork, `GITHUB_TOKEN` is read-only for every scope regardless of the workflow's `permissions:` block — so the render push 403s, and the sticky comment can't be posted either, since commenting is a write too.

The trigger that *does* hand out a write token, `pull_request_target`, is the wrong tool here: rendering previews means checking out the PR's own code and running its Gradle build, so a job holding that token would be executing untrusted code with write access to the repository.

This came out of reviewing [google/horologist#2783](https://github.com/google/horologist/pull/2783), which hit exactly this and landed on `pull_request_target` + `ref: head.sha` as the workaround.

## What

Split the action across two jobs instead of two triggers:

| `phase` | Runs | Needs |
|---|---|---|
| `all` (default, unchanged) | render + push + comment in one job | write token |
| `render` | pipelines only, stages the handoff, then stops | nothing — read-only token, no secrets |
| `publish` | restores the handoff, pushes and comments | write token, but runs the **base branch's** code and never executes anything from the PR |

`publish` implies `mode: comment`, installs no CLI, and picks the render job's pipeline selection back up, so a consumer passes `phase` and `pr-number` rather than repeating the whole invocation.

Nothing changes for existing consumers: `all` is the default and every existing step keeps its original `if:` condition, now wrapped with a phase guard.

## Design notes

**Where the seam is.** Steps up to and including `Run pipelines` are the render half; everything from `Push compose baseline / PR renders` onward is the publish half. Two surfaces (a11y, notifications) already re-render their comment body *after* the push to pin image URLs to the pushed SHA, so deferring the publish side was an established pattern rather than a new one.

**The artifact is treated as hostile.** It is produced by a job that ran the PR's own Gradle build, and the pipelines write their push metadata as each surface completes — so a later pipeline, still executing fork code, can rewrite what an earlier one staged. Push *control* therefore never comes from the artifact: destination branch, commit message and skip flag are rebuilt on the publish side from the same literals the single-job path uses, so a `_push_branch` rewritten to `main` aims nothing at the default branch. A `.github/` tree found in a staging dir is dropped before the push, since `push-branch.sh` commits the directory wholesale and a workflow file landing on a render branch would run on its own `push` trigger. The staging dirs contribute pixels; every decision is made from the action's own constants.

**What has to travel between the jobs**, because the publish job cannot recover it:

- `_pr_number` — `github.event.workflow_run.pull_requests` is empty for fork PRs, which is the one case this whole path exists for.
- `_scope_modules` — the publish job never checked the PR out, so it can't classify the diff. Without the render job's scope, a change-scoped run would report every unrendered module's baseline as a *Removed* preview.
- `_pipelines` — the resolved `only` / `skip` set. Re-resolving from the publish call's own inputs would default all four on, and the upsert for a pipeline that never ran would patch its sticky comment to "no changes". Parsed strictly: four `0`/`1` fields, anything else fails the run rather than being coerced into an enabled pipeline.
- `_ab_config` — read from the checkout, which on the publish side is the base branch, so a PR that adds or edits an A/B config would otherwise be graded against the old grouping.

Both sides of every comparison travel, not just the new renders, because the comparators fail closed on a missing baseline: an absent `_resource_baselines/renders` turns renderer anti-aliasing noise into a false resource diff, and a missing `_notification_baseline_findings.json` reads as an empty baseline, reporting every surviving preview as *Added* and losing removals.

**Why the envelopes get rewritten.** The CLI emits **absolute** `pngPath` values pointing into the render runner's Gradle build directories, which don't exist on the publish runner. New `stage-handoff` and `stage-handoff-resources` subcommands copy the renders the publish side still opens and rebase every path onto a workspace-relative `handoff-dir/current/`. Resource captures need this as much as composables do — `compare-resources` runs them through the same `_is_changed` perceptual filter.

**Why only some renders are staged.** `compare` reaches for a current PNG solely through `_is_changed`, and every call site guards on the key being present in the baselines. So a *new* preview is never re-read — its pixels already travel in `_pr_renders`, and staging them again would double the artifact on precisely the run (no baseline branch yet) where every preview is new. An *unchanged* row isn't read either, since `_is_changed` fast-paths on a sha match before touching disk. Both still get their path rewritten, so path resolution stays uniform.

## Test plan

- `python3 .github/actions/lib/test_compare_previews.py` — 110 pass, 11 new. Covers: changed rows staged and rebased, unchanged rows rebased but not copied, new rows not staged, no-baselines staging nothing, multi-capture `#N` keys lining up with `load_cli_output`, failed renders (no PNG) left alone, the resource variants of the same, and round trips that restore each bundle into a *different* workspace root and assert the relative paths resolve there.
- Simulated the `render` → `publish` handoff end to end with the bash steps extracted verbatim from `action.yml`: staged 1 of 3 renders (the sha-differing one), restored the bundle into a fresh workspace, and generated a correct comment body with commit-pinned `raw.githubusercontent.com` URLs and the change-scope note carried across.
- Replayed the push-control attack against the restore step: a handoff with `_push_branch` = `main`, `_skip_if_unchanged` = `1` and a planted `.github/workflows/evil.yml` comes out aimed at `compose-preview/pr`, with the workflow file dropped and the renders intact.
- `action.yml`, every bash step body, and all four README workflow examples parse.

Not exercised here: a real fork PR against a live repo. The reference workflows in the new README section are the intended first consumer.

## Docs

New **Fork PRs** section in `.github/actions/apply/README.md` with both workflows (including the concurrency groups that keep overlapping `synchronize` events from letting an older render's publisher overwrite the shared branches with stale pixels), a table of what travels in the handoff, and an explicit note on what the publish job does and does not make trusted — the render job still runs fork code, it just has nothing worth stealing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UzGvRNrQeHD1f4myvDX2iA)_

### DIFF
--- a/.github/actions/apply/README.md
+++ b/.github/actions/apply/README.md
@@ -154,6 +154,13 @@ permissions:
   contents: write
   pull-requests: write
 
+# Two `synchronize` events on one PR would otherwise race: the older render
+# can finish last, and its publisher then overwrites the shared render
+# branches and sticky comments with stale pixels.
+concurrency:
+  group: compose-preview-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   render:
     runs-on: ubuntu-latest
@@ -188,6 +195,13 @@ permissions:
   contents: write         # renders push to compose-preview/pr
   pull-requests: write    # sticky comment upsert
   actions: read           # cross-run artifact download
+
+# Publishers can finish out of order too, so serialize them per head branch.
+# NOT cancel-in-progress: a publisher that is already pushing should finish,
+# and the next one supersedes it on the same shared branch anyway.
+concurrency:
+  group: compose-preview-publish-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: false
 
 jobs:
   publish:
@@ -238,6 +252,15 @@ Two values are in there because the publish job cannot recover them itself:
 |---|---|
 | `_pr_number` | `github.event.workflow_run.pull_requests` is **empty for fork PRs** — the one case this whole path exists for. |
 | `_scope_modules` | The publish job never checked the PR out, so it can't classify the diff. Without the render job's scope, [change-scoped runs](#change-scoped-rendering) would report every unrendered module's baseline as a *Removed* preview. |
+| `_pipelines` | The resolved `only` / `skip` set. Re-resolving from the publish call's own inputs would default all four on, and the upsert for a pipeline that never ran would patch its existing sticky comment to "no changes". |
+| `_ab_config` | Read from the checkout, which on the publish side is the *base* branch — so a PR that adds or edits an [A/B config](#ab-comparison-of-preview-variants) would otherwise be graded against the old grouping. |
+
+Both sides of every comparison travel, not just the new renders: the
+comparators fail closed on a missing baseline, so an absent
+`_resource_baselines/renders` would turn renderer anti-aliasing noise into a
+false resource diff, and a missing `_notification_baseline_findings.json` reads
+as an empty baseline — reporting every surviving preview as *Added* and losing
+removals entirely.
 
 `_previews.json` is rewritten on the way in. The CLI emits **absolute**
 `pngPath` values pointing into the render runner's Gradle build directories, so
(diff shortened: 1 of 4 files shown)


yschimke:
All six Codex findings were real. Fixed in `3a475d7`.

**P1 — artifact-provided push branches.** This one defeated the premise of the PR, so thanks for it. Push control no longer comes from the handoff at all: the destination branch, commit message and skip flag are rebuilt on the publish side from the same literals the single-job path uses. The staging dirs contribute pixels; every decision is made from the action's own constants.

I also took the adjacent vector, which the finding implies but doesn't name — `push-branch.sh` does `git add -A` on the stage dir, so a `.github/workflows/*.yml` planted there would land on a render branch and run on its own `push` trigger. `GITHUB_TOKEN` is refused workflow writes unless the caller granted `workflows: write`, but that's the caller's setting to get wrong, so a `.github/` tree in a staging dir is now dropped before the push.

Verified by replaying the attack against the restore step: a handoff with `_push_branch` = `main`, `_skip_if_unchanged` = `1` and a planted `evil.yml` comes out aimed at `compose-preview/pr`, with the workflow file gone and the renders intact.

**P1 — notification baseline.** Correct, `_notification_baseline_findings.json` wasn't staged. Now is.

**P2 — resource images.** Correct, and it went one step further than the finding: `_resource_baselines` wasn't staged *and* resource captures carry absolute `pngPath` values that go through the same `_is_changed` perceptual filter, so they needed the rebasing composables already got. Added a `stage-handoff-resources` subcommand with tests, including a round trip that restores into a different workspace root.

**P2 — active pipelines.** Correct. The resolved `only` / `skip` set now travels as `_pipelines`, so the publish call still needs only `phase` and `pr-number`. It's parsed strictly — four `0`/`1` fields, anything else fails the run rather than being coerced into an enabled pipeline, since that file crosses the same trust boundary.

**P2 — A/B configuration.** Correct, staged as data and restored at the configured path, which matches single-job behaviour.

**P2 — concurrency.** Correct. Both reference workflows now carry groups: `cancel-in-progress` on the render side, and *not* on the publish side — a publisher mid-push should finish, and the next one supersedes it on the same shared branch regardless.

110 unit tests pass (11 new).

---

PR #3737, opened 2026-08-12
fix(apply): close publish-side escalation paths in the fork handoff

yschimke:
## Why

Follow-up review of #3736 (already merged) found the fork handoff reaching control surfaces the first hardening pass didn't cover. #3736 stopped the publish job from taking its *push target* out of the artifact; these are the routes that got there anyway.

All four P1s are live on `main` right now.

## The escalation paths

**`_scope_modules` reached `$GITHUB_OUTPUT` unvalidated.** It's `cat`'d from the handoff and written as `echo "modules=$MODULES" >> "$GITHUB_OUTPUT"`. A newline closes the `modules=` line and lets the remainder declare further outputs, so `full\nmode=baseline` flipped the run into baseline mode — whose push steps read their destination branch from `_baselines`, which the restore step does not rebuild. Full loop back to a write-scoped push at an attacker-named branch.

The first fix I wrote for this was wrong, and the test caught it:

```
grep -qE '^(full|[A-Za-z0-9_.:/-]+(,…)*)$'
  "full\nmode=baseline"  -> ACCEPT   # grep tests line by line; line 1 matches
```

Replaced with a whole-string shell glob, which sees the newline because `\n` isn't in the permitted set:

```
case "$MODULES" in *[!A-Za-z0-9_.:/,-]* | ,* | *, | *,,*) reject ;; esac
  "full"                 -> ACCEPT
  "app,feature:foo"      -> ACCEPT
  "samples/android,app"  -> ACCEPT
  "full\nmode=baseline"  -> REJECT
  "full\nmode<<X\n…"     -> REJECT
  "app;rm -rf /"         -> REJECT
  "app$(id)"             -> REJECT
```

**Publish accepted `mode: baseline`.** Documented as comment-only, but only `auto` was remapped, so passing `baseline` explicitly (or injecting it via the above) reached the baseline push steps. Publish is now comment mode unconditionally and anything else is rejected *before* the artifact is restored. The baseline staging dirs also get their push metadata stripped on restore — they're comparison inputs here, never push sources.

**`pr-number` wasn't bound to the run that produced the artifact.** Gradle returning doesn't mean the PR's processes exited, so one can outlive the render and rewrite `_pr_number` before the upload snapshots it — pointing a write-scoped commenter at an unrelated PR with attacker-authored content. The named PR must now have the triggering `workflow_run`'s head repo and branch, or the publisher refuses to comment.

**The render job shared the workflow's write scopes.** An organization can enable write tokens for fork pull requests; the reference workflow's single job requested `contents: write` + `pull-requests: write`, and this action exports `github.token` into the environment the fork's Gradle build runs in. Fork renders are now a separate job with job-level `permissions: contents: read`, which is what makes the isolation independent of that org setting.

## Also

- A `scope=none` render (docs-only PR) staged nothing, so the documented upload's `if-no-files-found: error` turned an intentional no-op into a red workflow that never reached the publisher. It now stages an empty handoff with all pipelines off, and the publisher is a clean no-op.
- The publish example distinguished fork from same-repo runs with a blanket `continue-on-error` on the download, which also swallowed genuine artifact-service and auth failures and let the workflow finish green having published nothing. It now decides from `workflow_run.head_repository` and lets real download failures fail.
- Documented that comment-shaping inputs (`comment-on-empty-diff`, `rerun-checkbox`, `ab-config`) are consumed publish-side and must be repeated there. `only` / `skip` are the exception — those travel in the handoff.

## Test plan

- `python3 .github/actions/lib/test_compare_previews.py` — 110 pass.
- Validation table above run against the actual `case` expression, covering newline-injection, output-delimiter injection, shell metacharacters and command substitution, plus the legitimate scope shapes `compute-scope.py` emits.
- `action.yml`, every bash step body (`bash -n`), and all four README workflow examples parse.

Not exercised here: a real fork PR against a live repo.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UzGvRNrQeHD1f4myvDX2iA)_

### DIFF
--- a/.github/actions/apply/README.md
+++ b/.github/actions/apply/README.md
@@ -147,12 +147,7 @@ on:
     types: [opened, synchronize, reopened]
   workflow_dispatch:
 
-# Needed by the paths that still publish inline: the baseline push on `main`
-# and same-repo PRs. On a fork PR GitHub downgrades the token to read-only no
-# matter what this block says — which is the whole reason the split exists.
-permissions:
-  contents: write
-  pull-requests: write
+permissions: {}          # nothing by default; each job asks for what it needs
 
 # Two `synchronize` events on one PR would otherwise race: the older render
 # can finish last, and its publisher then overwrites the shared render
@@ -162,8 +157,20 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  render:
+  # Fork PRs: render only, in a job that is READ-ONLY BY CONSTRUCTION.
+  #
+  # Don't fold this back into the job below by switching `phase` on a
+  # condition. GitHub normally downgrades a fork PR's token to read-only on
+  # its own, but an organization can turn on "send write tokens to workflows
+  # from fork pull requests" — and then a shared job's requested write scopes
+  # are granted for real, while this action exports `github.token` into the
+  # environment the fork's Gradle build runs in. Job-level `permissions` is
+  # what makes the isolation hold regardless of that org setting.
+  render-fork:
+    if: github.event.pull_request.head.repo.fork
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v7
         with:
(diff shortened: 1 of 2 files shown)


yschimke:
Good catch — fixed in `df528f5`.

The allowlist was the wrong shape. `compute-scope.py` echoes gradle project paths verbatim and Gradle permits `@`, `+`, spaces and non-ASCII in a project name, so a change-scoped fork PR touching `feature@demo` would render and stage happily, then have its own handoff rejected on the publish side and never post. Self-inflicted denial of service on a legitimate input.

Switched to rejecting what's actually dangerous at this sink. It's a single-line `key=value` write, so the risk is a control character — a newline ends the `modules=` line and lets the remainder declare its own outputs. Everything else is just text; the value reaches its consumer as a quoted argument.

```
case "$MODULES" in *[[:cntrl:]]* | ,* | *, | *,,*) reject ;; esac
```

Verified both directions:

| Input | Verdict |
|---|---|
| `full`, `app,feature:foo`, `samples/android,app` | ACCEPT |
| `feature@demo`, `lib+extras`, `my module`, `café:ui` | ACCEPT |
| `full\nmode=baseline` | REJECT |
| `full\nmode<<X\nbaseline\nX` | REJECT |
| `a\rmode=baseline`, `a\tb` | REJECT |
| `app,`, `a,,b`, empty | REJECT |

---

PR #3750, opened 2026-08-13
feat(pages): serve whole design pages as addressable SVG, not screen backdrops

yschimke:
## What changed

`/{system}/pages` used to show one composed **screen** from the design file as a flat PNG, with a rectangle over every component instance on it. It now shows a whole **page** of the design file — a specimen sheet — inlined as SVG, with this catalog's renders standing in for the design's own drawing of the components it implements.

The reason is a property of the data, not the code: in the Material 3 kit exactly one page (`Examples`) holds instances at all, most of each screen on it is hand-drawn rather than assembled from the kit, and the densest screen yields eleven placements of which two are OS chrome. The definition sheets — the other thirty pages — are where the design actually states what a component should look like.

**Inlining is the whole feature.** An `` is a picture; nothing can reach inside it. An inlined SVG exported with `svg_include_node_id=true` is a document: given a node id, the viewer finds the design's own drawing of `Shape=Circle`, hides it, and puts `Shape/Circle` in the hole it leaves. Clicking it opens that component's viewer.

## Visual evidence

Before — the screen backdrop, on `main`:

The retired screen backdrop: a phone screen with a rectangle per component instance

After — a design page, outlines measured off the inlined export and coloured by how each node was linked (blue `design-map`, amber name match, dashed red = nothing implements it):

A design page with an outline over every component node

The swap — "Show our renders" + "Hide the design's own". The circle, square and triangle are now this catalog's renders sitting exactly where the design drew them; the pill keeps the design's shape because the catalog publishes no sticker for it, and the two dashed-red nodes keep theirs because nothing implements them:

The same page with the catalog's renders standing in for the design's own shapes

"Only what we don't implement" — the coverage read:

Only the unlinked outlines, with everything the catalog implements muted

Both states are captured by the preview-harness (`serve-design-page` + its `swap` / `unlinked-only` fixture states), so every future change to the measuring, the swap or the hiding moves a baseline.

## Decisions worth reviewing

- **No geometry in the manifest.** The SVG knows where its nodes are, so `design-page.js` measures each `[data-node-id]` element and re-measures on resize. A recorded rectangle would be a second answer to the same question and a worse one — a Figma export box is the *render* box, effect bleed included. A node the manifest names that the export doesn't carry keeps its list row and gets no outline.
- **`SvgSanitizer` is a new trust boundary.** Inlining third-party markup at catalog load is the one place this server puts a delivery branch's markup into the document tree. Strict allowlist, `on*` dropped before the allowlist is consulted, URL attributes judged by value (`url(#clip0)` survives, an `http:` beacon does not, `data:image/svg+xml` is refused because that is markup the sanitizer never walked), DOCTYPEs refused. `/pages/<id>.svg` answers the same sanitized markup — two answers for one URL is how a check gets bypassed.
- **The contract is first-party now.** `DesignPagesManifest` v2 replaces the mirrored `@design-parity/page-backdrop` v1, which is refused rather than half-read (its raster has nothing addressable in it, so a stale delivery branch should show no pages rather than a dead page). Both ends now live here, so `DesignPagesParseTest` quotes the wire format instead of vendoring someone else's fixture.
- **The publish lane no longer talks to Figma.** A repo imports its own pages on the design file's cadence and commits `design/pages/`; the reusable workflow only re-keys and copies. A fork, a token-less run and an offline republish therefore publish the same pages. The `page-backdrop-version` input is gone with the npm package it pinned.

Pairs with yschimke/m3-catalog#58, which adds the importer and the shape stickers the Shape page swaps in.

## Test plan

- `./gradlew :cli:test :preview-data-api:test ktfmtCheck --rerun-tasks` — green. New: `SvgSanitizerTest` (14 cases, mostly the things that must *not* survive), `ServeDesignPageStoreTest`, `ServeDesignPageRoutingTest`, `DesignPagesParseTest`.
- `node --test scripts/design-artifacts/design-pages.test.mjs` — 17/17. (Four sibling test files fail on `main` too — missing native deps in this sandbox — and are untouched here.)
- `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot` — 104 passed, including the two new design-page states.
- Serve fixtures regenerated (`UPDATE_SERVE_WEB_FIXTURES=true`).

**Not verified end-to-end against the real kit export.** This sandbox has no egress to `figma.com` (both `www.` and `api.` are refused at the proxy with `403 CONNECT`), so every test above runs against a hand-written SVG fixture shaped like a Figma export. The first real import has to come from the `Refresh Figma pages` workflow in m3-catalog once that lands, and the thing most worth watching is whether Figma's `data-node-id` attributes carry the ids in the `123:456` form the join expects — the viewer already tries the hyphenated spelling as well, but a third spelling would need a fix here.


yschimke:
Fix the failure

PR #3753, opened 2026-08-13
fix(serve): name both axes on a cross-product component's subtree rows

yschimke:
Follow-up to #3748, from a review comment that arrived ~6 minutes before it merged.

When a component varies on state **and** props, a row labelled by one axis is ambiguous rather than terse. Standing on `pressed + RTL`:

- the row that resets the state (`default + RTL`) and the row that resets the props (`pressed + default`) both read **"Default"** — two different renders wearing one name;
- the render on screen was labelled by whichever axis pass reached its href first, so `pressed + RTL` read simply **"Pressed"**.

Rows are now collected as previews and labelled once *both* passes have run, because whether one axis is enough to name a row is a property of the whole set rather than of the row. With both axes in play every row names both coordinates; a single-axis component — nearly all of them — keeps the terse label, since repeating "· Default" on every row would be noise for a distinction it cannot make.

## Visual evidence

The new `serve-viewer-cross-product` fixture, entered on `pressed + RTL`, subtree open. Five distinct rows, and the toggle above names both axes it folded:

cross-product row labels

Before this change those rows read `Default`, `Pressed`, `Disabled`, `Default`, `Disabled` — the first and fourth indistinguishable, and the current render (second) not mentioning RTL at all.

## Why there was no coverage

No committed catalog bakes a state × props matrix, which is exactly why the ambiguity reached review unseen. So the shape gets its own page fixture plus a captured `subtree-open` state — the claim here is the row labels, and they are invisible while the subtree is folded (six rows, so it arrives folded). The visual-diff bot carries the naming rule from here without anyone remembering to check it.

## Test plan

- `./gradlew :cli:test --tests '*Serve*'` — green, run **without** `UPDATE_SERVE_WEB_FIXTURES`, so the golden-sync assertion actually executed. (Running with the flag is what hid a fixture drift on the parent PR: it regenerates instead of asserting.)
- Full `pages-snapshot` harness — green, 108 tests (up from 106: the new fixture and its open state).
- `ServeViewerDisclosuresTest` gains both sides of the rule: a cross-product component names both coordinates and no two rows share a name; a single-axis component keeps the terse label.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YDQPfAaByr2LRbpYLkNySU)_


yschimke:
`VS Code Extension E2E (interactive-a-d)` is red here on scenario **B. edit a @Preview source, refresh, expect new id**. Flagging rather than fixing, because I don't think it belongs to this PR and the failure mode moved under me.

Three runs, same test, two different symptoms:

| run | head | symptom |
| --- | --- | --- |
| initial | `e6776f2` (base `1affb54`) | `…Interactive1786624647154 capture(s) not on disk` — discovery named it, the render never landed |
| re-run of the same job | `e6776f2` | identical |
| after rebasing onto `bc98494` | `d5ff33c` | `no setPreviews for samples/cmp ever named Interactive1786625666855` — never discovered at all |

I rebased because `bc98494` ("test(e2e): wait for the settled render in the interactive edit scenario", #3752) landed on `main` at 12:20 and touches only `e2eInteractive.test.ts` — this branch was cut at 09:36 and was running the pre-fix test. That did change the outcome, but into a different failure: the added preview is now never discovered, rather than discovered-but-unrendered.

Why I don't think it's this PR: the diff is seven files — `ServeWeb.kt` (serve page HTML), two Kotlin test files, a preview-harness page fixture, the harness spec, and a render PNG + README. Nothing in the Gradle preview compile/discover/render path this scenario drives, and nothing in the electron suite references anything it touches. Scenarios A, C and D pass in every run; only B fails, and the whole job sits at ~7 minutes with A alone taking ~164s.

Happy to dig into `composePreviewDiscover` if you think the discovery miss is real rather than contention — just say so and I'll take it. Otherwise this looks like it wants another pass on the same test rather than a change here.

---

PR #3755, opened 2026-08-13
feat(serve): fold the default render into the component row, and flow a long axis into columns

yschimke:
Two things the merged subtree got wrong. Both only show up at the real catalog's shape, which is why neither was caught in #3748 or #3753 — the fixtures were smaller than the thing they stood for.

## The default render was listed twice

The component row already links it: in the viewer that row **is** a render, not a jump to a card. A `Default` child sat directly beneath it, same href, same destination, one line down.

The default now folds up into the component row, which also takes the `aria-current` pill when it is the render on screen; the children are only the ways this component *differs*. The landing tree keeps its synthetic `Default` child — there the component row is an in-page jump to a card, so the default genuinely has no entry otherwise.

## A long axis ran down the page

`iconbutton-outlined` bakes 22 renders across size × width × shape. Opened, that was a **550px single-column ladder** of two-word labels with the whole right side of the viewer empty — the subtree inheriting a shape that only makes sense in the landing's 216px sidebar.

The variant list now flows into columns (`column-width`, so the browser picks the count). Those 22 rows become four short columns, ~160px:

22-render subtree in columns

Note the component row: `Icon Button Outlined · default (light)`, count **22**, carrying the current pill — that is the default render, no longer repeated below.

Long labels keep the same treatment; the cross-product component's `Default · RTL` / `Pressed · Default` rows flow into three columns rather than four:

cross-product subtree

## Fixture honesty

`wideStatePreviews` goes from 10 states to the real **22**, so the committed capture shows what the open subtree actually costs on the catalog that motivated this work. At 10 it looked fine; at 22 it did not, and the fixture was the reason nobody could see that.

## Test plan

- `./gradlew :cli:test --tests '*Serve*'` — green, run **without** `UPDATE_SERVE_WEB_FIXTURES` so the golden-sync assertion executes.
- Full `pages-snapshot` harness — green, 108 tests.
- Assertions that pinned `aria-current="page">Default</a>` on the child row are updated to the component row across `ServeWebTest`, `ServeWebFixtureTest` and `ServeViewerDisclosuresTest`; the last also gains an explicit check that `Default` is **not** among the children.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YDQPfAaByr2LRbpYLkNySU)_


yschimke:
`VS Code Extension E2E (interactive-a-d)` is red here, and it is the same failure #3753 merged with — not something this PR introduces.

Confirmed on both heads of this branch (`4ac9663`, `720c4f8`), and byte-for-byte the symptom #3753 showed once it was rebased onto `bc98494`:

```
1) Compose Preview interactive scenarios (real Gradle)
     B. edit a @Preview source, refresh, expect new id; revert, expect it gone:
   Error: Timed out after 180000ms waiting for:
     setPreviews after adding Interactive1786629705497, with its capture on disk
   no setPreviews for samples/cmp ever named Interactive1786629705497
```

A, C and D pass in every run; only B, and it is always the newly-added preview never being discovered rather than an assertion about behaviour. This branch is based on `bc98494` (#3752, the settled-render fix for this very scenario), so it is not the stale-test problem I mistook it for on #3753.

Nothing in this PR can reach it: the diff is `ServeWeb.kt`, three Kotlin test files, `serve.css`, a harness page fixture, the harness spec, and two render PNGs plus a README. No Gradle preview discover/compile/render path, and the electron suite references none of it.

Full analysis and the three-run table are on #3753 (#3753 (comment)). Flagging rather than chasing it further — I've spent a re-run and a rebase on it already, and it wants a change to that scenario rather than to this branch.

---

PR #3756, opened 2026-08-13
feat(serve): list a catalog's design pages in its navigation tree

yschimke:
A catalog's design pages were reachable only through an `N pages` action chip in the landing's header row, beside `download all (.zip)` and the compare actions. Two things were wrong with that: a chip can say only *how many* pages there are — the names, which are the thing you actually choose between, were a page away — and it put the one entry in that row that is a **place** into a row of one-off actions.

They are now a branch of the navigation tree, at the foot: one row per page, named, under a `Pages` row that leads to the index.

## Before / after

The section-less catalog landing — the shape most published design systems are in, m3-catalog included — cropped to the header and the navigation column.

Before: a "2 pages" chip in the header row. After: a Pages branch at the foot of the tree, each page named

Dark:

The same comparison under the M3 dark roles

Full pages, both themes, and the note on how the "before" was produced: [`docs/design/evidence/serve-catalog-tree-pages/`](https://github.com/yschimke/compose-ai-tools/tree/96d07dad0a94b509294269d37b37d3ce3c139259/docs/design/evidence/serve-catalog-tree-pages).

## Why the branch behaves as it does

Two things make it unlike every other row, both deliberate:

- **It carries no `data-group`.** Every other tree row names an id on this page and is intercepted into a scroll; these navigate away, so the click handler's `if (!id) return` leaves them to the browser — the same treatment a variant row already gets.
- **It is always open.** With a handful of pages there is nothing to gain by hiding their names behind a twisty, and the open state is what makes the branch worth having over the chip. `reflectTree` and the `ArrowLeft` handler both now skip a row that names no target, which is what keeps it open — without those guards the first component click would collapse it, and Left on `Pages` would clear `openCard` and collapse whichever component *was* open.

A catalog too small to have a tree at all (no authored sections, too few previews to synthesize families from) keeps the chip — otherwise its pages would be published and unreachable.

## Coverage

`serve-landing-grouped` now publishes pages, so the branch has a committed baseline and every future PR diffs it without anyone remembering to. `serve-landing-public` has no tree and keeps the chip, holding the baseline for that fallback. Both are already in `STYLED_FIXTURES`, so the harness paints them with the production stylesheet.

## Test plan

- `./gradlew :cli:test :cli:ktfmtCheck` — green.
- `UPDATE_SERVE_WEB_FIXTURES=true ./gradlew :cli:test --tests '*ServeWebFixtureTest*'` — three landing goldens move (`grouped`, `sections`, `tree-depth`; the last two only pick up the `reflectTree` guard in their inlined script). `serve-landing-public` is byte-for-byte unchanged, which is the fallback's assertion.
- `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot` — 108 passed.
- New assertions in `ServeWebFixtureTest`: the branch's markup, the absence of a duplicate chip when a tree exists, the chip's survival when one doesn't, and the `reflectTree` guard.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q1RZC8kKiUs1XC2DzG67Eo)_


yschimke:
The visual-diff bot lists three changed captures; only two of them are this PR's.

`serve-landing-grouped` (light + dark) is the change — the `Pages` branch at the foot of the tree.

`serve-viewer-cross-product-subtree-open` is **inherited from #3753**, whose baseline the bot has not picked up yet: that fixture was last touched by #3753 ("name both axes on a cross-product component's subtree rows"), which is the commit this branch is based on, while the `vscode-preview/main` baseline is still at `b1db6a8`, before it landed.

Verified rather than assumed — rendering that fixture on this branch with `main`'s `serve.css` swapped back in produces a **byte-identical** PNG to this branch's own render:

```
branch render (1024, 1044) | main-CSS render (1024, 1044)
branch vs main-CSS identical: True
```

So nothing in this diff reaches the viewer's axes subtree, which is what you'd expect: the only stylesheet edits here add `.cp-tree-page` / `.cp-tree-pages-row` to existing selector lists without changing a single declaration.

---

Issue #1920, opened 2026-06-16
apply: cli-version defaults to `latest`, silently breaking the render on every release (CLI/plugin version skew)


yschimke:
## Summary

The `apply` action's `cli-version` input **defaults to `latest`**. Because pinning the action ref (`apply@vX`) does **not** pin the CLI, every new compose-ai-tools release auto-installs the newest CLI against the consumer's still-pinned Gradle plugin. The newer CLI doesn't discover the older applied plugin, so the compose-preview render silently breaks **repo-wide** for any consumer who didn't explicitly set `cli-version` — and the default path *is* the broken path.

## Symptom (from a real consumer)

`yschimke/meshcore-mobile` pins `apply@v0.15.8` and `composePreviewPlugin = "0.15.8"`. When **0.15.9** published, the render check started failing on every PR with:

```
compose-preview doctor
  ✓ compose-preview 0.15.9 (latest)        ← CLI auto-jumped to the new release
  [project]
  ✗ no modules have the compose-preview plugin applied
      → id("ee.schimke.composeai.preview") version "0.15.9"

compose pipeline: No preview modules discovered ... no _previews.json ... skipping.
##[error]One or more compose-preview pipelines exited non-zero (rc=1).
```

It's deterministic (failed on the original run and a manual re-run), and unrelated to the PR's actual change — it tracked the release, not the diff.

## Root cause

- `cli-version` defaults to `latest`, so the installed CLI floats to the newest published release independent of the action ref or the project's plugin pin.
- The Gradle plugin stays pinned in `libs.versions.toml`.
- A `latest` CLine newer than the applied plugin fails plugin discovery → "no modules have the plugin applied" → empty render → non-zero pipeline.

So the *recommended* setup (use the `apply` action, pin the plugin in the catalog) breaks on **every** release until the consumer notices and bumps the plugin.

## Workaround (what unblocked the consumer)

Set the CLI to track the plugin from one source of truth:

```yaml
- uses: yschimke/compose-ai-tools/.github/actions/apply@v0.15.9
  with:
    cli-version: catalog
    catalog-key: composePreviewPlugin
```

(meshcore-mobile PR for the fix: `build(deps): bump compose-ai-tools to 0.15.9; pin render CLI to the catalog`.)

## Proposed fix (pick one or more)

[...]

PR #3771, opened 2026-08-13
feat(cli): pin one compose-preview version for every entrypoint

### LINKED ISSUE #3738: Add a method to pin a version for all entrypoints.

yschimke:
Closes #3738.

## The problem

compose-preview reaches a project through several doors, and each one picked a version by itself:

| Entrypoint | Version it used |
|---|---|
| `compose-preview` CLI | whichever release was installed; auto-injected the plugin at that version |
| VS Code extension | the version compiled into the installed VSIX |
| `install` / `apply` actions | the `version:` input — `latest` by default |

So one project could render three ways: a developer's CLI on the newest release, a teammate's editor on an older VSIX, CI floating on `latest`. When those drift the failure isn't obvious — a newer CLI's Tooling model query doesn't discover an older plugin, and the render reports "no modules have the compose-preview plugin applied", which reads like a project misconfiguration. That's #1920 from the CI side; this generalises the fix to every door.

## The pin

One line in the project's `gradle.properties`, written by a new CLI command:

```sh
compose-preview pin --cli      # pin to the CLI you're running
compose-preview pin 1.1.0      # …or to a specific release
compose-preview pin            # report the pin, its source, and whether the CLI matches
compose-preview pin --remove   # unpin
```

```properties
composePreview.version=1.1.0
```

Resolution order, identical in every implementation: `--plugin-version` → `COMPOSE_PREVIEW_VERSION` → `gradle.properties`' `composePreview.version` → the catalog's `[versions] composePreviewCli` (the convention the actions already document, so projects pinned that way get this for free). **No pin → each entrypoint uses its own bundled version, exactly as before** — unpinned is a legitimate state, and `doctor` reports it as `ok` with a nudge, never a failure.

## Scope

The pin governs **auto-inject** — the zero-config path, which is how the large majority of projects run. A module that declares `id("ee.schimke.composeai.preview") version "…"` itself keeps its own version, and nothing here rewrites a build script or a version catalog: auto-inject already skips such a module (Gradle's `plugins {}` DSL rejects `id(…) version "…"` when the plugin is also on the buildscript classpath), so its declaration is the single source of truth there. `doctor` still reports both `project.version-pin` and `project.plugin-version` so a mixed project can see the two numbers side by side.

A pin also can't re-version a binary that's already running. A CLI on 1.1.0 driving a project pinned to 1.0.5 injects the **pinned** plugin, but the daemon and renderer that CLI ships stay at 1.1.0 — so it warns once, more strongly across a major (where the render/daemon wire format differs). `-SNAPSHOT` on either side is silent: a locally built CLI driving a pinned project is a deliberate dev flow.

## What changed per surface

| Surface | Change |
|---|---|
| CLI | New `pin` command (`setup` group). `autoInjectInitScriptArgs` bakes the pinned version into the init script instead of `BUNDLE_VERSION`. `init-script --plugin-version` now actually works (it was silently ignored before, despite `doctor` suggesting it). |
| `doctor` | New `project.version-pin` check — pin, source, CLI agreement; `warning` only on a real mismatch between clean releases. Remediation snippets elsewhere now name the pinned version. |
| VS Code extension | Injects the pinned version; logs the pin + source at startup, and notes when it differs from the bundled one. |
| `install` action | New `version: pin`. Fails when the project pins nothing rather than quietly installing `latest` — a silent fallback is the skew this prevents. |
| `apply` action | `cli-version: auto` checks the pin **first**, ahead of its existing catalog / build-script sniffing, so a pinned project is skew-proof in CI with no workflow change. |

The precedence list is implemented four times (Kotlin, TypeScript, two Python scripts) because a binary, a VSIX and composite actions can't share code. Each copy has its own tests, and `docs/AGENTS.md` now states that changing one means changing all four.

## Test plan

- [x] `./gradlew :cli:test` — full CLI suite green, including new `VersionPinTest` (23 cases: sources, precedence, file surgery, skew reporting) and `PinCommandTest` (9 cases: report / write / remove / `--json`, plus the end-to-end "a written pin is what auto-inject injects").
- [x] `npm test` in `vscode-extension` — 1687 passing, including new `versionPin.test.ts` (11 cases mirroring the Kotlin precedence).
- [x] `python3 .github/actions/install/test_resolve_version.py` — 14 new cases; wired into CI's `actions-tests` job.
- [x] `python3 .github/actions/apply/test_check_skew.py` — 30 cases (6 new for the properties pin, incl. that it outranks catalog + build-script detection).
- [x] `./gradlew :cli:ktfmtCheckMain :cli:ktfmtCheckTest` and `npm run format:check` clean.
- [x] End-to-end against the built binary in a scratch project: `pin` → `pin --cli` → `pin 1.0.5` → `pin --json` → `init-script --print` shows `val pluginVersion = "1.0.5"` → `pin --remove` restores the original `gradle.properties` byte-for-byte.

**No visual evidence:** this change touches version plumbing only — no `@Preview` output, webview, overlay, theme, or fixture is affected, and no render path changes shape. Text-only is the correct bar here.

## Docs

New [`docs/VERSION_PIN.md`](https://github.com/yschimke/compose-ai-tools/blob/agent/issue-3738-sx1nrl/docs/VERSION_PIN.md) (model, precedence, scope, per-entrypoint behaviour, Renovate recipe, the four implementations), plus consumer-facing sections in `site/install.md` and both action READMEs, and a `docs/VERSIONING.md § 1.1` pointer.

---
_Generated by [Claude Code](https://claude.ai/code/session_01K45YvcE1KxirpE58WPH66M)_


yschimke:
All five review findings were real; fixed in 67a52a7. Two of them were bugs the pin itself introduced.

**1. Pin masked an explicitly declared version (`check-skew.py`).** Correct catch, and it changed the design. `apply`'s `cli-version: auto` and the skew guard both answer "which plugin will this build apply", and for a module that declares the plugin the answer is the declaration — auto-inject *skips* declared modules, so the pin never reaches them. Putting the pin first let the guard certify a project as skew-proof while the declared module still failed discovery.

Declarations now win again (catalog `[plugins]` → build-script literal → pin), restoring the exact pre-pin precedence. The pin fills only the gap those scans never had an answer for: the auto-inject project, where nothing is declared anywhere and `auto` fell back to `latest`. A declaration disagreeing with a pin now emits a `::warning::` naming both — neither number describes the whole build, so surfacing it beats letting either win silently. New `conflicting_pin()` + 6 cases.

**2. Shared init-script path across VS Code windows.** A bug I introduced: before the pin every window used `BUNDLED_PLUGIN_VERSION`, so one fixed path in `globalStorageUri` was safe. Now materialised into a per-version subdirectory, mirroring the CLI's `defaultInitScriptStorageDir(version)`.

**3. Duplicate assignments.** `Properties.load` resolves the last; the writer rewrote the first — so `pin <version>` could report a version the reader never saw. Writes now collapse duplicates, `--remove` drops all of them.

**4. `--remove` with a catalog pin.** It now names what is still pinning and why `pin` won't touch it:

```
compose-preview: removed the version pin from gradle.properties.
compose-preview: still pinned to 0.9.0 via gradle/libs.versions.toml ([versions] composePreviewCli).
`pin` never edits a version catalog (Renovate owns it) — remove the composePreviewCli entry by hand to fully unpin.
```

**5. Properties grammar parity.** The sharpest one — the CLI's `Properties.load` accepts a bare `key value` separator that the three regex parsers rejected, so `composePreview.version 1.2.3` had the CLI injecting the pin while CI reported the project unpinned. Exactly the skew this feature exists to close. All four now share one grammar (`=`, `:`, whitespace) and the last-duplicate-wins rule, each with its own test.

Verified end-to-end against the built binary on a project carrying a whitespace-separated pin, a duplicate assignment, and a catalog entry: duplicates resolve to the last, re-pinning collapses to a single line with neighbours intact, and `--remove` reports the surviving catalog pin.

Test counts: CLI 23→31 pin cases, extension 1687→1690, `test_resolve_version.py` 14→17, `test_check_skew.py` 30→36. Docs updated for the changed `apply` precedence — the pin is no longer described as winning there.

---

PR #3827, opened 2026-08-14
feat(serve): add a Source lane to the viewer

yschimke:
The visual half of the source-tab work started in #3816. That PR derived the usage code and wired it into the playground handoff; this one puts it on the card.

Every catalog card could already answer *what does it look like?* and *where is it declared?* — the `source` link opens the preview's own Kotlin on GitHub, annotations, sticker frame, click tally and variant knobs included. Neither answers the question a developer actually arrives with: **what do I type to get this?**

## What it looks like

The Source lane in light theme: the chip pressed beside the Live preview chip, the hint reading "usage source — not a render", and the panel on the stage holding five imports and a nine-line @Preview composable, with a Copy button and links to the playground and the whole sticker

The panel's content is `PlaygroundSourceCleaner`'s output — the same derivation the playground seeds from, so the code you read and the code you can Run are one thing that cannot drift into two answers.

**Dark theme.** The stage's own backing exists to give a *render* a known field to sit on (the Day / Night / Transparent chips), and the backend badge names the engine that drew one. Over a code panel neither means anything — on a dark page the backing framed the panel in a white card and the badge sat on the first line of source — so both are suppressed for the lane's duration and the panel brings its own surface:

The same lane in dark theme: the panel on its own dark surface, with no white stage card behind it and no backend badge over the code

**At rest**, before the chip is pressed — the ordinary viewer, panel present in the markup but empty and hidden:

The same viewer at rest: the Source chip un-pressed beside the Live preview chip, the render on the stage, and no panel

## Design notes

**A chip, not a renderer option.** The combo beside it is headed "Switch renderer" and answers *which engine drew this?*; the design-spec chip answers *what was this specified as?*; Source answers *what do I type?* Three questions, three controls — the same reasoning that pulled the spec lane out of that combo. It does join the mode radio group like every other lane, so `?mode=source` is bookmarkable and Back returns to the lane it was entered from.

**Fetched on first press, not baked into the page.** Deriving a snippet costs a GitHub read on a cold cache and most visitors to a preview never open the panel, so a page load pays nothing. The panel is server-rendered *empty* only so the stage has a stable place for it and does not jump on first open. `GET /usage/<id>` takes no session lease — it is a registry read, and leasing would stand a render daemon up to answer a question about source text — and 404s for every "nothing to show" case (no resolver, unknown preview, no recorded source, cleaner declined), so the panel has one branch to handle and never renders half an answer.

**Independent of the playground.** This changes the seed resolver's scope: it is no longer gated on a playground service. Reading the code is useful on any host that can browse a catalog; *running* it needs one that can compile that catalog, which most of the public deployment's catalogs have none of. Only the handoff link needs a compiler, and `playgroundLinkFor` already checks for one — so the panel offers "open in playground →" only when the server says there is one.

**Honest about generic cleaning.** When a catalog has not declared what its own helpers mean, the panel says so rather than presenting machinery as usage — the same distinction #3820 added to the playground's seed note, driven from the same `scaffoldsDeclared` flag so the two cannot disagree about the same snippet.

**Every node is built with `textContent`.** The payload is Kotlin read from a catalog's repository — exactly the kind of content that must never be parsed as markup — and the fetch URL is origin-checked off its `data-` attribute like the spec raster and the Wasm frame.

## Wired in, not screenshotted once

Per the repo's rule about new visual surfaces: the `serve-viewer-source` fixture and its `source-panel` state are captured by the harness in **both themes**, so every future change to the chip or the panel is diffed by the visual-diff bot without anyone remembering to. The images above are those captures.

## Test plan

- `./gradlew :cli:test` — full suite green, including the new `ServeViewerSourceLaneTest` (7 cases: chip/panel/radio present exactly when the host can answer, nothing rendered when it cannot, blank href treated as absent, Source and playground independent in both directions, and the chip absent from the renderer combo).
- `npx playwright test pages-snapshot.spec.mjs` — **118 passed**, so the new fixture renders and no existing baseline moved.
- `./gradlew :cli:ktfmtFormat` clean; `node --check` on the touched JS.

## Follow-ups unchanged

`usage` on `catalog.json` (what agents and MCP read), upgrading `code-connect.json`'s `codeSnippet`, `@CatalogScaffold` replacing the JSON rules file, and the pixel gate — still the highest-value of them, for the reason #3820 demonstrated.

---


yschimke:
All five findings were real and are fixed in `919941f`. Two of them meant the lane did not work where it matters most, and one falsified a claim in the description above — worth recording rather than quietly patching:

| Finding | Verified | Fix |
|---|---|---|
| **Route not reserved** (P1) | `ServeSites.RESERVED_SYSTEMS` had no `usage`, so a site host read `/usage/<id>` as an unknown session and 404'd before the handler ran | Reserved. Every Source panel was failing on custom catalog domains — how the public deployment serves catalogs — while working on the canonical path |
| **Fetcher still gated on the playground lane** (P1) | `ServeCommand.kt:2113` supplied `playgroundSourceFetch` under `playgroundLane?.let` | Wired unconditionally. I decoupled the resolver from `playgroundService` *inside the server* and never followed it to the CLI that starts it, so "Source without playground" existed in one file and not in the product |
| **SVG toggle left a blank stage** (P2) | The direct-transition list named every other non-static lane but not this one, so `data-mode` flipped without `closeSource()` restoring the out-of-flow image | `sourceActive()` added to that list |
| **Copy fallback did nothing** (P2) | Only relabelled the button, selected nothing, and named the Mac key on every platform — and the path is reachable, since the Clipboard API needs a secure context a plain-HTTP LAN host isn't | Selects the code, tries the synchronous copy an insecure context still has, then prompts with the right key |
| **Cold reads not coalesced** (P2) | Check-then-fetch with no single-flight | Coalesced per cache key in the resolver. Fine while the playground page was the only caller; the panel is one click on a page anyone browsing a catalog is already on |

Each has a test. The single-flight one runs eight threads against a held fetch and asserts the source file is read once; the reserved-route one asserts the set directly.

Full `:cli:test` green, and 118/118 preview-harness captures pass with no baseline moved.

Separately, the earlier red on `Android (Robolectric) serve lane` and `Apply compose-preview` was a transient Maven Central **403** for `material-kolor-android-debug-4.0.5.pom` — unrelated to this diff (no build or dependency changes here). `main` was green ~12 minutes before that run and sibling branches passed. This push re-runs both.

---

yschimke:
Fix build issues

PR #3835, opened 2026-08-14
fix(serve): make a resident catalog authoritative for /usage locations

yschimke:
Follow-up to #3833, which auto-merged while review was landing. Four findings, all against the one mechanism it introduced — the remembered source location. That concentration is the actual signal: the mechanism was the wrong shape, so this restructures it rather than patching four symptoms.

## The fix in #3833 covered the second click, not the first

The P1. The map was populated lazily inside the resolver, so it only helped once a Source panel had already succeeded. A viewer rendered and then left in a background tab past the idle timeout suspends its session, and the *first* `/usage` request finds neither a live host nor an entry — exactly the failure #3833 was meant to remove.

It is now primed from the **leased viewer render**. That page draws the Source chip and is the only route to the panel, so priming there covers the first click by construction. It also re-primes on every page load, which shrinks the stale-entry window to one nobody can click through.

## A resident host is now authoritative in both directions

The old `else` branch also ran when a live host existed but no longer listed the preview. So a catalog refreshed under the same id with a preview dropped — or its source metadata cleared — would serve the *previous* publication's source instead of 404ing. A resident host's refusal now drops the remembered entry rather than falling through to it.

## A retired session drops its entries

A null host means "unregistered" as well as "suspended". A remembered location would keep answering for a withdrawn catalog — and keep re-fetching its old repository once the seed TTL lapsed — long after every other session-backed route for it had started 404ing. Gated on `sessions.isKnownSession`, with the entry dropped.

## The cap no longer blocks refreshes

`size < MAX` suppressed updates to keys *already present*. The cap bounds how many previews are remembered, not how fresh each is; refusing to overwrite at the cap is how a republished catalog keeps answering with its previous ref.

## Test plan

- `./gradlew :cli:test` — full suite green against current `main`; `./gradlew :cli:ktfmtFormat` clean.
- **Honest gap:** these four paths (suspended / retired / refreshed-and-dropped / at-cap refresh) are not directly unit-tested. The logic lives in private members of `ServeHttpServer` and reaching it needs a server with a session driven through suspension and re-registration, which no current test harness sets up. The behaviour is small and centralised in `sourceLocationOf` / `rememberSourceLocation` / the `locate` `when`, and I verified each branch by reading; but I would rather say so than imply coverage I did not add. A `ServeSessionRegistry`-level test that suspends and re-registers a catalog would cover all four and is the natural next step if you want it.

Non-visual — resolver and routing behaviour only, no markup or styling, so no render evidence applies.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt
@@ -306,9 +306,19 @@ class ServeHttpServer(
    * and unleased, and found nothing. The chip worked and its panel 404'd, on exactly the catalogs
    * that idle out — which is every quiet one on a shared host.
    *
-   * Remembering only what has actually been resolved keeps this small and self-targeting: an entry
-   * exists precisely because somebody opened that preview, which is who asks again. Refreshed
-   * whenever a live host answers, so a republished catalog overwrites rather than shadows.
+   * Three rules keep it from becoming a second, disagreeing source of truth:
+   * - **A resident host is authoritative, both ways.** Its answer is recorded; its *refusal* drops
+   *   the entry, so a catalog refreshed with a preview removed 404s rather than serving the
+   *   publication before it.
+   * - **It is primed from the leased viewer render**, not lazily on first use — that is what makes
+   *   the first Source click after an idle-out work, and it re-primes on every page load, so the
+   *   window in which a republished catalog could be answered from a stale entry is one that nobody
+   *   can click through.
+   * - **A retired session drops its entries**, so a withdrawn catalog stops answering here as it
+   *   already has everywhere else.
+   *
+   * Remembering only what has actually been resolved keeps it small and self-targeting: an entry
+   * exists precisely because somebody opened that preview, which is who asks again.
    */
   private val lastKnownSourceLocation =
     ConcurrentHashMap<Pair<String, String>, PlaygroundSeedResolver.Location>()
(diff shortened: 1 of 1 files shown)


yschimke:
The P1 was right, and it was the fourth defect in a row against one mechanism — so `802a83d` removes the mechanism rather than patching it again.

**The hole:** a catalog refreshed after a viewer had loaded left the primed entry untouched. If the replacement host then idled out before anything queried it, the old tab's first Source click served the previous publication's `repo/ref/path`. The replacement was authoritative in principle and never asked in practice.

**The fix:** stop keeping a private map. `ServeSessionRegistry.peekHost` says in its own KDoc that a caller needing a session's facts should hold a last-known snapshot via `addSuspendListener` rather than read absence as a verdict — and this server already does exactly that for catalog status (`catalogMetaSeen`), which is:

- written on suspension,
- refreshed on every resident read,
- replaced wholesale when a refresh re-registers a system,
- dropped with the session.

Those are precisely the four properties the separate map kept being patched towards. The source locations now ride in that snapshot.

The refresh case then resolves itself: suspension always runs the listener, so a catalog refreshed and later idled out snapshots the **replacement** host; while it is resident the live host answers anyway. That deleted the viewer-render priming, the per-preview map, its cap, and the special cases the cap and the retired-session check needed — a net simplification rather than another branch.

Test gap unchanged and still stated plainly in the description: the suspended/retired/refreshed paths need a `ServeSessionRegistry`-level harness that suspends and re-registers a catalog, which I have not added. The surface is now much smaller though — one snapshot field populated in `rememberCatalogMeta` and read in one `when` branch — and it inherits the lifecycle that mechanism is already trusted for.

`:cli:test` green, `ktfmtCheck` clean.

---

PR #3836, opened 2026-08-14
docs: audit the vendored embedded RC player and write the upstream work order

yschimke:
## Summary

Two companion docs on the vendored AndroidX embedded Remote Compose player (`third_party/rc-embedded-player`), plus their `docs/README.md` index entries.

**[`RC_EMBEDDED_PLAYER_UPSTREAM_SYNC.md`](https://github.com/yschimke/compose-ai-tools/blob/agent/embedded-rc-player-fixes-nzgsao/docs/design/RC_EMBEDDED_PLAYER_UPSTREAM_SYNC.md)** — the inbound audit: what we carry, what upstream has since done, what a refresh costs. Derived from a three-way tree diff (our tree vs. the pinned snapshot `c8e7d738`, the snapshot vs. `androidx-main` `2f88db18`, and our tree vs. `androidx-main`), not from re-reading `PROVENANCE.md`; the reproduction commands are in the doc.

**[`RC_EMBEDDED_PLAYER_UPSTREAM_FIXES.md`](https://github.com/yschimke/compose-ai-tools/blob/agent/embedded-rc-player-fixes-nzgsao/docs/design/RC_EMBEDDED_PLAYER_UPSTREAM_FIXES.md)** — the outbound work order: the nine fixes upstream has not taken, written to be handed to an agent working in an androidx checkout, each with its landing site, the current upstream code, the change, and the test to add. Includes the submission route (`compose:remote` does not accept GitHub PRs — these go through AOSP Gerrit) and an explicit do-not-send list.

### What the audit found

- **Upstream moved the player into a published library.** `compose/remote/integration-tests/player-compose-embedded` → `compose/remote/remote-player-compose` (`SoftwareType.PUBLISHED_LIBRARY_ONLY_USED_BY_KOTLIN_CONSUMERS`), an artifact the version catalog already pins. The alpha17 AAR carries zero `…/embedded/` classes, so the move post-dates it — the stated reason for vendoring expires with the next alpha. The new surface is `public` + `@RestrictTo(LIBRARY_GROUP)`, and upstream's `api/*.txt` files are not yet regenerated for it.
- **Four of our deltas are now converged with upstream** and retire on refresh: the `withInfiniteAnimationFrameMillis` frame loop, the dead `autoUpdate` knob, the gradient bound-colour resolution (`RcPlayerPaint.kt:411` upstream is byte-identical to ours), and the `DrawContent` double execution.
- **Upstream's rounded-clip work converges but is not a superset** — it has neither our `Path.addRoundRect` radius normalisation nor our clip-before-`DrawContent` ordering.
- **`Text` → `BasicText`** drops the ambient `LocalTextStyle` our text delta is built on (#3667), so the text refresh is a decision, not a merge.
- **Two upstream additions don't compile against the alphas we pin**: `Limiter` is absent from `remote-core` alpha17, and alpha17's `RemoteComposePlayerFlags` has no `isEmbeddedPlayerEnabled`, which upstream's `RcPlayer` now `check(…)`s on entry.
- **Migrating the Android lane to the artifact is gated**, and the gate is larger than it first looks: six upstream fixes, two host-side items, a deliberate call on the `CoreText` text features, a `compose-remote`/`wear-compose-remote`/`glance-wear` lockstep bump, and relocating the snapshot — `rc-embedded-player-jvm` currently reads its shared sources out of the Android module via `srcDir`, so the two halves of the end state conflict until it is moved.

### Three defects found in our own vendored code

Surfaced while writing the work order, all verified in the tree, none fixed here (docs-only PR):

1. **`isFloatOverridden` misses animated expressions.** The check sits in `GraphContext`'s leaf read, but `rememberRemoteFloatAsState` returns earlier for an id whose authored expression is animated — so a host write against one is still ignored.
2. **`ID_CONTINUOUS_SEC` is aliased to `ID_TIME_IN_SEC`** (`RcPlayerState.kt:133`, `GraphContext.kt:135`). Continuous seconds carries the sub-second fraction of the wall-clock position; the integral sibling makes a smooth phase input step. It advances, which is why the aliasing survived.
3. **Spring-backed expressions are invisible to every animated check.** `FloatExpression` declares both `mFloatAnimation` and `private SpringStopEngine mSpring` (verified in `remote-core-1.0.0-alpha17.jar`) and populates one or the other, so a spring expression falls through to the pure-target evaluator and renders frozen. This also affects upstream's own `hasAnimations` frame-loop predicate.

Also outstanding, and unrelated to any refresh: **the encoded-image parse flags are unwired in the daemon paths today** — `enableEncodedImageReferences()` never runs ahead of `RemoteComposeIrReplay.kt:39` or `RemoteOverridablePreview.kt:210,212`, so a document carrying a URL- or file-encoded bitmap fails to parse entirely. Invisible so far because no captured catalog carries one.

### Corrections to `PROVENANCE.md`

Its file counts are stale (42 upstream vs 53 here, not 44), and `RcPlayerDispatch.kt` / `RemoteImageSupport.kt` are undocumented local additions.

## Test plan

Docs only — no code, build wiring, or rendered output changes, so no visual evidence applies and no tests were affected. The factual claims were verified in-session rather than asserted:

- three-way `diff -r` over a blobless sparse `androidx/androidx` clone, re-run after correcting the script — 42/43/53 files, 29 drift and 25 local entries, matching the doc;
- `remote-player-compose-1.0.0-alpha17.aar` unpacked — `unzip -l classes.jar | grep -c embedded` → `0`;
- `remote-core-1.0.0-alpha17.jar` unpacked — `MatrixAccess` present, `Limiter` absent; `FloatExpression` declares both `mFloatAnimation` and `mSpring`;
- `javap` on alpha17's `RemoteComposePlayerFlags` (only `shouldPlayerWrapContentSize`) and `AndroidRemoteContext` (`getTypefaceResolver()` present);
- the work order's own verification loop was executed both ways round, confirming the `grep -q` form reports `present`/`MISSING` where the earlier piped form silently swallowed the failure.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LYPeT61uCaUmdCPrSbyNu4)_


yschimke:
All seven addressed in d4fbfa4. Two of them were real gaps in the inventory, and both resolve the same way — as **converged** fixes rather than retained ones:

- **Gradient bound colour.** Correct that it was missing; the review asked whether upstream has adopted it, and it has — head's `RcPlayerPaint.kt:411` carries the identical `register` bitmask and `read.getColor(word)` lookup (it additionally rewrote the ascending-stops check into a loop). So it retires on refresh rather than being explicitly retained, and `RcPlayerPaint.kt` is now flagged as *not* a pure seam re-extraction.
- **DrawContent.** Also correct, and the same conclusion: upstream had `toModifier(drawOpsList)` **and** a second `drawWithContent` in `RcPlayerComponent` at the pinned commit, which is the doubled chrome `rc-outlined-card-double-draw` records; upstream has since deleted its second path. Reclassified out of "new work for us". Since upstream's placement is position-in-the-wire-list where `RcPlayerDispatch` simply dropped the duplicate, the doc now calls for re-rendering that fixture on refresh rather than assuming equivalence.

That makes four converged deltas, not two.

The other five are applied as suggested: the `isEmbeddedPlayerEnabled` call-site list now names `RemoteOverridablePreview` and `RemoteComposeIrReplay` (via the `ExperimentalRemoteDocumentPlayer` wrapper) plus the module's own harnesses, and explicitly rules out `RcJvmRenderer` (its own `RcPlayerJvm`) and `RcPlayerBackend` (an enum); AAR consumption is now conditional on the four unmerged Android-lane fixes landing upstream, with the reasoning that a binary can't carry them; and the reproduction script pins the audited upstream head, captures the repo path before `cd`, and clears the extract directories first. Re-ran the corrected script end to end — 42/43/53 files, 29 drift and 25 local entries, matching the doc.

---

yschimke:
Both round-two findings were correct; fixed in 3abe727.

The classpath one is worth restating because it turned out stronger than "make the bump atomic". The vendored module doesn't just happen to depend on `remote-player-compose` — it *needs* it: `ExperimentalRemotePlayerApi`, `utils.getPath`, `utils.getTweenPath`, and `RemoteDocumentPlayer` in the comparison harness all live in that artifact and nowhere else. So the dependency can't be dropped to break the collision, which means there is no intermediate state that takes a post-alpha17 pin *and* keeps the patched sources. The bump and the source→artifact migration are one change. The doc now says that as a hard constraint in §4 and flags it against the "wait for the next alpha" option in §3.5, which had been written as if it were free.

On the prerequisite lists: three of them, disagreeing, and the shortest omitted exactly the delta most likely to be lost — the ambient-`LocalTextStyle` behaviour, since upstream's `BasicText` rewrite reads as a refactor rather than a behaviour change. Now one numbered gate of six in §4, cross-referenced from the sections that own each fix, with the `CoreText` reflection fields explicitly excluded as a feature addition that gates nothing.

---

yschimke:
Round three, all three correct — fixed in be10dc7.

The `BasicText` one is the sharpest: my option "keeps our behaviour, keeps upstream's material3 decoupling" was self-contradictory, since the symbol being read is `androidx.compose.material3.LocalTextStyle` (`RcPlayerTextLayout.kt:21`) and foundation has no ambient text style at all. That option keeps `implementation(libs.compose.material3)` and differs from today only in which composable draws. §3.2 now says so and adds the option that *does* drop material3 — a player-owned style input the host seeds — plus "accept losing the ambient style" as an explicit fourth choice rather than an implied non-option.

On the seams: right that only nine of the eleven have an upstream body to re-extract. `RemoteImageSupport.kt` and `GmsFontProviderCertificates.kt` are handwritten, and they're the two a mechanical refresh loses most easily precisely because nothing in a `diff -r` points at them — losing the first makes URL/file-encoded bitmap documents fail to parse outright, not degrade. Both now have explicit retain-or-replace steps.

And the font resolver belongs in the gate. It's a different *kind* of gate item, which is why it slipped: items 1–6 are things that must land upstream, whereas this is our wiring — so the gate is now eight, split by owner. Worth stating the failure mode plainly, which the doc now does: without a host-supplied `typefaceResolver`, downloadable-font documents render *successfully* in Roboto after the switch, so nothing announces the regression.

---

yschimke:
Round four, both correct — 5f3e95a. The image-flags one is the most valuable finding in this review so far, because it isn't about the doc.

**The encoded-image flags are unwired in production today.** I verified it rather than taking it on the review's word: `enableEncodedImageReferences()` is reached only from the `RcPlayer(CapturedDocument)` overload (`RcPlayer.kt:472`), this module's test harnesses, and the jvm renderer's own private copy (`RcJvmRenderer.kt:174`). The daemon paths parse the bytes themselves and never call it — `RemoteComposeIrReplay.kt:39` and `RemoteOverridablePreview.kt:210,212` (`remoteDocumentForPreview`, both branches). Both constructors parse on the spot while `Limits.ENABLE_IMAGE_URLS` / `ENABLE_IMAGE_FILES` still hold their shipped `false`, so a document carrying a URL- or file-encoded bitmap fails to parse and the preview renders nothing at all.

It has stayed invisible because it only bites documents that actually carry an encoded image reference, and the captured catalogs carry none — no lane has ever exercised it. The doc now records it in the present tense with both call sites. I have **not** fixed it here: this is a docs-only PR and that's a production change to the daemon connector, so it wants its own PR (and a fixture that would have caught it — the gap is really that nothing in the captured set has a URL-encoded bitmap). Say the word and I'll open it.

On the `CoreText` fields: right, "ride along" was wrong, and wrong in the direction that loses behaviour quietly. They live in vendored files that must leave the classpath at migration, and `RcPlayerTextLayout` consumes them — so autosize, justification, line break, hyphenation, the two extra ellipsis modes and the `StaticLayout` maxLines parity go with them. Now stated as a ninth gate item to decide deliberately rather than an exemption, with the note that it's the one most likely to be waved through precisely because it reads as a feature rather than a fix.

---

yschimke:
Both correct — f4ec355. These were contradictions my own earlier rounds introduced: I tightened §3.7 and §4 without going back to the summary sentences that had described the looser version, and both stale sentences failed in the direction of losing behaviour quietly.

- §3.8 still called the `CoreText` fields non-gating after §4 established they can't survive the migration; it now points at the ninth-item decision.
- The §4 preamble still budgeted all 11 added files as seams to re-extract after §3.7 split them nine-plus-two, so a reader working from the summary could have discarded the two handwritten workarounds — regressing encoded-image parsing or downloadable fonts. Now "nine seam files to re-extract plus two handwritten workarounds to retain".

Also worth noting for anyone reading the thread: this PR now carries a second doc, `RC_EMBEDDED_PLAYER_UPSTREAM_FIXES.md` — the outbound half, written as an implementable work order against an androidx checkout, with the nine fixes upstream hasn't taken. Several findings from these review rounds fed straight into it (the `BasicText` ambient-style question is filed there as a discussion to open rather than a CL to push, and remote-core aborting a whole document over one bad image reference is item 9).

---

yschimke:
All four in abe72fa. One of them found a gap in the **code**, not the write-up, which is the most useful thing this review has turned up.

**`isFloatOverridden` doesn't cover animated expressions.** Verified in our tree: the check lives in `GraphContext`'s leaf read, but `rememberRemoteFloatAsState` returns before ever reaching the graph when the id's authored expression is animated — straight to `rememberAnimatedRemoteFloat`, or to `rememberRemoteExpression` for an animated chain. So a host write against such an id is still ignored, exactly the failure the fix was meant to remove. The work order now carries this as a limitation to close while implementing rather than inherit, asks for the check at the top of the resolver, and requires both a plain and an animated expression in the test. Flagged honestly as found-by-review — plain in the code, unverified in behaviour.

The other three were instructions that would have misled whoever picks up the work order:

- **The clip screenshot test direction.** I used "background" for the surrounding surface two paragraphs after using it for the component's own chrome. Asserted the wrong way round, that test passes on the broken build and fails on the fix. Both colours are now named explicitly, with the correct direction spelled out and a note to get it right.
- **A `typefaceResolver` is not sufficient for fonts.** It feeds the canvas text ops via `AndroidRemoteContext`; the layout composables build a Compose `FontFamily` in `resolveFontFamily`, which is where our variable-axis handling lives (`GoogleVariableFontFamilies`, `RcPlayerTextLayout.kt:331`, returning ahead of the `GoogleFontFactory` fallback). Wiring only the resolver keeps downloadable fonts working while silently dropping `wght`/`wdth` on layout text. Gate item 7 now says so.
- **The item-5 grep answered "done" for a fix that hasn't landed** — `ID_CONTINUOUS_SEC` always matches `RcPlayer.kt`'s frame-loop scan, which is precisely the bug (loop runs, value doesn't move). Scoped to `state/RcPlayerState.kt`.

---

yschimke:
Four of the five applied in 79d9721. **The P1 is incorrect and I'm not acting on it** — see below.

### P1 (agent author/committer identity): not reproducible

The claim is that `abe72fa` records author and committer as `Codex <codex@openai.com>`. It does not:

```
$ git log -6 --format='%h | A: %an <%ae> | C: %cn <%ce>'
abe72fa | A: Yuri Schimke <yuri@schimke.ee> | C: Yuri Schimke <yuri@schimke.ee>
f4ec355 | A: Yuri Schimke <yuri@schimke.ee> | C: Yuri Schimke <yuri@schimke.ee>
5b04523 | A: Yuri Schimke <yuri@schimke.ee> | C: Yuri Schimke <yuri@schimke.ee>
5f3e95a | A: Yuri Schimke <yuri@schimke.ee> | C: Yuri Schimke <yuri@schimke.ee>
be10dc7 | A: Yuri Schimke <yuri@schimke.ee> | C: Yuri Schimke <yuri@schimke.ee>
3abe727 | A: Yuri Schimke <yuri@schimke.ee> | C: Yuri Schimke <yuri@schimke.ee>
```

No agent trailers on any of them either. Three independent things agree: the local `commit-msg` and `pre-push` hooks (which did fire earlier in this session and blocked a commit until the identity was fixed — that's why it's correct now), the `Reject agent attribution` CI gate on this PR, and the log above. Nothing to recreate.

### The rest

**`ID_CONTINUOUS_SEC` is not an alias of `ID_TIME_IN_SEC`** — correct, and it's a second semantic bug in our own delta rather than only in the write-up. In the remote-core mirror, continuous seconds is `minute*60 + second + millisOfSecond*1e-3` (wall-clock position within the hour, *with* the fraction) while `getTimeInSec()` is the integral `minute*60 + second`. Ours aliases both to `timeMillis / 1000f` (`RcPlayerState.kt:133`, `GraphContext.kt:135`), so a document using it as a smooth phase input steps instead of sweeping. It advances, which is exactly why it survived. The work order now says to derive it from the clock snapshot, warns against copying ours, and requires the test to assert phase and fraction rather than movement.

**The test commands couldn't run item 1's test** — `compileDebugKotlin` compiles production code and `:test` runs unit tests, while the only test that can see a draw-order bug is instrumented. Added `connectedDebugAndroidTest`, a note that item 1's `Test:` trailer must name it, and a pointer to the module's golden-image workflow.

**The greps read "done" on partial adoption** — items 1 and 2 shared one alternation and item 6 matched on any one of three fields. One command per item now, with item 6's eight behaviours checked individually, and an explicit rule that a partial hit means the item is still open with the landed part removed from its scope.

**The alpha bump is four parts, not two** — the catalog requires `compose-remote`, `wear-compose-remote` and `glance-wear` to move in lockstep to mutually built versions, with a documented `NoClassDefFoundError` from a prior independent bump and a red `main` from a version that hadn't published yet. Now recorded with the render step to prove the set agrees.

---

yschimke:
All five in 02bb3d7. Every one of them was an instruction that would have reached the implementing agent as wrong, which is the failure mode that matters most for this doc.

- **Item 5's landing site was incomplete.** `GraphContext` carries its own copy of the time branch — upstream's `GraphContext.kt:133` has the same `ID_TIME_IN_SEC`/`MIN`/`HR` `when` and the same omission — and a read *inside a computed expression* goes through the graph rather than the resolver, which is the common case for a sweep animation. Both files are now required, and the test has to cover both read paths.
- **The autosize units.** I'd written `TextAutoSize.StepBased(min, max, 0.5f)`, which reads as `sp` literals; the `CoreText` values are Remote Compose pixels, so all three convert through the current `Density`. As literals it compiles, looks right, and silently renders a different size range at any density ≠ 1.
- **The intro oversold the list.** It called this the §4 migration gate; it's the upstream portion only — the gate also has host-side items no upstream CL can satisfy (`typefaceResolver` wiring, variable-axis layout fonts, image parse flags ahead of the two `RemoteDocument` constructors).
- **The item-6 verification loop swallowed its own verdict.** Reproduced it before changing anything: with `git grep -c … | head -1 || echo MISSING`, Bash reports `head`'s status, so the `||` never fires and an absent field prints a blank line that reads as a pass. Replaced with `grep -q` in an `if` and verified both branches print — `mAutosize present`, `NoSuchSymbolXYZ MISSING`.
- **The item-9 parser test** covered only the URL encoding though the failure has two independently gated inputs; now parameterized over URL and file.

---

yschimke:
All four in 17efe5c. The spring one is the third code-level finding from this review series, and the widest.

**Spring-backed expressions are invisible to every animated check.** Verified in the artifact rather than only the mirror: `remote-core-1.0.0-alpha17.jar`'s `FloatExpression` declares both `public FloatAnimation mFloatAnimation` and `private SpringStopEngine mSpring`, and the constructor populates one *or* the other — a spring spec sets `mSpring` and leaves `mFloatAnimation` null. Consequences beyond what the comment claimed: it isn't only item 4's routing and our dependency walker, it's also `buildComputedOpIndex` and **upstream's own `hasAnimations` frame-loop predicate**, so a spring-only document may not keep the frame loop running at all. Item 4 now treats "is this animated?" as one predicate over both fields applied at all four sites, notes `mSpring` is private and needs the file's existing reflection pattern, and asks for direct and nested spring cases. Our delta has the same gap.

**The jvm/Android split can't build as the tree stands** — also correct, and structural. `rc-embedded-player-jvm` has no shared sources of its own: `build.gradle.kts:149` reads them out of the Android module via `srcDir("../rc-embedded-player/src/main/kotlin")`, with `:166` doing the same for test resources. So deleting the Android sources to dodge the artifact collision removes the jvm player's inputs, and keeping them keeps compiling the colliding classes. The migration has to relocate the snapshot to jvm-owned storage in the same change.

**Two more early-retirement traps in the appendix**, both the same shape as the ones fixed last round: item 3's single grep hits on the tracking field or the graph read even when the resolver-level check — the exact hole this PR documented — is absent, so it's three predicates now plus a note that the required *ordering* can't be grepped; and item 6's eight-token loop never covered the `StaticLayout` `maxLines` cap, which has no distinctive token, so eight `present` lines didn't mean done.

Worth stating plainly: a grep can show something landed, never that a behaviour is right. Both places now say so.

---

PR #3842, opened 2026-08-14
fix(serve): capture session snapshots as part of the registry's transitions

yschimke:
The structural fix for the two races review found on #3841. Both had one cause: the storage was the registry's business but lived outside it, coordinated by a listener *beside* the transition instead of being *part* of it.

## What could not be fixed from outside

**The suspend gap.** A suspend listener runs after the lock is released, so a reader could catch the moment in between — `peekHost` already null, `isKnownSession` still true, no snapshot yet — and get a transient 404 on the first Source click. Publishing before the hero bake only shortened it: the listener still ran late, and the projection has to walk every preview.

**Retirement being overtaken.** Because a listener writes storage the registry doesn't own, an admin retirement could lose to a slower writer still holding the removed host — a resident status read, or a concurrent suspend — recreating the entry the retirement had just evicted. The leak survived under churn.

## SessionSnapshots

`capture` runs **under the lock, immediately before the host is detached**, so "resident" and "snapshotted" are the only two states visible from outside. `discard` runs **under the same lock as the removal**, so nothing can be written back afterwards.

The price is the contract, and the interface says so: both run with the registry lock held — no I/O, no blocking, no re-entering the registry. Both implementations here are map operations over a list already in memory.

A side benefit: the source locations now have exactly one writer and need no resident-read refresh, which makes them *fresher* — taken from the host being detached rather than from whenever a status page last looked.

## This closes the test gap I've carried for three PRs

The reason the previous rounds were untested is that the logic lived in private members of `ServeHttpServer` and needed a server driven through suspension. As a registry transition it is a unit, so:

- **`a snapshot is captured as part of the detach, not after it`** — asserts from *inside* `capture` that `reg.peekHost(id)` is still non-null at that moment. If capture ever moves after the detach, this flips and the window is back.
- **`retiring a suspended catalog discards its snapshot`** — suspends, confirms the snapshot is held, retires, confirms it is gone.

Plus the retirement-signal test from #3841.

## What deliberately did not change

`catalogMetaSeen` (the status/home snapshot) keeps its listener-based eviction. It is also written from resident reads, so it has other writers **by design** and cannot claim this guarantee. Its eviction is still strictly better than the never-evicting behaviour that predates this work; giving it the same treatment would mean changing how the status page refreshes, which is out of scope here.

## Test plan

- `./gradlew :cli:test` green against current `main`; `ktfmtCheck` clean.
- Two new registry tests above; `ServeSessionRegistryTest` now covers capture ordering, retirement discard, and the unregister signal.

Non-visual: registry lifecycle and resolver plumbing, no markup or styling.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt
@@ -3309,16 +3309,20 @@ class ServeHttpServer(
    * catalog — the snapshot [ServeSessionRegistry.peekHost] tells a caller to keep rather than read
    * absence as a verdict.
    *
-   * Its own map beside [catalogMetaSeen] rather than a field inside it, for one reason: it is
-   * written *first*, before the hero thumbnail is baked. That bake decodes and rescales a PNG, and
-   * a session is detached from its host **under** the registry lock while listeners run **after**
-   * it is released — so a Source request arriving in between would find `peekHost` null, the
-   * session still known, and a snapshot not yet installed. Publishing the cheap half up front
-   * closes that window instead of widening it by a decode.
+   * Written and removed **only** by [ServeSessionRegistry.SessionSnapshots], under the registry's
+   * own lock, as part of the detach and retire transitions. That single-writer rule is what makes
+   * it correct rather than merely narrow: a reader sees the session resident (and uses the live
+   * host) or suspended-with-a-snapshot, never the gap between, and a retirement cannot be overtaken
+   * by a slower writer still holding the removed host.
    *
-   * Same lifecycle as [catalogMetaSeen] otherwise, and dropped in the same places: written on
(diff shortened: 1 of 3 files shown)


yschimke:
Right, and it's a gap in the mechanism this PR introduces rather than an older one — fixed in `90737c2`.

`reclaimIdleForked` is the **second** removal path and I only wired `discard` to `unregister`. Every suspended session is captured, forked revision hosts included, so the GC removed each entry and left its snapshot behind: one leaked per reclaimed revision, on the surface with the most churn, defeating exactly the bound that function exists to enforce.

Discarded under the same lock as the removal now, like `unregister`, with a test that suspends a forked session, reclaims it past the GC window, and asserts the snapshot went with it.

Also stopped storing anything for a host with no catalog source (a forked revision, a plain module) rather than an empty map per session. The GC discards those anyway, but an entry that can never answer a question isn't worth holding.

`:cli:test` green.

---

PR #3849, opened 2026-08-14
test: check generated usage snippets against a real compiler

yschimke:
The Source panel tells a visitor a snippet is "the plain Compose that produces this render", and the playground invites them to press Run on it. Nothing held that claim to anything: every existing cleaner test feeds it source *this repository wrote*, which can only prove the rules match the fixtures.

This adds the loop that checks it against real catalogs.

```sh
scripts/usage-corpus.sh ~/m3-catalog ~/meshcore-mobile
```

It samples previews from each checkout (5 default + 5 variant per catalog, `USAGE_CORPUS_SAMPLES=n` to change), runs the real `PlaygroundSourceCleaner` over them with that catalog's own rules, and compiles the output. Exit status is 0 when every snippet compiles.

### The compile bar is a consumer's classpath, not the catalog's

`:tools:usage-compile-check` compiles the corpus against **Compose and material3 only**. Compiling against the catalog's own classpath would let its internal helpers resolve, and the leakage worth finding is exactly the leakage that resolves there and nowhere else. The bar is "a developer pasted this into their own Compose app".

The module has no sources unless `-PusageCorpus=<dir>` points it at a corpus, and the generator test is a no-op without `-Dcomposeai.usageCorpus.<system>=<path>`, so a normal build pays nothing.

### What it found

| Catalog | Compiles | Rules |
|---|---|---|
| m3-catalog | **3 / 10** | 17 declared scaffolds |
| meshcore-mobile | **0 / 10** | none — generic path |

**Residue is not a proxy for "compiles."** `NumberBadge` was reported *clean* — no residue at all — and did not compile:

```kotlin
Badge { Text(previewOverrideString("label", "3")) }
```

`previewOverride*` is *this repository's* knob API, not a catalog's. It sits in no scaffold package and no catalog had declared it, so nothing flagged it, and every catalog's "plain Compose" was quietly leaking it into code a developer was invited to copy. Residue only ever reported *declared* scaffolding that survived a rule; it cannot see a helper nobody declared, which is precisely the case worth catching.

So this PR also fixes that: the override knobs are now `GENERIC` rules substituting the default argument (`previewOverride*(key, default, …)` → `default`), and **declared rules inherit the generic ones rather than replacing them** — before this, the catalogs that had gone to the trouble of declaring their scaffolding were the only ones *not* getting the shared rules. A catalog can still override any of them by naming the same helper.

m3-catalog's remaining seven failures are missing vocabulary, not machinery defects: `toggleable`/`editable` destructuring a `Pair` (2 — the known `$known-gaps` entry, wants a `DESTRUCTURE` kind), conditional `stringResource(if …)` (2 — the inliner correctly declines, but the snippet then keeps an unresolvable `Res`), and `CatalogFilledStars`, a catalog-owned `ImageVector` (4).

meshcore-mobile's 0/10 is a **structural mismatch, not a missing rules file**. m3-catalog is annotation-first — each `@CatalogComponent` wraps one library component, so there is a call site to reduce to. meshcore's `catalog.spec.json` names plain `@Preview` functions that compose the *app's own screens* over fixture data; there is no library call site underneath, and the honest reduction of "the chat screen with twelve fake contacts" is that screen with twelve fake contacts. For such a catalog the Source panel shows the preview sliced and import-pruned, and the panel already says so. Worth knowing before anyone tries to fix it with rules.

### Not wired to CI, deliberately

It needs checkouts of the catalog repositories, and the failure count is currently a baseline to drive down rather than a gate to hold. The shape it wants is a scheduled job that checks out both catalogs and reports the ratio — turning it into a gate once m3-catalog reaches 10/10.

### Test plan

- `./gradlew :cli:test` — green, including the new `declared rules inherit the generic ones` and `a preview override knob becomes the default the render was baked with`.
- `./gradlew ktfmtCheckAll` — green (the new module applies `composeai.base-conventions`, which the settings-derived aggregate task requires of every project with a build file).
- `./gradlew :tools:usage-compile-check:compileKotlin` with no `-PusageCorpus` — no-op, as intended.
- `scripts/usage-corpus.sh /home/user/m3-catalog <meshcore checkout>` — ran end to end, producing the numbers above.

No visual surfaces change: this is a test harness, a build module, and one rules-inheritance fix behind the Source panel. The panel's own rendering is unchanged.

Docs: `docs/design/USAGE_SNIPPET_CORPUS.md`.

---

### DIFF
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -441,7 +441,21 @@ tasks.register<Zip>("packageAndroidDaemon") {
   into("lib-daemon-android") { from(stageDaemonAndroidLibs) }
 }
 
-tasks.withType<Test>().configureEach { useJUnitPlatform() }
+tasks.withType<Test>().configureEach {
+  useJUnitPlatform()
+  // Catalog checkouts for the usage-snippet corpus (UsageSnippetCorpusTest). Absent by default, so
+  // the corpus is a no-op in a normal build; `scripts/usage-corpus.sh` supplies them. Forwarded
+  // rather than read from the environment so the paths show up in the build's own inputs.
+  for (key in
+    listOf(
+      "composeai.usageCorpus.m3-catalog",
+      "composeai.usageCorpus.meshcore-mobile",
+      "composeai.usageCorpus.out",
+      "composeai.usageCorpus.samples",
+    )) {
+    providers.systemProperty(key).orNull?.let { systemProperty(key, it) }
+  }
+}
 
 abstract class CheckCliDaemonLibraryBoundary : DefaultTask() {
   @get:Classpath abstract val runtimeClasspath: ConfigurableFileCollection
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundSeed.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundSeed.kt
@@ -1,5 +1,6 @@
 package ee.schimke.composeai.cli.serve
 
+import ee.schimke.composeai.cli.serve.UsageRules.Companion.declaresCatalogScaffolds
 import java.util.concurrent.ConcurrentHashMap
 
 /**
@@ -279,7 +280,7 @@ class PlaygroundSeedResolver(
         sliced = cleaned != null || sliced != null,
         cleaned = cleaned != null,
         residue = cleaned?.residue.orEmpty(),
-        scaffoldsDeclared = cleaned != null && rulesFor(where).scaffolds.isNotEmpty(),
+        scaffoldsDeclared = cleaned != null && rulesFor(where).declaresCatalogScaffolds(),
       )
     // Bounded, and deliberately not an LRU: entries are a few KB, a catalog has a fixed number of
     // previews, and a full cache means the ones people actually open are already served from it. A
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundSourceCleaner.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundSourceCleaner.kt
@@ -481,6 +481,47 @@ object PlaygroundSourceCleaner {
     return out
   }
 
+  /** A `name = value` argument, as distinct from a positional one. */
+  private val NAMED_ARG =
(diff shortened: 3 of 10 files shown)


yschimke:
Reviewed all six. Five were real and are fixed in 3836c23; the sixth is real, larger, and documented as a gap rather than fixed here.

**Fixed**

- **Named-argument substitution (P1).** Correct, and the worst of the six: `previewOverrideString(key = "title", default = "Shopping")` emitted `Text(default = "Shopping")` — Kotlin that looks right and does not compile, presented to a visitor as runnable. A `SUBSTITUTE` rule now declares the helper's `params`, so `$1` binds the way Kotlin binds it, in any argument order. A rule that declares no `params` keeps the positional reading and *declines* a named-argument call (reporting it as residue) rather than guessing.
- **`scaffoldsDeclared`.** Correct and the subtlest — I'd have missed it. `PlaygroundSeed.kt` computed it as `scaffolds.isNotEmpty()`, so a non-empty `GENERIC` had every catalog claiming a declaration it never made, and the panel making the stronger "this is usage code" claim on the strength of annotation-stripping alone. It now asks whether any scaffold is the catalog's own.
- **Missing helpers.** `previewOverrideFont` and both `previewOverrideChoice` overloads added. The two overloads differ only in their third parameter (`options`/`values`); an argument naming a parameter the rule's list omits is ignored rather than fatal, so one entry covers both.
- **Swallowed build failures in the script.** Correct, and the one that would have quietly turned the whole loop into a no-op: `|| true` plus an empty diagnostic grep reports every snippet as compiled and exits 0. A nonzero exit with nothing attributable is now a corpus failure that prints the log tail.
- **Per-snippet isolation.** Correct in both directions you named. Each snippet is now emitted into its own package. Both ratios are unchanged (m3-catalog 3/10, meshcore-mobile 0/10), which is itself worth having: the numbers were not an artefact of the snippets propping each other up.

**Not fixed here: seeded `@OverrideVariant` values**

Also correct, and it needs more than a rule change. `PreviewInfo.overrides` does carry the values, but the seed path resolves a *source location* (`PlaygroundSeedResolver.Location`) which has none, so this means plumbing them from the preview through the location into a key-aware substitution — a different change from "make the generic knobs compile".

Worth stating that it isn't a regression: before the knobs were rewritten at all, the `off` variant's snippet said `previewOverrideBoolean("checked", true)`, which reproduced that render no better *and* did not compile. It is now honest about the default render and still wrong about the variant. Recorded under "Known gap" in `docs/design/USAGE_SNIPPET_CORPUS.md`.

---

PR #3853, opened 2026-08-14
fix(serve): unqualify only declared scaffold packages, not package-shaped receivers

yschimke:
Third review round on the snippet corpus (#3849, #3851). Both merged before this round could land on them, hence the separate PR.

### A receiver chain is not a package

The fix in #3851 unqualified any call whose prefix was two or more lowercase-initial segments. That is exactly the shape of `state.metrics.counted { }` — an ordinary receiver chain ending in a name some catalog also declares as a scaffold. Stripping the receiver there hands the call to the scaffold passes and rewrites somebody's code, which is worse than the leak it was fixing.

Rules now declare `scaffoldPackages`, and only a prefix named there is stripped; `GENERIC` declares `ee.schimke.composeai.overrides`. An allow-list can only ever *miss* a qualified call, and a miss leaves it as residue — the direction this whole pass is built to fail in.

### A catalog overriding only a generic knob was reported as declaring nothing

`declaresCatalogScaffolds` compared keys against `GENERIC`, so a catalog whose `compose-usage.json` declares its own reading of `previewOverrideString` — a rule actively driving the cleaning — came out as undeclared, and the Source panel would have withheld a claim that catalog had earned. It now compares entries.

### A malformed `repos` property was indistinguishable from an absent one

Absent means "no checkouts wired", which is the normal build. Present-but-mistyped is a wrong invocation, and silently dropping the entry turned it into a successful no-op — the same class of bug as the `|| true` hole this loop keeps finding in itself. It now fails.

### Two checkouts with the same basename shared an output directory

The per-repository empty-corpus check then saw the other one's snippets and passed a catalog that was never sampled — precisely the silent success the check was added to prevent. Duplicate basenames are now rejected up front.

### The test-source filter matched `test` as a substring

`src/latestMain` and `src/contestMain` are production source sets containing `test`, and were excluded wholesale — dropping real previews just as quietly as leaking test ones. Now matched at the source-set name's boundary: ends in `Test` (`commonTest`, `jvmTest`, `androidUnitTest`), or is `test` / `testFixtures` outright.

### Test plan

- `./gradlew :cli:test` — green, including the widened receiver-chain test (`stats.counted` and `state.metrics.counted` both survive) and the fully-qualified-call test.
- `scripts/usage-corpus.sh /home/user/m3-catalog <meshcore checkout>` — ran end to end. Ratios unchanged: **m3-catalog 3/10, meshcore-mobile 0/10**, m3 still sampling 10 previews and reporting its 17 declared scaffolds.

Non-visual: a text-pass cleaner, a test harness, a shell script. Nothing rendered changes.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundSourceCleaner.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/PlaygroundSourceCleaner.kt
@@ -342,14 +342,17 @@ object PlaygroundSourceCleaner {
    * comes out marked cleaned with a repository-internal call still in it, which is the one outcome
    * this whole design is built to avoid.
    *
-   * The prefix must be **two or more** lowercase-initial segments — a package path, never a
-   * receiver. `overrides.previewOverrideString` (one segment, possibly a local val) is left alone;
-   * guessing there could rewrite a call on somebody's object.
+   * The prefix must be a package the rules **name** ([UsageRules.scaffoldPackages]), not merely
+   * something package-*shaped*. `state.metrics.counted { }` is two lowercase segments followed by a
+   * declared scaffold name, and it is somebody's ordinary receiver chain; stripping it on that
+   * resemblance would hand the call to the scaffold passes and rewrite it. An allow-list can only
+   * ever *miss* a qualified call, which leaves it as residue — the direction this pass fails in.
    */
   private fun unqualifyScaffoldCalls(text: String, rules: UsageRules): String {
-    if (rules.scaffolds.isEmpty()) return text
+    if (rules.scaffolds.isEmpty() || rules.scaffoldPackages.isEmpty()) return text
     val names = rules.scaffolds.keys.joinToString("|") { Regex.escape(it) }
-    val qualified = Regex("""(?<![A-Za-z0-9_.])(?:[a-z][A-Za-z0-9_]*\.){2,}($names)(?=\s*\()""")
+    val packages = rules.scaffoldPackages.joinToString("|") { Regex.escape(it) }
+    val qualified = Regex("""(?<![A-Za-z0-9_.])(?:$packages)\.($names)(?=\s*\()""")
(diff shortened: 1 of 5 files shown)


yschimke:
All three fixed in `f1b8f5a`. The P1 corrects a claim in this PR's own description, so worth stating plainly rather than burying.

**Unlisted package-qualified scaffold calls (P1) — the description above was wrong.** It said an allow-list "can only ever *miss* a qualified call, which leaves it as residue". It does not. The residue scan uses `mentionsWord`, which rejects a name preceded by `.` by design — the very reason a package-qualified call was invisible in the first place. So a scaffold whose package nobody declared was neither rewritten *nor* reported, and since every existing `compose-usage.json` deserializes `scaffoldPackages` as empty, that was the default case, not an edge one. The fail-safe framing was doing work the code didn't do.

`mentionsQualifiedCall` now reports a call through **any** qualifier. It deliberately does not try to tell a package from a receiver — it cannot, without the resolution this pass doesn't have — so it over-reports, and `state.metrics.counted(…)` lands in residue too. A false residue costs a note saying a helper may not have been rewritten; a false silence costs a snippet advertised as runnable that does not compile. Only one of those is worth avoiding.

Writing the test for it caught a second bug immediately: both call regexes required `(`, but a trailing-lambda call (`counted { }`) has no parentheses at all, and that is the shape most scaffolding wrappers are written in. Both now accept `{` as well.

**Explicitly empty `repos=`** — correct, and the same silent-no-op shape this loop keeps finding in itself. Absent means "no checkouts wired" and should pass quietly; present-and-empty is a wrong invocation and now fails.

**`commonTestFixtures` / `androidTestFixtures`** — correct, and my regression from last round: narrowing to boundaries dropped the `TestFixtures` coverage the previous filter had. Now matched at the boundary too, so fixture trees stay excluded and `latestMain` stays included.

Ratios re-measured after all of it, unchanged: m3-catalog 3/10, meshcore-mobile 0/10.

---

PR #3857, opened 2026-08-14
test: spike parse-only PSI for the usage cleaner, and measure it

yschimke:
Spike, not an implementation. It answers whether `PlaygroundSourceCleaner`'s text passes should become a real parse — with numbers rather than argument.

### Why ask

The cleaner scans text because the Kotlin frontend is deliberately kept off the CLI's runtime classpath. The snippet corpus made that decision's cost measurable: across four review rounds, nearly every defect it found was **parser-shaped**.

| Defect | What a parse knows |
|---|---|
| `previewOverrideString(key = …, default = …)` bound positionally | argument binding |
| `state.metrics.counted` taken for a package qualifier | qualified expression structure |
| `counted { }` missed — the regex required `(` | a call is a call |
| a qualified call neither rewritten nor reported | the callee, regardless of qualifier |
| `toggleable` destructuring a `Pair` (still open) | destructuring declarations |

Five symptoms, one missing thing.

### Findings

**1. Parse-only PSI works with no analysis.** `KotlinCoreEnvironment.createForProduction` with an *empty* `CompilerConfiguration` — no classpath, no source roots, no resolution — yields real `KtFile`s. The expensive half of a frontend is resolution, and none of the rewrites need it.

**2. It is cheap**, measured over the 35 snippets the corpus generates:

```
environment setup : ~420 ms   (once per process)
parsed            : 35 files, 41,824 chars, in ~81 ms
per file          : ~2.3 ms
```

Against a seed path that already fetches from GitHub raw, ~2.3 ms/file is free and the setup is one-off and lazy.

**3. The tree carries what the rewrites need**: 334 named arguments and 1380 qualified expressions across those snippets — each one something a text pass had to infer and got wrong at least once. On a fixture holding every shape the review rounds broke on, the parse separates all of them, including the destructuring form the `$known-gaps` entry is waiting on.

**And the constraint is not in the way.** The spike also loads the parser from the staged `lib-bta/` through a `URLClassLoader` parented to the platform loader — the same isolation `PlaygroundBtaCompiler` already uses for the compiler. Load + parse: ~520 ms, 13 jars, nothing added to the CLI's own classpath.

### One caveat, and it was mine

The isolated-loader test first failed *only* when `PlaygroundBtaCompilerTest` ran before it — which looked exactly like a prior in-process compile leaving global frontend state, and would have been a serious finding. It wasn't. `Class.getMethods()` has no specified order, so `methods.first { … }` was selecting a different `createFileFromText` overload between runs. Selecting by exact signature fixed it; stable across repeated runs since.

Recorded in the doc because the false version was the more interesting story, and only checking told them apart.

### What it does and does not buy

A parse fixes the *machinery*, not the ratio: m3-catalog's 2 destructuring and 2 conditional-`stringResource` failures become tractable, `CatalogFilledStars` (×4) still needs a substitution rule, and meshcore's 0/10 stays structural. The corpus is what makes a rewrite safe to attempt — **3/10 and 0/10 are a per-snippet baseline** it has to match or beat.

### If this is taken forward

The `testImplementation` on `kotlin-compiler-embeddable` is scaffolding for typed assertions and **should not survive**. A real implementation wants one small module compiled `compileOnly` against the frontend, jarred beside `lib-bta/`, entered through a single reflective call — rather than the reflection sprawl this spike uses to prove the route works.

### Test plan

- `./gradlew :cli:test` — green, whole module.
- Interference case (`PlaygroundBtaCompilerTest` + spike, same JVM) run 3× — clean each time.
- `:cli:installDist` staged `lib-bta/` (13 jars, 70 MB) for the isolated-loader test; it skips with a printed note when absent.

Non-visual: a spike test and a design note.

Docs: `docs/design/PSI_PARSE_SPIKE.md`.

---


yschimke:
All four correct, and two of them invalidated things this PR had already published. Fixed in `5756ef7`; the description above and the design note are corrected.

**The `params` claim was wrong.** I wrote that named-argument binding works "with no `params` list to declare", having asserted only the labelled call. Parse-only PSI has no resolution, so `previewOverrideString("subtitle", "Basket")` carries nothing saying which position is `default` — that still needs the callee's signature, which is exactly what `params` supplies. Both halves are asserted now: one labelled call binding by name, two positional calls carrying no labels at all. Parsing **narrows** what `params` is for; it does not retire it. The case for parsing rests on the other four shapes, which it does settle outright.

**The measurement read the wrong directory** — and this is the one I'd have kept repeating. A Gradle test runs from its project directory, so `build/usage-corpus` resolved to `cli/build/usage-corpus` while `scripts/usage-corpus.sh` writes to `<repo>/build/usage-corpus`. Not a null result: it silently measured whatever was there, which turned out to be 20 stale snippets **plus 15 fixture files `UsageSnippetCorpusTest` had written beside them**, and I published it as "35 snippets the corpus generates". It now honours `composeai.usageCorpus.out`, falls back to the repository root, skips the fixture tree, and prints the path it measured so the number can be checked rather than trusted.

Corrected figures: 20 files, 42,707 chars, ~540 ms setup, **~3.7 ms/file**, 330 calls / 334 named arguments / 1412 qualified expressions. The conclusion is unchanged; the arithmetic behind it now holds.

**The destructuring check could not fail** — it tested `ktFile.text`, which is the input echoed back, so it would have passed with PSI exposing nothing, which is precisely what it existed to demonstrate. Now collects `KtDestructuringDeclaration` and asserts entries (`[checked, onCheckedChange]`) and initialiser (`toggleable("on", true)`).

**The trailing-lambda assertion was satisfiable by the wrong shapes.** Each of the three `counted` forms is now identified structurally — trailing-lambda, parenthesised, qualified trailing-lambda — and all three asserted, rather than counting to two.

Also worth noting from re-running: the `lib-bta` isolated load is ~0.5 s warm but ~5 s cold on a 70 MB / 13-jar page-cache miss. The note now gives the range instead of the flattering number.

`:cli:test` and `ktfmtCheckAll` green.

---

yschimke:
All four correct. Fixed in `35f8854`.

**The isolated-loader test never built a tree.** It stopped at `getText()`, which returns the original view-provider buffer — PSI is lazy, so it would have passed with no AST constructed at all, and measured no parse either. Notable that this is the *same vacuous-check shape* the destructuring assertion had to be rescued from one commit ago, in this same file: I fixed the instance and not the habit. It now walks reflectively via `PsiTreeUtil.findChildrenOfType` and asserts the callees found — 9 call nodes (`Text`, `previewOverrideString`, `counted`, `toggleable`).

**Package-vs-receiver was never classified.** Right, and this is the one that would have mattered: both forms are the same `KtDotQualifiedExpression`, so asserting the receiver text is reachable passes for a cleaner that goes on to make exactly the wrong call again. The test now runs the `scaffoldPackages` allow-list over the extracted receivers and asserts the split — package-qualified gets unqualified, `state.metrics` is left alone. The doc's claim is narrowed to match: the parse supplies a clean whole-expression key, it does not make the decision.

**`params` survives labelled calls too.** This is a sharper version of last round's correction, which was still wrong. Extracting the label `default` doesn't say it's index 1, and the template is indexed (`plain = "$1"`) — so `params` is the name→index map for labelled calls *and* the signature for positional ones. Parsing retires it only if the rules format also moves from `$1` to a named placeholder like `${default}`, which is a separate change rather than a free consequence. Third statement of this claim, and the first one I believe.

**Neither design note was indexed.** Added both `USAGE_SNIPPET_CORPUS.md` and `PSI_PARSE_SPIKE.md` to `docs/README.md`'s `design/` section.

`:cli:test` and `ktfmtCheckAll` green; numbers unchanged (20 files, ~540 ms setup, ~3.6 ms/file).

---

yschimke:
Both correct. Fixed in `d5a57f1`.

**The warm-up never parsed.** It called `createFileFromText` and discarded the result — PSI is lazy, so no tree was built and no parser classes loaded, and the first `collectDescendantsOfType` inside the timed block was still paying initialisation. Exactly the cost the warm-up existed to move outside the measurement. It now walks each warm-up file. Per-file: **~3.6 ms → ~3.2 ms**. Conclusion unchanged; the number now means what it says.

**The deployment-route check was skipped in normal CI.** It looked for the *installed* `lib-bta/`, which `:cli:test` doesn't stage, so on any clean checkout the only test of the proposed route returned early and passed — a broken reflective signature would have sailed through, which is precisely what happened to me two rounds ago and was caught only by luck. The build now forwards the `composePreviewBta` artifacts (the same jars the install stages) through a `CommandLineArgumentProvider`, declared as a task input so the configuration cache stays valid, and the test asserts they arrived rather than skipping.

Worth naming the pattern, since it has now cost four findings on this PR: **an assertion over `text` proves nothing about a parse.** The destructuring check read `ktFile.text` and could not fail; that got fixed, and then the isolated-loader check was found doing the same thing at `getText()`. The shape outlived its own fix. Both walk the tree now, and the design note records it as a caveat rather than leaving it as a coincidence.

`:cli:test` and `ktfmtCheckAll` green.

---

Issue #3803, opened 2026-08-14
parity: add element selection to the focused comparison page


yschimke:
Workstream 3 of #3680. Design: [`docs/design/COMPONENT_PARITY_WORKFLOW.md` §5 and §6 steps 5–6](../blob/main/docs/design/COMPONENT_PARITY_WORKFLOW.md).

Click an annotated element, or drag a region; the selection rides into the prefilled report and later becomes an acceptance's `element` selector.

## Prerequisite status

This issue originally carried a two-part prerequisite. **One part is done:**

- ~~a server-computed tag index the element gates need~~ — **complete**, carved out to #3878 and delivered by #3830 / #3860 / #3864. Both producers (live daemon projection and published-catalog artifact) and the host API `ServeHost.tagIndexForPreview` exist.
- **semantics annotations on the focused comparison** — still outstanding, and still a prerequisite for the UI below.

The comparison page today receives only the *producer-authored* annotation lists from a bundle's `annotations/index.json`. The **derived** layers — typography and theme projected from the render's own semantics tree by `ServeDesignAnnotations` — are viewer-only, because `inspect.js` is coupled to the viewer's DOM ids and to `viewer.js`'s `data-cp-src` frame recording. Mounting them over the comparison page's Actual panel means generalising that machinery to accept a host element, which is its own change with real regression surface on a shipped page.

(#3830 did make the *theme* annotation kind reachable on the comparison page — it previously had a box and a legend row built for it and no toggle able to reveal them — but that is the authored layer, not the derived one.)

## What remains

1. **Derived semantics annotations on the focused comparison**, per above.
2. **Element selection**: click an annotated element, or drag a region.
3. **Selection must be drivable from the tag index, not only from annotation boxes.** A uniquely tagged node with neither typography nor container tokens produces no annotation at all, so a page that only makes annotation boxes clickable offers nothing to select for exactly the nodes best suited to a tag selector — and the reporter falls back to a drag rectangle. That silently downgrades the acceptance to geometric, with no element gate, so a tagged glyph that later vanishes or moves goes undetected. The index already carries `{count, bounds}` per tag, which is everything a selectable target needs.
4. Selection rides into the prefilled report.

## Constraints that still hold

- **Element identity is tag uniqueness, not `ref` path shape** — see #3878 for why, and for the `count` semantics a uniqueness check reads.
- Selection is a *report* affordance. Reporting a difference must never accept it (epic non-goal).

[...]

Issue #3810, opened 2026-08-14
parity: apply the same acceptance masks in the preview-server scorer


yschimke:
Workstream 10 of #3680. Design: [`docs/design/COMPONENT_PARITY_WORKFLOW.md` §4 — the invariants and the open-problems list](../blob/main/docs/design/COMPONENT_PARITY_WORKFLOW.md).

Apply acceptances in `format-compare.js`, reporting `raw` / `accepted` / `unaccepted` separately, **including status evaluation**.

**The design document deliberately does not specify a pixel algorithm here.** Earlier revisions carried a numbered pipeline and every version of it had a real defect — a gate placed before the data it reads existed, a resample that mixed what a previous step had separated, a delta computed in the wrong direction. What it gives instead is ten invariants that are not negotiable, and an explicit list of what this workstream has to *decide*:

- the portable pixel path (which resampler, since `drawImage`'s filter is not reproducible off-browser);
- whether mask pixels participate in `edgeMask`;
- the masked pass's denominator;
- what "accepted contribution" means, given it can legitimately go negative;
- sub-pixel rounding;
- the match metric shared by the candidate gate and the resolution test, and its aggregate-vs-per-pixel form.

**Facts about the existing scorer that constrain all of it** — `scorePlanes` is more particular than it looks, and this was misread three times before the whole function was read:

1. a per-plane edge mask, a pixel being an edge when its 4-neighbour luma gradient reaches `EDGE_GRADIENT_THRESHOLD = 12`;
2. each directed pass starts from the difference at the *same coordinate*;
3. it widens to the `EDGE_SEARCH_RADIUS = 5` displaced search **only** when the source pixel is an edge *and* the same-coordinate difference already exceeds `LUMA_TOLERANCE = 16`;
4. a candidate target counts **only if it is itself an edge**;
5. each displaced match is penalised by `√(ox² + oy²) × EDGE_POSITION_COST` (10);
6. the per-pixel charge is `max(0, best − LUMA_TOLERANCE) / (255 − LUMA_TOLERANCE)`, averaged over `width × height`, both directions averaged into `mismatch`;
7. the returned score is `max(0, min(100, (1 − mismatch) × 100))` — a clamped percentage, so a port that stops at step 6 publishes every number inverted.

`ssim` / `globalSsim` are still in the file and **have no callers**.

[...]

PR #3864, opened 2026-08-14
fix(parity): wire the tag index to the host, require its declared space, and cover unbridged images

yschimke:
Three review findings raised on #3860, which auto-merged before they could be applied. All three are correct; each is fixed here with a test that fails without the fix.

Refs #3680, #3803.

## 1. The reader had no production caller

`ServeBundleHost` loads the reference, annotation, page and activity stores but never the tag index, and `ServeHost` exposed no accessor — so `#3860`'s generation *and* staging both worked and the resulting file was read by nobody. The index was reachable only from its own unit test.

Adds `ServeHost.tagIndexForPreview`, loads the store in `ServeBundleHost`, and delegates it from both composite live hosts (`ServeCatalogLiveHost`, `ServePerPreviewLiveHost`) so a trusted live catalog keeps the published index its baked half carries — the same delegation those hosts already do for annotations.

## 2. A missing `space` read as a declared one

`ServeSemanticsTags.TagEntry.space` is defaulted, so an index entry that declared *no* coordinate space deserialised as though it had declared `render-pixels`, and the host could not distinguish the two. That defeats the entire purpose of putting the space on the wire — and a later element gate reading an undeclared index would compare or transform bounds in a plane nobody stated.

The manifest now decodes into a `WireTagEntry` where absence stays representable; entries are validated and converted to the producer type only once a supported space is actually present. An **unrecognised** value is refused as well as an absent one, so a future canonical-plane producer can't be silently read as render-pixel by an older host.

This is the same class of bug as the `@EncodeDefault` one on #3830 — a Kotlin default quietly standing in for something the wire never said. Worth noting the pair: there the default hid the field on the way *out*, here on the way *in*.

## 3. Images with no live alias were skipped

`bridgeLivePreviewIds` deliberately withholds `previewId` from an image whose function the Android-only supplement overrode. That is a statement about **which daemon may re-render it** — its baked pixels aren't what the primary desktop daemon would draw — not about whether a semantics tree for those pixels exists. It does, in the supplement's own bundle.

So reusing the live alias to find semantics denied a tag index to exactly those variants (compose-m3's inset focus-ring stickers among them), and an element gate could never have run on them.

Adds `resolveSemanticsIds` — an unfiltered resolution alongside `bridgeLivePreviewIds`, mirroring `stampPreviewDensities`, which already exists for precisely this reason and says so in its own KDoc ("static catalogs and intentionally unbridged supplement images still need the density"). It reuses the same `resolveFunction` + `pickVariantId` pair, applies no `overriddenFunctions` filter, and mutates nothing.

`catalogTagIndex` prefers `image.previewId` where present and falls back to the unfiltered map, because only the bridge reconstructs `@OverrideVariant` `_VARIANT_<state>` ids. `bridgeLivePreviewIds` itself is untouched — an overridden image must still get no live alias.

## Test plan

- **20 JS cases** (was 17): the live alias winning over the unfiltered map, an unbridged image indexing from the map alone, and no-map behaviour staying exactly as before.
- **12 Kotlin cases** (was 9): a missing `space` refused, an unrecognised `space` refused, a non-positive count refused. The existing fixtures gained explicit `space` values, which is itself the point — they were previously passing on the default.
- Full `:cli:test` green; `ktfmtCheck` green; both driver modules `node --check`ed.
- The **five pre-existing driver failures** (`catalog-themes`, `figma-rest-raster`, `rc-compare-pixels`, `rc-size-modifier`, `package-version`) are unchanged, and I checked the failing set by name to confirm no sixth appeared.

## No visual evidence

Nothing rendered changes — a data product, its reader, and host wiring.

## Still open from #3830

The canonical-plane question: the design doc says the index publishes bounds already transformed into an acceptance's canonical plane by the server, while both producers emit render pixels and declare the space. Finding 2 makes that declaration actually load-bearing, which sharpens the question rather than answering it. Resolution is still either (1) the transform belongs to the comparison and the doc needs correcting — my recommendation — or (2) the index needs a comparison-scoped producer.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeBundleHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeBundleHost.kt
@@ -203,6 +203,11 @@ class ServeBundleHost(
   override fun annotationsForReference(referenceId: String): List<DesignAnnotation> =
     annotations.forReference(referenceId)
 
+  private val tagIndex = ServeTagIndexStore.load(bundleDir, fileSystem)
+
+  override fun tagIndexForPreview(previewId: String): Map<String, ServeSemanticsTags.TagEntry> =
+    tagIndex.forPreview(previewId)
+
   /**
    * Per-preview `state`/`theme` from the catalog's `previews/variants.json` manifest (written by
    * [ServeCatalogStore]). Empty for a plain uploaded bundle that carries no manifest — every
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogLiveHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogLiveHost.kt
@@ -180,6 +180,9 @@ class ServeCatalogLiveHost(
   override fun annotationsForReference(referenceId: String): List<DesignAnnotation> =
     baked.annotationsForReference(referenceId)
 
+  override fun tagIndexForPreview(previewId: String): Map<String, ServeSemanticsTags.TagEntry> =
+    baked.tagIndexForPreview(previewId)
+
   // The catalog's published player comparison rides the baked staging dir, so it stays reachable
   // when a live daemon fronts this session.
   override fun rcCompare(): RcCompareManifest? = baked.rcCompare()
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHost.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHost.kt
@@ -48,6 +48,20 @@ interface ServeHost : AutoCloseable {
   fun annotationsForReference(referenceId: String): List<DesignAnnotation> = emptyList()
 
   /**
+   * The **published** tag index for [previewId] — `testTag → {count, bounds}`, the element identity
+   * a scoped parity acceptance resolves against. Empty by default and for any host that publishes
+   * none.
+   *
+   * This is the *static* half of the pair: a published catalog's renders happened in CI, so its
+   * index is computed there and read back by [ServeBundleHost] from `tags/index.json`. A
+   * daemon-backed [ServeRenderHost] instead projects the same shape live, per render, inside its
+   * `.annotations` response ([ServeSemanticsTags]) — where it can be keyed to the frame it came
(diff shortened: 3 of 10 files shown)


yschimke:
Two of the three findings are fixed in `58f2174`; the third I'm deliberately not actioning here, with reasoning.

**Supplement candidates must replace primary ones (P1) — fixed.** Correct, and the mechanism is worth restating: `previewsByFunction` dedupes by preview *id* and appends, so reversing the bundle list only reorders candidates. A primary `Foo_Dark` and a supplement `Foo` sharing one function both stayed in the list, and `pickVariantId` could score the primary higher — publishing bounds from a tree describing pixels the supplement drew. Wrong geometry as fact is worse than the absent entry it replaced. Added `previewsByFunctionReplacing`, and confirmed the new test fails against the reverse-and-append version rather than assuming it would.

Worth flagging beyond this PR: `stampPreviewDensities` claims the same "extra renders replace primary renders" property from the same construct, so it appears to have the same gap today. I've left it alone — it stamps densities that feed Figma export scale, so changing it is a behaviour change wanting its own change and its own evidence.

**Synthetic `@OverrideVariant` ids (P2) — fixed.** Also correct, and broader than described: `--source-module` is conditional on `live-rerender-source`, so a catalog built with neither it nor `--publish-live-bundle` never runs `bridgeLivePreviewIds` at all and has no `image.previewId` for *any* image. `resolveSemanticsIds` already covered that general case; the override variants needed the synthetic reconstruction, now mirrored from the bridge including the "only if it actually rendered" guard.

**No production consumer for the accessor (P2) — not actioning, and here's why.** The observation is accurate: nothing calls `tagIndexForPreview`, so loading `tags/index.json` cannot yet change an HTTP response.

But I'd draw a distinction from the finding this PR opened with. That one was a *broken chain* — staging wrote a file that no code path opened, so the artifact was unreachable in principle. That's fixed. What remains is *sequencing*: the chain is complete up to the host API, and its consumer is the element gate, which lands in #3810 with element selection in #3803. Neither exists yet.

Inventing an HTTP surface now would mean guessing the shape before the consumer that defines it. The natural home is the focused comparison page's embedded payload — that's where element selection will read it — but adding it there ahead of #3803 means shipping a payload nothing parses, plus fixture churn on a page this PR otherwise doesn't touch.

Happy to add it if you'd rather see it exposed now; say which surface and I'll do it in a follow-up. Otherwise it arrives with #3803.

---

Issue #3812, opened 2026-08-14
parity: document reporting, acceptance, verification, and closure


yschimke:
Workstream 12 of #3680. Design: [`docs/design/COMPONENT_PARITY_WORKFLOW.md` §6 step 13](../blob/main/docs/design/COMPONENT_PARITY_WORKFLOW.md).

Document the reporting → triage → acceptance → verification → closure loop in `docs/public-preview-server.md`, beside the existing parity view section.

This is the *consumer-facing* half. `COMPONENT_PARITY_WORKFLOW.md` is the design record — why each rule exists and what breaks without it — and it should stay that. What a catalog owner needs is shorter and different in kind:

- how to file a parity issue from a comparison, and what the locator block is for (and that editing it breaks the index);
- which labels to use — `area:{spec,component,preview,renderer,comparison}`, `parity:{regression,known-difference,verification-needed}` — and that there is deliberately no per-component label;
- how to author an acceptance: the directory layout, the mandatory tracking issue, the mask encoding rules, and the fact that a tolerance which needs to be large is evidence the acceptance is wrong;
- what each status means when it appears on the dashboard, and what to do about it;
- the closure step, and why deletion and issue-closing happen in one PR.

**Also needs setup documentation** for the credential path — the App must be installed on the *catalog* repo with **Contents: write**, and `actions/create-github-app-token` must be given `owner` and `repositories`. That default is the thing that looks right and silently is not, and a 403 there looks exactly like "no issues changed", so it belongs in prose where someone provisioning a new source repo will find it.

**Worth noting**: mask authoring has no UI in this plan — a human hand-writes `known-differences.json` and produces the two PNGs. That is probably acceptable for the first acceptances, since they should be rare and deliberate, but "export this selection as an acceptance stub" from the focused comparison is the obvious follow-up and is worth deciding on rather than discovering.

**Depends on** everything else being real enough to describe.

Issue #3811, opened 2026-08-14
parity: detect resolved, stale, and invalidated acceptances — and close the loop


yschimke:
Workstream 11 of #3680. Design: [`docs/design/COMPONENT_PARITY_WORKFLOW.md` §6 Phase 4](../blob/main/docs/design/COMPONENT_PARITY_WORKFLOW.md).

Surface the statuses the scorer already computes — `resolved`, `invalidated`, `refused` — plus **stale**, and then actually act on them.

**This step surfaces statuses; it does not compute them.** `resolved` detection lands with the scorer, because that engine's own fixtures require it.

**Constraints:**

- **Stale and unknown are a second axis, not two more statuses.** `status` is what a *comparison* concluded and is exactly what the two engines must agree on; stale/unknown come from the issue index, which one engine may have and the other may not. Folding them into `status` would make the conformance fixtures depend on a file outside the contract. The join happens here — at the dashboard and the gate — over an acceptance's mandatory `issue` URL, producing `open` / `closed` / `unknown` per acceptance. The evaluation result gains no field.
- **Only one combination is "stale"**: `closed` + status not `resolved`. `resolved` + `closed` is the loop having *completed*. `unknown` is never stale.
- **Stale requires positive evidence of closure, never inference from absence.** The index is fail-soft, capped, can lag between regenerations, and does not cover source repos whose dispatch credential nobody has wired. Treating absence as closed would mark live acceptances stale across an entire catalog the first time the file failed to parse — which is why the index publishes closed rows.

**And the loop needs an owner.** Nothing above actually *removes* anything: `resolved` means deleting the acceptance, and an issue closes once every linked acceptance has resolved, but surfacing a status does neither. Both are committed-file operations, so both belong in a PR, and the ordering matters — delete first and the issue loses the only record of what it was about; close first and the surviving siblings are briefly stale against a closed issue. So they happen **in one change**: a PR that deletes the resolved records and directories and closes the issue via a closing keyword in its description, so the merge does both atomically.

**With one restriction**: "every acceptance linked to this issue" is not evaluable from a run that reads one `known-differences.json`. `v1` gives an issue a single owning document, and a run that cannot establish ownership deletes its resolved records but **omits the closing keyword**, leaving the close to whoever can see the other referencing documents. Deletion is always safe locally; closure is the half that needs knowledge a single run does not have.

**Depends on** the scorer workstreams and the index reader.

Issue #3809, opened 2026-08-14
parity: publish known differences through catalog-export


yschimke:
Workstream 9 of #3680. Design: [`docs/design/COMPONENT_PARITY_WORKFLOW.md` §4 and §6 step 10](../blob/main/docs/design/COMPONENT_PARITY_WORKFLOW.md).

Carry `.design-parity/known-differences.json` and its artifact directories from the source repo through `@design-parity/catalog-export` into the published catalog, so a serving host can read the same acceptances the offline run does.

**Constraints:**

- **The artifacts are fetched, not just referenced.** Once a mask and an accepted-candidate PNG are published catalog assets they go through `ServeCatalogStore`'s ordinary staging, which applies `MAX_FETCH_BYTES` (25 MB per asset) — one of the reasons the schema caps artifacts at 8 MiB, comfortably below it, so the two engines agree well before a fetch would fail.
- **Path containment is checked at staging**, on a host that fetches third-party catalogs, so a traversal here is an escape from the artifact tree rather than a typo. The schema's portable-path grammar exists so the fetching host and the offline reader resolve the same file.
- **Publishing must not require a re-render**, the same property `parity/issues.json` relies on: `ServeCatalogRefresher` re-fetches on any branch head move and `ServeCatalogStore.load` re-stages the whole tree.
- Note the existing automation boundary — catalog and export-driver merges regenerate the delivery branches automatically; **renderer/plugin/CLI changes do not**, and need a manual `design-artifacts.yml` dispatch. Adding a new published path means checking `scripts/design-artifacts/scope-systems.sh` maps it, since a wrong mapping silently strands a catalog on stale artifacts.

**Depends on** the schema. **Cross-repo.**

yschimke:
Delivered by yschimke/design-parity#388 (merged), with the serving half in #4480.

`@design-parity/catalog-export` now carries `.design-parity/known-differences.json` and its artifact tree into the published catalog at `parity/known-differences.json` and `parity/known-differences/<id>/<file>` — the paths `ServeKnownDifferences` reads.

Against this issue's constraints:

- **The 8 MiB artifact cap is enforced at publish**, inclusive at the ceiling, so a bundle cannot carry a record its consumers would refuse. Over-size artifacts are reported in `skipped` rather than thrown, because a catalog with one broken acceptance is still worth publishing.
[...]

Issue #3808, opened 2026-08-14
parity: apply scoped acceptances in design-parity while preserving raw findings


yschimke:
Workstream 8 of #3680. Design: [`docs/design/COMPONENT_PARITY_WORKFLOW.md` §4 and §6 step 10](../blob/main/docs/design/COMPONENT_PARITY_WORKFLOW.md).

The offline half of the acceptance engine: `design-parity` reads `known-differences.json`, applies the same semantics the browser does, and reports `raw`, `accepted` and `unaccepted` separately.

**The raw finding is always preserved** (epic requirement, and the reason acceptance is not an ignore rectangle). An acceptance never removes a difference from the report; it moves it into the `accepted` bucket while `unaccepted` keeps everything else visible.

**Constraints:**

- **Separation precedes resampling.** The masked and unmasked regions of *both* inputs are split in their own pixel space, per acceptance, before any resample — never a pre-averaged composite. `drawImage` averages **signed** luma, so on a footprint straddling a mask edge an accepted difference can cancel an opposite unaccepted regression before scoring sees the pixel.
- **Only `valid` acceptances contribute a mask to the scoring union.** `resolved`, `invalidated` and `refused` suppress nothing. Keeping a `resolved` mask would remove its pixels as neighbourhood candidates for the pixels around it, which can hide a regression sitting next to the thing just fixed.
- **The union cannot be formed before the gates have run**, and the gates need canonical pixels — so the surviving-union split is a distinct, later stage, not the same pass.
- **Status precedence is strict**: `refused` → any non-`candidate-changed` invalidation → `resolved` → `candidate-changed` → `valid`. `resolved` is guarded on the candidate gate having actually *fired*, otherwise a tolerant metric marks an unchanged candidate resolved the moment it is authored.
- **This engine and the browser must agree bit for bit on which mask survived**, which is what the shared conformance fixtures exist to prove. A single aggregate status cannot express a mixed-validity case, so both engines could emit the same summary while disagreeing about the details.

**Depends on** the schema and its fixtures. **Cross-repo** — sequence after the schema is frozen so both sides build against one definition.

yschimke:
Implemented in yschimke/design-parity#391 and the post-merge review corrections in yschimke/design-parity#392. The follow-up preserves refusal reasons, accepts the v1 empty variant, strips candidate-only image semantics at the catalog boundary, and pins the compact conformance corpus identity/index. Full build and all 1,846 tests pass.

PR #3879, opened 2026-08-15
docs(parity): add implementable batch plans for the component-parity epic

yschimke:
Splits the twelve workstreams of #3680 into seven implementable batches under `docs/design/parity-batches/` — one branch's worth of work each, with what to build, what to read first, the traps already known, and what "done" means.

`COMPONENT_PARITY_WORKFLOW.md` stays the design record (why each rule exists). These files are the delivery plan (what to do, in what order, and what has already been decided).

## The batches

| Batch | Covers | Depends on | Visible payoff |
| --- | --- | --- | --- |
| 00 — decisions and corrections | — | nothing | no (unblocks 02/04/05) |
| 01 — locator + report body | #3801, #3802 | nothing | partly — issues become machine-identifiable |
| 02 — issue index end to end | #3804, #3805, #3806 | 01 | **yes** — open issues on the pages |
| 03 — element selection | #3803 | 01 | **yes** — click an element to report it |
| 04 — acceptance schema | #3807 | 00 | no (contract + conformance fixtures) |
| 05 — both engines + publish | #3808, #3809, #3810 | 03, 04 | **yes** — accepted vs unaccepted scores |
| 06 — resolution, closure, docs | #3811, #3812 | 02, 05 | **yes** — the loop closes |

**01 and 02 can run ahead of everything else** and deliver the epic's first four acceptance criteria. If only one batch gets done, do 01 — every issue filed after it lands is machine-identifiable, and issues filed before it need their bodies hand-edited later.

## Batch 00 exists on purpose

Four questions currently sit unsettled and each one blocks a later batch. Writing code against an unsettled question is how the design doc accumulated three wrong pixel pipelines, so they get recorded with options and a recommendation rather than discovered mid-implementation:

- **D1** — which plane the element tag index reports bounds in. The doc says canonical; neither producer can compute a canonical plane (it is resolved per comparison). Recommendation: the transform belongs to the comparison, and the doc needs correcting.
- **D2** — which surface carries the report form. The viewer's always-available number is `scoreSvgUrls` (PNG vs generated SVG) — a render-fidelity measurement that would feed the index as a mislabelled parity score. Recommendation: the focused comparison, which knows both `referenceId` and the real score.
- **D3** — the score rebaseline is versioned, and must not share a change with any acceptance-semantics edit.
- **D4** — the frame-vs-controls race in `refreshReportLink()`; take the crude fix first, because the better one can be got subtly wrong and pass review.

Plus three non-blocking loose ends: the `stampPreviewDensities` fold bug, the `vscode-preview/main` baseline instability, and the caller-less `ssim` / `globalSsim`.

## Inherited conventions

The README carries the conventions the already-delivered work established the hard way — fail-soft on anything read from a catalog; a published artifact needs a producer **and** a `ServeCatalogStore` staging call **and** a host loader (both of the other two have been missed here); never let a language default stand in for a wire fact; assert on the wire, not the round trip; confirm a new test fails without its fix.

## Test plan

Documentation only — no code, no build wiring, no rendered surface changes, so there is nothing to capture visually. Verified by reading: every issue number, file path, symbol (`scorePlanes`, `ServeParityActivity`, `ServeTagIndexStore`, `ServeIssueReport.withoutToken`, `refreshReportLink`, `scope-systems.sh`, `push-branch.sh`) and named constant (`EDGE_GRADIENT_THRESHOLD = 12`, `EDGE_SEARCH_RADIUS = 5`, `LUMA_TOLERANCE = 16`, `EDGE_POSITION_COST = 10`, `MAX_FETCH_BYTES` 25 MB) cited in the batch files was checked against the tree at `eb2f4db`, and the relative links between the README and the seven batch files resolve.

---

### DIFF
--- a/docs/design/parity-batches/00-decisions.md
+++ b/docs/design/parity-batches/00-decisions.md
@@ -63,9 +63,14 @@ Any resampler that is not `drawImage` produces different pixels, so the publishe
 **once**, deliberately. Invariant I10 is what stops them moving a second time. The decision is not
 *whether* but *how it is announced*:
 
-**Do:** bump the score schema version, regenerate committed baselines **in the same change**, and
-note it in the release notes. Never in a change that also alters acceptance semantics — a moved
-number and a changed verdict in one diff cannot be told apart.
+**There is no version to bump yet — that is part of the decision.** `scoreImages` returns
+`{percent, geometry}` and nothing else, and a repo-wide search finds no versioned parity-score
+carrier and no committed score-baseline format. So a downstream reader currently has no way to tell
+an old-kernel number from a rebaselined one.
+
+**Do:** identify the carrier or introduce one *first*; then bump it, regenerate committed baselines
+**in the same change**, and note it in the release notes. Never in a change that also alters
+acceptance semantics — a moved number and a changed verdict in one diff cannot be told apart.
 
 ## D4 — the frame-vs-controls race in `refreshReportLink()`
 
--- a/docs/design/parity-batches/00-decisions.md
+++ b/docs/design/parity-batches/00-decisions.md
@@ -1,0 +1,96 @@
+# Batch 00 — decisions and corrections
+
+**Issues:** none — this batch exists because writing code against an unsettled question is how the
+design doc accumulated three wrong pixel pipelines.
+**Depends on:** nothing.
+**Blocks:** [02](02-issue-index.md) (D2), [04](04-acceptance-schema.md) (D1), [05](05-acceptance-engines.md) (D1, D3).
+**Ships:** no user-visible change. Output is edits to
+[`../COMPONENT_PARITY_WORKFLOW.md`](../COMPONENT_PARITY_WORKFLOW.md) and to the affected issues.
+
+Each decision below is stated with the options, the recommendation, and what breaks if it is deferred
+past the batch that needs it. None is large. All are cheap now and expensive after two engines have
+been written against opposite readings.
+
+---
+
+## D1 — which plane the element tag index reports bounds in
+
+**Blocks:** 04 (the `plane` discriminant), 05 (the element gate), 03 (what a selection records).
+
+The design doc says the server publishes tag bounds already transformed into an acceptance's
+*canonical plane*. **Neither producer can do that.** The canonical plane is resolved per comparison,
+from a reference raster and an acceptance record; `ServeSemanticsTags` sees one daemon render and
+`scripts/design-artifacts/tag-index.mjs` sees one packed bundle. Both therefore emit `boundsInRoot`
+render pixels and declare `space: "render-pixels"` on the wire, and `ServeTagIndexStore` rejects any
+entry that declares anything else — so nothing can silently consume these as canonical today.
+
+| Option | Consequence |
+| --- | --- |
+| **(a) The transform belongs to the comparison.** Index stays render-pixel; the comparison maps into the canonical plane at gate time. | Doc §4/§5 need correcting. No producer changes. **Recommended.** |
+| (b) The index becomes comparison-scoped. | A third producer keyed by `(previewId, referenceId, acceptance)`, recomputed per comparison, and the published `tags/index.json` stops being a per-preview artifact. |
+
+(a) is recommended because a plane is a property of a *comparison*, and the index is a property of a
+*render*. Under (b) the published artifact would have to be regenerated whenever an acceptance
+changed, which is precisely the "publishing must not require a re-render" property batches 02 and 05
+both rely on.
+
+**Do:** amend the doc to say the index publishes render-pixel bounds and names its space, and that
+the canonical-plane transform is a step of the comparison. Then relax or keep `space` validation
+deliberately rather than by accident.
+
+## D2 — which surface carries the report form
+
+**Blocks:** 02 and 03 (both put things on "the page where you see the problem").
+
+[#3802](https://github.com/yschimke/compose-ai-tools/issues/3802) records that the form currently
+lives only on the viewer, whose always-available live number is `scoreSvgUrls` — PNG against the
+*generated SVG*. That is a render-fidelity measurement, not a design-parity one. Emitting it into an
+issue body labelled as a parity score produces a plausible, mislabelled number feeding an index.
+
+| Option | Consequence |
+| --- | --- |
+| **(a) Move the form to the focused comparison** `/{system}/compare/{previewId}`. | Always has a concrete `(previewId, referenceId)` pair *and* the real score, and is where element selection lands in 03. Viewer keeps a link to it. **Recommended.** |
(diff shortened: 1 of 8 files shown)


yschimke:
All ten review findings folded in as of b9aa4cb4 — each verified against the tree before writing it down. Two are worth calling out here rather than leaving in the diff.

**Confirmed against the code, and load-bearing:**

- `handleReferenceComparison` (`ServeHttpServer.kt:2452`) reads exactly `name` and `reference`. Moving the report form to the focused comparison therefore needs the normalised override map forwarded *first*, or the reporter files the default render against a locator describing something else — the same identity-vs-pixels mismatch batch 01 exists to prevent, arriving from the other side. Recorded as a prerequisite of D2.
- The `tagIndexForPreview` point is right and it corrects something I asserted the other way round while the index was being built. That API is the published static index and both live hosts delegate it to baked; the live index rides `.annotations`, which re-renders and so does not describe the fetched PNG either. Element gates need a shared render generation before they can be switched on. Selection and reporting are fine without it — a reported selection is a claim about what the reporter saw, not a verdict.
- `candidateTolerance` is specified integer at `COMPONENT_PARITY_WORKFLOW.md:349` and my summary dropped it; the hash rules at `:963-972` are genuinely two rules, and I had collapsed them into one fixture. Both corrected, with the extra fixtures named.

**One correction to the P1 on no-op acceptances.** The concern is real and now recorded as a decision to make before the batch-04 fixtures freeze, but the stated mechanism isn't: a later regression inside the mask changes the candidate, so the candidate gate *does* fire and the region *is* reported. What a no-op acceptance actually costs is a sub-tolerance regression suppressed in a region that never differed, plus its pixels being removed from the neighbourhood search of everything around it — which is the same objection the design doc already makes to keeping a `resolved` mask alive. That is enough to justify considering an `acceptance-is-noop` refusal alongside the existing `mask-empty` one; it isn't a hidden-regression hole.

---

yschimke:
Two red checks on `b9aa4cb4`, neither caused by this PR. Its diff is eight markdown files under `docs/design/parity-batches/` and nothing else.

**`ktfmt (Google style)` — broken on `main`, not here.** Two files fail:

- `renderers/android/src/main/kotlin/.../RobolectricRenderTest.kt` — modified by #3877 (`ef5d1baa`), which merged ~10 minutes before this push. CI checks the merge with `main`, so it is *that* version being formatted, not the older one on this branch.
- `renderers/android/src/test/kotlin/.../PreviewManifestLoaderStaleFanoutTest.kt` — byte-identical (`5039b84b`) on `main` and here.

The `format.yml` run on `main` itself confirms it: **failure at `ef5d1baa`**, success at `eb2f4db3` immediately before. So this is #3877's regression on `main`, and it will keep failing here until `main` is fixed. Deliberately not fixing it in this PR — a docs-only change should not be reformatting renderer Kotlin, and the fix belongs with the change that caused it.

**`CMP/Wasm Frame Pacing` — a flake.** `Cannot determine the version of Skiko for Compose '1.11.1'` while resolving `:rc-player-wasm:unpackSkikoWasmRuntime`. That is dependency resolution, which a markdown diff cannot affect, and the **same job passed on `main` at `ef5d1baa`**. Re-run is currently refused (`403 This workflow is already running`); I'll retry once the run settles.

I'll re-check when `main` goes green and merge it forward so CI re-runs against a fixed base.

---

yschimke:
Second round folded in as of `85a644ce`. All seven verified first; three of them found claims in these files that the repo does not currently support, which is exactly what a plan doc should not be shipping.

**Three unsupported claims, now fixed:**

- **D3 had nothing to bump.** `scoreImages` returns `{percent, geometry}` and a repo-wide search finds no versioned score carrier and no committed score-baseline format. "Bump the score schema version" was an instruction to modify something that does not exist — the decision now has to identify or introduce a carrier first.
- **Batch 05's no-rerender claim was wrong.** `design-artifacts.yml` is `bundle pack → catalog-export → publish out/`, so anything routed through it waits on a full render, and `scope-systems.sh` only *selects* that render. The issue index gets the property because §3 gives it a **separate one-file committer** — the doc says outright that the render workflow is "the wrong granularity for 'someone relabelled an issue'". An acceptance edit is the same granularity. The batch now has to pick: build the equivalent publisher, or drop the claim.
- **`compose-preview-issues/v1` has no field-level schema.** §3 says "shape as in the epic"; there is no manifest, row example or fixture in the tree. Batch 02 now opens by defining the rows and building the pure `parity-issues.mjs` producer plus the committed fixture the Kotlin reader's test loads — the arrangement `parity-activity.mjs` already uses, which I had failed to cite at all.

**Best catch of the round:** *"no tag index" must mean "no suppression"*. `ServeTagIndex.kt`'s KDoc — which I wrote — says a missing entry "degrades to no element gate, which is the safe direction". That is only true if the acceptance also stops suppressing. An element-scoped acceptance whose gate cannot run but whose mask still joins the valid union has silently become a plain ignore rectangle, so a tagged element that has vanished or moved goes on being hidden — the exact failure the element gate exists to catch. Recorded, with the KDoc correction attached to the same change.

Also folded in: the browser needs a catalog-wide orphan walk (a removed target is never scoped into any comparison, so it stays invisible in the browser while `design-parity` reports `orphaned-target`); conformance must pin intermediate planes and run **three** runners, since the JS suite consumes an already-built index and never exercises the Kotlin projector; and D2/D4 are hard prerequisites of batch 01 rather than optional, with pending-render and failed-render as their own tests.

CI status is unchanged from [my earlier comment](https://github.com/yschimke/compose-ai-tools/pull/3879#issuecomment-5301172413): `ktfmt` is red because `main` is red at `ef5d1baa`, and the Skiko resolution failure was a flake. Neither is caused by this diff, which remains markdown only.

---

yschimke:
Correcting my earlier diagnosis, and the `ktfmt` blocker now has a fix: **#3881**.

I attributed the breakage to #3877. That's where it first *appears* — its run was the first to complete — but not where it came from. The actual cause is **#3859 bumping `ktfmt` 0.26.0 → 0.27.0**: 0.27 wraps trailing-lambda chains and short named arguments differently, so two files stopped conforming without being edited. #3859's own `format.yml` run was cancelled rather than completed, as were #3871's and #3876's, so the rule change landed unverified. The giveaway is that `PreviewManifestLoaderStaleFanoutTest.kt` is byte-identical (`5039b84b`) before and after #3877 and fails regardless.

I could reproduce it once I dropped `--offline` (the new plugin version wasn't in the local cache — itself a symptom). `./gradlew :renderer-android:ktfmtFormat` produces exactly the two files CI names, whitespace only. That's #3881, on its own branch — this PR stays markdown-only.

Also correcting something I said in my [earlier comment](https://github.com/yschimke/compose-ai-tools/pull/3879#issuecomment-5301172413): merging `main` forward would *not* have helped here. It would have imported the unformatted files. The merge-forward only makes sense once #3881 lands.

`ktfmt` here will stay red until then. Nothing else on this PR is failing — the CMP/Wasm parity job has since reported green.

---

Issue #3480, opened 2026-08-07
serve viewer registers no @font-face, so the browser RC lane renders in the viewer's own fonts


yschimke:
The in-browser Remote Compose lane (`?mode=rc`, the `js` backend) draws text in whatever fonts the *viewer's machine* happens to have, not the fonts the baked reference was rasterised with. Text in that lane therefore renders in a different typeface — and at different metrics — depending on who is looking at the page.

## Why

`cssFontStackFor` in the vendored player names a concrete face ahead of the CSS generic, and its own comment is explicit that this is only a request:

> The concrete names are only a *request*: a page that has not registered the faces falls straight through to the generic and renders exactly as it did before […] The parity harness (`scripts/design-artifacts/rc-compare.mjs --fonts`) registers them from the same vendored files the snapshot renderer rasterizes with, which is what turns the request into a match.

So the design is sound — the player asks for `Roboto, sans-serif`, and whoever hosts it is expected to register the faces. `rc-compare.mjs` does exactly that via `rc-fonts.mjs` (`fontFaceCss` + `loadAndVerifyFonts`, 4 faces from `samples/cmp-wasm-catalog/src/wasmJsMain/resources/fonts`).

The served viewer does not. `serve.css` contains no `@font-face` rule at all — only page-chrome stacks (`system-ui, -apple-system, Segoe UI, Roboto, sans-serif` and a mono stack). Nothing registers the vendored faces for the canvas the player paints into.

## Effect

Measured in headless Chromium, per 100px:

| stack | fontBoundingBox ascent + descent |
|---|---|
| `Roboto` registered (as the parity harness does) | 93 + 24 = **117** (1.17em) |
| `Roboto, sans-serif` unregistered → container fallback | 91 + 21 = **112** (1.12em) |

Two consequences, both silent:

1. **Wrong metrics.** A ~4% line-box difference compounds down a card, on top of any layout divergence.
2. **Wrong weight.** With no registered Roboto, CSS weight matching for `500` finds no medium face and falls back to `400`, so text the document asks for at Medium renders Regular. This is visible on the Home Assistant catalog's tile/verticalstack titles, which are bold in the bake and regular in the browser lane.

Neither is a player bug — with the faces registered, both go away. I originally mis-attributed the weight symptom to the player before noticing my own harness wasn't registering fonts either.

## Suggested fix

[...]

PR #3905, opened 2026-08-15
refactor(serve): port the Remote Compose font preloader to Lit tooling

yschimke:
Fourth asset onto the typed, tested toolchain in `cli/serve-web`, after the Transparent toggle (#3838) and the backend badge + group memory (#3892). `rc-fonts.js` becomes `cli/serve-web/src/rcFonts.ts`.

## It stays a global, deliberately

The first three ports were custom elements. This one is not one, and forcing it to be would be wrong: `window.cpRcFonts.ready()` is called by the legacy IIFEs — `viewer.js`, `format-compare.js`, and the inline doc-player script — which read it at call time and already guard for its absence. Moving the implementation buys the type check and the tests. Changing the *seam* as well would mean editing three untyped callers in the same change, which is the opposite of one reviewable step; the seam goes when those callers do.

That's a third shape worth naming alongside the two the README already had, and it's written up there.

## Where the bundle is loaded from

The three pages that loaded `rc-fonts.js` now load `serve-components.js` instead. The viewer already loaded the bundle above its own scripts, so there the tag simply goes; the format-compare wall and the `.rc` permalink gain it in the same slot, ahead of the scripts that call the global.

None of those three pages emits a `cp-` element, so the bundle registers definitions that match nothing there and can move no pixels.

## What the tests found

It had no test of any kind. Ten now cover the two promises a Remote Compose lane actually leans on (#3480): that `ready()` walks the font registry exactly once however many lanes ask, and that it **never rejects** — a face that 404s, one the browser cannot even start, an engine with no Font Loading API, and a registry that cannot be iterated all have to degrade to the fallback face rather than stop the lane rendering.

Writing that last one down found a real gap. The original:

```js
return Promise.all(loads).then(
  function () { return document.fonts.ready; },
  function () {}
);
```

The rejection handler covers `Promise.all(loads)`, but not the `document.fonts.ready` returned by the success handler — so a rejecting registry would have rejected `ready()` in spite of the contract documented directly above it ("resolves once (memoized) and NEVER rejects"). The spec says `fonts.ready` does not reject, which is why nobody hit it. The port uses `.catch` on the whole chain so the guarantee actually holds, and `resolves when fonts.ready itself rejects` fails without it.

## Visual evidence

Byte-identical captures for every page this change touches — `serve-format-compare`, `serve-rc-lanes`, `serve-doc-remotecompose`, `serve-viewer-rc-players`, both themes and their runtime states, 12 PNGs, captured on this branch and on `main` and compared by hash:

```
base: 12 head: 12    differing: 0
  same  serve-doc-remotecompose.{dark,light}.png
  same  serve-format-compare.{dark,light}.png
  same  serve-rc-lanes.{dark,light}.png
  same  serve-rc-lanes-diff-baked.{dark,light}.png
  same  serve-viewer-rc-players.{dark,light}.png
  same  serve-viewer-rc-players-player-cmp-android.{dark,light}.png
```

No before/after images are embedded because there is nothing to show: the pairs are the same bytes, and a font preloader has no visual surface of its own beyond the faces the lanes were already meant to paint in.

**A targeted comparison rather than the full sweep, because the full sweep is red on `main` right now.** 15 of the 120 `pages-snapshot` tests fail on `origin/main` with my change reverted — 12 viewer captures plus 3 contract tests, all with `element is not visible` on a control-group summary. #3898 changed the viewer to open with its controls drawer closed (`class="cp-viewer"` with `aria-expanded="false"`, where it was `cp-viewer cp-controls-open`), and the harness states that reach straight into a group were not updated to open the drawer first. Not this branch's regression, and not fixed here — flagged separately.

## Test plan

- `cd cli/serve-web && npm run verify` — typecheck, 42 mocha tests (10 new), rebuild, committed-bundle check ✅
- `./gradlew :cli:test :cli:ktfmtCheck` ✅
- `pages-snapshot` narrowed to the four affected fixtures, on this branch and on `main` — 12/12 byte-identical ✅

---


yschimke:
`Android (Robolectric) serve lane honors overrides` is red, and it is not this branch — it fails on the base branch for the same reason, so I'm not fixing it here.

Both failures resolve their locator and then find it **hidden**:

```
locator resolved to <input type="text" class="cp-knob" ... data-knob-key="label" ...>
   - unexpected value "hidden"
```

`openOverridesGroup()` opens the inner `details.cp-group[data-cp-group="overrides"]`, but that group lives inside `.cp-controls`, and:

```css
.cp-viewer:not(.cp-controls-open) .cp-controls { display: none; }
```

#3893 dropped `cp-controls-open` from the viewer markup — `ServeWeb` now emits `<div class="cp-viewer" …>` where it emitted `<div class="cp-viewer cp-controls-open" …>` — so the drawer starts closed and everything inside it is `display: none` until something opens it. `git log -S'cp-viewer cp-controls-open'` points at `e949dca` (#3893).

Same cause takes out 15 of the 120 `pages-snapshot` tests, which I reproduced on `origin/main` with this branch's commit reverted (12 × `element is not visible` on `[data-cp-group="explode"] > summary`). So two harnesses are down on `main`, not one.

Worth noting the CSS comment right above that rule still reads:

> the open state is a class on .cp-viewer — cp-controls-open (**default on**) and cp-nav-open (default off)

which is no longer true, so it's worth confirming whether the closed-by-default overrides drawer was intended before deciding whether the fix belongs in `ServeWeb` (restore the class) or in the ~17 harness call sites (open the drawer first). I've asked; happy to push whichever, in its own PR rather than this one.

This branch itself is clean: `:cli:test` + `ktfmtCheck` green, and the four page fixtures it affects capture byte-identical to `main`.

---

PR #3911, opened 2026-08-15
test(harness): follow #3893's viewer, restoring 14 page captures

yschimke:
`pages-snapshot` has been red on `main` since #3893: **15 failed, 105 passed**. This takes it to **1 failed, 119 passed**.

Two separate casualties, both from the same PR, and both showing up as a locator that resolves (or doesn't) and then never becomes actionable — which is why they read as timeouts rather than assertion failures.

## 1. The controls drawer starts closed

#3893 changed the emitted markup from `class="cp-viewer cp-controls-open"` to `class="cp-viewer"`, and `serve.css` has:

```css
.cp-viewer:not(.cp-controls-open) .cp-controls { display: none; }
```

So every step reaching into a `details.cp-group` was acting inside a `display: none` subtree. Opening the inner `<details>` doesn't help — the drawer is the outer hidden thing:

```
locator resolved to <summary>Exploded 3D</summary>
  - element is not visible   (×115 retries)
```

A shared `openControlsDrawer` in `_viewer.mjs` opens the column first, from the six `pages-snapshot` call sites. It clicks the real `#cp-controls-toggle` rather than adding the class, because `viewer-drawers.js` moves more than the class — `aria-expanded`, the mobile scrim and the per-catalog fold preference travel together, and a capture taken after a class poke would show an open drawer above a toggle still claiming to be closed, a state no visitor can reach. Idempotent, because that control is a toggle.

`serve-lanes.spec.mjs` already got an equivalent inline fix on `main`, so it is untouched here. The logic now exists in two places; worth collapsing onto the shared helper sometime, but not by rewriting a just-merged file.

## 2. `#cp-axes-toggle` no longer exists

#3893 also deleted the axes fold outright — `#cp-axes-toggle` and `#cp-axes` are gone from the fixtures *and* from `ServeWeb`. Three states were clicking a control that isn't in the markup, so their locator never even resolved. The state axis is now the component nav's variant tree, whose rows arrive `aria-expanded`.

What each shot is evidence for is unchanged, so each now opens the thing that actually reveals those rows — the nav, which is closed at the harness's 1024px width. `axes-folded/nav-closed` simply drops its re-fold click for a fold that no longer exists.

## Visual evidence

There is no "before" for these: on `main` they produce no PNG at all, because the test dies before the screenshot. The evidence is that each shot now captures the thing it was written to diff, committed under `renders/harness-controls-drawer/`.

**The Exploded 3D camera sliders** — in the Overrides drawer, unreachable since #3893:

exploded controls

**The cross-product variant rows** — the whole claim of that shot is the LABELS, that each row names both coordinates (`Pressed · RTL`, `Disabled · Default`, …) rather than several rows all reading "Default":

cross-product subtree

**The disclosures opened** — the theme bar expanded, state axis reached through the nav:

disclosures open

These 14 captures rebaseline, since they are shots that were not being taken.

## The one still failing

`contract · the fit cap re-measures when the history strip lands` is red here and on `main`, for an unrelated reason: that fixture builds `.cp-history` at runtime from the fetched manifest, and the strip never lands, so `waitForSelector('.cp-history')` times out. No drawer or nav state affects it. Left for its own change rather than bundled in.

## Test plan

- `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot` — 119 passed, 1 failed (from 105 / 15 on `main`) ✅
- Same suite narrowed to `serve-viewer-exploded`, `serve-viewer-inspect`, `serve-viewer-catalog-knobs`, `serve-viewer-axes-folded`, `serve-viewer-cross-product` — each 7 passed / 1 failed, the failure being the history contract test in every case ✅

---


yschimke:
Superseded by #3913 — this shouldn't land as-is.

The two casualties this repairs are real and the diagnosis is right (drawer closed since #3893, `#cp-axes-toggle` gone with the axes block). #3913 carries the same repairs — the same six `openControlsDrawer` call sites, the same three axes states — plus the thing neither PR started out looking for: `viewer-history.js` anchored to `.cp-viewer-bar` too, so the render-history strip has not drawn on any viewer page since #3893. That's the `contract · the fit cap re-measures when the history strip lands` failure this PR leaves out as "unrelated"; it isn't a fixture quirk, it's a shipped product bug, and fixing it is what takes the suite from 1 failed to **158 passed**.

Both branches edit the same regions of `pages-snapshot.spec.mjs`, so they can't both merge.

One thing here is better and is worth carrying over separately: the helper belongs in a shared `_viewer.mjs`, not inline. `serve-lanes.spec.mjs` already grew its own copy of the same logic in #3907, so there are now three homes for one idea — a small follow-up to collapse them is the right shape, rather than either PR rewriting a just-merged file.

---

PR #3913, opened 2026-08-15
fix(serve): draw the render-history strip again, on the right side of the preview, and take the harness green

yschimke:
`npm run harness:snapshot` — the `Compare vscode-preview captures` job — has been failing **15 of 158** on every branch since #3893. Reproduced locally at exactly 15 failed / 143 passed, and it turned out to be three separate things, one of them a shipped product bug.

## 1. The render-history strip stopped drawing — on every viewer page

`viewer-history.js` anchored its timeline to `.cp-viewer-bar` and returned early when that element was missing. #3893 folded that bar's controls into the title row and stopped emitting it, so the script has been bailing at its second line ever since.

| before (`main`) | after |
| --- | --- |
| viewer with no history strip | viewer with the history strip restored |

It anchors to `.cp-viewer` now — the stage's own container, which is what the strip describes, and the one element here that cannot be reorganised out from under it without the page having no viewer at all.

**Nothing else noticed.** The single baseline that held the strip (`serve-viewer-history.light`) rebaselined without it on the merge, and the only check that went red was the contract test waiting for `.cp-history` — which is now the guard that keeps this fixed.

### …and on a phone it goes *under* the stage

Restoring it above the render is right on a laptop, where it reads as metadata about the thing below it. On a phone it is the one thing the mobile layout was rebuilt to forbid: the page there is **bar → title → preview**, and a strip in between costs 113px of exactly that. Measured at 412px — `.cp-stage` starts at 126px on every other preview and at 239px on one with a history.

| before (strip above, at 412px) | after (strip under the render) |
| --- | --- |
| phone viewer with the history strip pushing the render down | phone viewer with the render first and the strip below it |

`place()` picks the side from the breakpoint and re-picks it when the viewport crosses one, so the strip joins the disclosure and renderer rows `viewer-drawers.js` already moves below the stage — and immediately after the stage rather than after those rows, since this arrives asynchronously from a fetch and anchoring to a sibling another script may not have moved yet would make the order depend on which won the race. Desktop is byte-identical: `.cp-history` at 194px, `.cp-stage` at 278px.

That phone position is its own committed harness state (`serve-viewer-history` · `mobile`), because the desktop baseline can only ever hold one of the two answers.

## 2. Six states and contracts reached into a drawer that is now closed

The knobs, Size, Locale, the inspection layers and the exploded camera all live in the overrides drawer, and #3893 stopped serving `cp-controls-open` — deliberately, per its "default Overrides to collapsed" — so it starts closed at every width and `.cp-viewer:not(.cp-controls-open) .cp-controls { display: none }` hides everything in it. Each of those tests found its target *resolvable, invisible and unclickable* — a sixty-second timeout apiece. A shared `openControlsDrawer` opens the column through its toggle first, the same fix #3907 made for the serve-lane helper.

## 3. Three states drove a control that no longer exists

`#cp-axes-toggle` went with the axes block, again deliberately — #3893 "moved variants into Components". What those states exist to capture moved rather than vanished, so they follow it there and keep their claims: `subtree-open` still shoots the row labels that name both coordinates (asserting `.cp-nav-current .cp-tree-variants` is visible, so a nav that opens empty still fails), `disclosures-open` still shoots the only expanded-toggle treatment on the page.

**This supersedes #3911**, which repairs the same two harness casualties on its own branch. Same six `openControlsDrawer` call sites and the same three axes states; overlapping edits to the same regions of `pages-snapshot.spec.mjs`, so they cannot both land. One thing there is worth keeping and is not here: #3911 puts the helper in a shared `_viewer.mjs` rather than inline, which is the better home now that `serve-lanes.spec.mjs` carries its own copy of the same logic. Worth a small follow-up to collapse all three onto one helper.

## Test plan

- `npm run harness:snapshot` — **158 passed**, from 15 failed / 143 passed. The suite also drops from 10 minutes to 2.4, with sixty seconds of timeout gone from fifteen tests.
- `npm run harness:compare` — 13 passed.
- `./gradlew :cli:check` — green.
- Verified before the fix that the strip is genuinely absent on `main` (probe: `.cp-history` never created, `history.json` never fetched) and present after.
- Phone placement measured directly at 412px, before and after, and desktop re-measured to confirm it did not move.

## The one job this does not fix

`VS Code Extension E2E (interactive-a-d)` is also red on main, and it is a different animal: test **B. edit a @Preview source, refresh, expect new id** is a real-Gradle, seven-minute scenario that fails intermittently — 5 failures and 2 successes in the last 7 runs on main, across unrelated commits — and with **two different errors** (`no setPreviews … ever named Interactive…` at 13:14, `refresh reported 'failed' — no showMessage was posted` at 10:07). That is a flaky compile/discovery path, not a stale selector, and it wants its own investigation with the job's uploaded logs rather than a guessed timeout bump. Happy to take it next.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1AL72Dofi6oJ9ENPXFX51)_


yschimke:
`VS Code Extension E2E (interactive-a-d)` is red again on `d242f19`, and it's the pre-existing flake — not this branch. Not fixing it here.

It is test **B. edit a @Preview source, refresh, expect new id** again, and this run gives it a **third distinct signature**:

```
Timed out after 180000ms waiting for: setPreviews after adding Interactive1786802862575,
  with its capture on disk
preview com.example.samplecmp.PreviewsKt.Interactive1786802862575 capture(s) not on disk:
  samples/cmp/build/compose-previews/renders/Interactive1786802862575-00018008.png
```

Alongside the two already recorded in the description — `no setPreviews … ever named Interactive…` (13:14) and `refresh reported 'failed' — no showMessage was posted` (10:07). Three different points of the same real-Gradle discover→compile→render chain failing across unrelated commits is a flaky pipeline, not a stale selector. A, C and D pass; the other four E2E shards are green.

Nothing on this branch can reach it: the diff is `viewer-history.js`, a comment in `viewer.js`, `pages-snapshot.spec.mjs` and committed PNGs. No Kotlin, and nothing under `vscode-extension/src` references the serve assets — `e2eInteractive.test.js` drives `composePreviewDiscover` / `composePreviewRenderAll` through real Gradle and never loads a viewer page.

Worth its own change, with the `e2e-renders-interactive-a-d` artifacts from the failing runs rather than a guessed timeout bump.

---

Issue #3922, opened 2026-08-15
Published catalogs render no motion captures, so the motion axis is always empty


yschimke:
`samples/design-catalog-m3` declares an `@InteractionPreview`, but no published catalog carries a single `motion[]` entry. The whole motion chain — export publishing (#3917), serve store/host/route (#3921) — is therefore unexercised against real data, and the viewer control still to come has nothing to display.

## Evidence

- `SwitchOn()` carries `@InteractionPreview` at `samples/design-catalog-m3/src/main/kotlin/com/example/designcatalogm3/CatalogSelection.kt:36`.
- The `design-artifacts` run triggered by #3917 succeeded ([run 31890209677](https://github.com/yschimke/compose-ai-tools/actions/runs/31890209677), job `compose-m3`).
- The published `catalog.json` on `design-artifacts/compose-m3` has **zero** components with `motion[]`; `motion/` on that branch is a 404.
- The 2000-line tail of that job's log contains **no** occurrence of `apng`, `interaction`, `motion`, `InteractionPreview` or `AnimatedPreview`. (Tail only — the render itself is earlier in the log and I have not read the whole thing.)

The export pass behaved correctly throughout: `motionArtifactsFor` records only artifacts it matched against `bundle.entries`, found none, and so published none. The gap is upstream of the export.

## Ruled out

**The released-runtimes gate is not the cause**, despite being the obvious suspect. The render job does set `ORG_GRADLE_PROJECT_composeaiUseReleasedRuntimes=true`, and `samples/design-catalog-m3/build.gradle.kts` documents that gate as deliberately preventing a published preview from using an unreleased runtime API. But the substitution block covers only `:data-preview-overrides-runtime` and `:slot-preview-runtime` — **not** `preview-annotations`, where `@InteractionPreview` lives. The annotation resolves against HEAD, which is also why the catalog still compiles under the gate. The plugin and CLI are both built from source in that job.

`RenderPreviewsTask` also threads `interaction` captures through to the renderer (`RenderPreviewsTask.kt:555`, `:849`, `:971`), so the plugin's render path is not obviously missing support.

## Candidates not yet ruled out

[...]

Issue #3837, opened 2026-08-14
test(serve): four serve page captures differ between two runs of identical code


yschimke:
## What

`preview-harness/pages-snapshot.spec.mjs` produces different PNGs for the same tree across repeated runs. Found while establishing a before/after baseline for the `<cp-bg-toggle>` port — the comparison showed 6 of 192 captures differing, and 4 of those turned out to differ between **two runs of identical code**.

Measured on `main` (`37c74520c`), Chromium 1194, 1024×720, `--enable-unsafe-swiftshader --use-gl=angle`:

| Capture | Differing pixels | bbox |
| --- | --- | --- |
| `serve-landing-tree-depth-component-open.dark` | 5820 / 1756160 (0.33%) | (51, 352, 244, 402) |
| `serve-landing-tree-depth-component-open.light` | 5727 | (51, 352, 244, 428) |
| `serve-playground-multifile.dark` | 23 | (180, 192, 372, 641) |
| `serve-playground-multifile.light` | 6 | (306, 192, 372, 194) |
| `serve-viewer-catalog-knobs-scroll-full-page.light` | 22 | (765, 345, 767, 679) |

`serve-design-page-index.{dark,light}` is a fifth, and the nastiest, because it is **bimodal rather than noisy**: two isolated runs of the *unmodified* tree produced hash `27891711…` then `a97163d7…`, and three consecutive runs after that produced `a97163d7…` every time. A capture that is stable within a session and flips between them will read as a real diff on a PR that touched nothing near it.

## Why it matters

The visual-diff bot comments on every PR, and these are the pages a serve change is most likely to touch. A reviewer who sees `serve-landing-tree-depth-component-open` move has no way to tell a 0.33% layout regression from this, so the honest response is to re-run and compare — which is exactly the "is the net telling me anything?" erosion that makes a visual gate stop being trusted.

It also blocks the cheap version of the check the [`serve-web` porting loop](../blob/main/cli/serve-web/README.md) wants: "port a component, confirm zero captures moved".

## Repro

```
cd vscode-extension
npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot
cp -r preview-harness/out /tmp/run1
rm -rf preview-harness/out
npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot
# diff /tmp/run1 against preview-harness/out — same tree, differing PNGs
```

[...]

PR #3928, opened 2026-08-15
feat(serve): offer a preview's recorded interaction as a selectable Motion lane

yschimke:
A component page could only ever show its component at rest. Whether the component's own interaction plumbing actually drives its transition — and what shape that transition has — is exactly what a still cannot answer. Previews declaring `@InteractionPreview` / `@AnimatedPreview` have been publishing an animated capture to `motion/` on the delivery branch (#3917) and the serve store, host and route have been carrying it (#3921), with nothing yet to play it. This is that last piece.

## What changes

A **Motion** chip sits beside the design-spec and Source chips, answering a fourth question on that row. The combo asks *which engine drew this?*, the spec chip *what was it specified as?*, Source *what do I type to get this?* — and this one, *what does it do?* Pressing it puts the recording on the stage; pressing it again returns to the render.

Never the default frame, and that is the whole design. Most readers open a component page to look at the component, so a page that starts animating at them is answering a question nobody asked. It is also load-bearing for the bytes: an APNG begins playing the moment its `src` is assigned, so `src` is written on the lane's **first entry** and never at page load, and closing the lane drops it again rather than leaving a hidden capture looping and decoding for the rest of the visit. A preview publishing several captures gets a picker; a single one gets its caption — never both, since a pressed button already carries those words.

Motion joins the existing `cp-mode` radio group rather than inventing state of its own, which is what buys `?mode=motion`, `?motion=<id>`, restore-on-load and Back/Forward for free, and it is closed by every other lane transition exactly as the spec and Source lanes are.

## Visual evidence

The at-rest delta is a single chip, so here is the control row rather than a full page that would bury it.

Before:

Viewer control row before, with no Motion chip

After:

Viewer control row after, with the Motion chip beside Live preview

…and the lane open — chip pressed, per-capture picker revealed, the badge reading `▶ Recording` rather than crediting the still renderer, and the capture itself on the stage in place of the still:

Motion lane open in the light theme

Motion lane open in the dark theme

More in [`renders/serve-motion-lane/README.md`](https://github.com/yschimke/compose-ai-tools/blob/agent/serve-motion-control/renders/serve-motion-lane/README.md).

## Teaching the rest of the viewer what a recording is

Review found six issues and they share one root cause: a capture is neither a render nor current, and four subsystems assumed otherwise.

- **A pinned page no longer offers it.** A published capture is a file on the branch like the baked render is, so by the pinned-page rule it should stay *and take the pin* — but `/motion/` reads the branch tip with no revision to resolve against, and "a pinned request is never answered with current bytes" is the stronger half of that rule. Withdrawn whole until the route can pin, rather than playing today's recording beside a render from another commit.
- **The backend badge** said `▪ Snapshot` over animated pixels. It now reads `▶ Recording` and stays `data-live="false"` — playing is not interactive.
- **Render overrides** stayed enabled, re-rendering the hidden snapshot underneath or flipping a static catalog into Wasm while the recording went on saying something else. Folded into the same fixed-frame gate the spec lane already uses.
- **The inspection layers** stayed drawn. Every box is placed from the snapshot's natural size and offset, and nothing re-places them when a lane takes that image out of flow. Withdrawn by mode, so returning restores what the visitor had ticked instead of silently clearing it.
- **A failed capture** left its error over the next one, since the overlay is shared across lanes.
- **The picked capture rides in the URL** as `?motion=<id>`, hydrated before the bookmarked lane is entered exactly as `specView` is — a shared multi-capture link used to restore the first recording, so the page someone sent was not the page they were looking at.

## Wired into the harness, not screenshotted by hand

Both open shots come from the preview-harness's own capture path, so this surface is diffed from here on:

- the main `serve-viewer` fixture gains published captures, so the chip is in that baseline;
- a new `serve-viewer-motion` fixture carries an interacted `motion-lane` state that opens the lane over a stubbed capture. Its **own** fixture, not a state of the main one — the harness runs states in order against a shared page and this one leaves a lane open, which would otherwise have re-shot `serve-viewer`'s later `connecting` baseline with a recording where the render belongs. The Source panel is isolated for the same reason.

The stub is a real 14-frame APNG rather than another flat placeholder: a still served as `image/apng` would satisfy every selector in the shot while proving none of it. It plays once and rests on its final frame, which a real capture does not — `page.screenshot({animations: "disabled"})` reaches CSS animations and not an image's own, so a looping stub would have flaked the baseline by the full width of the switch's travel. Verified byte-identical across two runs.

## Test plan

- `./gradlew :cli:test --tests '*ServeWebFixtureTest*' --tests '*ServeCatalogStoreTest*'` — green, including the fixture drift guard against the regenerated goldens.
- `cd cli/serve-web && npm run verify` — 151 passing, committed bundle back in sync.
- New viewer test: the chip opens un-pressed, the stage image is neither shown nor `src`-ed at page load, each capture keeps its own extension end to end (an APNG served as a GIF renders one frame and stops), the declared caption labels its capture, a preview with no captures renders none of it, **a pinned page renders none of it**, and leaving the lane drops the `src` instead of hiding a still-playing image.
- New badge tests: a recording is named, not credited to the snapshot backend, and does not light the live accent.
- `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot -g "serve-viewer(-motion)? · (light|dark)"` — 4 passed, reproducible byte-for-byte.

## Not in this PR

Published catalogs still carry **no** motion captures — every declared annotation is rendered and then lost between the render task and the export, which is #3922. So this lane is exercised against fixtures rather than live published data, and it will show nothing on preview.coo.ee until that is fixed. The captures do exist locally (`samples/design-catalog-m3/build/compose-previews/renders/SwitchOn_Light-*.apng` is a real 114-frame APNG), which narrows #3922 to the pack/export leg.

The visual-diff bot's `serve-landing-grouped-pages-filtered` diff is not from this change — see the comment below and #3929.


yschimke:
Reading the visual-diff bot's five changed captures, because one of them is not mine.

**Four are expected.** `serve-viewer` (both themes) gains the Motion chip — that is the change. `serve-viewer-connecting` (both themes) is a *state of that same fixture*, so it inherits the chip; the diff bbox is `(184,263)–(678,303)`, which is exactly the control row and nothing else.

**`serve-landing-grouped-pages-filtered` `dark` is a flaked capture, not a regression.** The diff is 1758 px in `(23,353)–(249,407)`: the sidebar filter loses its `.cp-search:focus` outline and its clear ✕ between baseline and this run. The typed value is still there — I sampled the glyph pixels and they are identical in both — so only the *focus state* moved.

Nothing in this PR reaches that surface. The evidence that it is environmental rather than mine: rendering that fixture from this branch locally reproduces the **baseline**, not this run's capture.

```
serve-landing-grouped-pages-filtered.dark, local render of this branch
  vs baseline (5d1a9ce): bbox (184,2104)–(333,2112)   ← footer version string only
  vs this run (fe4c881): bbox (23,353)–(333,2112)     ← includes the filter region
```

Three consecutive local runs were byte-identical (`44ca92c0…`), so it is not flaky in this environment either — it flipped in CI. The likely mechanism is the preceding `pages-pane` state ending in a `page.click()`: `fill()` then focuses programmatically, and whether the ring survives depends on the pointer-interaction state Chromium was left in. That is the same family as the pointer-parking flake in #3837.

I have deliberately not "fixed" it here — it is a different fixture, and papering over it inside a Motion PR would hide a real nondeterminism worth its own issue. Flagging it so the changed-capture list is not read as collateral from this change.

---

yschimke:
All six Codex findings verified against the code and fixed in `43d3796`. They turned out to share one root cause worth naming: I added a lane whose frame is **neither a render nor current**, and four subsystems around it assumed otherwise.

| finding | verified how | fix |
| --- | --- | --- |
| P1 pinned pages serve current bytes | `docs/public-preview-server.md` — a published capture is a branch file like the baked render, so it should stay *and take the pin*; the route cannot pin, and "never answered with current bytes" is the stronger rule | axis withdrawn whole on a pinned page, beside every other pinned withdrawal at `ServeWeb.kt:8153` |
| P2 badge credits the still renderer | `BackendBadge.label()` has cases for wasm/live/svg/spec and falls through to `▪ Snapshot` — visible in my own screenshot | `▶ Recording`, `data-live="false"` |
| P2 overrides stay enabled | `syncServerControls` gates on `onSpec = specActive()` only | `onFixedFrame = onSpec \|\| motionActive()` |
| P2 inspection layers stay drawn | `inspect.js place()` derives geometry from `img.clientWidth`/`offsetLeft`, and nothing re-places on lane entry — so boxes keep their old coordinates over the capture | withdrawn by `data-mode="motion"` CSS |
| P2 stale error across captures | shared overlay, raised by this lane's own `error` handler | `clearModeError()` on the way in |
| P2 picked capture absent from URL | `specView` is the existing precedent for a lane sub-state | `?motion=<id>`, hydrated before the bookmarked lane, replacing rather than pushing |

Two I want to be explicit about rather than let the table flatten.

**The P1 is the one that mattered.** My original comment said the pin was omitted because the route cannot honour it — which was true, and was the wrong conclusion to draw from it. "The route can't pin, so don't pin" quietly becomes "serve current bytes on a permalink", and that is the exact failure the rule exists to prevent. Withdrawing the lane is right until `/motion/` can resolve a capture at a revision, at which point this becomes `withPin` like the design reference.

**The inspect one is not motion-specific.** `openSpec()` and `openSource()` take the snapshot out of flow the same way and nothing re-places the layers there either, so the spec lane looks like it has the same latent bug. I have fixed it only for `data-mode="motion"` — widening it is a separate change against lanes this PR did not add, and I would rather not fold that in silently.

Regenerated goldens and re-ran everything; the badge change moves `lane-open-*.png`, so the evidence in the description is refreshed to the new head.

---

yschimke:
Second round verified and fixed in `8ebcc18`. Codex was right on all three, and the first one is a fair hit on my previous commit rather than on the original design.

**The fixed-frame gate was half-applied, which is worse than not applying it.** I taught `canServerRender` and the Wasm-honoured trio about Motion and left `themeChoice`, `.cp-rc-knob` and `.cp-knob` gating on a spec-only flag sitting right beside the new one. So the theme select and the Remote Compose knobs stayed live over a recording while the code — and my own reply here — read as though the lane were gated. Two predicates for one question is how that happens.

`onSpec` is gone. `onFixedFrame = specActive() || motionActive()` is the single predicate every override family consumes, because they all ask the same thing: *is the stage showing a frame this server did not just render?* Both lanes answer yes. There is now a test asserting the spec-only flag has not come back, since a future lane will hit exactly this.

Two things worth calling out in that fix:

- **Author knobs had no lane gate at all.** Their enabled state is whatever the server rendered, restored on each pass, so a daemon-backed catalog left them live on a fixed frame. That was already true of the **spec lane** before this PR. I have fixed it for both, because a control family dead on one fixed frame is dead on the other and special-casing Motion would institutionalise the inconsistency. Flagging it rather than burying it: this changes spec-lane behaviour. (Contrast the inspection layers, which I still scope to `data-mode="motion"` — that one is a question about overlay geometry, not a one-word predicate.)
- **Back/Forward within the lane now replays.** The pop handler treats a same-lane restore as an override push, which Motion has nothing to do with — but a restored entry *can* name a different capture, and only the picker was following it. Codex's repro is real and doesn't need hand-edited history.

**Repeated labels are numbered**, and only where a label actually repeats — two caption-less captures of the same kind both read "Interaction", while a lone one should not grow a "1" counting nothing.

`main` moved twice more underneath this (#3927, #3931 — the latter deletes `url-state.js` for a chrome bundle, which is exactly the code I touched). Merged, regenerated, re-ran: `:cli:test` green, `cli/serve-web` 167 passing, full page harness 123 passed including the URL-state and spec-lane contract tests. The evidence renders are byte-identical, so the description still shows the current pixels.

---

PR #3930, opened 2026-08-15
fix(serve): land section links on an anchor that exists, and keep Enter navigating

yschimke:
Two defects in #3927, both correctly found by Codex review — which arrived after auto-merge had already landed it, so this fixes forward.

## P1 — every section link was dangling

Section links were built as `#cp-node-<setId>`. But the page view emits anchors while iterating `components = page.nodes.filter(PageNode::isComponent)`, and `isComponent` **excludes containers** — deliberately: nothing implements a component set, and drawing a hotspot for one would put an enormous target over the real components inside it.

So no set ever had an anchor, and every section link resolved to nothing. Clicking one opened the sheet at the top, silently.

The link now lands on the set's **first descendant component**, which is where the section visibly starts on the sheet. Nodes are in the design file's own order, so a set's descendants are the following run of deeper nodes:

```kotlin
nodes.asSequence().drop(i + 1)
  .takeWhile { it.depth > node.depth }
  .firstOrNull { it.isComponent }
```

A set with no descendant component — a truncated or malformed manifest — links to the sheet plainly rather than carrying a fragment that lands nowhere.

### Why no test caught it, and what does now

Each golden was **self-consistent**: the landing emitted the fragments, the page emitted the ids, and nothing compared the two. My own check was just as weak — I confirmed `id="cp-node-…"` appeared six times in the page fixture, without checking whether the ids the *links* used were among them.

There is now a cross-golden guard that extracts every `href="…#cp-node-…"` from the landing and every `id="cp-node-…"` from the page view and asserts no fragment dangles. I verified it bites by pointing the fixture back at the set ids — it fails, naming the dangling anchor — before keeping the fix.

The fixture now carries one section with an anchor and one without, so both branches of the behaviour are in a golden.

## P2 — Enter folded the row instead of following it

A keyboard-synthesized click has no pointer position, so `offsetX` reads `0` — inside the twisty's 14px region — and the handler called `preventDefault()`. Pressing Enter on a page row expanded it instead of opening the page.

The fold is pointer-only now (`if (!e.detail) return;`). The label is a link to the whole sheet and keeps working from the keyboard; the sections under it are their own tab stops, so nothing is unreachable by leaving the fold to the mouse.

## Test plan

- `./gradlew :cli:test` — green.
- `npm run harness:snapshot` — 159 passed.
- Guard verified to fail on the pre-fix ids, not just pass on the post-fix ones.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1AL72Dofi6oJ9ENPXFX51)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeHttpServer.kt
@@ -5464,13 +5464,31 @@ class ServeHttpServer(
    *   sidebar into the thing it replaced; past it the page's own row still leads to the whole
    *   sheet, which is where every section is anyway.
    */
-  private fun designPageSections(page: DesignPage): List<ServeWeb.PageSection> =
-    page.nodes
+  private fun designPageSections(page: DesignPage): List<ServeWeb.PageSection> {
+    val nodes = page.nodes
+    return nodes
       .asSequence()
-      .filter { it.isContainer && it.name.isNotBlank() }
-      .map { ServeWeb.PageSection(it.nodeId, it.name) }
+      .withIndex()
+      .filter { (_, node) -> node.isContainer && node.name.isNotBlank() }
+      .map { (i, node) ->
+        // Where a link to this section lands. The page view anchors COMPONENTS, and a set is not
+        // one — so the target is the set's first descendant component, which is where the section
+        // visibly starts on the sheet. Nodes are in the design file's own order, so "descendant"
+        // is the following run of deeper nodes, and the run ends at the next node that is not
+        // deeper than the set.
+        val anchor =
+          nodes
+            .asSequence()
+            .drop(i + 1)
+            .takeWhile { it.depth > node.depth }
+            .firstOrNull { it.isComponent }
+            ?.nodeId
+            .orEmpty()
+        ServeWeb.PageSection(node.nodeId, node.name, anchor)
+      }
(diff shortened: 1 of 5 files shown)


yschimke:
`VS Code Extension E2E (interactive-a-d)` is red here and it is the pre-existing flake, not this PR. Not fixing it in a change about section anchors — but this run says something new about it, so recording it rather than just waving it through.

Test **B** failed the same way as the 14:11 run: the capture never reaches disk inside its 180s budget.

```
Timed out after 180000ms waiting for: setPreviews after adding Interactive1786812550101,
  with its capture on disk
```

The new detail is the timing. Test **A**, the same discover→compile→render chain, took **211s** on this runner against **164s** three hours ago — a 29% spread on identical work. B's wait is a fixed 180s, and its `composePreviewDiscover` started at 16:49:53 with the failure logged at 16:52:26, i.e. it was still going at ~153s and ran out.

That reframes the diagnosis. Across the runs so far this test has produced three different messages, which reads as a flaky pipeline; but two of the three are "the artifact isn't there yet", and A's variance says the budget is simply marginal on a slow runner rather than the discovery path being nondeterministic. Worth measuring B's actual completion time on a few green runs before touching anything — if it lands in the 150–180s band, this is a too-tight timeout and not a race, and the fix is a budget proportional to the observed chain rather than a guessed bump.

Nothing on this branch reaches it: the diff is `ServeWeb.kt`, `ServeHttpServer.kt`, one landing golden, and harness specs. No Kotlin in the preview pipeline, and nothing under `vscode-extension/src`.

---

PR #3940, opened 2026-08-15
refactor(serve): port the render-history timeline to a `<cp-history-menu>` element

yschimke:
`viewer-history.js` (273 lines, no tests) becomes a Lit element plus two pure
modules, on the pattern `<cp-viewer-drawers>` and `<cp-page-zoom>` already use:
the calculation moves out of the DOM so it can be a table of cases rather than a
screenshot nobody reads.

| new file | what it decides | tests |
| --- | --- | --- |
| `viewer/historyUrls.ts` | which links can be built at all, and safely | 22 |
| `viewer/historyModel.ts` | which of them are worth showing | 9 |
| `components/HistoryMenu.ts` | what is left: a fetch and a template | 9 |

`historyUrls.ts` is the half that was worth extracting first. Every link the menu
draws is built from DOM text (`data-history-repo`, `data-history-blob-url`) and
lands in an `href` — the flow CodeQL reports as `js/xss-through-dom`, and the
old file's own comments record three earlier attempts getting it wrong the same
way: validate a string, then hand *the same string* onward. The rule that fixes
it is "match, then rebuild from the captured segments", and the only way to know
it still holds is to try the inputs it exists to reject — `javascript:alert(1)/x`,
`//evil.com/{blob}.png`, `renders/../../etc/passwd`, a 40-char non-hex blob, a
ref where a sha is required. Those are now 22 assertions instead of a paragraph.

The port also deletes `place()`. The old script built the menu at runtime and
then went looking for somewhere to put it — into `.cp-head-toggles` before the
Overrides toggle, with a fallback above the stage for a page whose toggle row
predated it. The server knows where the control belongs, so it emits
`<cp-history-menu>` there and the placement question stops existing. That
fallback is not hypothetical: `renders/header-status-and-history` records it
failing once already, when the script anchored to `.cp-viewer-bar` and #3893
stopped emitting it.

### The bug the port surfaced

The element reads its inputs off `.cp-viewer`, which sits *below* the row that
declares it. Reading them at connect time worked only because the script tag
happened to sit near the end of `<body>` — correct by accident, which any
reordering breaks and which is not true at all of a document assembled in one
`innerHTML` write. `<cp-history-menu>` waits for the parse instead, so the
ordering assumption is gone.

## Visual evidence

A port's visual claim is that there is nothing to see, so the evidence is the
before and after being *the same file*. All four `serve-viewer-history` captures
are byte-identical between this branch and `origin/main` — closed and open,
light and dark — so each image below is simultaneously the before and the after.

The identity row, closed. `History 3 versions unstable ▾` sits between Theme and
Overrides, with nothing between the title and the render:

toggle row, light

toggle row, dark

Open — the dated versions, `current` marked the way the revision menu marks a
pin, and a note saying what the list covers:

menu open, light

menu open, dark

The full 226-capture `pages-snapshot` run was compared across the two refs as
well. Two `serve-landing-*-filtered.dark` captures came back different; re-run in
isolation on this branch they match `main` byte for byte, so that pair is a
focus-state race between the two parallel workers, not this change — nothing
here renders on a landing page at all.

## Test plan

```
cd cli/serve-web && npm run verify          # 206 passing, bundles in sync
./gradlew :cli:test --tests '*ServeWeb*'    # green; 27 page fixtures regenerated

cd vscode-extension
npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot
# 121 passed, including
#   contract · the render history is a menu in the toggle row, and the fit cap agrees
```

That contract test is the one that matters here: it strips the inline payload so
the *fetch* path runs, then asserts the menu still lands inside
`.cp-head-toggles`, is not a sibling of the stage, ships closed, and leaves the
fit cap agreeing with the geometry after the row has taken another control.

## Remaining legacy assets

`parity.js` (124), `spec-compare.js` (290), `rc-lanes.js` (337),
`catalog-live.js` (400), `inspect.js` (424), `design-page.js` (647). The big two
— `format-compare.js` (1,686) and `viewer.js` (2,906) — want breaking into pure
modules the way the drawers were, not porting whole.

---


yschimke:
**CI red on `VS Code Extension E2E (interactive-a-d)` — failing on `main` too, not from this change.**

The failing case is `B. edit a @Preview source, refresh, expect new id; revert, expect it gone`:

```
Timed out after 180000ms waiting for: setPreviews after adding Interactive1786816263511,
with its capture on disk
```

The same shard failed on `main` at `688801d`, in a run that started four minutes
before this one, with the other four shards green in both. Recent `main` history
for this workflow is 5 failures in 12 runs.

The signature is a slow runner rather than a behaviour change: the wait is for a
real Gradle render of a preview the test just wrote to disk, and the preceding
`composePreviewDiscover --rerun-tasks --no-build-cache` on the same job took
2m31s on its own.

Nothing in this diff is reachable from that path — it touches `cli/serve-web`,
the `serve` assets, `ServeWeb.kt`, and the committed page fixtures. The
extension's Gradle/render lane is untouched.

Leaving it; I'll re-run against the base branch once it recovers.

---

yschimke:
**`Build CLI` and `Module Unit Tests` were one root cause, and it is a live break on `main`.** Pushed a fix — flagging it because it is not this PR's work.

```
CliFlagsRegistryTest > every flagValue call site is classified in CliFlags() FAILED
CliFlagsRegistryTest > flags read by commands are present in a command allowlist() FAILED

Flags read via flagValue()/flagValuesAll() but missing from CliFlags:
  [--catalog-feed-cache, --catalog-feed-idle-timeout]
```

Both flags arrived in #3933 (`07903a7`, the demand-activated catalog feeds) on
`ServeCommand` without being added to `CliFlags.VALUE_FLAGS` or `serve`'s
allowlist. `origin/main` still has neither. It has gone unreported because the
`main` CI runs for `07903a7`, `2c1b401`, `49e69b3` and `8000ca2` were each
cancelled by the next merge before this job finished.

Not just bookkeeping, which is why I fixed it rather than waiting:

- `findCommandIndex` did not know either flag consumes the following token, so
  `compose-preview --catalog-feed-cache <dir> serve` reads `<dir>` as the command.
- `serve`'s allowlist did not carry them, so validation would reject a flag the
  `--help` text documents.

Fix is `1eff81d`, four lines across `CliFlags.kt` and `CliFlagValidation.kt`,
kept as its own commit so it reads separately from the port. `:cli:test
--tests '*CliFlagsRegistryTest*'` and `ktfmtCheck` pass locally. Happy to split
it into its own PR if you'd rather land it ahead of this one.

The earlier `interactive-a-d` failure is unrelated to both and still tracks the
base branch.

---

PR #3948, opened 2026-08-15
refactor(serve): port the Remote Compose player wall to `<cp-rc-lanes>`

yschimke:
`rc-lanes.js` (337 lines, no tests) becomes a Lit element plus four pure modules.

| new file | what it decides | tests |
| --- | --- | --- |
| `rc/pixelDiff.ts` | the YIQ metric itself | 15 |
| `rc/rowPlan.ts` | which of a row's lanes get a number, and which need measuring | 17 |
| `rc/rowFilter.ts` | the shared search box, and what the status line admits | 8 |
| `rc/model.ts` | lane lookup, and validating `?ref=` against the model's own lanes | ↑ |
| `components/RcLanes.ts` | the observer, the canvas, and `window.cpRcLanes` | 13 |

## The metric, finally pinned

`rc/pixelDiff.ts` is the reason this file was worth splitting. It is pixelmatch's
YIQ metric hand-transcribed into nine magic constants and a threshold scale, it
produces **the only number on the compare page the offline run did not compute**,
and it had no test of any kind — so a drift in it makes the page quietly disagree
with the build while looking exactly the same.

It now has 15, including:

- a sweep of the RGB cube confirming the transform still tops out at the **35,215**
  its `threshold` option is expressed against. Worth stating: the maximum is red
  against cyan, **not** black against white — white only maximises the luminance
  term (32,857), and the missing 2,358 is the chroma pair. My first version of
  this assertion was the wrong one, and the code was right;
- the exact bytes of a flagged pixel (`255, 60, 60, 255`) and of the washed-out
  backdrop (`230, 230, 230, 255` for black at 10% over white);
- that the cutoff is `>` and not `>=` — one pixel on a boundary, but a flip moves
  every number on the page.

## The rest

`rc/rowPlan.ts` turns `scoreRow`'s promise chain into a list of steps: each one
either already knows what its chip says, or names the two images to measure.
Reading the original told you the order things happened in, not which case a
given cell was in — and the cases are the behaviour, since most exist to say
"there is no number here, and here is why". The one worth calling out is that a
blank baked capture must **not** short-circuit the whole row: two *player*
renders still compare fine, and killing the row would throw away the only
comparison left on exactly the documents where the baked lane failed.

`connectedCallback` installs immediately when the model is already parsed rather
than always deferring to `DOMContentLoaded`, because `format-compare.js` calls
`window.cpRcLanes.filter()` on its first pass — deferring unconditionally would
publish the global after its only caller had looked for it. The server emits the
tag directly after the inline model to make that true, and the deferred retry is
what keeps it from being an ordering accident.

## Visual evidence

All four pre-existing `serve-rc-lanes` captures — default and `diff-baked`, light
and dark — are **byte-identical** between this branch and `origin/main`. The port
moves nothing.

Reference = the baked PNG, the cheap path, replaying the offline run's own
numbers:

diff against baked

### The path nothing was watching

Picking a **player** as the reference is the in-browser canvas path — the one
question the build cannot answer — and it had neither a test nor a capture. An
in-browser pass that threw, or produced no image at all, would have left every
shot in the suite unchanged. New `serve-rc-lanes` · `diff-player` state:

diff against a player

the diff images it produces

That holds the pipeline end to end — load → canvas → `getImageData` → diff → data
URL → an `` that renders — plus the status line that admits an in-browser
number is not the build's measurement. The numbers read `0.00%` because the
harness serves one placeholder for every `/rc-compare/` URL, so every lane really
is identical; the arithmetic is covered directly in `pixelDiff.test.ts` instead.

## Test plan

```
cd cli/serve-web && npm run verify                            # 297 passing, bundle in sync
./gradlew :cli:test --tests '*ServeWeb*' --tests '*RcCompare*'  # green
./gradlew ktfmtCheck                                          # green

cd vscode-extension
npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot
# 124 passed
```

`serve-rc-lanes.html` and `serve-format-compare.html` are the only fixtures that
changed (a tag swapped for a script tag, and the blank line the removed tag left
behind). One `ServeWebFixtureTest` assertion checked for the `rc-lanes.js` href
and now checks for `<cp-rc-lanes></cp-rc-lanes>`; its claim is unchanged.

## Remaining legacy assets

`spec-compare.js` (373), `catalog-live.js` (400), `inspect.js` (424),
`design-page.js` (789). The big two — `format-compare.js` (1,706) and `viewer.js`
(3,127) — want breaking into pure modules the way the drawers were, not porting
whole.

---


yschimke:
Both Codex findings were right. Fixed in `b51dc9e`.

**1. The reference frame was re-decoded per lane.** `load()` caches the
`HTMLImageElement`, but every `pixels()` call still allocates a full-size canvas,
redraws and does another `getImageData` readback — four redundant full-frame
readbacks per row on the five-player wall, in exactly the lazy scroll path the
`IntersectionObserver` exists to keep smooth. `rc-lanes.js` held `refData` for the
row and I dropped that. Restored, as a promise so the concurrent lanes share one
decode, with a test that counts the decodes.

**2. The `diff-player` state captured mid-flight.** Waiting for the first diff
image meant the baseline held whichever subset of rows won that run's race, and
would have re-diffed itself on every future PR — a flaky new baseline is worse
than no baseline. `<cp-rc-lanes>` now marks each row `data-scored="pending"` then
`"done"`, and the state waits for every started row to finish. Three consecutive
local runs now produce byte-identical captures. The marker earns its place
independently: nothing on the page distinguished "still measuring" from
"finished, and this is all there is".

---

The two red checks are both outside this diff, which touches only
`cli/serve-web`, the `serve` assets, `ServeWeb.kt` and the harness:

- **`CMP/Wasm Parity`** — 6 of 50 documents fail in the *player*
  (`dimension type 2 is not implemented`, `dynamic color id 60 is not declared`,
  one 5s timeout with no ink), all in `samples/design-catalog-remote-m3`. Its own
  comment says it does not block merging. `main` has been churning here — #3934,
  #3939, #3944, #3945 and #3938 are all in that area in the last few hours.
- **`VS Code Extension E2E (interactive-a-d)`** — the same slow-runner timeout on
  `B. edit a @Preview source, refresh, expect new id` that reproduces on `main`,
  reported on #3940 earlier today.

The `serve-landing-sections-filter-focus` row on the preview-diff comment is the
focus-state race between the harness's two parallel workers, not a change here —
same capture that moved on #3946, and re-running it in isolation matches `main`
byte for byte. It's a real flake in the suite that will keep producing spurious
rows until someone pins the focus state; I have not touched it in this PR.

---

yschimke:
Second round of Codex findings, both right again. Fixed in `fa80fe4`.

**3. The element never rewired after a reconnect.** `install()` guarded on the
parsed model, so a detached-and-reinserted tag returned early — while
`disconnectedCallback` had already removed the picker's listeners, disconnected
the observer and dropped `window.cpRcLanes`. Everything this element owns lives
*outside* it, so the teardown has to be undoable. The guard now tracks
installation separately from the model (which an in-flight pass still reads), and
the pass is bumped on disconnect so that work abandons itself rather than writing
into a table nobody is observing. Test covers detach → reinsert → measure.

**4. The completion marker covered the computation, not the decode.** Right:
`data-scored="done"` is set straight after assigning each diff `img.src`, and the
runner's generic `document.images` wait runs *before* any state is applied — so a
screenshot between the two catches empty diff slots. The same race one layer
down from the one I fixed last push. `diff-player` now waits for both, and three
consecutive runs stay byte-identical.

Worth saying plainly: the first version of this state waited on the wrong thing
twice, and both times the review caught it rather than a red run — which is the
argument for the `data-scored` marker existing at all.

---

Still red, still outside this diff:

- **`CMP/Wasm Parity`** — 6 of 50 documents fail inside the player
  (`dimension type 2 is not implemented`, `dynamic color id 60 is not declared`,
  one 5s no-ink timeout), all in `samples/design-catalog-remote-m3`. The check's
  own comment says it does not block merging.
- **`VS Code Extension E2E (interactive-a-d)`** — the slow-runner timeout that
  reproduces on `main`.

Neither is reachable from `cli/serve-web`, the `serve` assets, `ServeWeb.kt` or
the harness, which is all this PR touches.

---

yschimke:
Correcting my own claim, now that there is data for it.

I said the racy version of the `diff-player` state "would re-diff itself forever".
The three CI captures produced so far are byte-identical:

```
005db0c402d3ce6dfdde4d64d7084289  fc0980f  (004651635 — original, first-image wait)
005db0c402d3ce6dfdde4d64d7084289  9c0239d  (b51dc9e  — after the all-rows-done wait)
005db0c402d3ce6dfdde4d64d7084289  e7e816b  (fa80fe4  — after the image-decode wait)
```

So both races were real defects but **unrealised on this fixture** — five rows of
400×400 data URLs settle inside one frame on the CI runner, and the first-image
condition happened to be true only once everything was already there. The fixes
remove genuine latent flakiness (and the marker is worth having regardless), but
I should not have implied the baseline was observably unstable. It was not.

The one finding that was a real, present regression rather than a latent one is
the first: I dropped `rc-lanes.js`'s per-row `refData`, so every lane re-decoded
the reference frame — four redundant full-frame `getImageData` readbacks per row
on the five-player wall, in the lazy scroll path. That one was costing something
on every visit, not just in principle.

---

yschimke:
**5. The `popstate` subscription leaked on reconnect.** Right, and it is a leak
*I introduced* one commit earlier: making the element rewire on reconnect turned
a dead subscription into a stacking one. Fixed in `9b5996af`.

`onPop` now returns its own unsubscribe, and the element puts it in `cleanups`
with everything else. That is a change to `chrome/installUrlState.ts` rather than
a local workaround, because registering a raw `popstate` listener in the element
would duplicate knowledge of how `onPop` is implemented. It is safe to change the
signature: `serve-chrome.js` has been the sole owner of `window.cpUrlState` since
`url-state.js` was removed, and the other consumer (`backgroundChoice.ts`) lives
as long as its page and ignores the return.

### A note on the test, because I got it wrong first

My first version of this test drove three reconnects, fired `popstate`, and
asserted on the chips left behind. **It passed with the fix removed** — four
stacked callbacks all compute the same reference and leave the same end state, so
there was nothing to see. A test that cannot fail is worse than no test, so it
now counts the live `popstate` listeners at `window`:

```
one subscription on first connect          → 1
still one after three reconnects           → 1   (4 without the fix)
and none once it is gone for good          → 0
```

Verified failing at `4 !== 1` with the unsubscribe removed. Doing that check is
also what surfaced that `window.cpUrlState` was never stubbed in these element
tests at all, so the entire Back/Forward path was unreachable from them — the
stub is in now.

301 serve-web tests, `ServeWeb*` green. The two red checks are unchanged and
still outside this diff.

---

Issue #3936, opened 2026-08-15
Bring the CMP Android (embedded) player to parity with the view player, then make it the default


yschimke:
## Goal

Make `rcPlayer=cmp-android` the default lane in preview.coo.ee and the `compose-preview` CLI. The payoff is the data tier, not the pixels: the view player is `AndroidView { RemoteComposePlayer }`, so a whole Remote Compose document reaches Compose as one interop leaf and `compose/figma-svg` exports it as a single `<image>` — a screenshot wearing an `.svg` extension. The embedded player emits real Compose nodes, so the same document exports editable `<text>` and the card's own geometry, and the semantics tree describes the card instead of a black box. That is the residual left over from #2937.

The blocker is that the embedded player still draws some documents wrong. This issue tracks closing that gap and then flipping the defaults — in that order.

## Measured today

A/B over the whole `homeassistant-remotecompose` preview set (229 renders, same Robolectric harness, same commit, only the player varying). **171 of 229 renders differ.** All 164 `.rc` sidecars are byte-identical across lanes, so the document is a controlled variable and every difference below is the interpreter.

Text-only cards are already at parity — the clock card matches:

![clock card, view vs embedded](https://raw.githubusercontent.com/yschimke/homeassistant-remotecompose/0603a26/docs/previews/before-after/rc-player-clock.png)

### Defect 1 — the tinted badge circle behind an icon is dropped

Affects button, tile, glance, grid and horizontal-stack cards.

![button card, view vs embedded](https://raw.githubusercontent.com/yschimke/homeassistant-remotecompose/0603a26/docs/previews/before-after/rc-player-button.png)

The useful detail: it is **not** a blanket loss of icon backgrounds. In the horizontal stack below, the blue sensor badge survives and the amber light-on badge does not — same card type, same layout, adjacent tiles. So this is a specific paint path (a colour/alpha/fill case the embedded draw dispatch mishandles) rather than "circles don't draw".

![horizontal stack, view vs embedded](https://raw.githubusercontent.com/yschimke/homeassistant-remotecompose/0603a26/docs/previews/before-after/rc-player-horizontal-stack.png)

### Defect 2 — the grid card loses a tile

Four-tile grid renders three; the fourth ("RH") is absent entirely, not clipped to a sliver. Suggests a layout/measure divergence in the grid or flow path rather than a draw bug.

[...]

PR #3949, opened 2026-08-15
feat(serve): answer replay-dropped overrides from the document, not an allow-list

yschimke:
Second step of #3936, and the first consumer of `RcDocumentCapabilities` (#3947).

## Why

`CatalogLiveRouting.irReplayDroppedOverrideNames` decided which overrides an IR replay cannot honour from a hand-maintained list of axes. Support is a property of the **document**, not of the axis — of `remote-m3`'s 27 published documents 16 declare colour state a palette can move and 11 declare none, so no single answer for `themeProvider` is right even inside one catalog.

## What changes

It now takes the preview's `RcDocumentCapabilities`, read off the captured `.rc`:

| axis | before | now |
|---|---|---|
| `themeProvider` | always dropped | dropped only where the document declares no colour slots |
| `rc.<name>` | dropped when the *value* is a string | that, **plus** dropped when the document doesn't declare the name |

The second row is new precision the list could not reach. An `rc.` seed aimed at a name no document carries was reported as applied — a silent false success, which is the exact failure this predicate exists to catch.

Null capabilities (no document, or bytes that don't decode) keep the conservative answer the list gave for every preview before it could check, so nothing claims an override applies on evidence it doesn't have.

## Deliberately unchanged

- `localeTag` and `knob.*` seed into a composition and have no document-side counterpart for the capabilities to consult.
- A **string** `rc.` seed stays named even where the document declares it — that gap is in the alpha player, not the document.
- `fontScale` and `uiMode` stay absent. Deciding them needs the player's *execution* semantics rather than its declarations, which is why `RcDocumentCapabilities` deliberately does not answer `uiMode` (see #3947).

## Test plan

- `:cli:test` green in full; `IrReplayDroppedOverridesTest` at 12 tests, 4 new: `themeProvider` landing vs dropped on two documents that differ only in their declarations, the null-capabilities fallback, an `rc.` seed for an undeclared name, and a declared string seed still dropped.
- Fixtures go through `RcDocumentCodec.encode`, so they exercise the same read path the serve layer uses rather than a hand-built object.
- An existing test caught a real regression in the first cut: the result order is part of the contract (callers render it as a list), and regrouping the axes broke `localeTag, themeProvider, knob.*, rc.*`. Restored, and the ordering is now stated in the code.

Capabilities are memoised per preview id — the documents are small, but this sits on the render-response path.

Text-only change; no rendered surface is affected, so there is no visual evidence to embed.

## Not in this PR

Routing a `?rcPlayer=` request through replay so the player choice actually takes effect for a catalog whose previews recompose. That is the remaining half of goal 1 on #3936, and it is where the answer to "does this override survive?" gets used to decide 200-with-dropped vs refusal.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JYeKdEQJAhTrkmEx8CFQAw)_

### DIFF
--- a/cli/build.gradle.kts
+++ b/cli/build.gradle.kts
@@ -254,6 +254,10 @@ dependencies {
   // (payload schema only; the alpha `androidx.compose.remote.*` deps live in the connector, not
   // here), so it stays off the renderer/daemon boundary the CLI guards.
   implementation(project(":data-remotecompose-core"))
+  // Reads a captured `.rc` document's override surface (`RcDocumentCapabilities`) so the serve
+  // layer can answer "does this override survive a replay?" per document instead of from a
+  // hand-maintained allow-list of axes. Pulls `:rc-player-protocol` transitively for the codec.
+  implementation(project(":rc-player-runtime"))
   // Public render-session library — the CLI consumes its own published API for daemon-driven
   // commands (`compose-preview a11y` etc.) instead of touching DaemonClient directly. We eat
   // our own dog food: anything the CLI can do, a third-party tooling consumer can do via the
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/CatalogLiveRouting.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/CatalogLiveRouting.kt
@@ -3,6 +3,7 @@ package ee.schimke.composeai.cli.serve
 import ee.schimke.composeai.daemon.protocol.PreviewOverrides
 import ee.schimke.composeai.daemon.protocol.RemoteNamedValue
 import ee.schimke.composeai.daemon.protocol.UiMode
+import ee.schimke.composeai.rcplayer.runtime.RcDocumentCapabilities
 
 /**
  * Shared override-routing logic for the two trusted-catalog live hosts:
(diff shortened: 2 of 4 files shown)


yschimke:
## Closing this — the approach is wrong at this call site, not just under-tightened

All four findings confirmed, and together they show the change cannot be salvaged by tightening it.

The flow I had not read: `ServeThemeReplay.expand` runs **before** the overrides reach `irReplayDroppedOverrideNames`. `ServeHttpServer.kt:4974` says it outright — *"an expanded provider is no longer in them"*. On success it removes `themeProvider` and injects `rc.<role>` seeds for the **entire** provider palette, from a host-level mapping that takes no preview id.

That breaks both halves of this PR:

**The undeclared-name check (#3790191873).** A document declares only the roles it uses — 3 of ~30 is typical. Every other expansion-generated role would be reported dropped, and `droppedOverridesAreTerminal` turns that into a **409 on a request that themed correctly**. A straight regression.

**The `themeProvider` check (#3790191872).** It only fires when expansion did *not* consume the provider — no mapping, so no seeds were sent. In precisely that case the provider cannot apply, so consulting `supportsThemeProvider` turns a correct "dropped" into a false "applied".

So the per-document capability is not the right lever *here*. The document-vs-palette decision is already made upstream by the expansion, and where the expansion declines, the answer is unconditionally "dropped". Reading the document at this site adds nothing correct.

The two remaining findings stand on their own and I'm not disputing them — the cache keyed only by `previewId` is unsafe across hosts (#3790191870), and seed type should be checked against the declared type (#3790191871). Both would only matter in a version of this that distinguishes caller-supplied seeds from expansion-synthesised ones, which needs the expanded name set threaded through `Seeding` — a larger change than this, and one that should carry its own tests.

`RcDocumentCapabilities` still has a correct consumer: the **routing** decision on #3936 — which lane a `?rcPlayer=` request takes, and whether a replay can serve it — rather than after-the-fact reporting on a path where the decision was already taken. That is where it goes next.

Closing rather than reworking in place, so the branch doesn't sit here looking like it's a tightening away from correct.

---

PR #3950, opened 2026-08-15
refactor(serve): name the Remote Compose lanes after their players

yschimke:
`baked PNG` → **AndroidX Java**, and `RC · embedded player` → **AndroidX
Embedded**. Both now say *whose player drew the pixels* rather than how the
pixels got here.

Before (`main`):

lanes before

After:

lanes after

- **AndroidX Java** is AndroidX's `RemoteComposePlayer` — an Android `View`
  painting to a framework `Canvas` — rendered offline under Robolectric/Skiko.
  "Baked PNG" named the file format it arrives in, which is the one thing a
  reader comparing it against four other players does not care about: every lane
  on the page is a PNG.
- **AndroidX Embedded** is AndroidX's `RcPlayer`, a pure-Compose interpreter of
  the same document. This repo vendors and updates a copy as
  `:third-party-rc-embedded-player`, but that is a fact about this build rather
  than about whose player it is.

## What did not change

The lane **ids** are untouched — `baked` and `embedded` on the wire. They key the
staged assets on the delivery branch (`rc-compare/baked/0.png`) and the `?ref=`
parameter, so renaming them would strand every published catalog and break every
bookmarked comparison. Only what a reader sees moved.

## Both sides, or neither

`ServeRcCompare.LANES` (the `serve` page) and `render-rc-compare-html.mjs` (the
published static `rc-compare.html`) carry the same lane table and are documented
as mirrors. Renaming one and not the other would leave the two pages disagreeing
about what the same column is called, so both moved together, along with the
prose on each that named the lane inline — the lede, the status line, the
blank-reference note and the summary lines.

## One thing to decide separately

The remaining three lanes keep their `RC · …` style, so the row now mixes two
conventions: **AndroidX Java** · `RC · JS player` · **AndroidX Embedded** ·
`RC · cmp-jvm player` · `RC · cmp-wasm player`. That is what was asked for, and
renaming the other three is a call I have not made. Say the word and they can
follow.

## Test plan

```
./gradlew :cli:test --tests '*ServeWeb*' --tests '*RcCompare*'   # green
./gradlew ktfmtCheck                                            # green
node --test scripts/design-artifacts/render-rc-compare-html.test.mjs   # 26 pass
cd cli/serve-web && npm run verify                              # 301 passing
```

`serve-rc-lanes.html` is the only fixture that changed. The assertions that
pinned the old labels — `ServeWebFixtureTest`'s published-order list and the
`render-rc-compare-html` column/summary matchers — now pin the new ones; their
claims are unchanged.

---


yschimke:
Codex is right about the mechanism, and I checked how far it reaches. It is a
real hazard but not currently reachable, so I have written the assumption down
rather than reverted the label — the rename was asked for explicitly.

**The mechanism is real.** `RemoteEmbeddedPreviewWrapper` sets
`RemoteComposePlayerKind.EMBEDDED`, so a preview carrying it bakes its
`previews/<id>.png` through the embedded `RcPlayer` rather than the View-backed
one (`data/remotecompose/connector/…/RemoteOverridablePreview.kt:66`). And
`rc-compare.mjs` records no per-row renderer provenance — it takes whatever
`previews/<id>.png` the bundle carries — so it could not tell the difference.

**It does not reach a scored catalog today.** The only published system with an
`ir/*.rc` corpus is `remote-m3` (`scripts/design-artifacts/scope-systems.sh`
lists `compose-m3 wear-m3 remote-m3`; the rc-compare jobs run with
`--system remote-m3`). `samples/design-catalog-remote-m3` uses
`PreviewWrapperProvider` theme catalogs and **no** `RemoteEmbeddedPreviewWrapper`.
The two previews that do carry it —
`RemoteIndeterminateCircularProgressIndicatorEmbeddedPreview` and
`RemoteAnimatedCircularProgressIndicatorEmbeddedPreview` — live in
`samples/remotecompose`, which is not a published design system and gets no
rc-compare run.

So today every scored baked row really is the View-backed player, and the label
is accurate. What is missing is anything that would *keep* it accurate.

`551ee8e4` records that at both lane tables: what the label assumes, what breaks
it, why it does not bite yet, and that provenance has to be carried per row
before a catalog mixing the two is scored. That is the honest state — a comment,
not a guarantee. If you would rather have the guarantee, the smallest version is
`rc-compare.mjs` refusing to score a bundle whose manifest does not name the
baked renderer, which is a change to the driver and its summary schema rather
than to a label.

---

yschimke:
`CMP/Wasm Parity` is red, and cannot be from this diff.

The job runs `rc-compare.mjs` + `rc-compare-means.mjs`. This PR touches
**neither** — the only file it changes under `scripts/design-artifacts/` is
`render-rc-compare-html.mjs`, which emits the published static page and is not
in that job's path. The rest of the diff is `cli/serve-web`, `ServeWeb.kt`,
`ServeRcCompare.kt`, one fixture and the committed renders.

The failure is the same one on `main` and on #3948: 6 of 50 documents fail
inside the CMP player itself (`dimension type 2 is not implemented`, `dynamic
color id 60 is not declared`, one 5s timeout with no ink), all in
`samples/design-catalog-remote-m3`. The check's own comment says it does not
block merging.

For the record, the *renaming* of a lane cannot move a parity number either way:
the scores come from `mismatchPct` in `rc-compare-summary.json`, which is keyed
by lane **id** (`baked`, `embedded`, …) — and this PR deliberately left every id
alone.

---

PR #3956, opened 2026-08-15
refactor(serve): port inspect.js to &lt;cp-inspect-layers&gt;

yschimke:
Fifth of the legacy-asset ports. `assets/inspect.js` (424 lines, one IIFE) becomes a Lit light-DOM element plus two pure modules, and the asset is deleted.

## Visual evidence

All three layers on — accessibility, typography, theme attributes — from the `serve-viewer-inspect` `layers` state:

Inspection layers, light

Inspection layers, dark

There is no before/after pair because **all four `serve-viewer-inspect` captures are byte-identical** between `origin/main` (`fdd01dd`) and this branch, including the runtime `layers` state that is the only one holding the feature at all. That identity *is* the evidence: same filenames, same bytes, nothing added or dropped.

## Why this one was worth the tests

Almost nothing the accessibility pass decides is visible in a screenshot. A rectangle over a component looks equally right whether it is the correct node, a duplicate of its parent, or a finding that was silently dropped — so each rule in `inspect/entries.ts` now has a test named after the wrong picture it prevents. Three carried real risk, and all three are legible in the capture above:

- **`merged` is ABSENT on the wire when true** — it is the Kotlin default. `isFocusStop` tests `merged !== false`, not `!merged`. Read the other way, every unmerged inner `Text` draws a second rectangle on exactly its focusable ancestor's pixels: two boxes, two legend rows, one thing. The fixture has such nodes; the capture shows four entries, not seven.
- **A finding whose bounds match no node is still surfaced**, as its own entry titled from `viewDescription`. Dropping it — the obvious reading of "join findings onto nodes" — reports the frame as clean on exactly the elements the hierarchy could not describe.
- **A too-small touch target is a warning even when nothing else flagged the node.** Entry 4 in the capture (`Share profile`, `24×16 dp`) is reachable, labelled and named; it is just too small to hit. Left at `info`, which is what a plain "did any finding match?" gives it, it reads as a pass.

Plus the colour rule: a flagged stop takes its hue from its level, and only the *un*-flagged ones get `PALETTE[index % length]` — a pastel on entry 3 would compete with the `SpeakableTextPresentCheck` error it is reporting, and one colour for everything merges adjacent focus targets into a block.

`inspect/layers.ts` holds the addressing. Typography and Theme are two layers over **one** `annotations` payload, so ticking both must not fetch it twice — on an override-bearing frame that second fetch is a second daemon render, which can come back describing different pixels. The data URL is derived from the frame on screen (`data-cp-src`) by swapping only the format suffix, so the overlay describes the pixels the visitor is looking at, with no second copy of the viewer's query rules to keep in step; a `scroll=long` frame falls back to the viewport-sized product rather than 500ing. `?inspect=` round-trips through `replaceState`, normalised to declared order — ticking a layer is a reading aid over the same frame, not a different render.

## What changed

- **New:** `src/inspect/entries.ts`, `src/inspect/layers.ts`, `src/components/InspectLayers.ts`
- **New tests:** 52 across `inspectEntries` (23), `inspectLayers` (17), `inspectLayersElement` (12) — 401 passing in `cli/serve-web`
- **Deleted:** `assets/inspect.js`, and its `ServeWebAssets` entry and `<script>` tag
- `serve.css` hides the new tag alongside the other render-nothing elements; the harness drops `inspect.js` from `SERVE_ASSETS`

## Test plan

```
cd cli/serve-web && npm run verify           # 401 passing
./gradlew :cli:test --tests '*ServeWeb*' --tests '*ServeUrlState*'
./gradlew ktfmtCheck
cd vscode-extension/preview-harness
HARNESS_FIXTURE=serve-viewer-inspect npx playwright test pages-snapshot.spec.mjs
# 10 passed on both refs; 4/4 PNGs byte-identical
```

---


yschimke:
The visual-diff row above was a real bug, not noise. Fixed in 8b2733d.

**What it showed.** On the branch the inspection boxes were drawn at a smaller scale and a shifted origin — the typography box that should trace the title pill instead sat inset from its left edge and above its top. Everything about the picture still looked deliberate; it was just describing the wrong pixels.

**Why the PR body's "byte-identical" claim was true and still missed it.** It is a race, and my local runs won it on both refs. `place()` ran exactly twice: once at the end of `draw()`, and once on the frame's `load`. That second call only fires when the frame was still in flight as the script ran. `assets/inspect.js` was a `<script>` emitted after `viewer.js`; `<cp-inspect-layers>` upgrades with the components bundle, and under that ordering the frame has usually decoded already — so the only `place()` measures whatever the stage happened to be mid-settle. The legend appearing beside the stage narrows it, and a web font landing narrows it again. Nothing re-measures afterwards, so the boxes stay wrong.

Worth being explicit that this fragility is **not** something the port introduced — `inspect.js` has the same two-shot placement on `main`. The port changed only which side of the race it lands on, which is why a pure refactor moved a picture.

**The fix** is to stop depending on the event. A `ResizeObserver` on the frame makes placement a consequence of its geometry rather than of arriving at the right moment: every reflow that moves or resizes it re-places, the first layout included. The stage is observed too, because a stage growing around an already-capped frame changes `offsetLeft` without changing the image's own box.

The new test drives a reflow through a stubbed `ResizeObserver` and asserts the resulting scale and origin. I checked it fails against the old two-shot placement rather than trusting that it would — `1) re-places the boxes when the stage reflows under them`, on `place()` never re-running.

All four `serve-viewer-inspect` captures are byte-identical to `origin/main` again locally; the bot's next run is the one that matters, since it is the environment that caught this.

---

PR #3960, opened 2026-08-15
refactor(serve): port design-page.js to &lt;cp-design-page&gt;

yschimke:
Seventh of the legacy-asset ports, and the largest. `assets/design-page.js` (789 lines, one IIFE) becomes a Lit light-DOM element plus four pure modules, and the asset is deleted.

**All 24 `serve-design-page` captures are byte-identical** to `origin/main` (`0713524`) — twelve states in both themes, nothing added or dropped.

## Visual evidence

The diff lane is the shot that matters, because it is the state that proves the port did not quietly break the thing hardest to notice (below):

The diff lane, scoring each node

A render the server could not produce, with the design's own drawing restored under it:

A failed render

The coverage filter:

Only what we don't implement

## The hazard the port had to design around

`design-page.js` read `window.ComposePreviewCompare` at IIFE time, and its `<script>` came *after* `format-compare.js`, which publishes that global. But the components bundle is emitted **before** `format-compare.js` on this page — so an element that cached the handle when it upgraded would cache `null`, and the diff lane would score nothing at all. Every badge stuck on a dash, no error anywhere, and a page that still looks like it is working.

So `<cp-design-page>` reads the handle when the lane is *entered* rather than when it upgrades. One property lookup per entry into the lane, and no script order can break it. The diff-lane capture above is the proof it resolves: `15.3% ⇲` in the red band, not a dash.

Worth remembering for `viewer.js`, which reads the same global the same way.

## What came out, and what each rule prevents

**`design/ink.ts` — fitting our drawn pixels onto the design's drawn box.** The two halves of the swap are boxes of different kinds: a design node's box is the tight bounds of the shape Figma drew, a render is a fixed canvas with the component inside it and transparent margin around. `object-fit: contain` fits CANVAS to INK, so the margin is spent inside the design's slot and our component comes out smaller — 4% (a circle, which nearly fills its canvas) to 42% (a semicircle, square canvas, 1.6:1 slot) on one catalog page. That reads as "everything scales when you flip the lane", which is precisely the reading the two lanes exist to make impossible. So ink is fitted to ink and the margin hangs *outside* the slot, which is why a placement can be 200% wide at a negative offset — and why that case has a test. Uniform and centred, never stretched: the aspect our render actually has is a finding about our code.

**`design/score.ts` — what a badge says.** Both ways of conflating the two numbers have already been shipped here once:

- `scoreImages` answers with a MATCH percentage — identical images score 100 — and this lane reports DRIFT. Getting the inversion backwards prints "100.0%" in red for a perfect match and green for a total mismatch: a readout that lies rather than one that is merely wrong.
- Taking `max(drift, geometry)` as the headline made every badge on this fixture read 52.4%, the aspect difference wearing the label of a pixel difference. The number is the drift alone; the geometry marks the badge (`⇲`), is spelled out in the tooltip, and counts for the BAND — so a component rendered at the wrong shape still triages red however well its pixels line up once normalised.

**`design/geometry.ts` — measured placement.** Slots are percentages of the ZOOM LAYER, not the stage: a zoomed canvas is larger than its stage and offset within it, so measuring against the stage would divide a zoomed distance by an unzoomed span. Against the canvas both share one transform and the ratio is identical at any zoom, which is why nothing is recomputed when a zoom is applied — there is a test asserting exactly that equality. Node ids are matched under both of Figma's spellings by comparing attribute values rather than building a selector out of one; a node id is text from a design file, and interpolating it into a selector has the same shape as an HTML injection.

**`design/lanes.ts` — the three lanes and two filters**, including the rule that switching the coverage filter on turns the resting marks on (a filter with nothing to draw on is a no-op the reader cannot see) while switching it off leaves them on (it was an explicit state to arrive at, and silently repainting the sheet plain would read as the filter having broken something).

## What no capture can show

A node the manifest names but the export does not carry says so (`[data-cp-missing]`) instead of vanishing; the design's own drawing is hidden only once ours has actually *arrived* and comes back if it never does; and the coverage filter takes what it mutes out of the tab order and the accessibility tree, not just out of sight — CSS alone cannot, so a keyboard user would otherwise tab onto an invisible rectangle with no focus ring. All three have tests.

## What changed

- **New:** `src/design/{ink,score,geometry,lanes}.ts`, `src/components/DesignPage.ts`
- **New tests:** 53 across `designInk` (18), `designScore` (7), `designGeometry` (15), `designPageElement` (13) — 508 passing
- **Deleted:** `assets/design-page.js` and its `ServeWebAssets` entry
- `serve.css` hides the new tag; the harness drops `design-page.js` from `SERVE_ASSETS`

## Test plan

```
cd cli/serve-web && npm run verify           # 508 passing
./gradlew :cli:test --tests '*ServeWeb*'
./gradlew ktfmtCheck
cd vscode-extension/preview-harness
HARNESS_FIXTURE=serve-design-page npx playwright test pages-snapshot.spec.mjs
# 10 passed on both refs; 24/24 PNGs byte-identical
```

---


yschimke:
Two red checks, both pre-existing on `main`. Neither is fixable from this branch, so I'm not pushing anything for them — details so you can judge that rather than take my word for it.

**`Apply compose-preview` — red on `main`, not caused here.** The compose-preview workflow has failed on every non-cancelled `main` run since `chore(deps): update Compose BOM to 2026.08.00` (#3955): the BOM bump itself (22:26), the 1.8.0 release commit (22:44), and #3959 (23:18). The last green run was #3954 at 21:41, immediately before the bump.

That also explains the preview-diff row above. `SharedElementFilmstripPreview` (samples:android) is a shared-element *animation*, and a Compose BOM upgrade moving one is expected. This PR's diff touches only `cli/serve-web/`, `cli/src/…/serve/`, `renders/`, and the preview harness — nothing under `samples/`, `renderers/`, or `gradle-plugin/`, so it cannot reach an Android Compose render at all.

**`VS Code Extension E2E (interactive-a-d)` — flaky on `main`, timing not correctness.** Scenario B failed with `Timed out after 180000ms waiting for: setPreviews after adding Interactive…, with its capture on disk`. That scenario edits a `@Preview` source and waits for a real Gradle compile *and* render to land; in the same run scenario A took 193s on its own and the suite took 7 minutes. The other three scenarios passed.

The same workflow fails intermittently on `main` on commits that cannot have caused it — `docs: bind npm trusted publishing…` (23:06) and the 1.8.0 release commit (22:44) both failed, while #3959 (23:18) and #3954 (21:41) passed. It's a loaded-runner timeout.

This PR's own gates are green: `Serve Web Components`, `ktfmt`, and the JS suite at 508 passing, plus 24/24 `serve-design-page` captures byte-identical to `origin/main` locally on both refs.

---

PR #3962, opened 2026-08-16
fix(rc-player): resolve ColorTheme indices and modes in every player

yschimke:
A `ColorTheme` operation records, per mode, an `android.R.color` **resource index** and a literal **fallback**. #3952 taught the vendored embedded player to read the index. This carries that to the CMP player and the jvm renderer, and fixes three ways the reading was still wrong.

None of them is visible in isolation, which is why each needed measuring rather than reading: a document whose index does not resolve renders as its fallback, and the fallback is exactly what a host with no palette would have drawn.

## The index table is transcribed, not derived

`AndroidColorThemeResolver` built its index → name map by reflecting over the creation-side `Rc.AndroidColors` constants. Those are wrong for **21 of the 196 indices** at `1.0.0-alpha17`:

- `SYSTEM_ACCENT2_200` is `30`, colliding with `SYSTEM_ACCENT2_1000` — so index `31` has no constant at all, and index `30` resolves to whichever field reflection yields: a real resource, and the wrong colour.
- Twenty names are misspelled against the resource they select — `SYSTEM_ERROR_620` at index `62` where the resource is `system_error_10`, and the entire `system_neutral1_*` run at 78–90 spelled `SYSTEM_NEUTRAL78_0`, `SYSTEM_NEUTRAL79_790`, … Those match no resource, so the colour quietly keeps its fallback.

The table now comes from `ThemeSupport.AndroidColors` in `remote-player-view`, which is what actually resolves a document's indices on a device.

## `SYSTEM` / `UNSPECIFIED` are resolved to a real mode

`theme` defaulted to `Theme.UNSPECIFIED` and was assigned straight to `paintTheme`. `ColorTheme.apply` selects light only for `Theme.LIGHT` and dark for everything else — so the default rendered every themed document **dark**, immediately after resolving the light palette correctly. Same document, same host:

| lane | pixel | resource |
| --- | --- | --- |
| embedded, before #3952 | `#E6E0E9` | captured fallback — index never read |
| embedded, `main` today | `#292A2D` | `system_surface_container_high_dark` |
| view player | `#E9E7EC` | `system_surface_container_high_light` |
| **embedded, this PR** | **`#E9E7EC`** | `system_surface_container_high_light` |

`Theme.SYSTEM` is now the default and is answered from `isSystemInDarkTheme()`, so every player resolves it identically. The view player's own rule (an SDK-33-guarded `Configuration.isNightModeActive`, with `UNSPECIFIED` falling through to dark) is where the divergence was found, not the specification it was fixed against.

## The requested mode reaches the cmp-jvm lane

`renderRemoteDocumentToPng` gained a theme while `renderRemoteDocumentToSvg` did not, so the SVG path silently took the new light default with no way to override — and neither `RcJvmRenderSpec` nor the worker frame carried a mode at all, so a `?uiMode=dark` request rendered the light branch. It is threaded end to end now: `uiMode` in serve, through `RcJvmServerRenderer` and the worker pool, into the request frame, out to both render entry points, plus a `--theme` flag on the one-shot path.

Carrying it changes the frame, so `PROTOCOL_VERSION` moves 1 → 2. That is what makes a stale `lib-rcjvm/` sidecar fall back to the one-shot path rather than read `theme` as `seedsLen` and desynchronise for good.

## The other players

- **CMP player** (`RcPlayerState` / `RcComposePlayer`, desktop + iOS + wasm) resolved nothing at all — `applyColorTheme` read only the fallbacks. It now resolves through a host-supplied `systemColorLookup`, defaulting to resolving nothing, which is the honest answer where there is no system palette: the document keeps the fallbacks it captured for exactly that case. `rcResolveSystemTheme` answers `SYSTEM`/`UNSPECIFIED` the same way. The callback is read through `rememberUpdatedState` rather than keyed into the state's `remember`, so a host's capturing lambda cannot tear down player state on an unrelated recomposition.
- **jvm renderer** had no theme handling whatsoever. It now shares the neutral half of the resolution and defaults to `Theme.LIGHT` rather than following the build machine's desktop theme, so a headless render stays reproducible.

Resolution requires the document's colour group to resolve **by name** to `android`. `colorGroupId = 0` names no table at all rather than naming Android's, and `GenerateBaselineFixture` writes exactly such an operation (group 0, index 0) — treating it as Android would repaint it `background_dark` in the CMP player while the embedded player kept its fallbacks.

The table exists twice — the vendored module is an Android library in an AOSP package, `:rc-player-protocol` is KMP with no Android target, and neither can depend on the other. `AndroidColorTableDriftTest` compares the two, and the sibling file is declared as a task input so the guard cannot silently go stale (it was UP-TO-DATE and not guarding until that was added).

## Visual evidence

Not embeddable for this change, and the reason is itself a finding. The only fixture that exercises `ColorTheme` (`SystemThemeSwatchesRemote`) lays out as a **1-pixel sliver in both players** while its baked Compose PNG shows three full swatches — so the pixel comparison above is a single pixel, and a screenshot of it would show nothing. That layout defect is independent of this change (it predates it and affects the view lane identically) and is worth its own fix; until then the numbers above, taken from the rendered PNGs of all three lanes, are the evidence. The CMP/Wasm parity bot reports 44 unchanged and 0 regressions, which is the independent check that nothing else moved.

## Test plan

- [x] `RcColorThemeResolutionTest` (CMP): per-mode index resolution, fallback on an unresolvable name, unknown *and* absent colour group left alone, and the table's contract pinned — including the four entries reflection gets wrong, so a future "simplify this" reintroduces the bug loudly.
- [x] `AndroidColorTableDriftTest`: both tables agree; verified to actually fail when one is perturbed.
- [x] `RcJvmWorkerPoolTest`: a dark request served by a warm worker that serves a light one after it. The stub worker reads and echoes the theme, so a dropped field fails rather than passes.
- [x] `RcJvmRenderWorkerProtocolTest`: the hand-built frame carries the new field.
- [x] Render lanes on a real `ColorTheme` document: embedded and view now agree on `#E9E7EC`.
- [x] Suites green: `:rc-player-protocol`, `:rc-player-runtime`, `:rc-player-compose`, `:third-party-rc-embedded-player`, `:third-party-rc-embedded-player-jvm`, `:cli`.

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/RcJvmServerRenderer.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/RcJvmServerRenderer.kt
@@ -81,6 +81,13 @@ internal object RcJvmServerRenderer {
     spec: RcJvmRenderSpec,
     seeds: Map<String, RemoteNamedValue> = emptyMap(),
     format: Format = Format.PNG,
+    /**
+     * Which branch a `ColorTheme` operation selects. Defaults to light rather than to the machine's
+     * desktop setting: this renderer is headless, so "the host's theme" is not a real question, and
+     * a render that changed colour with the build machine's OS would not be reproducible. Documents
+     * with no `ColorTheme` are unaffected either way.
+     */
+    theme: RenderTheme = RenderTheme.LIGHT,
   ): RenderResult {
     val cp = classpath()
     if (cp.isEmpty()) return RenderResult.Unavailable(unavailableReason())
@@ -99,7 +106,14 @@ internal object RcJvmServerRenderer {
     // cold would double the cost of every document that cannot be drawn.
     pool(cp)?.let { pool ->
       val pooled =
-        pool.render(docBytes, spec, seedLines(seeds).joinToString("\n"), format, secondsLeft())
+        pool.render(
+          docBytes,
+          spec,
+          seedLines(seeds).joinToString("\n"),
+          format,
+          theme,
+          secondsLeft(),
+        )
       when (pooled) {
         is RcJvmWorkerPool.PoolResult.Ok -> return RenderResult.Ok(pooled.bytes)
         is RcJvmWorkerPool.PoolResult.Failed -> return RenderResult.Failed(pooled.reason)
@@ -118,7 +132,7 @@ internal object RcJvmServerRenderer {
       }
     }
 
-    return renderOneShot(cp, docBytes, spec, seeds, format, secondsLeft())
+    return renderOneShot(cp, docBytes, spec, seeds, format, theme, secondsLeft())
   }
 
   /**
@@ -133,6 +147,7 @@ internal object RcJvmServerRenderer {
     spec: RcJvmRenderSpec,
     seeds: Map<String, RemoteNamedValue>,
     format: Format,
+    theme: RenderTheme,
     timeoutSeconds: Long = RENDER_TIMEOUT_SECONDS,
   ): RenderResult {
     val input = File.createTempFile("rcjvm-in-", ".rc")
@@ -160,6 +175,8 @@ internal object RcJvmServerRenderer {
         add(spec.density.toString())
         add("--format")
         add(format.wire)
+        add("--theme")
+        add(theme.wire)
         if (seedsFile != null) {
           add("--seeds")
           add(seedsFile.absolutePath)
@@ -339,6 +356,20 @@ internal object RcJvmServerRenderer {
   enum class Format(val wire: String) {
     PNG("png"),
     SVG("svg"),
+  }
+
+  /**
+   * The `ColorTheme` branch a cmp-jvm render selects.
+   *
+   * Deliberately this module's own type rather than `remote-core`'s `Theme` int: the CLI drives the
+   * player as a subprocess and carries no compile dependency on it, which is what lets the worker's
+   * classpath be staged independently. [wire] and [frame] are the two forms that cross the boundary
+   * — a `--theme` argument on the one-shot path, and an int in the pooled worker's request frame,
+   * whose values `RcJvmRenderWorkerMain` mirrors.
+   */
+  enum class RenderTheme(val wire: String, val frame: Int) {
+    LIGHT("light", 0),
+    DARK("dark", 1),
   }
 }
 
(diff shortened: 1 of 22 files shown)


yschimke:
Rebased onto `main` and pushed the review fixes. All three Codex findings were real; each is addressed and pinned by a test, and the PR body is updated.

One consequence worth calling out separately: carrying the theme changes the cmp-jvm worker frame, so `PROTOCOL_VERSION` moves **1 → 2**. Without that, a `lib-rcjvm/` sidecar staged from an older build would read `theme` where `seedsLen` used to be and desynchronise permanently instead of falling back to the one-shot path.

On CI — the two red checks reproduce on the base branch and predate this PR:

- `compose-preview.yml`: failing on `79f7655`, `cd5f589`, `87aa58b`, `6dac532`, `24fc87d3`
- `vscode-extension-e2e.yml`: failing on `79f7655`, `cd5f589`, `0f74e71`, `6dac532`

Not touching those here. The CMP/Wasm parity bot came back clean (44 unchanged, 0 regressions), which is the check that would have caught a rendering regression from the player changes.

---

yschimke:
Correction to this PR's description — the "visual evidence" section was wrong, and the merged body should not be trusted on that point.

I claimed the `ColorTheme` fixture "lays out as a 1-pixel sliver in both players" and that no visual evidence was therefore possible. **Both claims were false.** They came from a hand-rolled PNG reader I used to sample the renders: it skipped each scanline's filter byte but never reversed the filter, so almost every pixel decoded as noise and only a couple survived as plausible values. The renders were fine all along; my measurement was not.

Re-measured with a real decoder (`ImageIO`), all three lanes render three full 88×88 swatches — 23232 opaque pixels each, exactly 3 × 88²:

| lane | opaque px | swatch colours |
| --- | --- | --- |
| embedded, before | 23232 | `#65558F` `#1D1B20` `#E6E0E9` — the captured fallbacks |
| embedded, after | 23232 | `#1B1B1F` `#445E91` `#E9E7EC` |
| view player | 23232 | `#1B1B1F` `#445E91` `#E9E7EC` |

And the parity result is much stronger than the single pixel I reported. Pixel-for-pixel over the full 400×400 frame:

- **embedded (after) vs view player: 0 differing pixels of 160000.** Byte-identical.
- embedded (before) vs view player: 23232 differing pixels, max channel delta 33.

So the change does exactly what the PR claimed, with better evidence than the PR offered — the conclusion stands, the measurement behind it did not. The fixture also needs no layout fix; there is no defect there to file.

One real gap the corrected measurement exposes: the fixture is **not on `main`**. It lives only on the unmerged `agent/rc-colortheme-sample`, so nothing on `main` exercises `ColorTheme` in a render and the visual-diff bot cannot regress-test this path. Landing it is worth doing on its own, with the before/after above embedded as images.

---

PR #4002, opened 2026-08-16
fix(ci): resolve code scanning alerts

yschimke:
## Summary

Resolve the 106 open code-scanning alerts exported from GitHub, covering CodeQL and Zizmor findings across workflows, composite actions, and the VS Code extension security regression test.

## What changed

- pin privileged driver checkouts and third-party actions to immutable commits
- make Gradle caches read-only while executing PR or third-party code
- pass reusable-workflow inputs through environment variables instead of expanding them into shell syntax
- replace `secrets: inherit` with explicit secret mappings
- disable automatic package-manager caching in release jobs
- disable persisted checkout credentials where artifacts or agent code can carry the checkout forward
- narrow GitHub Pages deployment permissions to the deployment job
- encode generated review JSON with a JSON encoder instead of interpolating untrusted output
- make the hostile HTML-tag regression assertion case-insensitive
- document narrow Zizmor exceptions inline for coordinated same-repository release tags, locally derived `$GITHUB_ENV` values, the guarded PR-run reaper trigger, and required reusable-workflow permission ceilings

## Root cause

Several workflows treated reusable-workflow inputs as trusted configuration even though string inputs can alter shell syntax when expanded directly. Other alerts came from writable caches around checked-out external code, movable action refs in privileged workflows, broad secret inheritance, and scanner heuristics that could not infer local provenance or guarded trigger behavior.

## Validation

- exported alert scope re-scanned with Zizmor 1.15.2: none of the original rule/file pairs reproduce
- YAML parsed successfully for all workflows and composite actions
- `git diff --check`
- VS Code extension Prettier check
- VS Code extension test suite: 1,692 passing

The local `code-scanning-alerts.json` export was intentionally excluded from the commit.


yschimke:
Fix merge conflicts

PR #4069, opened 2026-08-16
feat(serve): report a bug in the preview server itself

yschimke:
## Summary

The viewer already has a **report an issue** link — but it files a bug about a *preview*, against the repo whose Kotlin declares it. A bug in the **server** — a knob that does nothing, a render that 500s, a page that draws wrong, a catalog that will not load — has no such owner, and routing it into a catalog's tracker sends it to people who cannot fix it.

This adds a second, server-wide affordance: **report a bug**, in the site footer of *every* browser-facing page (front door, `/status`, a 404, a catalog landing, a viewer), sitting beside the build number it is a bug in. It always targets `yschimke/compose-ai-tools`.

### The footer entry

| Before | After |
| --- | --- |
| Footer with source and /version only | The same footer with "report a bug" added |

That is the only change to the 60-odd existing page fixtures — one entry in the footer's links row, on every page.

### The `/report-bug` page

A page rather than a jump straight to GitHub. The per-preview report can be one click because its whole body is already on screen (a preview id, the overrides in the URL bar, a render). A server report is not: it carries the JVM the daemon runs on, which catalogs failed to load, and what the render lanes have been doing — none of which appears on any page. Posting that to a public tracker, in the reporter's name, without them having read it, is not something a footer button should do. So the page **shows the report before it is filed**, grouped, with the exact markdown underneath, and the button files what is on screen.

Captured from a viewer on a box that is *not* entirely healthy (a catalog that failed to load, a render that timed out) — a report filed from a perfectly healthy server is the one nobody sends.

| Light | Dark |
| --- | --- |
| The bug-report page, light | The bug-report page, dark |

## What the report carries

- **Server** — running build, public vs token-gated, uptime, and the Java + OS the renders actually happen on.
- **Page** — the path the visitor came from and, when that path names one, the design system, the preview, the catalog's `repo@branch`, the compose-ai-tools version that produced it, its trust verdict, its render lane, and any degradation in force. A page that belongs to no catalog (the front door) drops the section entirely rather than filling it with "unknown".
- **Catalogs not loaded** / **Recent failures** — the same load errors and daemon/render failures `/status` reports, capped at eight so the body stays readable. Fenced, with fence markers in the text neutralised, so a stack frame cannot shear the body.
- **Browser** — user agent, viewport, device pixel ratio, colour scheme. Filled in by the visitor's own browser via a `{{client}}` placeholder in the form's hidden `body`; the four facts the server cannot observe and the ones a "draws wrong" report turns on. JS off drops the section rather than shipping a literal `{{client}}`.

**Screenshots are two-sided.** From a viewer, that preview's render is embedded in the body (or linked — same *reachable* + *answerable* rules the per-preview report uses) and shown on the page. That covers "the render is wrong". It does not cover "the page is wrong", which is most server bugs, so the report leaves a Screenshot section for a pasted shot of the whole window.

## Safety

- `/report-bug` is gated exactly like `/status` — it reports the same catalog-load and daemon-failure detail.
- The `from` path is **browser-supplied** and reaches HTML, a link, and a public issue body. Accepted only as a same-origin absolute path: no scheme, no protocol-relative `//`, no fragment, no control characters, length-capped. Anything else costs the report a row and nothing else.
- The preview `from` names is resolved by **matching an existing preview id** (re-encoded), never trusted as one — and never percent-decoded here, which would repeat the `%2B`→space bug the usage route documents.
- `og:url` uses the bare route, not `externalPageUrl()`. Caught by the route test: the page URL carries the raw `from` query, so an `og:url` built from it echoed an unvalidated, possibly off-origin URL straight back into the markup.
- The session token is moved out of `from` into its own field by the page script and stripped again server-side: it reaches this server, never the issue body. The thumbnail on the page *does* keep it (fetched by the visitor's own browser), or a gated box would show a broken image on the page meant to prove what the reporter saw.
- Form, not anchor, throughout — the action is a server-rendered literal and the script only ever writes input values, so no page-derived string reaches a navigation sink.

## Wiring for future coverage

`serve-report-bug.html` joins the committed preview-harness page fixtures, so the new page **and** the footer entry on all 60-odd existing pages are rendered and diffed by the CI visual bot on every subsequent PR.

## Test plan

- [x] `ServeBugReportTest` — 14 tests: body shape, embed vs link, token stripping, `sanitizeFrom` rejections (protocol-relative, scheme, fragment, control chars, length), path parsing, fence escaping, uptime formatting.
- [x] `ServeBugReportRouteTest` — 7 tests against a real embedded server: the route, the token gate, `from` resolution, off-origin refusal, unresolvable preview, token-in-thumbnail-not-in-body, and the footer entry present on every page but not on the report page.
- [x] `bugReport.test.ts` — 8 tests: the footer form's `from`/`token` split, pipe escaping in the user agent, and that the visible `<pre>` and the hidden body stay identical.
- [x] `./gradlew :cli:test` — full module, green.
- [x] `./gradlew ktfmtCheckAll` — green.
- [x] `npm run verify` in `cli/serve-web` — typecheck, 799 tests, bundle rebuilt and committed.
- [x] `pages-snapshot` preview-harness run — 126 passed; the two new PNGs above are from it.

https://claude.ai/code/session_0145AmTxmFujKkwYmeyTovpj

---
_Generated by [Claude Code](https://claude.ai/code/session_0145AmTxmFujKkwYmeyTovpj)_


yschimke:
All six review findings were real and are fixed in `ca3b918`. The P1 in particular: the feature was **silently inert end to end**, and my unit tests missed it because they built the DOM before calling the installers rather than in the order the real document does.

| Finding | Fix |
| --- | --- |
| **P1** — `serve-chrome.js` runs before the footer is parsed | Both installers now wait for `DOMContentLoaded` (the shape `installPageTheme` already uses). New test drives the real order: script first, footer after, assert unfilled, then dispatch. |
| **P2** — render URL dropped viewer overrides | Override params carried from `from` through `ServeOverrides.isOverrideParam` + `normalizeOverrideParams` into both the issue-body render URL and the page thumbnail; routing-only params and the token still excluded. |
| **P2** — root-form viewers (`/p/Red`) unresolved | Falls back to the path's `?session=`, else `defaultSessionId`. |
| **P2** — suspended sessions read as missing | Falls back to the registry's last-known snapshot (`catalogMetaSeen`) for provenance, trust and preview ids, and reports the lane honestly as `suspended (idle)` rather than mislabelling it baked. |
| **P2** — failures concatenated, not merged | Raw timestamped records merged and sorted newest-first *before* the cap. |
| **P2** — colour scheme reported the OS, not the page | The page's pinned `cp-scheme-*` is captured on the reported page (footer form carries it, since `/report-bug` has a scheme of its own) and **both** are reported, labelled apart. |
| **CodeQL high** — incomplete escaping | The user-agent cell escaped `\|` but not `\` or a backtick. Now escapes backslash **first** (or it doubles the escapes added after it), then pipe, then backtick. |

Three new browser tests and three new route tests cover the regressions. Local: `:cli:test` green, `ktfmtCheckAll` green, 802 serve-web tests green, 126-test preview-harness run green.

On the unrelated red checks on the previous commit — `Playground compiles + redeems a snippet end to end` and `Apply compose-preview` both died in Maven Central dependency resolution:

```
Could not GET '…/io/github/alexzhirkevich/compottie/2.2.4/compottie-2.2.4.pom'.
Received status code 403 from server: Forbidden
```

No test body ran (no `playground-lane.log` was created at all), and `compottie` is the Lottie dependency, which also explains the 58 vanished RemoteCompose renders in the preview-diff comment. Nothing to do with this diff — this push re-runs them.

---

yschimke:
Both new P2s were real, and both were regressions I introduced in `ca3b918`. Fixed in `849ae6a`.

**Stale context at submit.** The fields were filled once at `DOMContentLoaded`, but `installUrlState` rewrites the address bar on every knob/device/theme/history change — so a visitor who adjusted anything before pressing the button submitted the URL the page was *opened* at. That's especially bad here because the server then propagates those overrides into the render URL it embeds, so the fix I'd just made to carry overrides was being fed stale input. Both `from` and `scheme` are now re-read from a `submit` handler, which is the last moment before they're used. My own KDoc claimed "the address bar is the honest answer to what were you looking at" and then froze it at load — the comment was right and the code wasn't.

**Session fallback too broad.** Correct on both counts. `fromSessionId` returns `defaultSessionId` even with no `?session=`, and I applied it whenever no *known* path-prefixed system was found — so `/`, `/status`, `/docs/…` and a 404 were attributed to the default catalog with its provenance, trust and render lane, and a path naming an *unknown* system could fall through and match a same-named preview there, attaching a render that was never on screen. Now guarded on `ref.system == null && ref.previewSegment != null`; both halves are pinned by tests.

New tests: two route tests (server pages claim no system; an unknown system is not re-attributed) and one browser test that drives the real pushState-then-submit order.

---

**On the two remaining red checks** — `Apply compose-preview` and `VS Code Extension E2E (interactive-a-d)`. I don't believe either is this PR's, but I'm flagging rather than dismissing:

- They failed identically on `35bc950` and `ca3b918` — two independent runs — and this diff touches only `cli/serve`, the serve-web bundle, docs, and page fixtures. Neither job exercises any of that: `Apply compose-preview` fails inside a real-Gradle `samples:compose` render (`Failed: compose (exit 2)`), and the E2E failure is test B doing a real `:samples:cmp:composePreviewCompile` (`refresh reported 'failed'`). The page fixtures I changed feed `pages-snapshot.spec.mjs`, a different harness from the electron E2E.
- The `compottie` Maven Central 403 that broke the earlier run has cleared: the preview-diff bot on `ca3b918` shows **1 changed variant and no removed previews**, versus 58 removed RemoteCompose previews on `35bc950`. So that infra blip is no longer a confound, and these two are still red.

`849ae6a` gives a third data point. If those same two are red again with everything else green, I'll treat them as pre-existing and say so rather than keep pushing at them — but I'd rather state that once it's actually established.

---

PR #4074, opened 2026-08-16
fix(catalog): publish the motion axis the render already produced

yschimke:
## Summary

No catalog has ever published a Motion section — not m3-catalog's five expressive components, not this repo's own `compose-m3` dogfood sheet — and no run has ever said why. The captures themselves were fine the whole time: `previews/SwitchOn_Light-c16813d5.apng` is sitting in the published `design-artifacts/compose-m3` bundle right now, 550 KB of correctly rendered 60fps APNG, reachable by anyone who unzips it.

Two silent drops, in series.

**1. The axis never reached `catalog.json`.** The join sets `source.motion`, then hands the sources to `buildCatalog`, which assembles each component from an allow-list of fields it knows — `buildComponent` copies the ones it knows, `toCatalogManifest` copies them again. `motion` was not among them. What made it invisible rather than merely broken: the publish pass reads `manifest.components[].motion`, found nothing to copy, reported nothing missing — there were no declarations left to miss — and its log line is guarded on `published > 0 || missing.length > 0`, so it printed nothing at all. A green run, a complete-looking catalog, and no trace.

Fixed at the source: `@design-parity/catalog-export` learned the field in [design-parity#354](https://github.com/yschimke/design-parity/pull/354), released as **0.1.52**, and this bumps the pin onto it. The axis now reaches `catalog.json` the same way every other component field does, with no workaround landing here.

**2. The captures were unjoinable to their stickers.** `bundle pack` keyed a motion entry by the render's own leaf — `SwitchOn_Dark-d0f22b72.apng` — which is `<readable>-<digest>` (docs/RENDER_FILENAMES.md) and deliberately *not* the preview id, while the still beside it is keyed `previews/com.example…CatalogSelectionKt.SwitchOn_Dark.png`. Two namespaces. Both downstream joins are by name: `catalog-motion.mjs` reads a capture's theme off the still sharing its stem, and `catalog-motion-publish.mjs` names the published file after that still. Both declined, silently.

That one wouldn't have failed loudly either — it would have published *wrongly*. A capture with no theme is pinned to **every** card of its component (`ServeCatalogStore`: "a capture with no theme belongs to every card"), so the light card would have played the dark recording, under a filename derived from a render digest. Motion entries now share the stills' id space — `previews/<id>[_interaction|_anim].<ext>` — with the naming rule extracted to `motionBundleEntryPath` and pinned by its own test.

Fixing #1 alone gets bytes onto the branch; fixing both is what puts the right recording on the right card.

**And a check, because the silence was the real defect.** `buildCatalog` assembles from an allow-list, so *any* field it has not been taught vanishes between the join and the written manifest with no error anywhere — that property didn't go away, it just isn't biting `motion` today. `motion-carried.mjs` compares what the join resolved against what `catalog.json` ended up with: it logs the captures carried, and names every component whose axis was dropped plus the pin to check. A downgraded pin or a regressed allow-list now fails loudly instead of publishing a catalog that merely looks complete.

## Test plan

- `node --test` in `scripts/design-artifacts` — 752 pass, 0 fail, including 8 new `motion-carried.test.mjs` cases (carried, wholly dropped, partially dropped, resolved-for-a-component-not-in-the-manifest, never mutates the manifest). `package-version.test.mjs` reads its expected version from the lockfile, so it tracked the bump on its own.
- `./gradlew :gradle-plugin:test --tests '*MotionBundleEntryPathTest*' --tests '*PreviewBundleFormatTest*' --tests '*SanitizeBundleEntryIdTest*'` and `./gradlew ktfmtCheck` — green; 10 new cases pin the entry naming, the `_interaction` / `_anim` disambiguation, and that a still is declined.
- **Full chain replayed against real published data** — the actual `design-artifacts/compose-m3` render bundle and published `catalog.json`, through the actual published 0.1.52 engine, with no re-stamp anywhere in the chain:
  - join resolves 2 `Switch/On` captures, both theme-tagged (`light` / `dark`);
  - `buildCatalog` + `writeCatalog` put them in `catalog.json`, `images[]` untouched and still all `.png`;
  - `checkMotionCarried` → `{ declared: 1, carried: 1, captures: 2, dropped: [] }`;
  - publish plan → `motion/switch-on/ideal__default__light.apng` and `…__dark.apng`, **`unresolved: 0`**.
  
  Those flatten to the route ids `switch-on__ideal__default__light` / `…__dark` — exactly the pages that show nothing today. Before the fixes the same replay gives 2 untagged captures and `unresolved: 2` targeting `motion/switch-on/SwitchOn_Dark-d0f22b72.apng`.

## Visual evidence

The surface this restores can't be rendered in this container: producing it needs a full catalog render plus a delivery-branch publish, and until this lands no published catalog contains a `motion/` directory to point at. The nearest committed proxy is m3-catalog's own capture sheet from the commit that declared these five interactions, which is what the Motion lane is meant to play:

The five expressive components' interaction captures

How this gets verified visually for real: after merge the driver pin refresh (below) plus a `compose-preview` release land the fix in the `design-artifacts` lane; the next run publishes `motion/` onto the delivery branch, and https://preview.coo.ee/m3-catalog/p/switch-on__ideal__default__light grows the Motion chip it should have had since m3-catalog#99. The `` the chip paints into is already built by `ServeWeb` and already covered by `ServeWebFixtureTest` — nothing on the serving side changed here, it has simply never been given a capture to show.

## Follow-ups this needs to reach preview.coo.ee

1. **Driver pin refresh.** `design-artifacts-reusable.yml` pins the export driver to a SHA (`df2b08c9…`), deliberately, so the generator half does not reach consumers until that pin moves — the usual `fix(ci): refresh design artifact driver pin` follow-up.
2. **A release.** The `bundle pack` half ships in the published plugin/CLI, so it reaches m3-catalog (pinned at `composePreviewPlugin = 1.9.0`) only via a new release plus the routine renovate bump.

https://claude.ai/code/session_01TNe5b6rJ5Wm8actAqGLifG


yschimke:
All three failing checks are red on `main` at this branch's base commit (`0024cd2b`), not caused by this PR.

**Build CLI** and **Module Unit Tests** are the same one failure, twice — both run `:cli:test`:

```
ServeWebFixtureTest > serve web fixtures are in sync with ServeWeb() FAILED
    org.opentest4j.AssertionFailedError at ServeWebFixtureTest.kt:6162
2366 tests completed, 1 failed
```

Byte-identical on `main` at `0024cd2b` ([Build CLI](https://github.com/yschimke/compose-ai-tools/actions/runs/31972122278/job/95226070482), [Module Unit Tests](https://github.com/yschimke/compose-ai-tools/actions/runs/31972122278/job/95226070711)) — same test, same line, same 2366/1 totals. **Apply compose-preview** likewise fails on that same base commit ([run](https://github.com/yschimke/compose-ai-tools/actions/runs/31972122321)), and on the four `main` pushes before it.

Nothing here touches `ServeWeb` or its fixtures — the diff is the export driver (JS), the `bundle pack` motion entry naming, and `motion-carried.mjs`. The jobs that *do* cover this change are green: **Design Artifacts Driver Tests** ✅, **ktfmt** ✅, and `:samples:cmp:composePreviewRender` rendered its 68 previews including the animated captures inside the same failing job.

Since `ServeWebFixtureTest` is a serve-web fixture sync check and `main` has had a run of component-browser fixes landing (#4055, #4071, #4072, #4075, #4077), this looks like it belongs to that work rather than anything to fix here. I'll merge `main` in and re-run once it's green, and pick it up again if the failure survives that.

---

PR #4091, opened 2026-08-17
fix(serve): stop the render-server badge painting as a bar in Catalog mode

yschimke:
The render-server badge is a count **on** the header's Status link (`ServeWeb.siteHeader`) — the number is the summary, the link is where the detail is. Catalog mode's header carries no nav, so it has neither that link nor the badge's server-rendered slot, and no `/status` page to read the detail against.

`presenceScript` covered that case by creating the span itself and appending it to `<header>`. That header is a two-column grid (`.cp-site-header`, `minmax(0, 1fr) auto`) whose row already holds the brand and the Catalog/Dev toggle, so the appended span flowed into an implicit second row, in the `1fr` track, stretched by the grid item default `justify-self: stretch` — and the instance count painted as a full-width green bar under the brand, pushing the whole page down.

The slot is now the whole feature switch: no slot, no badge, and no poll either. Catalog mode had been fetching `/api/daemons` every 20s for an answer it has nowhere to display.

## Before / after

Catalog mode with a render server up, light:

| Before | After |
| --- | --- |
| The Catalog header with a full-width green bar reading "1" beneath the brand | The same header with no bar: brand, breadcrumb, title |

…and dark, where the bar is louder still:

| Before | After |
| --- | --- |
| The dark Catalog header with a full-width dark-green bar | The same dark header with no bar |

Both columns are the same new harness state, the `before` one run against the pre-fix `presenceScript`. Dev mode is untouched — it has the slot, and its `daemon-connected` / `daemon-idle` baselines are unchanged.

## Automatic coverage from here

This surface existed but was never shot with the poller on it, which is why the bar shipped. Closed in this PR rather than left as a one-off screenshot:

- `serve-component-browser-component.html` now carries a `presenceUrl`, so the real poller ships on the Catalog page the harness shoots.
- `pages-snapshot.spec.mjs` gains `serve-component-browser-component-daemon-connected`, stubbing a running daemon the same way the Dev landing's existing state does.

That state's claim is a **negative** one — a live render server must move nothing in the Catalog header — which a screenshot can hold and an element-waiting assertion cannot, because on a correct build the element never arrives. It waits on the presence heartbeat (which always fires) rather than the daemon poll (which correctly no longer does).

## Test plan

- `./gradlew :cli:test` — 2376 tests, only the two failures already present on a clean tree at this base (`ServeUrlStateTest › back and forward restore the whole selection without reloading`, `ServeWebFixtureTest › the in-browser Wasm lane is reachable from the renderer combo…`). Verified by stashing this branch's changes and re-running.
- New `ServeComponentBrowserTest › catalog mode paints no render-server badge`: Catalog mode keeps the presence heartbeat but emits no badge slot and no code path that manufactures one; Dev mode still server-renders the slot inside the Status link.
- `npm run harness:snapshot` for both touched fixtures — `serve-component-browser-component` (light + dark, new state) and `serve-landing-declared-themes` (light + dark, all existing states) — all pass.
- `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest` — no changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CMxDEybsRyvpWadPFgn7Xn)_


yschimke:
On the preview-harness diff above: the two **new** captures are this PR's (`serve-component-browser-component-daemon-connected`, light + dark) — that's the point of the change, and they show the Catalog header staying put with a render server up.

The one **changed** capture, `serve-landing-tree-depth-component-open`, is not from this PR. The only visual delta is the sticky header sitting ~24px lower over the title — the page is scrolled slightly at capture time — and everything below it is identical. That fixture carries no presence script and this branch doesn't touch its HTML (`serve-landing-declared-themes` and `serve-component-browser-component` are the only two fixtures in the diff), so nothing here can reach it.

It's run-to-run variance in a state already known to be non-deterministic: `component-open` clicks a tree row, and #3837 covered the hover half of that (fixed with `parkPointer`). Where the click leaves the scroll offset is the same class of variance, still unpinned. Left alone here rather than folded into an unrelated fix — worth its own change if it keeps reappearing.

---

yschimke:
Fix failure

PR #4093, opened 2026-08-17
fix(serve): open Catalog mode on the baked snapshot

yschimke:
Every component page in Catalog mode came up blank on the public deployment ([`preview.coo.ee/compose-m3/`](https://preview.coo.ee/compose-m3/)), while the same preview's PNG rendered fine.

## What was happening

Catalog mode was auto-enabling the in-browser CMP Wasm lane: the inline tab script ticked `#cp-wasm-toggle` at parse time, which is what rewrote the URL to `&mode=wasm` and swapped the stage.

`viewer.js` gates a boot-time lane switch on the initial snapshot having **landed**, because entering an interactive lane cancels the in-flight snapshot. Ticking the toggle directly bypassed that gate, so the stage's `` never got a `src` at all. The Wasm iframe is sized to that ``'s box — and a src-less `` still reports one: the browser's ~104×20 alt-text placeholder. So the app booted, initialized WebGL and posted its `cp-wasm-ready` first-frame signal, rendering correctly into a 104×20 iframe parked mid-stage, with the snapshot it was supposed to fall back to never having loaded:

```
IFRAME [{ src: "/wasm/compose-m3/?id=button-filled&uiMode=dark#…", w: 103.56, h: 20 }]
STAGE  { img: { src: "", nw: 0, nh: 0, w: 104, h: 20, vis: "hidden" } }
```

## The change

Catalog mode now opens on the same baked snapshot Dev mode does — the artifact the catalog publishes, and the one every other surface here compares against. The inline script is left doing only what it is named for: keeping the Preview / Source tab pair in sync. `?mode=wasm` still pins the in-browser lane and Dev mode keeps the renderer combo, so nothing is unreachable; it is just no longer the default. The two withdrawn items are marked as such in the component-browser checklist.

`positionOverlay` is hardened alongside it: mirroring the snapshot's box now requires a **decoded** render (`naturalWidth > 0`), not merely a non-empty rect, so a late or 404'd snapshot puts an overlay on the stage's content box instead of on a placeholder.

## Visual evidence

Live public deployment, `/compose-m3/p/button-filled__ideal__keyboard-focus__dark?chrome=catalog`, headless Chromium at 1400×1000, 35s settle per page. The "after" shot proxies that same deployment with this branch's `viewer.js` and rewritten inline script swapped in, so it is real catalog data with only the changed assets replaced.

**Before** — blank stage, URL rewritten to `&mode=wasm`:

Catalog mode before: blank stage

**After** — the baked snapshot, clean URL, and the keyboard-focus ring the Wasm lane cannot render:

Catalog mode after: the baked snapshot

**For contrast** — the Wasm lane entered the gated way (`?chrome=dev&mode=wasm`), which sizes the iframe to the render's 351×210 and draws. It was never the app that was broken:

Dev mode on the Wasm lane

The captures and the reasoning behind them are committed under [`docs/design/evidence/catalog-mode-snapshot-default/`](docs/design/evidence/catalog-mode-snapshot-default/README.md). The shipped markup is covered by the regenerated `serve-component-browser-component` page fixture, so the visual-diff bot diffs this surface on every future PR.

## Test plan

- `./gradlew :cli:test` — green (includes `ServeComponentBrowserTest`, which now asserts Catalog mode opens on `data-mode="snapshot"` and emits no `getElementById("cp-wasm-toggle")`, and `ServeWebFixtureTest`, refreshed with `UPDATE_SERVE_WEB_FIXTURES=true`)
- `cli/serve-web`: `npm run typecheck`, `npm test` (815 passing), `npm run build`, `format:check`
- `./gradlew ktfmtFormat`
- Manual: the three captures above, plus a confirmation that the standalone Wasm app renders correctly at every size once it is given a real box

---
_Generated by [Claude Code](https://claude.ai/code/session_01FxLTzz2N5LQP5C7HkSUejg)_


yschimke:
`Compose Preview` is red here, but it fails on the base branch too — not something this PR introduced.

Same failing test, same cause, on `main` at this PR's base commit `a2cf3e7` ([run 6955](https://github.com/yschimke/compose-ai-tools/actions/runs/32002150635)) as on this PR ([run 6959](https://github.com/yschimke/compose-ai-tools/actions/runs/32005105806)):

```
Failing tests in :samples:sdk-matrix:composePreviewRender (1):
  RobolectricRenderTest.renderPreview[com.example.sdkmatrix.MatrixPreviewKt.SdkMatrixPreview_SdkMatrixPreview]
    java.lang.IllegalStateException: compose-preview:
    -Dcomposeai.render.linkBufferComposer=true was requested, but this render's Compose
    runtime has no androidx.compose.runtime.ComposeRuntimeFlags.isLinkBufferComposerEnabled.
```

`:samples:sdk-matrix` renders against the older `compose-bom-compat` line by design, so the `linkBufferComposer` opt-in has no flag to set there. Fixing it means either dropping the flag for that module or moving it forward a Compose line — either way it is out of scope for a change that touches only the serve viewer's HTML/JS, and this diff contains no sample, renderer, or plugin code.

The large `samples:design-catalog-wear-m3` churn in the preview-diff comment above comes from the same place, for the same reason: nothing in this diff can change a rendered preview.

I'll re-check and merge the base branch in once `main` is green again.

---

Issue #4086, opened 2026-08-16
`@OverrideVariant.kitAxis` / `kitValue` are dead metadata — nothing downstream reads them


yschimke:
Follow-up to #4073, and a correction to its "Also worth having" section. That section said `@OverrideVariant` "already has `kitAxis` / `kitValue` for exactly this" and proposed adding them to `@CatalogVariant`. **Adding them there would produce a second pair of fields nothing reads**, because the existing pair is already inert.

## The finding

The fields are declared, documented, discovered and serialised:

- `OverrideVariant.kt` declares them, with a worked example (`kitAxis = "Show avatar"`, `kitValue = "True"`).
- `PreviewDiscovery.kt` reads them into `OverrideVariantSpec`.
- Both `PreviewData.kt` mirrors carry them, so they reach `previews.json`.

And then nothing consumes them:

```
$ grep -n "kitAxis\|kitValue" design-map/design-map.mjs design-map/emit-design-map.mjs
(no matches)

$ grep -ro "kitAxis\|kitValue" @design-parity/kit-index@0.1.51
(no matches)
```

`design-map.mjs` never projects them into the sidecar, so they cannot reach a resolver even in principle — and the resolver keys purely on `seeds` regardless. The annotation's own KDoc says they are "metadata only: they do not change the preview override seed," which is accurate; what it does not say is that no reader exists.

## Why it matters

This is the difference between a catalog naming the Compose-side value and having to spell the kit's. m3-catalog#130 folded `DatePicker/Range` onto `DatePicker/Modal` and had to write:

```kotlin
props = ["type=full-screen (range)"]
```

`type=range` resolves to nothing against the kit's `Type=Full-screen (range)` — silently, dropping the node. The natural fix is exactly what `kitAxis` / `kitValue` are described as doing, and it is unavailable: a kit spelling, parentheses and all, is now load-bearing in catalog source, where it will rot the next time the kit renames a variant value.

## What the fix actually needs

Three steps, and the middle one is in another repo — worth knowing before anyone plans this as a one-line annotation change:

1. **`design-map.mjs`** — project a declaration's `kitAxis` / `kitValue` into the sidecar beside its `seeds`, so a resolver can see them at all.
2. **`@design-parity/kit-index`** — prefer an explicit kit axis/value over the seed's own spelling when matching a declaration to a component-set variant. This is the load-bearing half, and it ships as a published package that catalogs pin by version, so it lands on its own release cadence.
[...]

Issue #3899, opened 2026-08-15
preview-annotations: let a component or variant name the design-kit axis its knob maps to


yschimke:
## The gap

A catalog joins its rendered variants to a design kit's variant nodes by matching knob names against kit axis names. Today that match is a guess made downstream, because nothing in the annotations carries the correspondence.

`@OverrideVariant(name = "avatar", strings = ["content=avatar"])` says the sticker seeds `content=avatar`. It does not say that `content` means the kit's `Show avatar` boolean property on this component and its `Configuration` axis on that one. The author knows; the annotation has no field for it; so every consumer re-derives it heuristically from the strings.

## What that costs in practice

m3-catalog resolves the join in [`scripts/kit-variants.mjs`](https://github.com/yschimke/m3-catalog/blob/main/scripts/kit-variants.mjs) — 35 axis-alias entries, 27 value-alias entries, and several layers of fuzzy word-matching over kit axis names. Measured on the current tree:

- **109 of 647 authored variant renders (16.8%) resolve to no kit node** and are silently left uncompared (yschimke/m3-catalog#101)
- **17 distinct knobs** (`state`, `labels`, `content`, `leading`, `trailing`, `header`, `handle`, `dividers`, `fab`, `menu`, `caret`, `footer`, `headline`, `nav`, `inset`, `action`, `configuration`) all alias to the single kit axis `Configuration`

The heuristic's own comments record what happens when a knob name is the only signal available: `footer=true` matched `Show back=True`; `supporting=on` matched `Leading icon=True`; `inset` matched `Middle-inset` before `Inset`. Each was fixed by adding a table entry. There is no reason to expect the table to converge, because the ambiguity is real — one knob name genuinely does mean different kit axes on different components, and only the author knows which.

This is the failure mode m3-catalog's own AGENTS.md predicts: *"If you find yourself writing a lot of mapping config to express something, that is a signal the upstream libraries are missing an annotation — raise it in compose-ai-tools and add the annotation there rather than growing a JSON file here."* This issue is that raise. The config grew in a `.mjs` file rather than a JSON one, which is why it went unnoticed for a while.

## Proposed

An optional field on the annotations that already exist, naming the kit-side axis or property a knob maps to:

```kotlin
@OverrideVariant(name = "avatar", strings = ["content=avatar"], kitAxis = "Show avatar")
```

and, for the common case where one knob maps consistently across a whole component, the same field on `@CatalogComponent` as a default for its cells.

Design notes:

[...]

PR #4094, opened 2026-08-17
feat(design-map): project a variant's declared kit axis and value

yschimke:
Steps 1 and 3 of #4086. Step 2 — the resolver half that makes these fields mean something — is design-parity#356, which lands on its own release cadence.

## The finding this closes

`@OverrideVariant.kitAxis` / `kitValue` have been declared, documented, discovered and serialised since #3899, and read by **nothing**. The projection never put them in the variant sidecar, so they could not reach a resolver even in principle. That is what #4086 recorded, and it is why adding the same pair to `@CatalogVariant` *first* would have produced a second pair nobody read.

## What changes

- **`design-map.mjs`** carries a declaration onto the seed it belongs to, so it travels in the sidecar beside `key` / `raw`.
- **`@CatalogVariant`** gains `kitAxis` / `kitValue`, so a folded component can name both sides too — discovery reads them, and both `CatalogEntry` mirrors carry them through to `previews.json`.

Together with the resolver half, that is what lets a catalog keep the Compose word and hand the kit its own spelling. m3-catalog#130 had to write:

```kotlin
props = ["type=full-screen (range)"]
```

because `type=range` finds nothing against the kit's `Type=Full-screen (range)`. A kit spelling, parentheses and all, load-bearing in catalog source — and it rots the next time the kit renames a value. It now reads:

```kotlin
@CatalogVariant(of = "DatePicker/Modal", props = ["type=range"],
                kitAxis = "Type", kitValue = "Full-screen (range)")
```

## What it deliberately does not do

**Projecting is not translating.** Nothing here checks a declaration against a kit, because there is no kit here to check against — that is the whole reason the split falls where it does. This only puts the author's words where a resolver can read them.

The one thing it *does* judge is whether a declaration can be **placed**. The annotation carries one pair per variant, so a cell seeding two knobs gives no way to say which one the axis names. Those are reported (`diagnostics.unplacedDeclarations`, printed by the emitter) and dropped rather than guessed at — a guess pins the wrong axis and resolves, confidently, to the wrong node. A COMPONENT-level `kitAxis` stays a *default* and is silent when it cannot be placed: a blanket that does not cover everything is not the same kind of mistake as an assertion that cannot be honoured.

Additive throughout — a variant that declares neither field projects exactly as before.

## Test plan

- `npm test` in `design-map/` — 31 tests, all passing, 7 new: both annotation forms carrying a declaration, either field standing alone, the component default being inherited and overridden per cell, the unplaceable-declaration diagnostic, the silence of an unplaceable default, and the fields arriving in the sidecar.
- `./gradlew -p gradle-plugin :preview-discovery:test` — green, with a new bytecode-level case asserting a `@CatalogVariant`'s `kitAxis` / `kitValue` are read off the CLASS-retained annotation table (beside the existing `@CatalogComponent` / `@OverrideVariant` one).
- `./gradlew :preview-data-api:test` — green, with a new wire test pinning both fields through manifest decoding.
- `ktfmtCheck` on the two `api/` modules — green. `:preview-discovery:ktfmtCheck` reports pre-existing drift across nine files on an unmodified `main` in this environment (a local ktfmt version skew), so that module's formatter was deliberately not run: only the two intended edits are in the diff.

Text-only change: no rendered surface is affected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QMqCAjq4C7mZHRcnfKY3tG)_


yschimke:
All three taken, pushed in 8d8decd with a test each.

- **Value-only cell declarations inherit the component axis.** That is the case the component-level default exists for — the component names the shared property once, cells name the values they sit at — and dropping the axis left a seed carrying a value with nothing to hang it on.
- **A driven interaction no longer makes a one-knob cell ambiguous.** The synthetic `state` seed comes from the harness, not the annotation, so it does not count towards "which knob does this declaration name". A cell seeding `size=l` and pressing the component keeps its declaration; an interaction-only cell still has the state seed as its subject.
- **`CatalogEntry` keeps its pre-`kitValue` constructor descriptor.** Appending a defaulted parameter preserves source compatibility only, and a consumer compiled against 1.11.0 would have failed with `NoSuchMethodError` — late, since the type is decoded far more often than constructed. Added the overload plus an ABI test, following `PreviewResult`'s pre-`projectDirectory` constructor two declarations down in the same file.

---

yschimke:
On the `copy` follow-up: the observation is correct — adding `kitValue` does change `copy` / `copy$default` as well as the constructor — but I'm not adding a bridge for it, for two reasons.

**A hand-written 14-parameter `copy` overload breaks source compatibility to fix binary compatibility.** Kotlin's generated `copy` takes all 15 parameters with defaults, so an overload taking the old 14 (also all defaulted) is applicable to every named-argument call site. `entry.copy(caption = "x")` then fails to compile with an overload-resolution ambiguity — for every consumer, including this repo. Trading a link error that needs a stale artifact for a compile error that needs nothing is not an improvement.

**The repo's own precedent draws the line at the constructor.** `PreviewResult`'s pre-`projectDirectory` overload two declarations down (#4075) restores the constructor descriptor and nothing else, and its KDoc says so explicitly. `CatalogEntry` now matches that, with the same kind of ABI test. Widening the policy to `copy` is a decision about the whole published API — and one that would need a different mechanism than an overload — rather than something to settle inside a feature PR.

Worth noting what the exposure actually is: `CatalogEntry` is a wire DTO. Consumers decode it through the module's own generated serializer, which is compiled together with the class; constructing or `copy`ing one from outside is the rare path, which is exactly why the constructor bridge is cheap insurance rather than a full compatibility programme.

---

yschimke:
The preview-diff bot's one "Changed" entry — `SharedElementFilmstripPreview` — is **not** from this PR. Raised as #4097 with the evidence; recording it here so the diff is not read as this change's.

This PR touches the design-map projection, two annotation fields and their discovery. Nothing in the renderer, nothing in `samples/android`. What the render actually shows:

```
bbox             (32, 511) – (441, 672)      one contiguous band, ~one panel of five
changed px       57,317 / 1,922,629 = 2.98 %
max channel diff 164                          a different frame, not a re-encode
```

And that same PNG changed in ten of the most recent render commits across five unrelated PRs (#4092's CI repair, #4090's serve fix, #4074, #4069 four times over, #4053). The cause looks like `LaunchedEffect { seekState.seekTo(…) }` — `seekTo` suspends, so the *destination* is deterministic but the capture does not wait for the panel to arrive there.

---

yschimke:
`Module Unit Tests` is red, and it is **not this PR**: the same assertion fails on `main`.

```
WearFocusedPressPixelTest > pressed capture changes the button container beyond the focused state FAILED
    com.google.common.truth.AssertionErrorWithFacts at WearFocusedPressPixelTest.kt:30
16 tests completed, 1 failed
```

Same test, same first assertion (`pressed != focused`), on `main` at `a2cf3e7` — [run 32002150575](https://github.com/yschimke/compose-ai-tools/actions/runs/32002150575), 06:39, reported there at `:28` because #4092 has since grown the comment above it by two lines.

Worth flagging separately: this branch merged `main` at `c7c5b5b`, so its run **includes #4092's repair** — `@FocusedPreview(indices = [0], pressed = true)` in place of the seeded `MutableInteractionSource` — and the press still does not reach the pixel. That makes this run the first completed post-repair result (the two `main` runs since #4092 landed were both cancelled), and the answer is that the repair has not taken. Raised as #4098.

Nothing in this PR touches the renderer, the Wear catalog, or the capture path — it is the design-map projection plus two annotation fields. I'll re-run when `main` recovers.

---

yschimke:
Base recovered — #4095 landed on `main` while I was writing the previous comment, and I filed #4098 against a `main` I had not re-fetched. That issue is closed as superseded; the correction that matters is that #4095 **reverts** #4092's approach (back to the seeded `MutableInteractionSource`, `@FocusedPreview(pressed = true)` removed) and diagnoses the failure as the specimen depending on the shard layout — not either of the causes I guessed at.

Merged `main` into this branch as `d60f5a0`; CI is re-running against `ba309d9`.

---

yschimke:
`VS Code Extension E2E (interactive-e-g)` failed on `d60f5a0`, on scenario **E** (rename a `@Preview` function). Root cause is in the log and is not the scenario's subject:

```
composePreview: module 'cmp' has 18 source file(s) declaring @Preview but its active
class outputs contain 0 .class files. A compile task may have restored an empty
build-cache entry; rerun with --no-build-cache --rerun-tasks.
  classDir: samples/cmp/build/classes/kotlin/main (exists=true, classFiles=0)
```

With no classfiles, discovery emitted `0 code preview(s) and 3 asset preview(s)`, so `composePreviewRender` matched nothing, the render failed, and the panel resolved to its empty state — which is exactly what the assertion reported. The plugin's own message names the mechanism: an empty build-cache entry restored for the compile task. The scenario also cancels `composePreviewCompile` mid-flight by design (`terminating gradlew pid 4107` in the log), which is the obvious way to end up with a half-written class output directory.

Not from this diff: nothing here touches compilation, caching, or the render path. `main` is green on this workflow at `ba309d9` ([run 32008346648](https://github.com/yschimke/compose-ai-tools/actions/runs/32008346648)), and the same workflow has failed intermittently on `main` through the day (`d2ccc99`, `a2cf3e7`, `e09ecc5`, `bb19254`, `1c47a48`).

Re-running the failed job — the run still had jobs in flight when I tried, so I'll trigger it as soon as it settles. If scenario E fails the same way twice, that is a different conversation and I'll raise it rather than re-run again.

---

yschimke:
A second E2E shard failed too — `interactive-a-d`, scenario **B**, and with a different symptom from `interactive-e-g`'s:

```
Timed out after 180000ms waiting for: setPreviews after adding Interactive1786954564656,
with its capture on disk
  capture(s) not on disk: samples/cmp/build/compose-previews/renders/Interactive…-a8eb6ffe.png
```

So one shard rendered nothing because the class outputs were empty, and the other waited three minutes for a capture that never arrived. Scenario A in that same shard took **115s** and D took 36s, against a job budget of six minutes for four scenarios — the shape of a slow or contended runner rather than of a broken assertion.

Neither symptom can come from this diff: it adds two annotation fields, their discovery, and a JSON projection. Nothing in it participates in compilation, caching, or rendering — and if discovery were broken, the failure would be deterministic rather than one shard here and one scenario there.

Re-run of the failed jobs queued (the run had settled by then). Standing by what I said earlier: if the same scenarios fail again, I stop re-running and raise it — twice is a reproducible cache/timing problem in the harness, which deserves a fix rather than another attempt.

---

yschimke:
`Module Unit Tests` is red again on `d60f5a0`, and I was wrong to call the base recovered — **#4095 did not fix the Wear test either.** `main` fails the identical assertion at `ba309d9`, #4095's own merge commit ([run 32008346698](https://github.com/yschimke/compose-ai-tools/actions/runs/32008346698), `WearFocusedPressPixelTest.kt:28`).

I've reopened #4098 with the full picture. Four attempts, one assertion:

| Approach | Where | Result |
| --- | --- | --- |
| seeded `MutableInteractionSource` | `main` pre-#4092 | FAILED `:28` |
| `@FocusedPreview(pressed = true)` (#4092) | this PR (`c7c5b5b`) | FAILED `:30` |
| back to seeded source (#4095) | `main` (`ba309d9`) | FAILED `:28` |
| — inherited | this PR (`d60f5a0`) | FAILED `:28` |

The new lead is in that `main` run: `composePreviewRender` and `composePreviewDiscover` were both **FROM-CACHE**, so the PNGs the test read were not produced by the run. If the render task's cache key does not cover the pressed-capture spec, then both fixes were fine and neither was ever rendered — which would explain two *opposite* changes producing an identical failure.

Either way this is a base-branch failure that predates and outlives the diff here. Everything else on this PR is green.

---

yschimke:
`interactive-a-d` failed again — and as promised, I'm not re-running it a third time. Raised as #4101.

It failed in a **different scenario**, which is the point:

| Run | A | B | C | D |
| 1 | ✓ 115s | ✗ timeout 180s | ✓ **1.7s** | ✓ 36s |
| 2 (re-run) | ✓ 109s | ✓ 112s | ✗ **timeout 480s** | ✓ 35s |

Same commit, same shard. Scenario C completes in 1.7 s on one run and cannot finish in 480 s on the next; B fails then passes. A scenario that swings from 1.7 s to >8 min is not failing on its own logic, and a diff cannot be responsible for a failure that moves to a different test when you run it again.

Both symptoms in #4101 are upstream of the assertions — renders never arriving, and class outputs coming back empty from a cancelled compile (the plugin's own error text names the mechanism). The suite cancels real Gradle invocations by design, which is plausibly manufacturing the condition the next scenario trips over.

Status of this PR: everything green except two known base/infra failures — `Module Unit Tests` (#4098, `main` is red on it too) and this shard (#4101). Nothing left that points at the diff.

---

PR #4100, opened 2026-08-17
fix(renderer): share the pressed-ripple settle with focus GIFs

yschimke:
## Summary

Review follow-ups to #4096, which merged before they could be folded in. Three points, all from the Codex review on that PR.

**1. Pressed focus GIFs kept the old window.** `handleFocusGifCapture` still idled for `FocusController.SETTLE_MS` on a pressed step, so a `@FocusedPreview(gif = true)` frame stayed shard-dependent in exactly the way #4096 fixed for static PNGs. Both paths now call one `settlePressedRipple`, so a pressed GIF frame cannot settle differently from a pressed PNG.

**2. `WearFocusedPressPixelTest` did not catch what the settle gets wrong.** It asserted inequality, but an under-sized window degrades the ripple *gradually* — measured `#C2B5DB` → `#D5C8EC` → `#D4C8EC` as the sandbox is reused — and only the last of those is *equal* to the focused fill. The middle one passed: a capture one channel step away from focus-only, published as `pressed`. The comment I left on it claimed a merely-close result would fail, which was wrong. It now asserts a minimum per-channel distance, sized off the settled values (a settled press sits 17–19 steps from focused and ~36 from resting; the under-settled one sat 1) and reports both fills as hex on failure.

**3. The settle drains unrelated main-looper work.** Correct, and I could not make it not do that. Idling the window runs whatever a pressed preview had scheduled inside it — a `postDelayed`, a widget's own timer — before the capture. Two narrower forms were measured against the full catalog at `shards = 1`:

| attempt | pressed fill | settles? |
| --- | --- | --- |
| 1 s ceiling instead of 5 s | `#D4C8EC` | no |
| stop early once `ShadowLooper.getNextScheduledTaskTime()` reports nothing due within a frame | `#D4C8EC` | no |
| the merged form (5 s, stepped) | `#C2B5DB` | yes |

So the window stays as it is. What changed is that the cost is now written down rather than implied, along with the alternative I did not get to: ending the settle on the ripple's own state (`Animatable.isRunning()` on the composition's `RippleHostView`s), which needs a way to reach those drawables from the renderer that does not exist yet. The blast radius stays bounded by the gate — pressed captures only, a handful of catalog specimens.

## Test plan

Local, with a real Android SDK + Robolectric:

- `./gradlew :samples:design-catalog-wear-m3:composePreviewRenderAll :samples:design-catalog-wear-m3:testDebugUnitTest --rerun-tasks` at **`shards = 1`** (the CI-failing configuration) — **BUILD SUCCESSFUL**, 16 tests, 0 failures, `ButtonPressed` `#C2B5DB`. Same at **auto sharding** — **BUILD SUCCESSFUL**.
- The strengthened assertion was confirmed to fail on a genuinely under-settled render, not just to pass on a good one: the 1 s and early-exit attempts above both failed it.
- `./gradlew :renderer-android:test` and `./gradlew ktfmtCheckAll` — both clean.

## Visual evidence

No pixel change: `ButtonPressed` renders the same `#C2B5DB` this PR and last, and the evidence renders committed in #4096 under [`docs/design/evidence/wear-pressed/`](https://github.com/yschimke/compose-ai-tools/tree/main/docs/design/evidence/wear-pressed) still describe the current output. This PR moves the settle into a shared helper, extends it to the GIF path, and tightens a test — the only surface whose pixels *could* move is a pressed `@FocusedPreview(gif = true)` capture, and no preview in this repo uses that combination today, which is why the GIF path's drift went unnoticed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01K75KTvGkd2NRMZX6AXmaMB)_

### DIFF
--- a/renderers/android/src/main/kotlin/ee/schimke/composeai/renderer/RobolectricRenderTest.kt
+++ b/renderers/android/src/main/kotlin/ee/schimke/composeai/renderer/RobolectricRenderTest.kt
@@ -1595,25 +1595,12 @@ abstract class RobolectricRenderTestBase(
               // deterministic timing; the Android clock and Compose's `mainClock` are
               // independent, so this advances only platform animations.
               //
-              // [FocusController.SETTLE_MS] is NOT the right window for it, which is the
-              // bug this replaced. That constant sizes a Compose-side settle; the platform
-              // ripple needs far more looper time, and how much more depends on how long
-              // the sandbox has already been running — so a 250ms window silently
-              // under-settles the ripple in proportion to the number of previews rendered
-              // before this one. Measured on `:samples:design-catalog-wear-m3`'s
-              // `ButtonPressed` at `shards = 1`, the pressed container faded from #C2B5DB
-              // (3 preview rows ahead of it) through #D5C8EC (11 rows) to #D4C8EC (the
-              // full 59-row catalog) — the last of those pixel-identical to the
-              // focus-only capture, i.e. a "pressed" specimen showing no press. Because
-              // `shards` auto-sizes off a `previews.json` a cold CI checkout does not
-              // have, the same commit rendered pressed pixels in one job and focus-only
-              // in another. [PRESS_SETTLE_MS] is sized for the settled end state instead,
-              // and stepping keeps the `Choreographer` delivering a callback per frame
-              // across it rather than one for the whole jump.
+              // This used to idle for [FocusController.SETTLE_MS], which is a Compose-side
+              // window and silently under-settles the ripple in proportion to how long the
+              // sandbox has been running — see [PRESS_SETTLE_MS] for the measurements and
+              // for why that made a pressed specimen depend on its shard layout.
               if (focus.pressed) {
-                val looper = org.robolectric.Shadows.shadowOf(android.os.Looper.getMainLooper())
-                val step = java.time.Duration.ofMillis(PRESS_SETTLE_FRAME_MS)
-                repeat((PRESS_SETTLE_MS / PRESS_SETTLE_FRAME_MS).toInt()) { looper.idleFor(step) }
+                settlePressedRipple()
               }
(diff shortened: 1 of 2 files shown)


yschimke:
Triaging the two rows in the preview-diff comment above, since neither is this PR's doing and the body says "no pixel change".

**`ButtonPressed` — stale baseline, not a change from this PR.** The "before" side is the *pre-#4096* render: `md5 c15ba6ac9f71515606e131c2a970b77f`, container fill `#E9DDFF` at (30, 68), byte-identical to what `compose-preview/main` published from 2026-08-14 until #4096 landed. The "after" is `#C2B5DB`, which is what #4096 already produces on `main`. So this row is #4096's change surfacing against a baseline captured before it, not something #4100 does — this PR moves that settle into a shared helper without altering its behaviour for the static PNG path.

**`SharedElementFilmstripPreview` — unstable, and independent of this PR.** It has no `pressed` capture, so the settle this PR touches (gated to `focus.pressed`) cannot reach it. Verified per the stability guidance by rendering it twice on the same commit locally:

```
336074be3c20e08d261bb07d5509b350  SharedElementFilmstripPreview_Shared_Element_Filmstrip-a8af1cd8.png
7598dd7775cfa54b45d5e0710d27ae86  SharedElementFilmstripPreview_Shared_Element_Filmstrip-a8af1cd8.png
```

Different bytes, same commit, no code change in between. The bot's before/after pair differs in **7.83% of pixels with a max channel delta of 216**, so it is swinging between substantially different frames rather than jittering at the edges — consistent with the preview's own construction: five `SeekableTransitionState`s each seeked once on first composition, where which frame you capture depends on whether the seek has been applied by capture time.

Flagging rather than fixing — it wants its own change, and it will keep showing up as noise on unrelated PRs until then.

---

PR #4120, opened 2026-08-17
fix(vscode): render a @Preview added to the file already on screen

yschimke:
`VS Code Extension E2E (interactive-a-d)` scenario **B** has been failing intermittently on `main`. It is not a flaky test. It is a real user-facing bug, and the test was right.

## The bug

Add a `@Preview` to the file you are currently looking at. Discovery finds the new id and the panel draws its card — and the card never gets pixels. It stays empty until you navigate away and back, or something restarts the daemon.

The daemon view-open warm-up was suppressed by a bare `Set` of `(module, file)` scope keys, so *"this file has been warmed"* was its entire memory. A scope, once warmed, stayed warmed for the life of the daemon — the only things that cleared it were `onClasspathDirty` and `onChannelClosed`, i.e. daemon death.

Editing a source file leaves the daemon alive. So `warmShownPreviewsForFile` returned early and never asked the daemon to render the preview that had just appeared.

## Why it looked like flakiness

The scenario fails only when the daemon survives the edit. When the render instead goes through Gradle's `composePreviewRenderAll`, the PNG lands and the test passes — hence the intermittency.

The scenario's own comments had already got halfway there, noting that after `composePreviewDiscover` "the extension then issues no further Gradle command … it is wedged, not slow". That was correct and the conclusion was the near-miss: the missing command was never Gradle's. It was a daemon render nobody requested. The CI transcript matches exactly — `composePreviewDiscover` at 20:20:24, silence, timeout at 20:27:09.

Worth noting the failure predates this line of work: the same job failed on `0568bcc99` (a CI-pin refresh) and `c9903031d`, both before any of it.

## The fix

The guard now records **which previews** a warm-up covered, so the question becomes "has this file been warmed for *these* previews" rather than "has this file been warmed". A focus bounce answers the same id set and is still suppressed — that is what the guard exists for — while a file that just grew a preview gets its render.

Extracted as `DaemonWarmScopeLedger` rather than patched in place. The decision is pure logic and was untestable inside a 5,000-line module, and this is a bug about a decision, not about wiring. Fingerprints are order-independent (discovery does not promise ordering) and use a separator that cannot occur in a preview id, so `["a", "b"]` cannot collide with `["a b"]`.

## Test plan

```
npm test          # 1711 → 1721 passing
npx tsc -p ./     # clean
npm run format:check
```

Three of the ten new tests fail against the old set semantics — including the one that is the bug:

```
1) warms again when a preview is added to a warmed file
2) warms again when a preview is removed
3) cannot be fooled by an id that contains the separator
```

The rest pin the behaviour that must survive: focus-bounce suppression, order independence, scope isolation, forgetting a scope whose warm-up threw, clearing a dead daemon's module without touching neighbours, and not letting `:samples:cmp` clear `:samples:cmp-wasm`.

## What I could not verify here

**I could not run the E2E scenario itself.** `@vscode/test-electron` downloads the VS Code binary from `update.code.visualstudio.com`, and this environment's egress proxy refuses that host:

```
connect_rejected  update.code.visualstudio.com:443  gateway answered 403 to CONNECT
```

So the end-to-end proof is CI's to give, not mine. What stands behind the fix here is the unit-level regression test plus the traced code path; if scenario B fails again after this lands, the diagnosis above is wrong and I would want to know.

Text-only PR: no rendered surface changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VTL61d6Gj1HAXJ4c1Kfpgn)_

### DIFF
--- a/vscode-extension/src/daemonWarmScopes.ts
+++ b/vscode-extension/src/daemonWarmScopes.ts
@@ -1,0 +1,83 @@
+/**
+ * Which (module, file) scopes have had their previews pre-rendered by the
+ * daemon, and **for which previews**.
+ *
+ * ### Why this is a ledger and not a `Set`
+ *
+ * The daemon view-open warm-up exists to stop a focus bounce re-rendering cards
+ * that are already on screen. It was suppressed by a bare set of scope keys, so
+ * "this file has been warmed" was the whole memory — and a scope, once warmed,
+ * stayed warmed for the life of the daemon. Only a daemon death (classpath
+ * dirty, channel closed) cleared it.
+ *
+ * That is wrong for the case where the file's previews *change* under a living
+ * daemon, which is the ordinary act of adding a `@Preview` to the file you are
+ * looking at. Discovery found the new id and the panel drew its card; the
+ * warm-up then declined to render, because the scope was already warm. The card
+ * sat there with no pixels behind it until the user navigated away and back, or
+ * something restarted the daemon. It also made the
+ * `B. edit a @Preview source, refresh, expect new id` end-to-end scenario fail
+ * intermittently — whenever the daemon happened to survive the edit, the PNG the
+ * test waits for was never written.
+ *
+ * Recording the id set turns the question from "has this file been warmed" into
+ * "has it been warmed for *these* previews", which suppresses the bounce (same
+ * ids, same answer) without swallowing a real change.
+ */
+export class DaemonWarmScopeLedger {
+    private readonly warmed = new Map<string, string>();
+
+    /**
+     * Whether [ids] still need a warm-up render in [scopeKey].
+     *
+     * True when the scope has never been warmed, or was warmed for a different
+     * set of previews. Order is not significant — discovery is free to reorder
+     * a manifest without that meaning anything changed.
+     */
+    shouldWarm(scopeKey: string, ids: readonly string[]): boolean {
+        return this.warmed.get(scopeKey) !== fingerprint(ids);
+    }
+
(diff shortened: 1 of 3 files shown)


yschimke:
**Correction: this did not fix scenario B.** The PR body above claims it does, and that claim is wrong.

`VS Code Extension E2E (interactive-a-d)` failed again on main at `de7b61b09`, which contains this change (verified: `daemonWarmScopes.ts` present, four `DaemonWarmScopeLedger` references in `extension.ts`). Identical error and identical shape:

```
Error: Timed out after 480000ms waiting for: setPreviews after adding Interactive1787000880222,
  with its capture on disk
preview …Interactive1787000880222 capture(s) not on disk: …/Interactive1787000880222-748d6ac4.png

21:09:18  composePreviewDaemonStart
21:09:39  composePreviewDiscover
          ← 6m42s of silence
21:16:21  B. FAILED
```

What went wrong in my reasoning: the job passed on this PR, and I let one green run of a job with a *known intermittent* history read as confirmation. It never was — I noted the caveat and then leaned on the result anyway.

**The bug this fixes is still real** and worth keeping: a warm scope, once warmed, stayed warmed for the life of the daemon, so a `@Preview` added to the file on screen was discovered and never rendered. Three of the ten tests fail against the old set semantics. What is not true is that it owns the scenario-B symptom.

The remaining lead is the 6m42s of silence after `composePreviewDiscover`. This run uploads `e2e-renders-interactive-a-d.zip` (726 kB) including `vscode-extension/.vscode-test/user-data/logs/` — the extension host log covering that window. That is the evidence to work from, rather than a third hypothesis.

---

PR #4145, opened 2026-08-17
feat(serve): persist warmed theme renders across restarts and regenerations

### LINKED ISSUE #4130: feat(serve): persist the theme cache across restarts and regenerations
## What's wrong

`CatalogThemeCache` is a process-local `LinkedHashMap` with no disk tier, and it is dropped by **two** independent events:

1. **Server restart.** Nothing is persisted.
2. **Delivery-branch regeneration.** Every catalog load builds `ServeSessionState(… catalogThemeCache = CatalogThemeCache() …)` (`ServeCommand.kt:3446`, also 3144/3347). The `state.catalogThemeCache ?: CatalogThemeCache()` at `ServeCommand.kt:1342` only preserves the cache across *host* replacement within one generation — daemon suspend/resume — not across a reload.

For `m3-catalog` on `preview.coo.ee`, counting both over three days:

| day | regenerations | releases | resets |
|---|---|---|---|
| Aug 15 | 7 | 3 (v1.6.0–v1.8.0) | 10 |
| Aug 16 | 5 | 4 (v1.9.0–v1.11.0) | 9 |
| Aug 17 | 3 | 4 (v1.11.1–v1.14.0) | 7 |

The catalog needs roughly **28 hours of wall-clock** to warm 10,120 targets at the observed rate (~6 entries/min; ~1–2 s/entry of batch time, measured from `reply` 902 ms, `remote-m3` 1,057 ms, `meshcore-mobile` 2,082 ms). It has never once had a window that long. It is not warming slowly — it is being reset 7–10 times a day and starting over.

## The hard part: invalidation

A disk cache is only safe if the key covers **everything that can change the pixels**. The current key — `ServeOverrides.cacheKey(previewId, overrides)` — identifies *what was asked for*, not *what produced it*. That's sufficient in memory, where one process holds exactly one generation, and insufficient the moment an entry outlives the thing that rendered it.

You cannot enumerate every input with confidence. So the design should be: key on the coarsest cheap thing that provably covers content and renderer, then **verify a sample against ground truth on load** so an input nobody thought of is caught rather than served.

### Fingerprint the generation, not the entry

```
<cacheRoot>/<system>/<fingerprint>/<sha256(cacheKey)>.png
```

where `fingerprint = sha256(deliveryCommit ‖ composeAiToolsVersion ‖ backend ‖ renderConfigDigest)`:

- **`deliveryCommit`** — the delivery-branch commit this load actually read. Already resolved at `ServeCatalogStore.kt:332`, no new work and no jar hashing. A regeneration is always a new commit, so this covers catalog content, component code, theme definitions and resources exactly. For source-built and locally-bundled catalogs, where there is no commit, fall back to a digest of the bundle jar (hash it while staging, since the store is already streaming those bytes).
- **`composeAiToolsVersion`** — covers renderer changes. Also serves as the proxy for the container image, and therefore for the JVM, Skia and the installed fonts, because the image is rebuilt per release. That is an *assumption worth writing down in the code*, not a proof: a base-image bump without a release would slip through it.
- **`backend`** — desktop / Android-Robolectric, from the daemon descriptor. Different renderers, different pixels.
- **`renderConfigDigest`** — the server-side render defaults that are *not* in `cacheKey`: density/scale, default device, default font scale, image encoding. These are the easiest thing to forget, precisely because they never appear in the key today.

The property that matters: the fingerprint is **derived from inputs, not asserted**. Nobody has to remember to bump a cache version when they change the renderer. A changed input yields a different directory, which is simply never read, and swept later.

### Sample-verify on load, because the enumeration will be wrong eventually

On adopting a cache generation, re-render ~5 random cached entries and compare bytes. On any mismatch, drop the whole generation and log loudly. Cost is a few seconds per catalog per startup against rebuilding 10,120 entries — and it converts "we forgot an input" from *silently serving wrong pixels* into a self-healing cache miss.

This matters more here than in a typical build cache. A stale build artifact gets caught by a test; a stale rendered preview is shown to an agent as ground truth, which is the exact failure this repo exists to prevent.

### Also needed

- **Sweeping.** A byte cap over the cache root and a sweeper that deletes fingerprint directories no loaded catalog references. Without it, ~7–10 generations a day accumulate on disk.
- **Observability.** Record the fingerprint inputs in a manifest beside each generation and surface them on `/status`, so a drop is explainable — "renderer 1.13.0 → 1.14.0 invalidated 8,412 entries" — rather than another unexplained return to zero.
- **Write-through, not write-back.** Persist on `put` so a crash keeps whatever was earned; the current failure mode is losing everything.

## Test plan

- fingerprint changes with each input in turn (commit, version, backend, render config) and is stable when none change;
- a cache generation written by one process is adopted by the next with the same fingerprint;
- a regeneration under the same server version adopts nothing and starts clean;
- a seeded byte mismatch in the verify sample drops the generation;
- the sweeper removes unreferenced generations and honours the byte cap.

## Context

Admission starvation is a separate, more urgent regression — filed as #4129. Fixing that gets `m3-catalog` warming again; this issue is what lets the work survive to completion.

yschimke:
Closes #4130.

## What was broken

The theme cache was process memory, and **two** separate events dropped it: a server restart, and a catalog reload — every load builds a fresh `ServeSessionState` and therefore a fresh `CatalogThemeCache`. For `m3-catalog` on the public box:

| day | regenerations | releases | resets |
|---|---|---|---|
| Aug 15 | 7 | 3 | 10 |
| Aug 16 | 5 | 4 | 9 |
| Aug 17 | 3 | 4 | 7 |

Against a catalog needing roughly **28 hours** of lane time to warm its 10,120 targets. It had never once had a window long enough to finish, on any day — which is what a `cached` count that keeps returning to zero actually means. #4143 got it warming again; this is what lets the work accumulate.

## Invalidation — the hard part

A cache key says **what was asked for**. A persisted entry outlives what rendered it, so it needs the other half: **what produced it**. You cannot enumerate every input with confidence, so the design is not "list the inputs and hope":

**Key on the coarsest cheap thing that provably covers content and renderer, then verify a sample against ground truth on load** so an unenumerated input surfaces as a cache miss rather than a wrong pixel.

`ThemeCacheFingerprint` covers:

- **The classpath's bytes**, not its paths — a load stages the same bundle into a fresh directory every time, so paths churn while content does not. Hashing the jars covers the catalog's code, theme definitions, resources *and* dependencies in one move. This is the one input derived from the thing itself rather than asserted about it.
- **The daemon variant** — desktop and Android/Robolectric read the same classpath and don't agree pixel-for-pixel.
- **The tool version** — covers the renderer, and stands proxy for the container image, hence JVM, Skia, fonts. **That proxying is an assumption, not a proof**, and it's the one a base-image bump without a release would defeat. It's recorded in the generation manifest, and the sample verification is what actually catches it.
- **The render config** — the server-side defaults that never appear in a cache key, and are therefore easiest to forget precisely *because* they're absent from it.

Nothing is asserted by hand: change the renderer and the version moves, change the catalog and the classpath moves. There is no cache-version constant to remember to bump, because a constant someone must remember is one that will eventually be wrong. An identity that can't be established returns null and **declines to persist** — every wrong pixel this cache could serve begins with two generations agreeing on a name.

Each `(system, fingerprint)` is its own directory, so invalidation is structural: a new revision or build writes a new generation and reads none of the old one.

**The verification.** The first optimizer pass re-renders a small sample and compares. A mismatch discards the *whole* generation, not the offending entry — a missed input makes every entry under that name suspect. It renders through `live` rather than `renderPrefetch`, because the prefetch path consults this very cache and the comparison would pass by construction; only a `DAEMON`-generation render counts as evidence, and a daemon that can't answer yet verifies nothing rather than wiping a warmed catalog.

## Two tiers, not a copy

Memory stays a 128 MB window onto disk rather than a preload — a fully warmed m3-catalog is several times that, so preloading would thrash the LRU and report `cached` as whatever happened to fit. Lookups fall through to disk and promote; every count asks both tiers. That also fixes a related trap: an eviction now only un-completes a catalog when the entry is gone from *both* tiers, or a fully warmed catalog would sit at `paused` forever and the optimizer would re-render what was already on disk.

## Sweeping

Dead generations are reclaimed once the catalog pass finishes — the only moment the live set is known; sweeping earlier would delete a generation a catalog further down the list was about to adopt. A live set over `--theme-cache-max-bytes` is **reported, never evicted**: deleting what is being warmed to fit a cap turns the cap into a treadmill where the optimizer re-renders exactly what the sweep discarded, forever, and the box looks busy while making no progress. That's a configuration answer, not something to paper over.

## Deployment

No temp-directory fallback, deliberately: a theme cache in `/tmp` is written once, read never, and thrown away with the container — disk and render time for nothing, while reporting itself as working. Where there's no durable location the server runs memory-only and logs it.

`deploy/vps/docker-compose.yml` gains `theme_cache:/config/theme-cache`, alongside the existing `catalog_feed_cache` volume — named volumes survive the container recreation a `docker compose pull` deploy performs, which is the entire premise. Both entrypoints pass `SERVE_THEME_CACHE_DIR` / `SERVE_THEME_CACHE_MAX_BYTES`. The default location is beside `catalogs.json`, matching the feed cache.

`/status.json` gains `themeCache` — generations, bytes, hits, writes — so "is warming accumulating?" is answerable directly instead of inferred from a count that keeps resetting.

## Test plan

`./gradlew :cli:test` green (2,450 tests), `./gradlew ktfmtCheckAll` clean, both entrypoints pass `bash -n`, the compose file parses.

New `ThemeCachePersistenceTest` — most of it asserts *when a generation must not be reused*, since that's where the risk is:

- the same bytes staged in a different directory are the same generation (the property the whole design rests on — a path-based fingerprint would call every load new and buy nothing);
- a changed jar, a changed version, a changed variant and a changed render config are each a different generation;
- classpath order doesn't invent one;
- exploded class directories are hashed by content, not skipped (a from-source catalog puts one on the classpath);
- an unreadable classpath declines to name the generation at all;
- a render written by one process is read by the next; a different fingerprint starts clean; two catalogs with identical keys don't share;
- sweeping reclaims dead generations, keeps live ones, and reports rather than evicts an over-cap live set;
- a name that could escape the store root is refused, not sanitised;
- a target evicted from memory still counts as cached while on disk;
- verification drops the generation on mismatch, keeps it on match, and does nothing when the renderer returns null.

Two flags-registry gates (`CliFlagsRegistryTest`) caught the new flags being unregistered; both are now in `CliFlags.VALUE_FLAGS` and the serve allowlist.

Text-only evidence: this changes background caching, a status block and deployment config — no rendered surface.

## What this does not do

The disk tier only helps where the directory is durable. On the public box that means the new volume has to be present — until the compose change is applied there, the server will log `theme cache disabled` or run memory-only and behave exactly as it does today.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VTL61d6Gj1HAXJ4c1Kfpgn)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/CliFlagValidation.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/CliFlagValidation.kt
@@ -136,6 +136,8 @@ internal object CliFlagValidation {
             "--bundles",
             "--catalog-branch-prefix",
             "--catalog-feed-cache",
+            "--theme-cache-dir",
+            "--theme-cache-max-bytes",
             "--catalog-feed-idle-timeout",
             "--catalog-max-images",
             "--catalog-refresh-interval",
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/CliFlags.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/CliFlags.kt
@@ -107,6 +107,8 @@ internal object CliFlags {
       "--catalog-max-images",
       "--catalog-feed-idle-timeout",
       "--catalog-feed-cache",
+      "--theme-cache-dir",
+      "--theme-cache-max-bytes",
       "--catalog-source-root",
       "--wasm-dir",
       "--rc-player-wasm-dir",
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt
@@ -868,6 +868,9 @@ class ServeCommand(args: List<String>, private val browseProject: Boolean = fals
               ThemeCacheFingerprint.renderedClasspath(
                 launch.classpath,
                 launch.systemProperties,
+                // Same preview manifest, reachable by a second route on descriptors that name it
+                // here rather than in a system property.
+                extraPayloads = listOfNotNull(launch.manifestPath),
               ),
             variant = launch.variant,
             toolVersion = BUNDLE_VERSION,
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/ServeCommand.kt
@@ -5,6 +5,7 @@ import ee.schimke.composeai.cli.serve.CatalogLoadTracker
 import ee.schimke.composeai.cli.serve.CatalogRefreshResult
 import ee.schimke.composeai.cli.serve.CatalogThemeCache
 import ee.schimke.composeai.cli.serve.DaemonStartupLog
+import ee.schimke.composeai.cli.serve.GenerationInputs
 import ee.schimke.composeai.cli.serve.GitWorktrees
 import ee.schimke.composeai.cli.serve.GradleRevisionBuilder
 import ee.schimke.composeai.cli.serve.LiveSeatLimiter
@@ -68,6 +69,8 @@ import ee.schimke.composeai.cli.serve.ServeTrustAdmin
 import ee.schimke.composeai.cli.serve.ServeTrustStoreFile
 import ee.schimke.composeai.cli.serve.ServeUrls
 import ee.schimke.composeai.cli.serve.ServeWeb
+import ee.schimke.composeai.cli.serve.ThemeCacheFingerprint
+import ee.schimke.composeai.cli.serve.ThemeCacheStore
 import ee.schimke.composeai.cli.serve.TrustStore
 import ee.schimke.composeai.cli.serve.declaredThemesFromPreviews
 import ee.schimke.composeai.cli.serve.detectedFeaturesOf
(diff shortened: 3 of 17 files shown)


yschimke:
We use docker image, not vps. Update docs to clarify that

yschimke:
Fixed in `07a6c33e6` — and it was more than a docs error, so thank you for catching it.

I'd put the durable volume and the opt-in env var in `deploy/vps/docker-compose.yml`, and the new docs section named that file as where the cache lives. Since `preview.coo.ee` runs `deploy/image`, the feature was unreachable on the box it was written for: `deploy/image` had no volume for it, and I'd just defaulted it to `none`. So this PR would have shipped a theme cache your deployment could not turn on.

What changed:

- **`deploy/image/docker-compose.yml`** now mounts a dedicated `preview_theme_cache` volume at `/theme-cache`. Deliberately not on `preview_config` — that volume holds `catalogs.json` and `producers.json`, and an up-to-8-GB render cache has no business crowding them. The mount is unconditional so enabling it is one line in `.env`:

  ```
  SERVE_THEME_CACHE_DIR=/theme-cache
  ```

  While it stays `none` the volume is simply empty.

- **The docs** now say which profile is which, that unset does *not* mean off (an unset value derives a directory beside `catalogs.json`, i.e. onto `preview_config` — which is why `none` exists as an explicit sentinel), and how the rolling update's overlapping replicas relate to the sweeper's grace window.

The `deploy/vps` additions stay as they are — still correct for that profile, just not yours.

Two things still open on this PR that are yours to call rather than mine:

1. **Whether the tier ships off.** It currently defaults to `none` everywhere. Given the review history below — eight rounds, each finding at least one render input the fingerprint did not read, and rounds 4–9 finding defects in the previous round's fix — I'd keep it opt-in for the first release and enable it on `preview.coo.ee` explicitly once it has had real exposure. Flipping it on is the one `.env` line above.
2. **Two known-open review findings** I stopped rather than fixed: verification calls `live.render` directly, which returns `NotFound` for supplement-module previews in multi-module catalogs (so those generations stay quarantined for the host's life), and sweeps only run at startup and post-publication, so a generation spared by the grace window is never reclaimed if the catalog then stays stable. Both are recorded in the review threads.

---

yschimke:
And add some basic hit rate tracking so we know if it's working or just cause disk io for nothing.

Or maybe one part of the key stops it working?

yschimke:
Added in `207cacac9`. Both halves of your question now have a number on `/status.json`.

**"Is it working, or just disk I/O for nothing?"** — the per-catalog `renderCache` row gains `memoryHits`, `diskHits`, `misses` and `hitRate`, plus `withheld` for reads the adoption quarantine deliberately refused. `withheld` is kept out of `misses` *and* out of the rate on purpose: during a cold start it can outnumber real misses, so folding it in would report the cache as failing at precisely the moment it is being careful.

**"Or maybe one part of the key stops it working?"** — that failure mode was invisible, and it is the one I'd bet on. Three things now expose it:

- `renderCache.persisted.fingerprint` — the generation directory this catalog reads and writes. Differs across two restarts that changed nothing ⇒ the key moved.
- `renderCache.persisted.adopted` — renders already on disk when this process opened the generation. This is the only evidence anything survived a restart. `adopted: 0` with `writes` climbing on every boot *is* the write-amplification case, in two numbers.
- `themeCache.generationsBySystem` — a catalog the box has only ever served one way should have **one** generation on disk. A count that climbs per restart means some input the digest reads is unstable, and every write is being left for a sweep to reclaim.

One related gap I fixed while in there: every path where a catalog silently falls back to memory-only — unreadable launch descriptor, a classpath entry the fingerprint couldn't digest, a generation directory that couldn't be opened — was indistinguishable from running the server without `--theme-cache-dir` at all, and each is silent and permanent for the life of the host. They now name themselves in `renderCache.persistenceOff`.

Docs: new "Telling a working cache from write amplification" subsection in `docs/public-preview-server.md` with a table of what a bad value means for each field. Tests: five new cases (tier-split hit counting, withheld-is-not-a-miss, adopted across a restart vs. a moved fingerprint, the named fallback reason, and the per-system census). `:cli:test` and `ktfmtCheckAll` green locally.

Unchanged from my last comment — still yours to call, not mine: whether the tier ships off by default, and the two known-open review findings I stopped rather than fixed (verification calling `live.render` directly returns `NotFound` for supplement-module previews in multi-module catalogs; sweeps only run at startup and post-publication, so a generation spared by the grace window is never reclaimed if the catalog then stays stable). Codex has since posted a ninth round; I haven't acted on it.

---

Issue #4068, opened 2026-08-16
publish: ship the iOS player as an XCFramework consumable from Swift


yschimke:
Part of publishing the Remote Compose player stack. **Gated behind the API work** — #4056 and the parameter workstreams. Independent of the Maven issues (#4063–#4066).

## Today

`:rc-player-compose` builds a static framework per iOS target (`baseName = "RcComposePlayer"`), and `RcComposeViewController.kt` is a purpose-built UIKit entry point returning a `UIViewController`. So the iOS story is *designed* — it just has no delivery mechanism. The repo contains no XCFramework assembly task, no `Package.swift`, and no podspec.

A Swift consumer today would have to add this Gradle build to their project and depend on the KMP module directly.

## What

1. An XCFramework task bundling `iosArm64` + `iosSimulatorArm64`. Note `iosX64` does not exist for this module — CMP 1.11 dropped the Intel simulator variant (see the module's build comment and #4066), so the XCFramework covers device and Apple-silicon simulator only. Intel Macs are out, and that should be stated rather than discovered.
2. SPM and/or CocoaPods distribution — decide which. SPM needs a `Package.swift` referencing a binary target plus somewhere to host the zip with a checksum; GitHub Releases is the obvious candidate given the existing release chain.
3. CI coverage. `ci.yml` currently compiles the iOS *test* targets (`:rc-player-compose:compileTestKotlinIosArm64` and friends) but never links or packages a framework, so framework assembly would be newly exercised — and the CI comment at `ci.yml:377` notes `linkDebugTestIosSimulatorArm64` has died before. Expect this step to need real work.

## Surface question

`RcComposeViewController` takes `theme: Int` and `fontFamilies: Map<String, RcFontFaces>` today; #4058 and #4060 change both. A Kotlin enum and a Kotlin interface each cross the Swift boundary differently from an `Int` and a `Map` — worth checking how `RcPlayerTheme` and `RcTypefaceLoader` actually read from Swift *before* those land, since it may be the one place where the nicer Kotlin API is the worse Swift one. If so, say so on #4058/#4060 while they are still open.

Also: `forwardIosPlayerEvent` exists as an `internal` seam, and `RcPlayerEvent` is a sealed interface — sealed hierarchies flatten unhelpfully in Objective-C interop. Check what a Swift consumer actually sees in an `onEvent` callback.

Issue #4066, opened 2026-08-16
publish: `:rc-player-compose` to Maven Central


yschimke:
Part of publishing the Remote Compose player stack. **Gated behind the API work** — #4056, #4058, #4059, #4060, #4062. Depends on #4065 (`:rc-player-runtime`).

Artifact: `ee.schimke.composeai:rc-player-compose`. The one most consumers actually want — it carries `RcComposePlayer`, and it is the module whose entire public surface #4056 is rewriting. Land it last.

## Known blockers specific to this module

**1. Its target set does not match its siblings, and that is visible to consumers.** Per the module's own build comment, there is no `iosX64()` here: CMP 1.11 dropped the Intel-iOS-simulator variant, while `trace`/`protocol`/`runtime` still publish it. So a consumer on `iosX64` resolves three of the four artifacts and fails on the fourth. Decide before release whether to drop `iosX64` from the whole stack for consistency or document the asymmetry — silently publishing a stack that only partly resolves on one target is the worst of the three options.

**2. Compose Multiplatform version in the POM.** This is the only module in the stack depending on CMP, and the compose-compiler/runtime pairing is a hard constraint for consumers. Check what the published POM records and whether it belongs in the compatibility matrix at `docs/RENDERER_COMPATIBILITY.md` alongside the existing AGP × Kotlin × Compose rows.

**3. `jvm("desktop")` → unnamed `jvm()`** — see #4063. Higher impact here: `desktopTest` is the whole render-test suite.

## Sequencing

Every parameter of this module's entry point is being retyped by #4058, #4059 and #4060, and #4061 may still change the `RcTypefaceLoader.Default` contract afterwards. Publishing before #4061 settles means either shipping a `Default` that resolves nothing and changing it later, or holding the first release for it. Prefer holding — a default that starts resolving fonts it previously didn't is a behaviour change consumers cannot see coming from a version bump.

`RcComposeViewController` (iosMain) is published as part of this artifact but is not usable from Swift without #4068 — worth a KDoc note pointing at it.

yschimke:
Landed on `main` in ff7c303b2 (#4189), which published all four `:rc-player-*` library modules together. GitHub only auto-closed #4063 — the PR body wrote `Closes #4063, #4064, #4065, #4066`, and the keyword only binds to the first number.

What this issue specifically asked for:

[...]

Issue #4061, opened 2026-08-16
feat(rc-player): ship a real default RcTypefaceLoader and move manifest loading into shared code


yschimke:
Follow-up to #4060, which introduces `RcTypefaceLoader` with `Default` temporarily equal to `Empty`. This issue decides what `Default` actually is, and moves the one real loader implementation out of the Wasm host. Not on the critical path for the API work.

## Why

After #4060 the interface exists but the only implementation that does anything is still 60 lines inside one host: `loadHostFontFamilies()` at `rc-player/wasm/src/wasmJsMain/…/Main.kt:200` fetches a `fonts.json` manifest, groups faces by family, lowercases keys, caches per base URL, and separately registers the `default`-role family under the literal key `"default"`.

That last part is a protocol fact — what a document means when it names no family — living in a browser fetch helper. iOS has no equivalent, so an iOS host rendering the `remote-m3` catalog gets Compose's built-in face for all body text. The `google:Roboto Flex` name being unresolvable is the same class of bug, and per the comment at `:229` it has already been hit and fixed once, in this one host.

## What

- `RcManifestTypefaceLoader` in shared code: a suspend factory taking a base URL/path and a byte-fetching function, returning a synchronous `RcTypefaceLoader`. The Wasm host keeps only its `fetchBytes`; the parsing, grouping, aliasing and caching move.
- Keep the per-base cache (`Main.kt:249`) — it is what stops a document swap re-fetching a whole catalog's fonts.
- Then decide `RcTypefaceLoader.Default`.

## The three questions to settle first

**1. What is `Default` per platform?** Desktop and iOS could expose system fonts; a browser has no system font access at all. `expect`/`actual` per target is the obvious mechanism, but "Default resolves nothing on Wasm and system fonts elsewhere" is a surprising contract. The alternative — `Default` stays `Empty` everywhere and hosts always supply a loader — is more honest and may be the right answer. **Settle this before writing code**; it determines whether this issue is a refactor or a design change.

**2. Can a lazy loader still enumerate `families`?** #4060 keeps `families: Set<String>` on the interface so `composeSupportReport` can gate renderability before showing a document. A manifest loader satisfies that — it has parsed the manifest. A hypothetical system-font loader may not. If `Default` needs lazy resolution, the pre-render check needs rethinking, and that is a bigger change than it looks: today an unsupplied family fails loudly at load rather than silently rendering in a fallback.

[...]

Issue #4056, opened 2026-08-16
Umbrella: the RcComposePlayer public surface — one signature, better types


yschimke:
**Superseded in approach.** This was originally "split into a simple and an advanced overload". That was the wrong fix: the problem was never the parameter *count* — `Text` has fifteen defaulted parameters and reads fine — it was that four of the types were wrong. Two overloads would have papered over that by hiding the bad types in the advanced one.

**This issue no longer carries code.** It is the agreed target signature that #4058, #4059 and #4060 each implement one parameter of. They are now unblocked and fully parallel — nothing has to land first.

## Target

```kotlin
@Composable
public fun RcComposePlayer(
  document: RcDocument,
  modifier: Modifier = Modifier,
  theme: RcPlayerTheme = RcPlayerTheme.System,
  namedValues: SnapshotStateMap<String, RcNamedValue> = rememberRcNamedValues(),
  onEvent: (RcPlayerEvent) -> Unit = {},
  typefaces: RcTypefaceLoader = RcTypefaceLoader.Default,
  systemColors: (name: String) -> Color? = { null },
)
```

Plus the existing `ByteArray` convenience overload, unchanged in role.

The simple call stays `RcComposePlayer(document)`. Every parameter is now a domain concept with a working default, rather than a wire value the caller has to look up.

## Against today

| Parameter | Today | Target | Issue |
|---|---|---|---|
| `theme` | `Int` (`-2` is dark) | `RcPlayerTheme` enum, 3 cases | #4058 |
| `systemColorLookup` | `(String) -> Int?` ARGB | `systemColors: (String) -> Color?` | #4058 |
| `namedValues` | `Map<…>`, in the `remember` key | `SnapshotStateMap<…>`, reactive | #4059 |
| `fontFamilies` | `Map<String, RcFontFaces>` | `RcTypefaceLoader` + default | #4060 |

## Why these are still parallel

Each changes one parameter of one function. The PRs touch adjacent lines and will conflict trivially; that is cheaper than serialising three workstreams behind a skeleton PR that does nothing on its own.

## Notes

Nothing in `rc-player/*` is published today, so there is no source- or binary-compatibility burden — change the signature outright rather than deprecating. Publishing is tracked separately (#4063–#4068) and is deliberately gated behind this work.

yschimke:
All three parameter workstreams have landed, so the target signature is in place on `main`:

```kotlin
@Composable
public fun RcComposePlayer(
  document: RcDocument,
[...]

PR #4178, opened 2026-08-18
feat(rc-player): replace the host font map with an RcTypefaceLoader interface

### LINKED ISSUE #4060: feat(rc-player): replace the host font map with an `RcTypefaceLoader` interface
Implements one parameter of the target signature in #4056. Parallel with #4058 and #4059 — nothing blocks this.

Reshaped: this was "move `fontFamilies` to an advanced overload". With one signature (#4056) there is no advanced overload to move it to, and the map was the wrong shape anyway.

## Why an interface rather than a map

`Map<String, RcFontFaces>` makes every host reimplement the same lookup rules, and the rules are *protocol* facts, not host preferences. They currently sit in `resolveFontFamily` (`RcComposePlayer.kt:3188`) and in one host's fetch helper:

- a document naming no family asks for the literal key `"default"` (`:3197`);
- named families arrive lowercased, and `google:`-prefixed names have the prefix stripped (`:3202`);
- generic families (`sans-serif`, `serif`, `monospace`) try the host first, then fall back to Compose's built-ins (`:3198`–`:3200`).

A second host either reimplements all of that or renders body text in the wrong face. Per the comment at `Main.kt:229`, that has already happened once for the `"default"` alias.

## Shape

The good news is the map is used in exactly two places, so the interface is small:

```kotlin
public interface RcTypefaceLoader {
  /** Families this loader can resolve. Gates renderability before a document is shown. */
  public val families: Set<String>

  /** The family at [settings], or null. Called during composition and draw — must not block. */
  public fun typeface(family: String, settings: FontVariation.Settings? = null): FontFamily?

  public companion object {
    /** Resolves nothing; documents render in Compose's built-in faces. */
    public val Empty: RcTypefaceLoader
    /** Platform default — see #4061. */
    public val Default: RcTypefaceLoader
  }
}
```

`typeface(...)` is exactly today's `namedFonts[name]?.family(settings)` at `:3195`, and `families` is exactly today's `fontFamilies.keys`.

**Synchronous by design.** `resolveFontFamily` is called during composition and draw (`:956`, `:1020`), so the interface cannot suspend. Async work stays in *construction* — which is already how it works: `loadHostFontFamilies()` is a suspend function that produces the map before `LoadState.Ready`. Keep that split.

## Scope for this issue

Behaviour-identical and mechanical:

- Add the interface plus `RcBundledTypefaceLoader(faces: Map<String, RcFontFaces>)` wrapping today's map, so existing hosts keep working by wrapping what they already build.
- `RcTypefaceLoader.Default` is `Empty` for now — deciding what a real platform default is belongs to #4061 and is a genuine design question, not a detail.
- Swap the parameter and the `LocalRcNamedFonts` CompositionLocal (`:1132`) to carry the loader.
- Move the `"default"` / `google:` / generic-family rules into the loader contract or a shared helper, so they are stated once rather than per host.

Preserve `RcFontFaces`'s per-axis instance cache (`RcHostFonts.kt:44`). A document draws the same family at the same axes every frame; re-parsing per frame is visible on a text-heavy watch face. `RcFontFace`/`RcFontFaces` stay public and otherwise unchanged — the keep-the-bytes design is load-bearing for variation axes and is documented in place.

## What this must not break

`composeSupportReport(…, availableFontFamilies = …)` is how the Wasm host (`Main.kt:107`) and the iOS entry point (`RcComposeViewController.kt:23`) decide a document is renderable *before* showing it. A document naming an unavailable family currently fails loudly at load. `families` on the interface exists to keep that check working — if a later loader cannot enumerate up front, that is a design change to argue on #4061, not something to quietly drop here.

## Call sites

- `rc-player/wasm/src/wasmJsMain/…/Main.kt:134` — the real user; wraps its map in `RcBundledTypefaceLoader` for now.
- `rc-player/compose/src/iosMain/…/RcComposeViewController.kt:17`.
- `rc-player/compose/src/desktopTest/…/RcFontAxisRenderTest.kt:90`.

## Test plan

- `RcFontAxisRenderTest` is the sharpest net: it asserts ink width *differs* across `wght` axis values, so a broken loader or a lost instance cache shows up as equal widths rather than a compile error.
- `./gradlew :rc-player-compose:allTests`, `:rc-player-wasm:wasmPlayerDist`, `:rc-player-compose:compileTestKotlinIosArm64`.
- `serve-lanes-e2e` against the real `remote-m3` `fonts.json`.
- Visual: `text-serif` and a variable-axis preview from `compose-m3` through the Wasm lane, before/after. A regression here is a fallback face — it passes tests and looks wrong.

yschimke:
Closes #4060. Implements one parameter of the #4056 target signature.

**Stacked on #4176** (`RcPlayerTheme` / `Color` lookups), which is itself stacked on #4173. Bases retarget automatically as each merges.

## Why an interface rather than a map

`Map<String, RcFontFaces>` made every host reimplement the same lookup rules, and those rules are *protocol* facts rather than host preferences — a document naming no family asks for the literal key `"default"`, names arrive lowercased, a `google:` prefix is a source marker and is stripped, generic families try the host before Compose's built-ins. A second host either reimplements all of it or renders body text in the wrong face, which per the comment at the Wasm host's manifest loader has already happened once.

They now live in `rcResolveTypeface`, stated once, and a loader answers only what it is the authority on:

```kotlin
public interface RcTypefaceLoader {
  public val families: Set<String>
  public fun typeface(family: String, settings: FontVariation.Settings? = null): FontFamily?
  public companion object {
    public val Empty: RcTypefaceLoader
    public val Default: RcTypefaceLoader  // == Empty; #4061 decides
  }
}
```

`families` stays on the interface so `composeSupportReport` keeps gating renderability before a document is shown — a document naming an unavailable family still fails loudly at load rather than rendering silently in a fallback. Whether a lazy loader can satisfy that is #4061's question, not something dropped here.

**Synchronous by design.** `typeface` runs during composition and draw. Async work stays in *construction*, which is already how every host works: the Wasm host still fetches and decodes its manifest, then wraps it in `RcBundledTypefaceLoader`.

## Scope kept to what the issue asked

Behaviour-identical and mechanical. `RcFontFace` / `RcFontFaces` are untouched, per-axis instance cache included. The generic-family fallback chain stays in the player rather than moving onto the loader — #4061 Q3 decides whether a host should be able to override it, and pre-empting that here would be a behaviour change disguised as a refactor.

## Visual evidence

The failure mode here is a *fallback face*: it compiles, it passes, and it looks wrong. So the check is pixels against the repo's vendored Roboto Flex, rendered on the CMP desktop lane on this branch and on its base branch with an otherwise identical harness. **Every pair is byte-identical.**

| | before (`fontFamilies: Map`) | after (`typefaces: RcTypefaceLoader`) |
|---|---|---|
| `google:Roboto Flex` @ `wdth 25` | wdth25 before | wdth25 after |
| `google:Roboto Flex` @ `wdth 151` | wdth151 before | wdth151 after |
| no family named → `"default"` key | default before | default after |
| name the host cannot supply | unresolvable before | unresolvable after |

The two `wdth` rows are visibly different widths from each other, which is what proves the axis instance still reaches the font engine through the new interface — a loader that lost the instance cache or the settings would render them identically and still pass a compile.

## Test plan

- `RcFontAxisRenderTest` — the sharpest existing net, asserting ink width *differs* across `wdth` values. Migrated to the loader and still green, so the instance cache survived.
- New `RcTypefaceLoaderTest` (7 cases) in `commonTest` — the `"default"` key, the `google:` strip, the generic chain in both directions (host absent → built-in, host present → host wins), embedded-font precedence, `RcBundledTypefaceLoader` key lowercasing, and `Default` resolving nothing. **This is the first time these rules are asserted on a non-Wasm target**, which is the point of moving them.
- `./gradlew :rc-player-compose:allTests :rc-player-compose:checkKotlinAbi :rc-player-wasm:wasmPlayerDist ktfmtCheckAll` — green; distribution 23,567,624 / 23,780,000 bytes.
- `compileTestKotlinIosArm64`, `compileTestKotlinIosSimulatorArm64` — green; the iOS entry point takes the loader and reads `typefaces.families` for its support report.

## Note for #4061 and #4068

- `RcTypefaceLoader.Default` being `Empty` is load-bearing for #4066's sequencing: a default that starts resolving fonts it previously did not is a behaviour change a consumer cannot see coming from a version bump, so #4061 should settle it before the first release rather than after.
- Across the Swift boundary a Kotlin interface with a defaulted parameter exports as a protocol plus a synthetic `typeface(family:settings:)`, so a Swift host implements it rather than building a dictionary — worth confirming against #4068's concern that the nicer Kotlin API is sometimes the worse Swift one.


yschimke:
fix merge conflicts

Issue #4058, opened 2026-08-16
feat(rc-player): type the theme and color surface — `RcPlayerTheme` enum and `Color` lookups


yschimke:
Implements two parameters of the target signature in #4056. Parallel with #4059 and #4060 — nothing blocks this.

Both changes are the same defect on the same code path: raw wire ints in the public API, on the theme/`ColorTheme` resolution path. `RcPlayerState.applyColorTheme(colorTheme, requestedTheme)` consumes both together, so they are one review.

## 1. `theme: Int` → `RcPlayerTheme`

A caller has to know that `-2` is dark and `-3` is light, and nothing rejects `7`. The constants live on the protocol *operation* (`rc-player/protocol/…/RcModel.kt:352`):

```kotlin
public data class RcTheme(val theme: Int) : RcOperation {
  public companion object {
    public const val SYSTEM: Int = 0
    public const val UNSPECIFIED: Int = -1
    public const val DARK: Int = -2
    public const val LIGHT: Int = -3
  }
}
```

**The name `RcTheme` is taken** and must stay taken — it is a document operation the decoder produces. `RcPlayerTheme` in `…rcplayer.compose` is the natural home; this is a host concept, not a wire one.

**Three cases, not four.** `SYSTEM` and `UNSPECIFIED` differ on the wire but are identical at the host boundary — `rcResolveSystemTheme` (`RcSystemTheme.kt:27`) maps both to `isSystemInDarkTheme()`. Exposing both asks callers to choose between two spellings of one behaviour.

```kotlin
public enum class RcPlayerTheme { Light, Dark, System }
```

One thing to verify rather than assume: `RcComposeViewController` defaults to `UNSPECIFIED` while the Compose entry defaults to `SYSTEM`. After this both are `System`. `RcSystemTheme.kt`'s doc comment says they already resolve identically — confirm with a render test, don't take the comment's word for it.

## 2. `systemColorLookup: (String) -> Int?` → `systemColors: (String) -> Color?`

Same raw-ARGB problem. The conversion belongs at the compose-module boundary: `RcPlayerState` (`rc-player/runtime/…/RcPlayerState.kt:94`) keeps its `Int`, because the runtime module has no Compose UI dependency and those are wire values there. That boundary is correct and should not move.

[...]

PR #4181, opened 2026-08-18
fix(rc-player): named-value changes rebuild RcPlayerState and discard animation and touch state

### LINKED ISSUE #4059: fix(rc-player): named-value changes rebuild RcPlayerState and discard animation and touch state
Blocked on #4056 (overload skeleton). Parallel with the theme-enum and host-hook workstreams.

**This is a behaviour bug as well as an API change** — it is the substantial one of the three parallel workstreams.

## The bug

`rc-player/compose/src/commonMain/kotlin/ee/schimke/composeai/rcplayer/compose/RcComposePlayer.kt:331`:

```kotlin
val state =
  remember(document, namedValues) {
    RcPlayerState(document, namedValues, …)
  }
```

`namedValues` is a `Map<String, RcNamedValue>` key of the `remember`. Any change to it — including a caller rebuilding an equal map in a parent recomposition — constructs a **fresh `RcPlayerState`**, throwing away:

- running `RcFloatExpression` animation timelines and their phase,
- `RcTouchExpressionRuntime` gesture state mid-drag,
- every variable a host action already mutated,
- marquee/impulse timeline position.

So the one API a host uses to drive a live document is also the one that resets the document. Any host doing the obvious thing — a slider bound to a named float — restarts every animation in the document on each frame of the drag.

Note the care already taken immediately below, at `:354`, to keep `systemColorLookup` *out* of the `remember` key for exactly this reason, with a comment explaining that rebuilding the state "discard[s] variables an action changed, touch-expression state and running animation timelines". `namedValues` has the same hazard and was not given the same treatment.

## The mechanism already exists

`RcPlayerState.setNamedValue(name, value)` (`rc-player/runtime/…/RcPlayerState.kt:1057`) applies a single named value incrementally against `variableNames`, with type checking against the AndroidX variable type. Nothing on the public path calls it.

## What

Make named values a snapshot-backed, caller-owned holder, and key `remember` on the document alone:

```kotlin
@Composable
public fun rememberRcNamedValues(
  vararg initial: Pair<String, RcNamedValue>,
): SnapshotStateMap<String, RcNamedValue>

@Composable
public fun RcComposePlayer(
  document: RcDocument,
  modifier: Modifier = Modifier,
  theme: RcPlayerTheme = RcPlayerTheme.System,
  namedValues: SnapshotStateMap<String, RcNamedValue> = rememberRcNamedValues(),
  onEvent: (RcPlayerEvent) -> Unit = {},
)
```

- `remember(document)` — drop `namedValues` from the key.
- Bridge changes into the existing `setNamedValue` via `snapshotFlow`/`LaunchedEffect` so a mutation updates the live state rather than replacing it.
- Seed the initial values through the `RcPlayerState` constructor as today.

Decide during implementation how a *removal* from the map behaves — `setNamedValue` has no inverse, and "revert to the document's recorded value" is the only sensible reading. If that turns out to need runtime support, say so on the issue rather than silently treating removal as a no-op.

Also worth folding in while here: `RcNamedValue.Color(argb: Int)` (`RcPlayerState.kt:1142`) has the same raw-int problem as the theme constants. The runtime type should keep its `Int`, but the compose module can offer a `Color`-taking factory. Coordinate with the host-hook workstream, which is doing the same conversion for `systemColorLookup`, so the two agree on where the conversion helper lives.

## Test plan

- New test: mutate a named value mid-animation and assert the timeline phase is continuous — this is the regression the current code fails. It should fail before the fix.
- New test: mutate a named value and assert `RcPlayerState` identity is unchanged across the mutation.
- `./gradlew :rc-player-compose:allTests`.
- `rc-player/wasm/…/Main.kt:132` passes `state.namedValues` from a `LoadState.Ready` data class — migrate it and confirm `?value=` URL params still drive the document.
- Visual: the `serve` state-toggle lane (`docs/design/SERVE_STATE_TOGGLE_SPEC.md`) is the user-visible surface for this; capture before/after of a toggle driving an animated document.

yschimke:
Closes #4059. The substantial one of the three parallel workstreams — a behaviour bug as well as an API change.

**Stacked on #4178 → #4176 → #4173.** Bases retarget automatically as each merges.

## The bug

`namedValues` was a key of the player's `remember`, so *any* change to it — including a caller rebuilding an equal map in a parent recomposition — constructed a fresh `RcPlayerState` and discarded running `RcFloatExpression` timelines, `RcTouchExpressionRuntime` gesture state mid-drag, marquee position, and every variable a document action had already mutated.

So the one API a host uses to drive a live document was also the one that reset it. The care taken twelve lines below — keeping `systemColorLookup` out of the `remember` key, with a comment naming this exact hazard — was never extended here.

## What changed

- `namedValues: SnapshotStateMap<String, RcNamedValue> = rememberRcNamedValues()`, caller-owned.
- `remember(document)` — `namedValues` dropped from the key.
- A `snapshotFlow` bridge applies each change through `RcPlayerState.setNamedValue`, the incremental type-checked mechanism that already existed and had no caller on the public path.

Errors are deliberately **not** swallowed in the bridge: an unknown name or a type mismatch threw from the `RcPlayerState` constructor before this change, and a host naming a variable the document does not declare should still hear about it rather than watch a value quietly not apply.

## Removal semantics — it did need runtime support

The issue asked to decide this and to say so rather than silently no-op. "Revert to the document's recorded value" is the only sensible reading, and it needs the pre-override value, which by the time a removal arrives has been overwritten. So `RcPlayerState` captures what the *document* recorded for each named variable at construction — **before** the host's seed is applied, so removing a seeded value falls back past it rather than to it — and `clearNamedValue` restores it. `namedValue(name)` is the read-back counterpart.

**Text is the exception, and is handled as one.** `beginFrame` rebuilds `texts` from `baseTexts` and re-applies `textOverrides` on top, every frame. Writing the old string back as a "restore" would install a fresh override holding the document's text, and the document could never change it again — so clearing a text *removes* the override. `RcNamedValueLifecycleTest.clearingATextRemovesTheOverrideRatherThanWritingTheOldStringBack` steps a frame specifically to catch that.

## `RcNamedValue.Color`

Folded in as the issue suggested: `rcNamedColor(Color)` and `RcNamedValue.Color.color` in the compose module, going through the same `RcColorConversion` helper #4058 introduced for `systemColors`. The runtime type keeps its `Int`.

## Visual evidence — this is the regression

A 300 ms layout animation, started by a click, while a host writes a **new** named-float value every frame (a slider bound to a named value, the case the issue is about). The value drives nothing the animation reads. Frames 75 ms apart.

| t | before | after |
|---|---|---|
| 75 ms |  |  |
| 150 ms |  |  |
| 225 ms |  |  |
| 300 ms |  |  |
| 375 ms |  |  |

Before, the red panel never grows — each write rebuilt the state, resetting the float the click had set and discarding the timeline, so the animation restarted from zero every frame and never arrived. After, it animates to completion while the same writes land. Only the first pair matches, because at 75 ms the two have not yet had time to diverge.

## Test plan

- New `RcNamedValueRenderTest` (2 cases, desktop render):
  - `changingANamedValueMidAnimationLeavesTheTimelineRunning` — the regression. **Verified it fails before the fix**: restoring `remember(document, namedValues.toMap())` makes exactly this test fail, and nothing else.
  - `settingAndThenRemovingANamedValueDrivesTheLiveDocumentBothWays` — a host write reaches the live document, and removing it brings the document's recorded value back, both observed as ink.
- New `RcNamedValueLifecycleTest` (5 cases, `:rc-player-runtime` `commonTest`) — clear-restores-recorded, clear-falls-back-*past*-a-seeded-value, the text-override removal above, colour restore, idempotence, and that an unknown name still fails loudly in both `clearNamedValue` and `namedValue`.
- `./gradlew :rc-player-compose:allTests :rc-player-runtime:allTests` — green, including the existing exact-pixel render corpus.
- `checkKotlinAbi` on both modules, `compileTestKotlinIosArm64`, `ktfmtCheckAll` — green.
- `:rc-player-wasm:wasmPlayerDist` — 23,582,992 / 23,780,000 bytes. The host now seeds a `SnapshotStateMap` keyed on the document, so a swap starts from `?namedValues=` again rather than carrying the previous document's overrides into one that may not declare them; `?value=` behaviour is otherwise unchanged.


yschimke:
fix merge conflicts

yschimke:
fix merge conflicts

PR #4195, opened 2026-08-18
feat(design-catalog): unlist compose-m3 / wear-m3 and scope both to pipeline features

yschimke:
## What

`compose-m3` and `wear-m3` are compose-ai-tools' **own preview harness**, not design-system references — [m3-catalog](https://github.com/yschimke/m3-catalog) is the exhaustive Material 3 one and `wear-m3-catalog` will be the Wear one. So:

1. Both are **`listed: false`** on preview.coo.ee — still reachable at `/compose-m3/` and `/wear-m3/` as a live smoke target, off the front-page index (the same treatment `cadence` already gets).
2. With that framing the inventory bar becomes **exhaustive by feature, not by component**: one or two carriers per pipeline capability, and nothing that only re-proves a capability another sticker already carries.

|  | before | after |
| --- | --- | --- |
| `compose-m3` | 27 components / 86 captures | **13 / 52** |
| `wear-m3` | 26 components / 58 captures | **22 / 49** |

## Visual evidence

Every **before** capture, outlined green where it survives and red where it's removed. Nothing that survives moved — this change only deletes entries, so each kept sticker is byte-identical to what it rendered before.

### compose-m3 — cut hard

Five button emphasis levels collapse to one (which also hosts the pressed / focused / disabled / icon-label / font-scale captures), three plain cards collapse to the slotted one (the only card carrying `PreviewSlot`), and the four boolean selection controls collapse to the checkbox and the switch.

compose-m3 sticker sheet, before captures outlined green where kept and red where removed

### wear-m3 — cut narrowly, deliberately

`samples/design-catalog-remote-m3` declares `compareWith: "wear-m3"` and authors a `parallel` into **19** of these ids. An unresolved parallel still builds — it just renders an unpaired row (a `console.warn`) — so deleting the by-component redundancy here would *silently* degrade the published remote-m3 cross-system compare page. What went is the four entries nothing pairs against and no feature needs: `Layout/List`, `Template/EdgeButton` (a second `@ScrollingPreview(END)` × breakpoint capture — compare it to `EdgeButtonSticker` in the sheet), `CheckboxButton/Checked` (a second `@OverrideVariant` boolean) and `ListHeader`.

All 19 parallels still resolve against the new inventory, verified before committing. The rest waits for remote-m3's `compareWith` to move to `wear-m3-catalog`.

wear-m3 sticker sheet, before captures outlined green where kept and red where removed

The sheets and their rationale are committed under [`renders/catalog-feature-scoping/`](https://github.com/yschimke/compose-ai-tools/tree/e561776ec83a710e76fd27a679f1ec528a99e530/renders/catalog-feature-scoping), and the scoping policy is written up in [DESIGN_CATALOGS.md § What belongs in an in-repo catalog](https://github.com/yschimke/compose-ai-tools/blob/e561776ec83a710e76fd27a679f1ec528a99e530/docs/design/DESIGN_CATALOGS.md).

## Also in here

- **Latent test bug fixed.** `ServeBundleDaemonTest`'s slug-sharing SVG pairs named `*Off` / `*Disabled` / `*Unchecked` wrapper functions that were deleted when those states moved onto `@OverrideVariant` — so every pair `continue`d past its assertion. They're now the real `<fn>_VARIANT_<name>` capture ids.
- Unused string resources pruned: 12 × 18 locales on the m3 side, 3 × 18 on Wear.

## ⚠️ Operator follow-up after merge

`publish-config-to-box.sh` reconciles **additively** — a system already registered on the box comes back 409 — so flipping `listed` in `catalogs.json` does **not** unlist a live entry on its own. After this merges:

```
DELETE /admin/catalogs/compose-m3
DELETE /admin/catalogs/wear-m3
```

…and let the next `publish-preview-config` run re-add them from the committed file with `listed: false`. This is noted inline in `catalogs.json` too.

## Test plan

- `./gradlew ktfmtCheckAll` ✅
- `:samples:design-catalog-m3-shared:desktopTest` ✅ (interactivity tests reworked: the removed `switch-off` / `checkbox-unchecked` slugs are now seeded through `PreviewOverrideController`, the way the renderer does)
- `:samples:design-catalog-wear-m3:testDebugUnitTest` ✅ (incl. `WearFocusedPressPixelTest`)
- `:cli:test --tests '*CatalogUsageRulesTest*' --tests '*PlaygroundSourceCleanerTest*'` ✅ — the playground source-cleaner still expands `CatalogButtons.kt#FilledButton` and `CatalogSelection.kt#CheckboxChecked`
- `:samples:design-catalog-m3:composePreviewRenderAll` → 52 previews ✅; `:samples:cmp-wasm-catalog:compileKotlinWasmJs` ✅
- `validate-catalog-spec.mjs` on both specs: `OK — 0 warning(s), 0 errors` ✅


yschimke:
Fix failures

PR #4197, opened 2026-08-18
fix(release): publish Apple targets from macOS, and make the Swift tag resolvable

yschimke:
Codex reviewed #4189, #4193 and #4194 as each merged, so every finding below landed on `main` unaddressed. This clears the release-infrastructure ones.

## Release blockers

**Apple targets were being published from Linux.** `:rc-player-{trace,protocol,runtime,compose}` apply `composeai.maven-publishing` and declare `iosArm64` + `iosSimulatorArm64`, but the root `publishAndReleaseToMavenCentral` aggregate runs on `ubuntu-latest`, where Kotlin/Native has no Apple SDK or linker. The next release would have failed while staging them, or shipped the stack without the iOS artifacts it advertises. Those four are now excluded from the Linux aggregate and published by `publish-player-stack` (release) / `publish-player-snapshot` (snapshot) on `macos-15`. The exclusion lists and the job task lists are the same four modules, and both files say so.

The snapshot version arithmetic moved out of the workflow into `scripts/snapshot-version.sh` so the two snapshot jobs cannot compute different coordinates and publish half a stack under a version nothing resolves. It carries a self-test, wired into CI beside the `Package.swift` one.

**The Swift tag was not a version SwiftPM could see.** `swift/<version>` is not a semantic-version ref — SwiftPM strips a leading `v` and nothing else — so `.package(url: ..., from: "1.15.1")`, the only syntax `Package.swift` and RC_PLAYER_SWIFT.md advertise, resolved no versions at all. The tag is now a bare `<version>`: resolvable, distinct from the `v<version>` release tag, and not matched by `release.yml`'s own `v*` push trigger.

**The XCFramework was built from `main`, not the tag.** `main` can advance between the tag being cut and `publish-xcframework` starting, so the framework attached to — and labelled with — a release could be newer source than the Maven artifacts and source release that `publish-gradle-plugin` builds from the same tag. It now checks out `TAG_NAME` and takes a fresh `main` only for the `Package.swift` commit that has to land there.

**The documented re-release path aborted itself.** `workflow_dispatch` on an existing tag finds the work already done — the zip is reproducible, so the rewriter computes the same URL and checksum — and `update-package-swift.sh` returned 1 on `before == after`, killing the job under `set -e`. A no-op rewrite is now success (with its own self-test case), the commit is skipped on a clean tree, and an existing tag at the right commit is accepted. A tag pointing somewhere *else* is still a hard error: that is exactly the asset/tag disagreement this scheme exists to prevent.

**`publish-wasm-player` could not upload its asset, and its fallback never fired.** The job held `contents: read` while ending in `gh release upload`, and — being `continue-on-error` — would have swallowed the authorization failure and shipped a release without `rc-player-wasm.zip`. The fallback upload also inherited the implicit `success()` step condition, so an npm outage, the one thing it exists to cover, was precisely when it was skipped. Now `contents: write` and `if: ${{ !cancelled() }}`, guarded on the bundle actually having been staged.

## Docs corrected against the code

- **`namedValues` is a JSON array, not NUL-delimited rows.** `flattenNamedValuesFromLocation` runs `JSON.parse` and wants `{kind, name, value}` objects; the NUL rows are the internal shape used to cross the JS/Wasm boundary afterwards. A host following the old table hit the caught parse error and silently got no overrides. RC_PLAYER_EMBED.md now shows the real shape, with a worked example and the silent-drop semantics written down.
- **The Swift sample did not compile.** `RcComposeViewController` is a top-level Kotlin function, so it exports onto the `RcComposeViewControllerKt` file facade with all five arguments required, and `KotlinByteArray` has no `Data` initializer. Both are now written out as read from the generated Objective-C header (built locally via `linkReleaseFrameworkIosSimulatorArm64`), including the `Int8(bitPattern:)` copy. The KDoc on the entry point also no longer claims the framework is unpublished.
- **Two stale `iosX64` claims** in RC_CMP_WASM_PLAYER.md contradicted its own target table.

## Test plan

- `scripts/update-package-swift.sh --self-test` — passes, including the new no-op-rewrite case.
- `scripts/snapshot-version.sh --self-test` — passes (main, branch-suffix, explicit suffix, pre-release tag, and the empty-after-sanitization error).
- `./gradlew :rc-player-compose:ktfmtCheck` — clean.
- `./gradlew :rc-player-compose:linkReleaseFrameworkIosSimulatorArm64` — built locally; the Swift signatures in the docs are read out of the resulting header, not guessed.
- Both edited workflows parse as YAML; job graphs are `publish-player-stack` alongside the existing release jobs and `publish-player-snapshot` alongside `publish-snapshot`.

The macOS publish lanes themselves can only be exercised by a real release — they reuse the runner, the Xcode selection step, and the Gradle invocation shape that `publish-xcframework` already proves on `macos-15`.

Non-visual: workflows, shell scripts, and documentation.


yschimke:
Fix failures

PR #4199, opened 2026-08-18
fix(deploy,serve): stop reporting success over surviving replicas, held locks and stale caches

yschimke:
Codex reviewed #4168 as it merged, so all four findings landed on `main` unaddressed. They are the same shape: a failure path that reports **success** and leaves behind exactly the state the code exists to prevent — and none of them shows up in a log.

## Old replicas could survive a "successful" rollout (P1)

`docker-rollout` ends with `docker stop $OLD_CONTAINER_IDS || true` and `docker rm ... || true`. `|| true` is right for the benign race — a container already exited and reaped reports failure — but it swallowed *every* failure, so old replicas still standing were reported as a completed rollout. That restores the doubled replica count and memory pressure the rolling swap exists to avoid, on a box already tight enough to need `mem_limit`.

The `|| true` stays; after it, every old id is inspected and a survivor is a hard error naming the containers.

## An interrupted runner released its lock and kept rolling (P1)

`trap release_rollout_lock EXIT INT TERM` — but a POSIX shell **resumes** after a trap handler returns unless the handler exits. On `INT`/`TERM` the runner therefore removed the lock marker and carried straight on with the rollout, so another runner could acquire the lock and overlap the remaining work. `EXIT` keeps the plain cleanup; the signals get handlers that clean up and `exit 128+signal`, so a supervisor can tell an interrupted rollout from a failed one.

## Admission measured the host's memory, not the container's (P1)

`LinuxHostResourceSampler` read `/proc/meminfo`, which inside a container reports the **host's** memory. `deploy/vps/docker-compose.yml` caps preview at 3 GiB and `deploy/oracle` at 6 GiB on far larger hosts, so a replica sitting just under its own OOM limit still read as having ~75% available — and the pressure gate went on admitting optimizer work at precisely the moment it needed to stop.

It now also reads the cgroup and reports the **smaller** of the two headrooms:

- cgroup v2 `memory.max`/`memory.current`, and v1 `memory.limit_in_bytes`/`memory.usage_in_bytes`, because the deployment targets do not agree on a generation.
- `inactive_file` is subtracted from usage: that is page cache the kernel reclaims under pressure, not memory this process cannot give back. Counting it would make a container that has merely read a lot of files look permanently full and the gate would then refuse optimizer work forever — the opposite failure, and a quieter one.
- An unlimited cgroup — v2's `max`, v1's near-`Long.MAX_VALUE` sentinel — contributes nothing, so a bare-metal host reads exactly as it did before.

## A contended discard was still trusted (P2)

`CatalogThemeCache.verifySample` calls `store.discard()` when a cached render no longer matches, then sets `persistenceTrusted`. It ignored the return value. `discard` takes the generation's write lock with `tryLock`, so a concurrent foreground render made it return `false` — leaving the PNGs *just proved wrong* on disk while the cache was marked trusted, ready to serve them.

`discard` now retries the try-lock for ~1s (the lock is held for one PNG write), and trust follows the discard rather than the detection. If it still fails, `persistenceTrusted` stays false — which the read path already implements as exactly the right quarantine: adopted entries are withheld, `contains` still reports them so the optimizer does not re-render, and the next verification pass tries again.

## Test plan

- `deploy/image/test-rollout-cleanup.sh` — new, 7 assertions, runs offline against a stub `docker`. Covers both shell fixes: the already-gone race staying a success, a survivor failing, a signalled runner exiting rather than resuming (and its 143 status), plus greps pinning that neither script drifts back. Wired into `ci.yml` beside the other `deploy/image` self-tests.
- `./gradlew :cli:test --tests '*HostOptimizerAdmission*' --tests '*ThemeCache*'` — green, including four new sampler cases driving temp `proc`/cgroup trees: host-vs-cgroup minimum, page-cache discount, unlimited v2 and v1, and an honoured v1 limit.
- `./gradlew :cli:ktfmtCheck` — clean.
- `sh -n` on both shell scripts.

Non-visual: deployment scripts and server-side admission/caching.


yschimke:
Fix failure

PR #4212, opened 2026-08-19
fix(rc-player): follow a replacement named-value holder, and seed its bookkeeping from the state

yschimke:
Two of the four P2s Codex raised on [#4181](https://github.com/yschimke/compose-ai-tools/pull/4181), which merged before the review landed. A third is documented rather than fixed; the fourth I could not reproduce and did not ship a change for.

## A replacement holder was ignored, and the detached one kept driving

The collector reads `namedValues` inside `snapshotFlow`, so a `LaunchedEffect` keyed only on `state` stayed bound to whichever map instance it started with. A parent that swapped in a different `SnapshotStateMap` while keeping the same document got the worst of both: the replacement's values — current and future — were ignored entirely, while later mutations of the *detached old map* went on driving the player.

Keying on the holder as well re-binds the flow without rebuilding `RcPlayerState`, so running animation timelines, mid-drag touch state and action-changed variables all survive the swap. That is the whole reason the holder exists, and rebuilding the state to follow it would have given the swap the exact cost the holder was introduced to remove.

## Bookkeeping seeded from the wrong map

`applied` was initialised inside the effect from `namedValues`, not from the map `RcPlayerState` was actually built with. When those differ — a sibling `SideEffect` or a concurrent host update landing between composition and the effect starting — `applied` holds the newer map while the state holds the older snapshot, the first emission compares equal and is skipped, and the player stays stale until that entry happens to change again.

The seed is now captured beside the state in the same `remember`, and the applied-set lives in a `remember` that survives the effect restarting. That second part matters for the fix above: re-seeding on every restart would leave a key applied through the *old* holder uncleared, because the diff would no longer know it had ever been set. The test walks exactly that path.

## Documented, not fixed: read-back

`rememberRcNamedValues`'s KDoc said ownership is the caller's because "a host usually wants to read a value back (a document action can change one)". The bridge cannot deliver that: `snapshotFlow` copies edits *in*, and `executeActionsUnchecked` mutates the state's own maps without touching this one. So a host reads back its own last override, or nothing — never the value an action moved it to.

I did not add reverse sync. It is a genuine design question, not an oversight: writing back into a caller-owned map would mutate host state from inside a draw pass, and a host driving a value from its own model — a slider, a form field — would find the player fighting it. The KDoc now states the gap and points at `onEvent`, which already reports the actions themselves.

## Not reproduced: legacy-branch semantics

Codex's remaining finding is that `invalidationVersion` is read during composition only inside the `layout != null` branch, so on the legacy path a named-value change behind `RcRootContentDescription` would redraw pixels while leaving the semantics label stale.

The reasoning is sound but I could not make it fail. I built a canvas-only document with a named root description and drove it three ways: with the clock free-running the label updates; with the clock pinned and no frame advanced, nothing updates *with or without* the fix (no recomposition is applied at all); with a frame advanced after the mutation it updates either way — a frame tick recomposes the player on that path regardless.

So the hoist changed nothing observable, and the test I wrote for it passed against unfixed code. I reverted both rather than ship an unfalsifiable change with a test that gives false confidence. If someone can construct a case where the recomposition genuinely does not happen, the one-line hoist above `semanticsModifier` is the fix.

## Test plan

- `./gradlew :rc-player-compose:jvmTest` — green, with a new `replacingTheNamedValueHolderRebindsTheCollector` covering the swap, the detached map going quiet, and a removal through the replacement still restoring the document's own value.
- Confirmed that test fails against the unfixed player (`3 tests completed, 1 failed`).
- `./gradlew :rc-player-compose:ktfmtCheck :rc-player-compose:checkKotlinAbi` — clean; no ABI change.

Non-visual: the rendered output is unchanged for a host that never swaps its holder. The swap case is asserted as pixels in the new test.

### DIFF
--- a/rc-player/compose/src/commonMain/kotlin/ee/schimke/composeai/rcplayer/compose/RcComposePlayer.kt
+++ b/rc-player/compose/src/commonMain/kotlin/ee/schimke/composeai/rcplayer/compose/RcComposePlayer.kt
@@ -334,38 +334,48 @@ private fun RcComposePlayerResolved(
   // had changed — the same hazard the `systemColors` comment below describes, on the one API a host
   // uses to drive a live document. Changes are applied incrementally instead; see the
   // `LaunchedEffect` under this block.
-  val state =
+  // The seed is captured BESIDE the state, not recomputed. The collector below has to diff its
+  // first emission against the map the state was actually built from: if `namedValues` changes
+  // between this composition and the effect starting — a sibling `SideEffect`, a concurrent host
+  // update — reading it again there would initialise the bookkeeping from the NEWER map while the
+  // state still held the older snapshot, so the first emission would compare equal and be skipped,
+  // leaving the player stale until that entry happened to change again.
+  val (seededNamedValues, state) =
     remember(document) {
-      RcPlayerState(
-        document,
-        // Seeded once, from whatever the map holds when this document is first composed.
-        namedValues.toMap(),
-        eventSink = { latestEventSink(it) },
-        onInvalidated = { invalidationVersion += 1 },
-        effectSink = { effect ->
-          when (effect) {
-            is RcPlayerEffect.HapticFeedback ->
-              latestHapticFeedback.performAndroidXHaptic(effect.type)
-            is RcPlayerEffect.WakeIn -> {
-              val current = wakeIntervalSeconds
-              if (!effect.seconds.isNaN() && (current == null || effect.seconds < current)) {
-                wakeIntervalSeconds = effect.seconds
+      val seed = namedValues.toMap()
+      seed to
+        RcPlayerState(
+          document,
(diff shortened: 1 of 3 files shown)


yschimke:
fix merge conflicts

yschimke:
Superseded by #4213, which landed the same three changes while this was in review — the holder-identity key with `appliedNamedValues` outliving the restart, the seed captured from the map the constructor received, and the `RcNamedValues` KDoc corrected to state the bridge is one-way.

#4213 also carries the fourth finding I could not reproduce and therefore reverted: it hoists the `invalidationVersion` read above the semantics modifier so a legacy canvas document rebuilds its root description. My reasoning for dropping it was that a frame tick recomposes that path anyway — worth knowing that the change landed regardless, so if the gap is real it is now closed either way.

Closing rather than rebasing; nothing here is left to add.

Issue #4202, opened 2026-08-19
compose-preview: no way to settle a time-driven reveal before a still capture


yschimke:
A component whose content animates in from a `LaunchedEffect` captures as its **first frame**, and there is no annotation that means "settle, then capture".

Concrete case, from [yschimke/wear-m3-catalog](https://github.com/yschimke/wear-m3-catalog): Wear's `ConfirmationDialogContent` starts its children at `alpha = 0` and animates them in after a delay —

```kotlin
val alphaAnimatable = remember { Animatable(0f) }
LaunchedEffect(Unit) {
    animatedDelay(DurationShort2.toLong(), reduceMotionEnabled)
    alphaAnimatable.animateTo(1f, textOpacityAnimationSpec)
}
```

— so the published sticker is an empty container ring: no icon, no curved text. Swapping in a plain `Icon` changes nothing, which rules out the icon composable as the cause.

**What doesn't solve it today**

- `@AnimatedPreview` produces a motion artifact *beside* the same unsettled still, so the component's own card still shows the empty frame.
- `@ScrollingPreview(END)` settles post-scroll animations, which is the closest existing precedent — it's exactly this shape of problem for the Wear edge button, and it works. There's just no equivalent for "this reveal is driven by time, not by a gesture".

**What would**

The capture model already carries `advanceTimeMillis` (it's in `previews.json`), so this looks like a missing annotation surface rather than a missing capability — something like `@SettledPreview(afterMs = 600)`, or an `advanceMs` on the existing capture annotations.

Catalogs hit this wherever a design system animates a reveal: confirmation and success dialogs, snackbars, expanding sheets. Meanwhile the affected set is recorded as an exclusion in that repo's coverage file with this reason, so it comes back when the knob exists.

yschimke:
Second reproduction, from [yschimke/m3-catalog](https://github.com/yschimke/m3-catalog) — same shape, different design system, and it suggests the affected set is wider than reveal animations.

Material 3's `DateInputTextField` starts its field **empty** and writes the initial date in a `LaunchedEffect`:

```kotlin
var text by rememberSaveable(stateSaver = TextFieldValue.Saver) { mutableStateOf(TextFieldValue()) }
...
LaunchedEffect(initialDateMillis) { initialDateMillis?.let { text = TextFieldValue(...) } }
```

[...]

PR #4246, opened 2026-08-19
fix(preview): honour an exact settle coordinate, and widen the motion collision

yschimke:
Follow-up to #4242 (issue #4202), which merged while these were still in validation. Five Codex findings on that PR, all reproduced in the code before fixing.

## An exact `afterMs` was not exact, on three paths

`@SettledPreview(afterMs = N)` names an instant — the same promise `advanceTimeMillis` makes. Three places moved the shutter off it:

* **Android still** — the job still had `advanceTimeMillis == null`, so it was classified for the adaptive pixel-quiescence probe. That probe always spends at least one more 16ms frame and up to the sample budget, so an "exact 350ms" capture was actually written from somewhere in **366–414ms**. `shouldAdvanceClockForVisualSettling` now takes `hasExactSettle` and declines, exactly as it declines for an explicit `advanceTimeMillis`.
* **Desktop focus/hover** — `renderFocusPreview` advanced its two setup frames and *then* the whole window on top, landing at **N+32**. It now receives a coordinate rather than a window, and spends only the remainder; the setup frames count toward it, mirroring `CAPTURE_ADVANCE_MS` on the Android side.
* (The `+32` on the plain Android path was fixed in #4242 itself.)

Auto mode is untouched in all three: it names a bound, not a moment, so it keeps the probe and keeps walking its full window.

## Desktop auto settle could stop one frame early

`settleScene`'s quiescence check consulted the delay queue and scene invalidations, but not runnables already dispatched. `scene.render(...)` can queue a newly-introduced `LaunchedEffect` *after* the preceding `advanceTo` drained — at that instant nothing is scheduled and nothing is invalidated, so the walk returned on the pre-reveal frame. The render's own dispatches are now drained and counted (`hasPendingWork`).

## The motion collision guard was too narrow

#4242 dropped the settle when `@AnimatedPreview` shared the function, because both products replay from one composition against one paused clock and virtual time does not rewind. `@InteractionPreview` and `@FocusedPreview(gif = true)` collide identically — they replay from the same clock and are deliberately sorted after the stills for the same reason. All three now drop the settle and name themselves in the warning.

## A settled still was costed as a one-pass capture

Both backends walk the settle window a frame at a time, so a default 1000ms settle is ~62 rendered frames — but `captureCost` still stamped `STATIC_COST`. That let a catalog of settled previews stay in the on-every-save fast tier and concentrate into a single Android shard. `settleCaptureCost(windowMs)` scales with the window, putting the default just above `HEAVY_COST_THRESHOLD` while a short explicit window stays cheap.

## Not fixed, deliberately

**Preserving `Capture`'s pre-settle `copy` / `copy$default` descriptors.** The descriptor analysis is right, but `copy$default` is compiler-synthesised from the primary constructor — it cannot be hand-written or overloaded from Kotlin source, and a `fun copy(...)` of the old arity does not recreate the descriptor existing binaries call. The only real remedies are dropping `data` from `Capture` or accepting the change. `preview-data-api` already accepts exactly this break for `CatalogEntry` (pre-`kitValue`) and `PreviewResult` (pre-`projectDirectory`) — both preserved constructors only, and the module contains zero `copy` overloads. Changing that is a module-wide policy call, worth its own PR alongside `binary-compatibility-validator`, which the repo does not currently run.

**Consumer docs in [`yschimke/skills`](https://github.com/yschimke/skills)** (`skills/compose-preview/references/capture-modes.md`) still need `@SettledPreview` and its motion-product incompatibility added. That repo is not reachable from this session, so it remains a follow-up — same note as on #4242.

## Test plan

Run in this session on the cloud container (Linux, headless), green:

| Command | Result |
| --- | --- |
| `:gradle-plugin:test` | 459 tests, 0 failures |
| `:gradle-plugin:functionalTest` | 95 tests, 0 failures |
| `:renderer-android:testDebugUnitTest` | 252 tests, 0 failures — includes the new `an exact settle is a snapshot and skips the quiescence probe` |
| `:renderer-desktop:test` | 155 tests, 0 failures — includes the new `work dispatched but not yet run counts against quiescence` |
| `:daemon:core:test` | 552 tests, 0 failures |
| `:preview-data-api:test` | 30 tests, 0 failures |
| `ktfmtCheckAll` | clean |

**No render moves.** `:samples:android:composePreviewRenderAll` re-rendered all four committed settled fixtures **byte-identical** to what #4242 landed — including `DeferredValueSettledPreview`, which is the `afterMs = 300` case the probe-skip touches. The exact-mode fixes change where the shutter lands for coordinates the probe was overshooting, not the sample's resting frames, so this is a text-only change by the repo's visual-evidence rule.

---
_Generated by [Claude Code](https://claude.ai/code/session_011BnQSKS218vZqT8GtpfvS1)_

### DIFF
--- a/api/preview-annotations/src/commonMain/kotlin/ee/schimke/composeai/preview/SettledPreview.kt
+++ b/api/preview-annotations/src/commonMain/kotlin/ee/schimke/composeai/preview/SettledPreview.kt
@@ -45,8 +45,9 @@ package ee.schimke.composeai.preview
  * Only still captures are settled — a `@ScrollingPreview` LONG/GIF product, an `@AnimatedPreview`
  * GIF and an `@InteractionPreview` recording all drive the clock themselves and are left alone.
  *
- * Pairing this with `@AnimatedPreview` on the **same function** is reported and ignored, rather
- * than half-honoured. Both products render from one composition against one paused clock and want
+ * Pairing this with a motion product on the **same function** — `@AnimatedPreview`,
+ * `@InteractionPreview`, or `@FocusedPreview(gif = true)` — is reported and ignored, rather than
+ * half-honoured. Both products render from one composition against one paused clock and want
  * opposite things from it: the GIF needs the timeline from its start, the settled still needs a
  * coordinate near the end, and virtual time does not rewind. Put them on separate preview functions
  * — each then owns its own composition — and both get what they asked for. Serving both from one
--- a/docs/HOW_IT_WORKS.md
+++ b/docs/HOW_IT_WORKS.md
@@ -89,7 +89,8 @@ It applies to **still** captures only — a scroll drive runs its own settle, an
 themselves, and an explicit `advanceTimeMillis` is a snapshot of a coordinate
 the author chose.
 
-Pairing it with `@AnimatedPreview` on the *same function* is reported as a
+Pairing it with a motion product on the *same function* — `@AnimatedPreview`,
+`@InteractionPreview`, or `@FocusedPreview(gif = true)` — is reported as a
 discovery warning and the settle is dropped, rather than half-honoured. Both
 products render from one composition against one paused clock and want opposite
 things from it: the GIF needs the timeline from its start, the settled still
--- a/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewData.kt
+++ b/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/PreviewData.kt
@@ -673,6 +673,20 @@ const val SCROLL_GIF_COST: Float = 40.0f
 const val FOCUS_GIF_COST: Float = 40.0f
 const val ANIMATION_COST: Float = 50.0f
 const val INTERACTION_COST: Float = 60.0f
+
+/**
+ * Extra cost per second of `@SettledPreview` window, on top of [STATIC_COST].
+ *
+ * A settle is not a one-pass capture: both backends walk the window a frame at a time, so a default
+ * 1000ms settle is ~62 rendered frames. Sized so that default lands at 6.0 — just above
+ * [HEAVY_COST_THRESHOLD], which keeps it out of the on-every-save fast tier and lets shard balancing
+ * see the work — while a short explicit window (say `afterMs = 300`, ~19 frames) stays cheap.
+ */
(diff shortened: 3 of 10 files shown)


yschimke:
Three findings this round. I'm not pushing for any of them — reasoning per item, and a note on the pattern at the end.

**P2 — frame rounding vs. clock bookkeeping (Android).** Correct, but **not introduced here, and not specific to the settle**. `advanceTimeBy` rounds up to the frame duration while `currentTime` records the unrounded target — and that is the same path `@RoboComposePreviewOptions(advanceTimeMillis = …)` has always used. The repo already ships non-aligned values through it: `SharedElementFilmstripPreview` at 600 (`% 16 = 8`), `SpinnerTimelinePreview` at 500 and 1500 (`4`, `12`). So exact settle inherits precisely the semantics `advanceTimeMillis` has, which is what it set out to do; "exact" has always meant "the first frame at or after N" on this backend.

Worth fixing, but as its own change: normalizing the bookkeeping is safe and moves no pixels, while making the final advance unrounded **moves bytes on every published non-aligned render above** and needs a baseline pass. Filed as #4247 with both options separated.

**P2 — desktop focus and `afterMs` in 1..31.** Also correct, and also #4247. The two setup frames are load-bearing rather than incidental: they let first composition and layout settle *before* the focus walk, and the walk needs a laid-out tree to find anything focusable. Bounding them by a sub-frame coordinate would trade a precise timestamp for a capture with nothing focused in it. The right resolution is probably to clamp-and-warn at discovery or document the floor — a design call, not a patch, and no sample exercises the combination.

**P1 — consumer docs in `yschimke/skills`.** Standing answer, unchanged from #4242: that repo is not reachable from this session, so the update cannot be made here. It is called out in this PR's body and in #4242's. It does need doing — `capture-modes.md` needs `@SettledPreview` plus the full motion-product incompatibility list — but by someone with access to that repo.

---

On the pattern: this is the fifth review round across #4242 and #4246, and every round has found something in the code the previous round added. Rounds 1–4 were unambiguously worth it — they caught a real ABI break, a `require` that would have thrown, a settle that never reached the desktop focus path, and one round where my own fix was the bug. This round is different in kind: two sub-frame precision arguments about a clock that only ticks at 16ms, one of which describes behaviour the repo has shipped for years. That reads like diminishing returns rather than an unfinished job, so I've stopped pushing and written the remainder down in #4247 instead.

---

Issue #4247, opened 2026-08-19
compose-preview: "exact" capture coordinates are frame-rounded, and the clock marker records the unrounded value


yschimke:
Split out of #4246 (Codex review), which is where the two symptoms below were reported. Neither is introduced by that PR — the first is a long-standing property of the exact-snapshot path, the second is a corner of the focus renderer. Both are about the same thing: what "exact" can actually mean on a clock that only ticks every 16ms.

## 1. Frame rounding vs. clock bookkeeping (Android)

`RobolectricRenderTest` advances with `rule.mainClock.advanceTimeBy(delta)`, which aligns to the frame duration and rounds **up**, then records `currentTime = target` — the *unrounded* value. So an exact coordinate that is not a multiple of 16ms captures one frame late while the local marker says otherwise, and the discrepancy accumulates across a preview's jobs.

This is **not** specific to `@SettledPreview(afterMs = …)`. It is the same code path `@RoboComposePreviewOptions(manualClockOptions = [ManualClockOptions(advanceTimeMillis = …)])` has always used, and the repo ships plenty of non-aligned values through it:

| Sample | Value | `% 16` |
| --- | --- | --- |
| `SharedElementFilmstripPreview` | 600 | 8 |
| `SpinnerTimelinePreview` | 500 / 1500 | 4 / 12 |
| `samples/wear` gesture stickers | 800 | 0… (800 % 16 = 0) |

So `@SettledPreview`'s exact mode inherits exactly the semantics `advanceTimeMillis` already has, which is the intent — but "exact" is really "the first frame at or after N".

Two ways out, and they are not equivalent:

* **Normalize the bookkeeping** — set `currentTime` to what the clock actually reached rather than the requested target. Removes the drift; does not change any captured pixel, because the clock already lands where it lands.
* **Make the final advance unrounded** (`advanceTimeBy(..., ignoreFrameDuration = true)` or an explicit frame at the coordinate). Genuinely lands on N — and **moves bytes on every already-published non-aligned render** above, so it needs its own baseline pass.

The first is safe and probably wanted regardless. The second is the behaviour change and should be a deliberate call.

## 2. Desktop focus captures cannot honour `afterMs` below 32ms

[...]

PR #4248, opened 2026-08-19
fix(preview): record the clock position the render loop actually reached

yschimke:
Part 1 of #4247.

`MainTestClock.advanceTimeBy` aligns to the frame duration and rounds **up** — it steps whole 16ms frames until it has covered at least the requested amount. So the render loop works in two spaces that do not coincide: the **requested** coordinate a capture asks for, and the **physical** clock, which lands on the next frame boundary at or after it.

It was measuring each hop as the difference between two *requested* coordinates while spending it on the *physical* clock, so the rounding excess was paid again at every hop instead of being absorbed, compounding across a fan-out. The loop now keeps the two apart: the marker stays in requested space, while each advance spends only what the physical clock still owes.

## The one render that moves

`SpinnerTimelinePreview` is the only sample that shows it, and it needs all three of its jobs — `advanceTimeMillis = 0, 500, 1500`, none frame-aligned:

| job | requested | physical clock, before | physical clock, after |
| --- | --- | --- | --- |
| 1 | 0 | 0 | 0 |
| 2 | 500 | 512 | 512 |
| 3 | 1500 | 512 + 1008 = **1520** | 512 + 992 = **1504** |

The requested marker reads 0 / 500 / 1500 throughout, before and after — it is the *physical* column that changes. The third capture was landing 20ms past the coordinate it asked for and now lands 4ms past, the closest a 16ms frame clock can sit to 1500.

| Before — hop measured in requested space | After — hop measured against the clock |
| --- | --- |
| spinner at 1520ms | spinner at 1504ms |

The visible difference is the indeterminate arc one frame further along.

## Two wrong turns on the way here, both caught

Worth recording, because the final shape is a reaction to both:

**1. Recording the rounded position in the requested space.** The first commit did that. It made two jobs that legitimately share a coordinate look like time running backwards — discovery crosses the timing rows into both `captures` and `dataProducts`, so an `advanceTimeMillis = 500` preview that also emits a `@ScrollingPreview(LONG/GIF)` product has two jobs at 500ms, the first leaves the clock at 512, and the second failed the ascending-order `require` and aborted the whole preview. Codex called this before CI did.

`RedToBlueScrollTimedPreview` is added as exactly that pairing. I verified it actually catches the regression rather than assuming: reverting the renderer and re-rendering produces `output advanceTimeMillis must be ascending (got 500 after clock was at 512)` — the predicted failure verbatim — and it passes again with the fix.

**2. Letting internal advances absorb into the requested space.** That same first version also let the `@FocusedPreview` settle window and the press settle consume the time budget of later captures, which took `WearFocusedPressPixelTest` down on CI — the pressed state layer lost the time it needed to crossfade, so the pressed capture stopped differing from the focused one. Internal advances now sit *on top of* the requested timeline, as they always did.

I chased that one wrongly at first: a local re-run against stale renders reproduced the failure and I nearly concluded the current version broke it too. Three clean `--rerun-tasks` runs say otherwise — the failure belongs to the first commit, which is what CI tested.

## Scope

Deliberately only part 1. The other half of #4247 — making the final advance genuinely unrounded so an exact coordinate lands on `N` rather than the next frame boundary — is a larger behaviour change that would move more baselines, and stays open, as does the `afterMs < 32` question on the desktop focus path.

## Test plan

| Command | Result |
| --- | --- |
| `:renderer-android:testDebugUnitTest` | pass |
| `:samples:android:test` | pass |
| `:gradle-plugin:test` | pass |
| `:samples:design-catalog-wear-m3:testDebugUnitTest` | pass — 3× with `--rerun-tasks`, the suite CI caught the first commit on |
| `ktfmtCheckAll` | clean |
| `:samples:android:composePreviewRenderAll` | full sweep vs unmodified `main`: **1 moved** (SpinnerTimeline at 1500ms, intended), **1 new** (the fixture), **162 byte-identical** |


yschimke:
Both findings on this PR were right, and both are fixed. Recording what happened, since the shape of the change is a reaction to them.

**P1, "Preserve equal targets after frame rounding"** (on `284d3d4`) — correct, and it was a crash rather than a drift. Discovery crosses the timing rows into both `captures` and `dataProducts`, so `advanceTimeMillis = 500` plus `@ScrollingPreview(LONG/GIF)` puts two jobs at 500ms; the first left the rounded clock at 512 and the second failed the ascending-order `require`, aborting the whole preview.

`a9919cb` keeps the two coordinate spaces apart, as suggested: the marker stays in requested space, each advance spends only what the physical clock still owes, and a job whose coordinate the clock has already passed is owed nothing.

Rather than assume the new fixture would catch it, I checked: `RedToBlueScrollTimedPreview` added, renderer reverted, re-rendered — `output advanceTimeMillis must be ascending (got 500 after clock was at 512)`, the predicted failure verbatim — then green again with the fix.

That first commit also took `WearFocusedPressPixelTest` down on CI, which the review didn't mention but is the same root cause: internal focus/press settles were being absorbed into the requested space, so the pressed state layer lost the time it needed to crossfade. They sit on top of the requested timeline again.

Worth admitting one detour: a local re-run reproduced that Wear failure against stale renders and I nearly concluded the *current* version broke it too. Three clean `--rerun-tasks` runs say otherwise — it belongs to `284d3d4`, which is what CI tested.

**P2, "Describe the requested marker accurately"** (on `a9919cb`) — also correct, and a good catch. I rewrote the implementation to the two-coordinate design but left the evidence README describing the first version, so it claimed the marker becomes 512 and "tracks the clock" when it deliberately stays at 500. `dec2c6a` rewrites it to show the requested and physical columns separately, and records what is deliberately *not* absorbed. The PR body had the same stale description and is updated.

---

PR #4266, opened 2026-08-19
feat(serve): inspect typography on published catalogs and per-preview daemons

yschimke:
The two follow-ups noted on #4256: the lanes that still could not answer `/render/<id>.annotations`.

## Published catalogs

A static `ServeBundleHost` has no daemon to capture a semantics tree from, so the lane reported `NotFound` — but a published catalog already ships `annotations/index.json`, whose preview layer is typography measured over the very PNG that host serves. It now answers from there.

This is not an approximation: the host never re-renders, so no override can move the pixels the bounds describe (an override-bearing request gets the same baked frame, and the HTTP layer reports what it dropped). Typography only in practice — theme attributes are projected from a live semantics tree and no producer authors them into a bundle.

Because the two lanes carry different layers, availability became two questions:

| | Typography row | Theme attributes row |
| --- | --- | --- |
| daemon-backed session | `hasDesignAnnotationsFor` | `hasDesignAnnotationsFor` |
| published catalog | **`hasPublishedTypographyFor`** (new) | — |

Folding them into one flag would either hide a layer that works or offer one with nothing behind it.

Both composite catalog hosts now fall back to that baked layer for anything their daemon lane cannot answer — an unmapped variant, an unresolvable daemon, a backend with no semantics lane — gated on the request routing to baked pixels in the first place (`CatalogLiveRouting.overridesAffectRender`, the predicate `render` already routes on). Under a font scale the daemon draws different pixels, and published bounds over those would misplace every box while looking entirely deliberate. A `Failed` from the daemon still travels unchanged; only `NotFound` falls through.

## Per-preview daemons

`ServePerPreviewLiveHost` routed neither inspection lane, so its viewer could inspect nothing at all. Both now route to the preview's own daemon with the viewer's overrides intact, advertised from alias membership — the same basis as the knob controls and the live stream, and for the same reason: asking a per-preview daemon what it can produce would materialise a bundle and a JVM on every page render, for a checkbox nobody has ticked. An id that turns out not to resolve reports `NotFound`, exactly as `renderScrollPng` and `subscribeStream` already do.

The response encoding moves into one `ServeAnnotationsPayload`, so the two producers cannot drift into shapes the overlay parses differently.

## Visual evidence

The `serve-viewer-published-typography` fixture — a **static** catalog viewer, no live daemon — plus its `layers` capture state, both committed, so the CI visual-diff bot renders and diffs this surface on every future PR. Shots are the harness's own server and Playwright flow: open the Overrides drawer, then tick Typography where there is one to tick.

**Before** — the lane 404s, so the drawer has Size and Locale & text and stops there; the catalog's annotations fed only the compare page.

static catalog viewer with no Inspect group

**After** — the published annotations answer it, and only Typography is offered: no semantics tree means no Theme attributes row.

the same viewer with Typography ticked, boxes on the frame and a legend beside it

The per-preview daemon half has no capture: `ServePerPreviewLiveHost` has no production construction site yet (it is referenced only from its own tests), so there is no page to shoot. Its routing is covered by unit tests instead, and it will be diffed by the existing `serve-viewer-inspect` states once a deployment builds one.

## Test plan

- [x] `ServeBundleHostTest` — published typography answers the lane with its tag index; `layout` records are filtered out (there is no layout layer to group them under); layout-only, unannotated, and unknown ids all stay `NotFound`.
- [x] `ServeCatalogLiveHostTest` — an unmapped variant inspects the published layer without waking the daemon; a mapped one whose daemon has no semantics lane falls through to it; a font-scaled request gets neither.
- [x] `ServePerPreviewLiveHostTest` — both lanes route by daemon id with overrides intact; an unmapped id reaches the baked layer without resolving a daemon; a11y has no baked half and stays unavailable.
- [x] `ServeWebTest` — a published-typography session renders the Typography row and not the Theme one; a daemon session still renders both.
- [x] `./gradlew :cli:test` — 2575 tests, 0 failures.
- [x] `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot` — 136 passed, including the new fixture's two themes and its `layers` state.
- [x] `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest` and `npm --prefix vscode-extension run format` — no changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RYt3qzYeCt91pc8YpkVVVx)_


yschimke:
On the diff bot's `serve-landing-tree-depth-component-open` (dark) row: that one is **not** this PR. Nothing here touches that page — the only fixture HTML this branch adds is `serve-viewer-published-typography.html`, and the four `New` rows are its expected captures.

The two frames hold identical content (same tree expanded, same cards, same variant rows); they differ only in the sticky header's vertical offset, i.e. where the page was scrolled when the shutter fired. It reproduces on this branch with no code change at all — two consecutive runs of

```
npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot -g "serve-landing-tree-depth"
```

give different bytes for that capture:

```
398f8ffe0bd41f66897e4ddf735a1a92ea4de0b894848075ea947f347983aff9
a4b9e3d01abc08caf56331d8edc417f78514c08194e48583c1fada32c9af7e1f
```

So it is a third instability in a capture that has already been stabilised twice (issue #3837's pointer parking, and the raster flags in `playwright.config.mjs`) — this time on scroll position rather than hover or rasterization. `scrollY` is stably `2` when the state is driven on a fresh page, so the varying part is likely leaking in from the states that run before it against the same page, but I have not pinned it down.

Deliberately not fixed here: a speculative fix would rebaseline an unrelated fixture inside a typography PR, and I would rather not guess at the mechanism. Worth its own issue.

---

PR #4270, opened 2026-08-19
fix(serve): read a default-valued uiMode as the default, not a pin

### LINKED ISSUE #4218: preview server: a default value spelled out in the URL (uiMode=light) suppresses the Figma comparison
Reported downstream as [yschimke/m3-catalog#145](https://github.com/yschimke/m3-catalog/issues/145), and refiled here because nothing in that repository decides it — the query-string handling belongs to the preview server.

## What's wrong

`uiMode=light` in a viewer URL suppresses the Figma comparison panel, even though `light` **is** the default. A user who has only ever taken the default lands in a state that looks like an explicit mode selection.

Reproduce:

```
https://preview.coo.ee/m3-catalog/p/button-filled__ideal__default__light?uiMode=light&mode=spec&specView=diff
```

The Figma % / comparison panel does not appear. Drop `uiMode=light` and it does.

## Why it is easy to reach

The light/dark toggle writes `uiMode` into the URL on the way through, so clicking dark and then back to light leaves `uiMode=light` behind. The user has made no net choice, but the URL now reads as an override and the comparison goes away. Copying that URL into an issue — which is exactly how m3-catalog#145 was filed — propagates the dead state.

## The ask, in two halves

1. **Don't serialise default values into the URL.** A control returning to its default should remove its parameter rather than spell it out. This keeps shared links minimal and makes "has the user chosen anything?" answerable from the URL.
2. **Where a default value is present anyway, recognise it as the default.** An explicit `uiMode=light` should be treated as "light, which is the default" rather than "the user pinned a mode", so features gated on *not having pinned a mode* — the Figma comparison being the one observed — stay available.

Half (2) matters on its own even with (1) in place, because old links, bookmarks and pasted URLs will carry the parameter indefinitely.

## Environment

- Design system `m3-catalog`, catalog `yschimke/m3-catalog@design-artifacts/m3-catalog`
- Rendered by compose-ai-tools 1.12.0 at the time of the report; the URL behaviour reproduces on the currently deployed server.

yschimke:
Fixes #4218.

`?uiMode=light` on a `…__light` preview suppressed the Figma comparison. The light/dark toggle writes `uiMode` into the URL on the way through, so clicking Dark and then Light again leaves the parameter behind — a visitor who has made no net choice lands on a page whose spec chip has dropped its match score and whose readout says the spec is "baseline-only, clear the overrides to compare", with no override present to clear.

## Before / after

The reported URL shape, captured through the `pages-snapshot` harness. Read the chip and the readout; the delta map either side is the same picture, which is the point.

| before | after |
| --- | --- |
| The spec lane at ?uiMode=light before the fix: the chip reads a bare Figma and the readout says the imported spec is baseline-only | The same URL after the fix: the chip reads Figma 90.3% and the readout says 90.3% match, 92.15% pixels differ |

`Figma` → `Figma 90.3%` on the chip, and `92.15% pixels differ · the imported spec is baseline-only …` → `90.3% match · 92.15% pixels differ` in the readout. Notes on both are committed alongside them in `renders/spec-default-theme/README.md`.

## What changed

Both halves the issue asks for.

**Don't serialise a default into the URL.** The server publishes the preview's baked theme on the theme select as `data-default-theme`, and a new shared predicate — `pinsTheme(choice, default)` in `cli/serve-web/src/viewer/themeChoice.ts` — forgives a choice that merely names it. `activeThemeChoice` then answers `""` for such a pick, so `query()` stops emitting `uiMode`, `syncUrl` drops it from the address bar, and the render URL goes back to being a plain published PNG.

**Recognise a default that is present anyway.** The sticky bootstrap in `ServeWeb.viewerThemeStickyScript` mirrors the same predicate, so a `uiMode` arriving from an old link, a bookmark or a paste is *displayed* (the Light chip reads pressed, as it should) without setting `data-theme-active="1"` — which is what `data-spec-baseline` is published from three lines later. The remembered-choice path is guarded identically. Back/Forward hydration and the pinned-page parameter carry follow the same rule, so stepping back through a dark→light round trip restores "nobody picked" rather than a pick that agrees with the default.

**A preview whose theme the catalog cannot name claims no default.** `data-default-theme=""` is "the server could not say", deliberately not "light": the select still displays Light, but the baked pixels are not known to be a light render, so `uiMode` there stays a real override that has to travel. This keeps the client's forgiveness a subset of the server's own baked-no-op rule (`CatalogLiveRouting.withoutBakedNoOps`), so no request that used to re-render now silently gets baked pixels instead.

No pixels move for a theme that *is* pinned: the "baseline-only" readout is correct there and is unchanged, because the imported spec is exported once and never re-exported per theme. What changed is only which values count as pinning one.

## Coverage

A new `serve-viewer-spec-default-theme` page fixture (a Figma-backed catalog on a preview whose id names its theme) is registered with the snapshot harness and navigated at `?uiMode=light&mode=spec&specView=diff`, so this exact regression is rendered and diffed by the CI visual-diff bot on every subsequent PR rather than depending on anyone remembering it. `serve-viewer-path` keeps the untokened case, so the pair covers both sides of the rule.

## Test plan

- `cd cli/serve-web && npm run typecheck && npm test` — 853 passing, including new `pinsTheme` cases and the default-aware `activeThemeChoice` branches (default-valued pick, frozen frame, dark-baked preview, unknown default)
- `./gradlew :cli:test ktfmtCheck` — 2,579 tests; new `ServeViewerThemeBarTest` cases cover the attribute for `__light` / `__dark` / dark-first / unnameable previews and both guarded seeding paths in the bootstrap
- `cd vscode-extension && npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot` — 138 passed, including the new fixture in both themes
- Page goldens regenerated (`UPDATE_SERVE_WEB_FIXTURES=true`); the committed `viewer.js` bundle is rebuilt and in sync

Rebased onto cc0010f (#4267) after the first CI round: the `CMP/Wasm Parity` failure it reported was `IndeterminateCircularProgressRemote` drifting 0.58% → 1.14%, which is the preview #4267 exists to fix — it makes the RC player load the player-supplied system variables (time) every frame, and ships its own `indeterminate-progress-before/after.png` evidence for exactly that spinner. The base had moved under this branch; nothing here reaches the RC player.


yschimke:
On the preview-harness diff above — the six **new** captures are expected, the four **changed** ones are not this branch's.

`serve-viewer-spec-default-theme` (both themes) is the fixture this PR adds. `serve-viewer-published-typography` and `-layers` came in with #4266; the baseline branch just hasn't caught up to `main` yet.

The four changed captures are pre-existing nondeterminism in the harness. Rendering `serve-landing-tree-depth` and `serve-viewer` **twice from the same commit of this branch**, exactly those four differ run-to-run while stable captures stay byte-identical:

```
serve-landing-tree-depth-component-open.dark:  DIFFERS run-to-run
serve-landing-tree-depth-component-open.light: DIFFERS run-to-run
serve-viewer-connecting.dark:                  DIFFERS run-to-run
serve-viewer-report-open-bundle.dark:          DIFFERS run-to-run
serve-viewer.light:                            identical run-to-run
```

They also can't be this diff by construction. `serve-landing-tree-depth.html` is untouched here, carries no `#cp-theme` select and doesn't load `viewer.js`. `serve-viewer.html` gains only `data-default-theme=""` on the visually-removed select plus the guard in the bootstrap — and with an empty default `pinsTheme(value, "")` is true for every value, so that page's behaviour is byte-for-byte the old one.

Comparing the bot's before/after pairs, all four differ in the same way: a sticky header or a scroll container caught at a slightly different offset (`serve-viewer-connecting` reveals a second Issues row in one and not the other), which is the same class of flake the spec's own notes record for the pointer-position captures. Worth a separate look at some point, but nothing to fix on this branch.

---

PR #4276, opened 2026-08-19
feat(serve): browse a catalog's motion captures on one page

yschimke:
## What

A **motion browser** at `/motion` (and `/{system}/motion`): every recorded interaction and animation a catalog publishes, on one page.

A capture was per-component surface only — the viewer's Motion lane. That is the right home for *reading one* and the wrong home for the catalog-wide question: **does this design system move consistently?** Two containers that morph on the same spatial spring and a third that cross-fades is a system bug, and it is invisible from three separate component pages. It was also undiscoverable — captures are rare, so the way to find one was to open components until a Motion chip appeared.

## Where the entry point is

A chip in the catalog landing's `⋯` menu, beside the Theme pill — where this catalog's other **destinations** already live (the comparison views, the parity dashboard, the playground). It carries its count for the same reason the pages chip does: one recording and thirty are different offers. That is the same count the route gates on, so the chip never leads to a 404 and a catalog that records nothing is never offered an empty page.

| before | after |
| --- | --- |
| the actions menu with one entry | the same menu with a 4 motion captures chip |

## The browser

Nothing plays until it is asked to. Assigning an `` `src` **is** starting an APNG, so a grid that shipped with the recordings in place would decode all of them at a reader who opened the page to compare them one at a time. Every card ships with its component's baked still and swaps only on a press or on **Play all**; stopping restores the still, which is what actually stops the decoding. There is therefore no autoplay for `prefers-reduced-motion` to suppress, and the control is a button rather than a hover, so it works from a touch screen and a keyboard.

**At rest** — one card per *capture*, not per component (`Switch · On` appears twice, once as *Toggle on* and once as *Thumb settle*), grouped under the landing's own sections, named by the same `MotionCaptureLabels` split the viewer's picker uses:

| light | dark |
| --- | --- |
| the motion browser at rest, light | the motion browser at rest, dark |

**After Play all** — every card swapped to its recording, the ▶ cue gone, the button reading Stop all:

| light | dark |
| --- | --- |
| the motion browser playing, light | the motion browser playing, dark |

Each card deep-links to `?mode=motion&motion=<id>`, which opens the viewer already on **that** recording rather than the component's first one — an `` cannot be paused, sought or rate-controlled from script, so the browser hands off to the canvas transport instead of trying to be it.

## Decisions worth a reviewer's eye

- **Suppressed in Catalog mode**, with the other destinations. A close call — the browser is browsing surface, not tooling, and Catalog mode deliberately *keeps* the per-component Motion control. What settled it is the affordance: every other entry in that menu is stripped there, so keeping this one would give the mode a `⋯` menu that exists to hold a single link. Recorded in [the component-browser checklist](docs/design/component-browser-mode-checklist.md) so it is one line to flip.
- **Inline script rather than a `cli/serve-web/` bundle.** Swapping `src` is the page's whole behaviour; a per-page file in that build would cost a round-trip to buy nothing.
- **`/motion` joins `ServeBugReport`'s server segments.** In query mode (`/motion?session=…`) it would otherwise have been read as a design system named "motion" and the report filed against one that does not exist — the same trap already documented there for `pages` and `parity`.

## Test plan

- `./gradlew :cli:test` — **2601 tests, 0 failures**.
- New `ServeMotionIndexTest` (8): both URLs on a card, one card per capture, deep links, label parity with the viewer's picker, the counts, the chip's gating and singular/plural, Catalog-mode suppression, section grouping.
- New `ServeMotionIndexRoutingTest` (4) over real HTTP — a page fixture only proves `ServeWeb` draws something when handed arguments the *test* chooses; this is the half that proves the handler hands them over. Includes the one way this route could have been wrong by construction: `/motion` sits one segment above `/motion/<id>.apng`, so both are asserted on the same server.
- `ServeBugReportTest` extended for the new server segment.
- `npm run harness:snapshot` in `vscode-extension/` — **178 passed**. New `serve-motion-index` fixture plus a `motion-index-playing` state, so the resting grid *and* the swap are baselines the visual-diff bot carries from here on: a page that started autoplaying, or a Play all that stopped reaching the cards, now moves pixels.


yschimke:
Fix ci failures

PR #4277, opened 2026-08-19
fix(preview): diff a PR against the baseline for its own base, not the branch tip

yschimke:
## The bug

The compose preview comment pinned `Before` to whatever `compose-preview/main` pointed at **when the PR job happened to run**. That tip is not the PR's base — there was no merge-base logic anywhere. A baseline lands 5–20 minutes *after* the merge that produced it, and `main` keeps moving while a PR is open, so the diff was taken between two unrelated points in history and everything in the gap was attributed to whichever PR happened to render there.

Both directions showed up on `wear-m3-catalog`:

**Baseline behind the PR's base** — the previews a just-merged PR changed get replayed as the *next* PR's own. [wear-m3-catalog#24](https://github.com/yschimke/wear-m3-catalog/pull/24) changed one file (`Dialogs.kt`) and its comment reported 16 new / 9 changed / 4 removed. The function names are exactly the union of #20, #21 and #23, merged 13 minutes earlier with their baseline still rendering:

| Section | #24 reported | Actually from |
|---|---|---|
| New | `ArcProgress`, `ButtonRowGroup`, `CircularProgress`, `Scheme colours`, `Shape shapes`, `Type type styles`, `WearList`, `WearScaffold` | #21 |
| New | the six `… theme` entries | #23 |
| Removed | `ButtonPlaceholder`, `CardPlaceholder`, `TextAction` | #21, verbatim |
| Changed | `LevelRail` / `ScrollRail`, `TextAction` | #20, #21 |

Only `AlertDialogSticker` was #24's own.

**Baseline ahead of the PR's base** — previews `main` gained after the PR branched are absent from its render, so they surface as deletions. [wear-m3-catalog#38](https://github.com/yschimke/wear-m3-catalog/pull/38) reported 36 "Removed" entries, none of which it deleted.

## The fix

`select-baseline.py` walks the baseline branch for the newest commit whose **source** `main` commit — the one recorded in its `Update preview baselines from <sha>` subject — is an ancestor of the state the render was built on (the merge ref's first parent when the checkout has one, else the event's `base.sha`). That removes the second case outright and bounds the first.

Whatever gap is left is genuine — the baseline for this PR's base hasn't published yet — so it's measured rather than guessed at, and the comment declares it:

> [!NOTE]
> Compared against the baseline rendered from `abc12345`, which is **2 commit(s) behind** this PR's base (`def67890`) — the newest baseline published when this ran. Any preview those commits changed is attributed to this PR below. Re-run this check once the baseline for `def67890` has published to see only this PR's own changes.

Two properties worth calling out, both covered by tests:

- **Best-effort by construction.** No `main` history in a shallow checkout, an unparseable subject, a source commit whose object isn't local → outputs unwritten, pipeline falls back to the branch tip. That fallback *is* the behaviour this replaces, so it can never do worse than not running.
- **The base branch is fetched blobless; the baseline branch deliberately is not.** Filtering the baseline branch would satisfy selection and leave the render trees behind a lazy promisor fetch that `git archive` performs moments later — whose failure mode is an empty baseline and a comment claiming every preview is new. A plain fetch of a branch a shallow repo doesn't have yet already brings full history, so there's nothing to gain by filtering it.

## Not addressed

The resource / a11y / notification comments still compare against their own branch tips and carry the same skew. Also unrelated and unchanged: preview IDs are the diff key, so a PR that adds a size dimension or renames a frame rewrites every key and reports the same function as both New and Removed (wear-m3-catalog#35, #39). That's honest output, just verbose — a rename-pairing pass would be a separate change.

## Test plan

- `python3 .github/actions/apply/test_select_baseline.py -v` — 13 tests, new. The selection rule against a synthetic ancestry oracle (including "walk back past a baseline the PR doesn't contain" and "fall back to the tip when nothing qualifies"), plus end-to-end against real repositories: merge-ref checkout, genuinely shallow checkout via `file://`, exact-base, unknown-base, and an assertion that the selected commit's `renders/` are local afterwards. Wired into the `actions-tests` CI job.
- `python3 .github/actions/lib/test_compare_previews.py -v` — 123 tests pass, 6 new covering the NOTE block: drift declared with both commits, silent at drift 0, silent with no skew file, malformed JSON degrades instead of breaking the comment, note precedes the sections it qualifies, and silent on a clean diff.
- Full `actions-tests` suite run locally — `test_compare_previews`, `test_vscode_preview_diff`, `test_a11y_report`, `test_check_skew`, `test_post_comment`, `test_compute_scope`, `test_path_scope`, `test_select_baseline`: all green.
- `bash -n` on `pipelines/compose.sh`, YAML parse on `action.yml` and `ci.yml`.
- Rendered a sample comment end-to-end through `compare-previews.py compare --baseline-skew` and confirmed the NOTE renders and the `Before` URLs are pinned to the selected baseline commit rather than the tip.

No visual evidence: this is CI plumbing — the comparator's ref selection and one Markdown block. Nothing renderable changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EeWhyVpezQtSSfYYPEpw3R)_


yschimke:
Fix ci failures

PR #4279, opened 2026-08-19
fix(serve): fold a component's breakpoints onto one card

### LINKED ISSUE #41: So many duplicate components
### What's wrong

Image

### Screenshot

alertdialog__ideal__default__192dp

### Which preview

| | |
| --- | --- |
| Preview | `alertdialog__ideal__default__192dp` |
| Source | https://github.com/yschimke/wear-m3-catalog/blob/main/catalog/src/main/kotlin/ee/schimke/wearm3catalog/sections/Dialogs.kt |
| Rendered by | compose-ai-tools 1.18.0 |

[Open this preview](https://preview.coo.ee/wear-m3-catalog/p/alertdialog__ideal__default__192dp)

yschimke:
Fixes yschimke/wear-m3-catalog#41 ("So many duplicate components").

## What was wrong

wear-m3-catalog renders every full-screen component at the five screen sizes its Figma kit declares (`192dp`, `204dp`, `216dp`, `225dp`, `240dp`). Its published `catalog.json` is correct — **53 components**, each carrying one image per size — but the served sheet showed **109 cards**, with 14 components appearing five times apiece under one name.

The export was never the problem. It tags each image with the breakpoint it rendered at (`"size": "192dp"`, resolved from the spec's `breakpoints` table by `catalog-breakpoints.mjs`). `ServeCatalogStore` then dropped that tag: it never reached `VariantMeta`, so `ServePreview` had no size and `ServeWeb` saw only five ids differing by a trailing token. The grid folds `state` and `props`; with nothing to key a size fold on, every breakpoint stayed a card.

Folding on the *id* was never an option, which is why this needed the metadata — a catalog names its own breakpoints (`192dp` here, `smallRound` elsewhere), and `previewSizeVariantLabel`'s hard-coded token list didn't recognise `192dp`, so even the collision-qualifier fallback left all five cards plainly named "Alert Dialog".

## What changed

`size` now rides `catalog.json` → `variants.json` → `ServePreview`, and:

- **the landing** folds every non-primary size onto the component's one card, drawn at the size the catalog declares **first** — which is also the size a design catalog's references are mapped against;
- **the viewer's component subtree** gains a size row per breakpoint, keyed so it holds the state and props fixed (from `no-buttons` you get `no-buttons` at 204dp, never the default);
- **the comparison page** folds the same way, but only for a size carrying no reference of its own — a kit that publishes a second size (Wear's `Picker`, at its `Larger Screen (BP)` cell) keeps that row, which is the whole point of the page.

A catalog published before the export recorded `size` is unchanged: folding those renders would leave them with no switcher to be reached from, so each keeps its card and its size qualifier. Same for a plain module bundle, which has no manifest to name its device fan-out.

## Visual evidence

Rendered from the **real** published `yschimke/wear-m3-catalog@design-artifacts/wear-m3-catalog` `catalog.json` (367 images) with its committed render PNGs routed in, screenshotted with the preview-harness's Chromium. "Before" is the same page built from previews with `size` stripped — exactly what the server saw until this change.

### Containment — the tab the issue was filed from

Five "Alert Dialog" cards and five "Open On Phone Dialog" cards become one each.

**Before**

Containment tab before, showing five identical Alert Dialog cards and five identical Open On Phone Dialog cards

**After**

Containment tab after, showing one Alert Dialog card and one Open On Phone Dialog card

### Actions

"Edge-hugging buttons" drops from five identical cards to one; the section count goes 20 → 16.

**Before**

Actions tab before, showing five identical Edge Button cards

**After**

Actions tab after, showing one Edge Button card

### The folded sizes stay one hop away

The viewer's component subtree gains a row per breakpoint beside the state rows — the "Alert Dialog" count goes 3 → 7.

**Before**

Viewer before, component subtree listing only Edge button and No buttons

**After**

Viewer after, component subtree listing Edge button, No buttons, 204dp, 216dp, 225dp and 240dp

The same pictures, with the full write-up, are committed at `docs/evidence/serve-breakpoint-fold/`.

## Future coverage

`serve-landing-breakpoints.html` and `serve-viewer-breakpoints.html` join the committed page fixtures, so the visual-diff bot captures this surface on every future PR without anyone remembering to.

## Test plan

- `ServeWebBreakpointFoldTest` (new, 5 cases): the grid folds a component's other breakpoints into one card; the nav tree names each component once; the viewer offers every other breakpoint in one hop; a size row holds the state fixed; a catalog that declares no sizes keeps a card per render.
- `./gradlew :cli:ktfmtCheck :cli:test --tests 'ee.schimke.composeai.cli.serve.*'` — green (281 tests).
- Page fixtures regenerated with `UPDATE_SERVE_WEB_FIXTURES=true`; the only pre-existing golden that moved is called out below.

## One unrelated commit

`serve-viewer-spec-default-theme.html` is the one page fixture #4271 changed without regenerating — its floating report launcher and the reworded "report a server bug" title reach every viewer page. `ServeWebFixtureTest` has been red on `main` ever since (verified by stashing this branch's changes and re-running against `origin/main`), so every open PR inherits that failure. Refreshed in its own commit; the diff is exactly the launcher markup and the title.

---
_Generated by [Claude Code](https://claude.ai/code/session_019FnnYDMBa6NXnEwNguEB2p)_


yschimke:
Two red/noisy signals on this run, neither caused by this diff. Details so nobody has to re-derive them:

**CMP/Wasm Parity — `IndeterminateCircularProgressRemote` (0.29% → 0.78%).** That preview is a *continuously animating* remote-clock arc (`RemoteCircularProgressIndicator` with no progress argument), and it is already on the CMP/Wasm allowlist in `.github/ci/remote-m3-cmp-wasm-allowlist.json` with the reason "CMP/Wasm currently draws no pixels for the Remote Material 3 indeterminate clock animation". Its mismatch is a function of which frame of that animation the two lanes land on. This diff touches `cli/src/main/kotlin/…/serve/*.kt`, two docs, and page-fixture HTML/PNGs — no renderer, no sample, no Remote Compose code — so there is no path from it to that number. The check says itself that it does not block merging; not pushing anything for it.

**vscode-preview-diff — "252 changed".** That is the `vscode-preview/main` baseline being behind `main`, not this branch. Pixel-diffed `serve-notfound.dark` (a page this PR cannot touch) between the baseline and head commits the bot linked: the only delta is the footer reading "report a bug" → "report a server bug" and the new floating report launcher in the corner — both #4271's, which also explains why `serve-viewer-report-menu` and `serve-viewer-published-typography` show up as *New* rather than changed.

The two captures that *are* this PR's are in that comment's **New** table, and they show what they should: `serve-landing-breakpoints` draws one "Alert Dialog" card from 15 previews across five breakpoints, and `serve-viewer-breakpoints` lists `No buttons`, `204dp`, `216dp`, `225dp`, `240dp` under it.

---

Issue #4285, opened 2026-08-19
fix(serve): paint live frames newest-wins in the viewer and the catalog grid


yschimke:
Sub-issue of #4159.

## What's wrong

`drawFrame` in `cli/serve-web/src/viewer.ts` builds a fresh `Image` from a `data:` URL for every frame and paints it in `onload`. There is no `seq` check and no newest-wins queue.

Decode time varies per frame, so an earlier frame can finish decoding *after* a later one and paint over it — and then stick until the next frame arrives, which at the current 4 fps cap can be 250 ms later. On a stream of a mid-animation ripple, that reads exactly like the artifact reported in #4159, independently of the daemon-side cause.

The VS Code client already does this correctly — `StreamFrameQueue` / `StreamClient` in `vscode-extension/src/daemon/streamClient.ts`, with the three rules written up in `docs/daemon/STREAMING.md` § "Client model". The serve viewer never got the port.

## Scope

- Drop frames whose `seq` is older than the last painted `seq`.
- Hold at most one pending frame; a newer arrival replaces it.
- Decode via `createImageBitmap` rather than an `Image` + data URL, so the visible canvas never tears down its current bitmap before the next is ready.
- Apply to `<cp-catalog-live>` too — it has its own painter with the same shape.

Ideally the queue lands as a shared module under `cli/serve-web/src/live/` (next to `pointerMap.ts` / `session.ts`), unit-tested the way `liveSession.test.ts` and `livePointerMap.test.ts` are, so the viewer and the grid cannot drift.

## Done when

Frames paint strictly in `seq` order on both the viewer and the grid, with a unit test that feeds out-of-order arrivals and asserts the stale one is dropped.

Issue #4283, opened 2026-08-19
perf(daemon): make the live interactive frame loop adaptive, and stop re-reading each frame's PNG


yschimke:
Sub-issue of #4159.

## What's wrong

`JsonRpcServer.INTERACTIVE_FRAME_INTERVAL_MS = 250L` drives `startInteractiveFrameLoop`, and the in-flight guard means real cadence is `max(250 ms, renderTime)`. Nothing modulates it in either direction:

- **Too slow while animating.** A Material motion spec of 200–500 ms gets sampled once or twice, so live interaction reads as chunky even once the frozen-ripple bug (sub-issue for the platform clock) is fixed.
- **Far too fast while idle.** An idle live viewer burns a full Robolectric capture every 250 ms, indefinitely. The tick also refreshes `lastUsedAtMs`, so `AndroidInteractiveSession`'s 60 s idle lease can never fire while a socket is open.

Separately, each frame does three to four full PNG passes:

1. `captureRoboImage` encodes and writes the PNG
2. `WrappedFrameCrop.cropTopLeft` reads, decodes, re-encodes and rewrites it (wrap-content previews)
3. `JsonRpcServer.hashFrameBytes` reads it again to SHA-256 it
4. `FrameStreamRegistry.consumeForPreview` reads it a **third** time to base64 it

The bytes are never carried in memory between those steps.

## Proposed change

**Adaptive cadence.** Render as fast as the backend manages while pixels are changing (capped — 15–20 fps is a plausible starting ceiling), back off geometrically to ~1–2 s once N consecutive frames dedup identical, snap back to the fast cadence on any input or any changed frame. The daemon already SHA-256s every frame, so the "nothing changed" signal is free.

On desktop, `ComposeScene.hasInvalidations` gives a real quiescence signal rather than one inferred after the fact. It is currently unused anywhere in the repo.

**Collapse the redundant reads.** Carry the frame bytes in memory from capture through hash and base64 instead of re-reading the file, and fold the wrap-content crop into the capture rather than a decode/re-encode round trip.

## Done when

An idle live session costs near-zero renders; an animating one runs visibly closer to the renderer's natural cadence; per-frame disk I/O is one write instead of two writes and three reads. Numbers before and after, from the telemetry sub-issue.

yschimke:
**Half of this landed independently as #4274**, which was merged while the investigation was in flight.

[...]

Issue #4159, opened 2026-08-18
Preview server issue: optimise live mode frame rate


yschimke:
### What went wrong

Clicks in the live mode show ripple artifacts. Can we optimise both

a) the daemon running each frame
b) sending each frame change - possibly a delta if localised changes are common?

Also we should Make "Snapshot" clearer that it toggles a live session. Or maybe double click to activate on the static preview? With a hint tooltip?

### Screenshot

![render](https://preview.coo.ee/m3-catalog/render/toolbar-horizontalfloating__ideal__default__light.png?themeProvider=ee.schimke.m3catalog.LightMediumContrastTheme)

### Server

| | |
| --- | --- |
| compose-preview | `1.14.1` |
| Mode | public (open) |
| Uptime | 8h 20m |
| Server JVM | `21.0.11 (Eclipse Adoptium)` |
| Server OS | `Linux 7.0.0-28-generic (amd64)` |

### Page

| | |
| --- | --- |
| Page | [`/m3-catalog/p/toolbar-horizontalfloating__ideal__default__light?themeProvider=ee.schimke.m3catalog.LightMediumContrastTheme`](https://preview.coo.ee/m3-catalog/p/toolbar-horizontalfloating__ideal__default__light?themeProvider=ee.schimke.m3catalog.LightMediumContrastTheme) |
| Design system | `m3-catalog` |
| Preview | `toolbar-horizontalfloating__ideal__default__light` |
| Catalog | `yschimke/m3-catalog@design-artifacts/m3-catalog` |
| Catalog rendered by | compose-ai-tools 1.14.1 |
| Trust | branch:yschimke/m3-catalog@design-artifacts/m3-catalog |
| Render lane | live daemon |

### Browser

| | |
| --- | --- |
| User agent | `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36` |
| Viewport | 1280×683 CSS px |
| Device pixel ratio | 2 |
| Page colour scheme | light |
| OS colour scheme | light |

yschimke:
## Investigation

[...]

PR #4293, opened 2026-08-19
fix(daemon): run looper-scheduled work in a held interactive session

### LINKED ISSUE #4282: fix(daemon): run looper-scheduled work in a held interactive session
Sub-issue of #4159.

> **Rescoped.** This issue originally claimed to be the fix for #4159's ripple artifact. It is not — that was a wrong diagnosis, corrected [here](https://github.com/yschimke/compose-ai-tools/issues/4159#issuecomment-5348522275) with per-frame measurements. The ripple animates and settles whether or not this bug is present. What is described below is a *separate*, real bug that the investigation turned up on the way, and it is fixed. #4283 and #4285 are the ripple fix.

## What's wrong

A held interactive session has two clocks. Compose's paused `mainClock` drives composition, `LaunchedEffect` delays and Compose-drawn animation. Robolectric's **looper** clock is a separate thing, moved only by `ShadowLooper.idleFor`, and it is what any `Handler.postDelayed` comes due against — the app's own timers, a platform widget's, anything scheduled onto the main looper.

`RobolectricHost.SandboxRunner.runHeldInteractiveSession` only ever advanced `mainClock`:

- on `InteractiveCommand.Render` (`advanceTimeMs ?: HELD_CAPTURE_ADVANCE_MS`)
- on `InteractiveCommand.Dispatch` (`POINTER_HOLD_MS`)

`waitForIdle()` is the only other looper contact in `:daemon:android`, and it idles without moving the clock. So the looper clock sat frozen at zero for the whole life of a held session: **a live preview never ran a single delayed callback, however long it was left open.**

`AndroidRecordingSession.runLiveTickLoop` drives frames through the same `interactive.render`, so delivery-branch motion captures had the same gap.

## Fix

`advanceHeldClocks(rule, totalMs)` steps both clocks together in 16 ms slices (one `Choreographer` frame, the same `PRESS_SETTLE_FRAME_MS` the static lane steps a pressed capture by) and every `advanceTimeBy` site in `SandboxRunner` routes through it. Stepped rather than jumped because `idleFor` runs the callbacks that came *due*, not one per frame — idling a whole tick at once runs everything scheduled inside it simultaneously, which is a jump to the end state rather than motion through it.

The cost — looper work due inside the window runs ahead of its wall-clock moment — is the same one the static lane [documented and accepted](https://github.com/yschimke/compose-ai-tools/blob/main/renderers/android/src/main/kotlin/ee/schimke/composeai/renderer/RobolectricRenderTest.kt#L2589) for pressed captures. In a live session it is closer to right than the alternative: a live preview is supposed to be running the app's timers, and it was running none.

## Test

`AndroidInteractiveSessionTest.heldRenderAdvancesThePlatformAnimationClock`, against a new `PostDelayedSquare` fixture that flips red→green from a `Handler.postDelayed` at 100 ms. Verified to fail without the fix (0.00% green — the square stays red) and pass with it. The first frame is asserted to still be *red*, so the test also pins that the clock is stepped in real time rather than settling every pending callback at once.

yschimke:
Closes #4282.

A held interactive session has two clocks. Compose's paused `mainClock` drives composition, `LaunchedEffect` delays and Compose-drawn animation. Robolectric's **looper** clock is a separate thing, moved only by `ShadowLooper.idleFor`, and it is what any `Handler.postDelayed` comes due against — the app's own timers, a platform widget's, anything scheduled onto the main looper.

`RobolectricHost.SandboxRunner.runHeldInteractiveSession` only ever advanced `mainClock`. `waitForIdle()` is the only other looper contact in `:daemon:android` and it idles without moving the clock. So the looper clock sat frozen at zero for the whole life of a held session: **a live preview never ran a single delayed callback, however long it was left open.** `AndroidRecordingSession.runLiveTickLoop` renders through the same `interactive.render`, so delivery-branch motion captures had the same gap.

`advanceHeldClocks(rule, totalMs)` steps both clocks together in 16 ms slices — one Choreographer frame, the same step the static lane's `settlePressedRipple` uses — and every `advanceTimeBy` site in `SandboxRunner` routes through it. Stepped rather than jumped because `idleFor` runs the callbacks that came *due*, not one per frame: idling a whole tick at once runs everything scheduled inside it simultaneously, which is a jump to the end state rather than motion through it.

## Visual evidence

Two consecutive held renders of the new `PostDelayedSquare` fixture, which flips red→green from a `Handler.postDelayed` at 100 ms. Frame 1 covers the 32 ms bootstrap settle; frame 2 advances 200 ms.

**Before** — red, red. The callback never comes due.

postDelayed before

**After** — red, green. Frame 1 is still correctly red, which is the half that shows the clock is stepped in real time rather than settled to its end state in one jump.

postDelayed after

## This is not the #4159 ripple fix, and it was originally claimed to be

Worth reading before reviewing, because the framing changed mid-PR. The starting diagnosis was that Material's pressed `RippleDrawable` rides this same looper clock, so a live click painted a half-expanded ripple that never progressed. The first half is true; the conclusion is not.

`AndroidRippleFrameTest` (committed here) presses a real M3 filled button in a held session and films the frames — a filled button is 0 dp elevation pressed and unpressed, so the ripple is the only thing that can move. The ripple animates and settles **whether or not this bug is present**, converging on the same final frame either way. What the filmstrips did show is the real cause, and it is what the title of #4159 said all along:

**Filmed at 50 ms** — the ripple is plainly visible mid-expansion in frame 3, then settles.

ripple at 50ms

**Filmed at 250 ms**, which is what `INTERACTIVE_FRAME_INTERVAL_MS` actually ticks at — the ripple never appears in any frame. The animation completes between two ticks.

ripple at 250ms

| filmed at | rest→1 | 1→2 | 2→3 | 3→4 | 4→5 |
| --- | --- | --- | --- | --- | --- |
| 50 ms, with this fix | 0.0% | 11.2% | 5.1% | 0.05% | 0.0% |
| 50 ms, without it | 0.0% | 5.7% | 8.0% | 4.8% | 0.04% |
| 250 ms — the live loop | 0.0% | 12.5% | 0.0% | 0.0% | 0.0% |

So this PR stands on its own — a live session that runs none of the app's timers is broken regardless — but the ripple artifact belongs to #4283 (cadence) and #4285 (out-of-order painting). #4159's plan and #4282's scope have both been corrected on the issues.

`AndroidRippleFrameTest` is therefore committed as a **characterisation test**, not a regression test for this change: it pins that the ripple animates across several frames and converges, which is the property #4283 has to deliver, and the numbers above are its baseline.

## Test plan

- `AndroidInteractiveSessionTest.heldRenderAdvancesThePlatformAnimationClock` — new. **Verified to fail without the fix** (0.00% green, the square stays red) and pass with it.
- `AndroidInteractiveSessionTest` — all 26 pass, covering the other held-session paths the shared helper now routes (semantics actions, click fast-path, typing, rotary, focus).
- `AndroidRippleFrameTest` — new, passes.
- Evidence regenerable from `AndroidRippleFrameTest` (writes to `build/ripple-frames/`) with `STEP_MS` retimed; see `docs/evidence/live-held-session-clock/README.md`.


yschimke:
fix merge conflicts

PR #4294, opened 2026-08-19
fix(serve): score a comparison over its drawn content, not the whole canvas

### LINKED ISSUE #4290: The diff %-age doesn't make sense
### What went wrong

So different, but marked as 93%

### Server

| | |
| --- | --- |
| compose-preview | `1.21.0` |
| Mode | public (open) |
| Uptime | 1h 14m |

### Page

| | |
| --- | --- |
| Page | [`/wear-m3-catalog/compare?format=reference`](https://preview.coo.ee/wear-m3-catalog/compare?format=reference) |
| Catalog rendered by | compose-ai-tools 1.18.0 |
| Render lane | live daemon |

### Browser

| | |
| --- | --- |
| User agent | `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36` |
| Viewport | 1280×683 CSS px |
| Device pixel ratio | 2 |
| Page colour scheme | dark |
| OS colour scheme | dark |

### Recent failures

```
2026-08-19 20:39 UTC  remote-m3: render failed after 4561ms — render failed: NoClassDefFoundError: androidx/compose/remote/player/view/RemoteComposePlayer
```

yschimke:
Fixes #4290 — *"So different, but marked as 93%."*

## What was wrong

`scorer/planes.ts` averaged each pixel's cost over **every pixel of the frame**, so the empty backdrop the two sides share was most of the average. On a 384×384 round watch screen — black in both frames whatever is drawn on it — that is ~80% of the canvas agreeing before anything has been compared, and almost every pair landed in the nineties whatever it looked like. Two compounding halves:

- **The denominator.** The same missing mark scores 85.7% on a 9×7 canvas, 97.1% on 21×15 and 99.3% on 41×31. The number described how much room a component was rendered into.
- **The cost curve.** A gap was charged linearly to 255, so a filled control sitting where the reference has bare background cost about a fifth of a pixel — a component that had lost its fills entirely still read as "mostly matching".

## What it does now

The cost is divided by the pixels that carry **content** — where either frame has detail (the union of their edge masks, widened a pixel), plus wherever the two disagree — so the answer is the share of the pair's content that agrees, at any canvas size. A gap of more than half the luminance range (`FULL_DIFFERENCE_DELTA = 128`) is charged in full, so a control that lost its fill costs what an absent control costs. The edge-tolerant search, the both-directions average, the content-box normalisation and the fixed backdrop are untouched.

## Visual evidence

wear-m3-catalog's published references against its published renders, scored with the shipped `format-compare.js` before and after. Left is the imported Figma reference, right is the render:

wear-m3-catalog reference/render pairs with their before and after scores

Top row is the pair from the issue: a stepper whose reference has two filled buttons, a level-indicator arc and wrapped text, against a render with two hairline chevrons, no fills and a line of text running off the screen. **93.7% → 70.0%.** The pairs that genuinely agree (outlined icon button, body text) stay where they were.

The spec lane, same page and same frames — `serve-viewer-path` · `spec-diff`, captured with `npm run harness:snapshot`:

**Before** — the verdict reads `90.3% match` beside its own readout saying `92.15% pixels differ`:

the spec lane's diff view reporting 90.3% match beside 92.15% pixels differ

**After** — same frames, same delta map, `71.5% match`:

the same view reporting 71.5% match beside the same 92.15% pixels differ

Both, plus the distribution table, are committed under [`docs/design/evidence/compare-score-content-normalisation/`](https://github.com/yschimke/compose-ai-tools/tree/agent/issue-4290-diff-percentage/docs/design/evidence/compare-score-content-normalisation).

## What it does to a real catalog

All 186 published wear-m3-catalog pairs, scored both ways through the shipped asset in headless Chromium (the same path `design-reference-score.mjs` uses; the "before" column reproduces every published `match` value exactly):

| | min | p10 | p25 | median | p75 | p90 |
| --- | --- | --- | --- | --- | --- | --- |
| before | 21.9 | 50.6 | 93.7 | 97.3 | 99.8 | 100.0 |
| after | 4.0 | 17.7 | 78.9 | 91.1 | 98.5 | 100.0 |

The SVG lane moves the same way (380 export/render pairs, median 99.7 → 96.6): its worst cases are exports that rasterise **blank**, and they now read 10–15% instead of 53–66%.

Bands:

- The compare wall's `grade` 90/75 splits those 186 rows **115 / 38 / 33** after versus **159 / 5 / 22** before — the triage band it was built for barely existed under the old score — so those thresholds are unchanged.
- The spec lane's `matchBand` 99.5/97 only ever made sense against the compressed range and becomes **95 / 85**, taken from the distribution above and mirrored in `ServeWeb.specMatchBand` and the export driver's `matchBand`.
- `MATCH_FLOOR` (90) is unchanged — "a tenth of the drawn content disagrees" is still the right floor for "worth opening".

## Known limitation (unchanged, not introduced here)

The metric is luma-only, so `shape-materialshapes` pairs — a saturated purple reference against a pale lilac render at nearly the same luminance — still score ~91. Worth a follow-up; out of scope for this fix.

## Test plan

- `cli/serve-web`: `npm run typecheck` ✅, `npm test` ✅ (889 passing), `npm run build` ✅ (committed bundle regenerated). `scorerPlanes.test.ts` gains the #4290 regression — the same absent mark scores identically on 9×7, 21×15 and 41×31 — plus `pixelCost` and `contentMask` tables.
- `vscode-extension`: `npm run harness:compare` ✅ (6 passed). The near-empty fixture's mark used to sit 16px off in one frame and pass on blank-canvas dilution; it is now the same mark in both, so the test still isolates what it is about (the content-box fallback) rather than the dilution that is gone.
- `scripts/design-artifacts`: `node --test design-reference-score.test.mjs` ✅ (4 passed).
- `./gradlew ktfmtCheckAll` ✅. `./gradlew :cli:test --tests "…ServeWebTest"` could not complete in this sandbox — Maven Central returns `429 Too Many Requests` through the agent proxy — so the updated band assertion there is covered by CI.

## Follow-up after merge

The `match` numbers baked into `references/index.json` on each `design-artifacts/<system>` branch were minted by the old metric and stay stale until a `design-artifacts` run republishes them; a renderer/CLI change does not trigger one, so this wants a manual `workflow_dispatch` (or the Monday cron) to bring the published chips back in line with what the lane computes live.

---
_Generated by [Claude Code](https://claude.ai/code/session_01URRQ8yHeqhCwrXzwWwtfVG)_


yschimke:
fix ci failures and merge conflicts

yschimke:
Both done in `fad6a2c`.

**Merge conflict** — `main` had moved on by three commits (#4296, #4298, #4295). The single conflict was `assets/serve-components.js`, which both sides had rebuilt (main for the compare wall's report launcher, this branch for the scorer). Resolved by regenerating it from the merged sources with `npm run build` in `cli/serve-web`, not by merging minified output. No semantic overlap: #4295 touches the bug-report launcher, nothing in `scorer/`, `compare/` or `spec/`.

Verified on the merged tree: `npm run typecheck` ✅, `npm test` ✅ (899 passing), `./gradlew :cli:test` ✅ (full suite), and `UPDATE_SERVE_WEB_FIXTURES=true … ServeWebFixtureTest` produced no further golden drift — the auto-merged `serve-format-compare.html` already carried both the new subtitle and the launcher markup.

**CI failure** — the only red was `CMP/Wasm Parity`, on `IndeterminateCircularProgressRemote` (0.66% → 1.02%, limit 0.25 pp). It has now come back **green on the merged head** (`0 regression(s), 51 unchanged`), which confirms the diagnosis: the same preview was reported as a 0.66% → 0.27% *improvement* two commits earlier, and the only delta between those commits was a Kotlin test fixture and two golden HTML files. An indeterminate progress animation captured at an arbitrary phase varies by more than the 0.25 pp limit; nothing in this diff reaches the Remote Compose players. No code change was made for it.

Everything else that has finished on `fad6a2c` is green; `Build CLI`, `Module Unit Tests` and the integration matrix are still running.

---

PR #4297, opened 2026-08-19
feat(design-artifacts): publish a design reference on the ground its sticker stands on

yschimke:
A `@Preview` that declares `showBackground` is drawn on an opaque ground — a full frame, or, on a round Wear device, the watch face inscribed in it. Its Figma counterpart is a node export, and a node export carries only the node.

Both sides are cropped to their content box before scoring, so a kit cell that draws a clock strip on a specimen board rather than on a watch gets **enlarged to the size of a watch face and compared with one**. The published verdict on those rows was a statement about the missing ground.

On wear-m3-catalog that is `TimeText`, `ScrollIndicator`, `LevelIndicator`, both `PageIndicator`s and `CircularProgressIndicator`: references 0.2–12% opaque against stickers that are 78.3% (the inscribed disc), scoring **21.87–35.66%, every one `off`**.

`--reference-backdrop '#000000'` (workflow input `reference-backdrop`) lays that ground under the reference before it is published. Empty stays the default and changes nothing for any existing catalog.

## Before / after

Checkerboard so transparency is visible. The percentage under each pair is the **real publish-time score** — the same `design-reference-score.mjs` that bakes `match` into `references/index.json`, driving the viewer's own `format-compare.js`.

Reference backdrop before and after, with the published match score under each pair

The last two rows are the controls. `AlertDialog`'s kit cell already draws the face, so its score does not move; `Button/Filled` is a transparent component sticker, declares no ground, and comes back untouched.

## The colour is declared, the shape is recognised

Only the catalog knows what its previews asked for, and a colour read off the sticker would be read off the very image the reference is about to be compared with — so the caller names it.

The shape comes from the sticker's alpha, and `stageOf` accepts **exactly the two shapes `showBackground` can produce** — the whole frame, and the disc inscribed in a square frame — rather than whatever it finds. That is the safety argument, not a detail: a transparent component sticker has an alpha channel shaped like the component, matches neither, and is left alone. Across wear-m3-catalog's 380 published stickers the classification is unanimous:

| | count |
|---|---|
| device disc — every breakpoint of every full-screen component | 215 |
| no ground declared — left exactly as they were | 165 |
| full-screen stickers missed, component stickers touched | **0** |

Anything else — the Wear scroll capsule's vertical stadium, a partial scrim — is skipped and counted, never approximated, and the step says so rather than leaving it to be inferred from unchanged bytes:

```
design-references: backdrop #000000 laid under 35 device-mask and 0 full-frame reference(s);
  151 sticker(s) declare no ground and were left transparent
```

## Why `max(0, ground − reference)` and not source-over

The two sides describe the same silhouette, so `over` stacks a coincident edge against itself. A reference that already carries its own watch face has a rim alpha of 47 where the sticker's is 27, and `over` promotes that to **69** — a harder edge than either image has, on ~1300 rim pixels of every such reference. An earlier revision of this branch did exactly that and cost all 17 already-correct references 0.02–0.04 points of match. Filling only what is missing leaves their published number where it was.

## Measured

Publish-time scorer, over wear-m3-catalog's 35 device-mask references:

| | before | after |
|---|---|---|
| 17 whose kit cell already draws the face | 82.31 – 98.50% | **±0.00, unchanged** |
| 18 whose kit cell does not | 21.87 – 35.66%, all `off` | 96.48 – 100%: 13 `match`, 4 `close`, 1 `off` |

Mean **58.92% → 96.43%**. The one that stays `off` is `CircularProgressIndicator`'s default cell at 96.48%, which is now a statement about the ring rather than about the ground it was missing.

Every scored pair is tabulated in [`docs/design/evidence/reference-backdrop/`](https://github.com/yschimke/compose-ai-tools/blob/agent/wear-m3-catalog-figma-backgrounds-vvaagg/docs/design/evidence/reference-backdrop/README.md), with a no-render reproduction straight off the delivery branch.

## Test plan

- `node --test reference-backdrop.test.mjs` — 12 tests: the three colour spellings and the off-switches, a refused typo, both recognised shapes, a transparent component sticker and a capsule both refused, the antialiased rim tolerated, the fill, the soft-edge blend, and the two invariants that keep it safe to turn on for a whole catalog (no-op on a reference that already covers the stage; the rim tracks the sticker's coverage rather than hardening).
- `node --test png-resample.test.mjs design-references.test.mjs reference-backdrop.test.mjs` — 59 pass, no regressions in the siblings this sits between.
- The before/after and every number above were produced by running the real `stageOf` / `applyBackdrop` and the real `design-reference-score.mjs` against the current `design-artifacts/wear-m3-catalog` delivery branch — published bytes, not fixtures.
- Behaviour is inert without the flag: `reference-backdrop` defaults to empty, and no caller in this repo sets it.

The consumer is yschimke/wear-m3-catalog, which needs this input to exist and so merges after this one.

---
_Generated by [Claude Code](https://claude.ai/code/session_018wx4PdYHyT19r7r7L6PdGT)_


yschimke:
fix ci failures

PR #4339, opened 2026-08-20
feat(serve): make the catalog blob cache clearable and legible

yschimke:
Phase 4 of [the catalog content cache plan](https://github.com/yschimke/compose-ai-tools/blob/main/docs/design/CATALOG_CONTENT_CACHE.md). Follows #4303 (phase 1), #4316 (phase 2) and #4322.

Three operator surfaces the cache was missing.

## `DELETE /admin/catalog-cache`

Discards every cached byte, gated by the admin token like the rest of that surface. Responds with the pool's state afterwards, in the same shape `/status.json` reports, so an operator can see it took.

**Whole-pool, not per catalog** — and that is not a shortcut. Blobs are named by their own digest and shared between systems on purpose: a font fetched for one catalog is the same file the next one reads. So no blob has an owning catalog to delete it by, and partitioning the pool to make a `/<system>` route possible would give up the deduplication, which is worth more than the button. Everything dropped is re-fetchable, so the cost of using it unnecessarily is bandwidth.

## `POST /<system>/refresh?force=1`

Re-reads a catalog even when its branch head has not moved, by dropping the recorded head exactly as trust revocation already does via `forgetHeads`. The short-circuit in `checkOne` is what makes polling cheap and is right almost always, but it left no way to say *read it again anyway*. The two routes compose into the full prove-it-from-the-branch sequence: clear the pool, then force a refresh.

## `catalogCache` on `/status.json`

A whole-box row: `blobs`, `bytes` against `maxBytes`, `hits`, `misses`, `writes`, `evicted`, `corrupt`. Whole-box like `themeCache`, and for the same reason — a site host scoped to one catalog has nothing of its own to report.

The pair to watch is `branchFetch.cached` against `catalogCache.hits` — the same event counted from two ends — and both against `branchFetch.attempted`. Hits climbing while `attempted` flattens across a restart is the feature working. `corrupt` above zero says a volume is losing bytes, since every blob is re-verified against its own digest on read.

Null on a server publishing no catalogs: its pool isn't merely empty, nothing will ever touch it, and a row of zeroes reads as a cache that's failing rather than one nobody asked for.

## Two names in the plan I did not ship

The plan promised `?flush=1` and `DELETE /admin/catalog-cache/<system>`. Both overstate what's expressible here, and shipping them as written would have been a lie in the API. The design doc now records why:

- **No per-system delete**, for the dedup reason above.
- **`?flush=1` became `?force=1`.** A per-catalog *flush* would mean "ignore the cache for this load" — a bypass threaded down to every read — and it buys very little: the cache is commit-addressed and every blob re-verifies against its own digest, so "cached bytes are wrong" is structurally prevented rather than merely unlikely. What an operator actually lacks is the force. A route named `flush` that evicts nothing would read as a cache control and behave as a poll, and the next person to reach for it would be reaching for the wrong thing.

## Test plan

- `./gradlew :cli:test ktfmtCheckAll` — **2,685 tests, 0 failures**, BUILD SUCCESSFUL.
- **`ServeAdminRoutingTest`** (2 new, against a real `CatalogBlobPool` on disk) — clearing drops the blobs and reports `"blobs":0`, with the bytes actually gone rather than merely unreferenced; and the route 404s without the admin token (and with a wrong one) on a `--public` server, having cleared nothing.
- **`ServeHttpRoutingTest`** (1 new) — `?force=1` reaches the refresher through both the path and `?session=` forms, and anything other than an explicit `1` is an ordinary refresh, so a stray query cannot force one.

No visual evidence: admin routes, a status field and docs — nothing renders.

## What's left

Phase 5 (background audit: adopted-commit reachability plus a sampled byte comparison) and phase 3 (generation snapshots), in that order of confidence. Phase 3 is still the one I'd hold — with phases 1–2 warm, a boot pays one `git ls-remote` per catalog (irreducible: that *is* the freshness check) plus staging-directory assembly from local bytes, and nobody has measured the second. This PR ships the row that would measure it.

---
_Generated by [Claude Code](https://claude.ai/code/session_011PoRMxwHiYqJq2WdXfEN64)_


yschimke:
Fix pr comments

PR #4348, opened 2026-08-20
feat(serve): link the Source panel's snippet to its KDoc pages

### LINKED ISSUE #4331: Can we link to Kdocs and maybe show the API docs as another tab?
### What went wrong

This is useful https://developer.android.com/reference/kotlin/androidx/wear/compose/material/Button.composable

### Base render

render

### Server

| | |
| --- | --- |
| compose-preview | `1.22.0` |
| Mode | public (open) |
| Uptime | 2h |

### Page

| | |
| --- | --- |
| Page | [`/wear-m3-catalog/p/button-imagebackground__ideal__default?fontScale=1`](https://preview.coo.ee/wear-m3-catalog/p/button-imagebackground__ideal__default?fontScale=1) |
| Preview | `button-imagebackground__ideal__default` |
| Catalog rendered by | compose-ai-tools 1.22.0 |
| Render lane | live daemon |

### Browser

| | |
| --- | --- |
| User agent | `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36` |
| Viewport | 2560×1330 CSS px |
| Device pixel ratio | 1 |
| Page colour scheme | light |
| OS colour scheme | light |

yschimke:
Fixes #4331.

## What this does

The viewer's Source panel shows the plain Compose behind a card, but nothing said what `Button` in it actually *is* — a reader wanting the API had to leave and search for it.

The snippet already knows. `PlaygroundSourceCleaner` prunes its imports to exactly what the usage code touches, which is the only place this server can tell that a Wear catalog's `ImageBackgroundButton` renders `androidx.wear.compose.material3.Button`. New `ApiDocLinks` reads those imports back off the snippet, resolves each to its reference page on `developer.android.com`, and `/usage/<id>` carries them so the panel lists them under the code — composables first, so the component the card is about leads.

**Not the docs tab the issue also asked about**: `developer.android.com` refuses to be framed, so a tab could only ever hold links anyway — and links belong beside the code that names them, where the reader has just met the symbol.

## The part that needed care: two page shapes, and the wrong one 404s

A top-level `@Composable` is published at `<pkg>/<Name>.composable`; a class / interface / object / annotation at `<pkg>/<Name>`. An import carries no signature, so the kind is inferred from how the snippet *uses* the name:

* qualifier (`ButtonDefaults.buttonColors()`), annotation (`@Composable`) or type (`: Modifier`) ⇒ declaration page — a composable is never any of those;
* called in **statement position** ⇒ composable page. Statement position is decided by the character *before* the call, not by the start of a line: a constructor called for its value is always part of a larger expression (`color = Color(0xFF…)`), including when it wraps onto a line of its own;
* neither ⇒ no link. `shape = CircleShape` names a property, and dokka gives properties no page.

Comments and string literals are blanked first — a `Slider.kt` mention in a KDoc read as a qualifier, and `contentDescription = "Add"` made an `Icons.Filled.Add` import look like a symbol the code used by name.

**Measured, not asserted**: run over 244 live snippets from every catalog on the public preview host, the resolver produces 220 distinct pages and **zero 404s**. `ApiDocLinksTest` pins each call-site shape that was wrong before it was added, so a later simplification fails rather than silently starting to publish dead links.

## Visual evidence

The harness's existing Source-panel capture now stubs `apiDocs`, so this surface is on a diffed baseline from here on rather than only in this PR. Before / after, same base commit, both themes:

**Light — before**

Source panel, light, before

**Light — after**

Source panel, light, after

**Dark — before**

Source panel, dark, before

**Dark — after**

Source panel, dark, after

## Test plan

* `./gradlew :cli:test` — green, including the new `ApiDocLinksTest` (11 cases, each a page shape that 404'd before the rule that fixes it).
* `./gradlew :cli:ktfmtCheck` — green.
* `cd cli/serve-web && npm run verify` — typecheck + 929 mocha tests (4 new in `viewerApiDocs.test.ts`, covering the pinned-host check that keeps a catalog from putting an arbitrary href on the page) + the committed bundle asserted in sync.
* `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot -g serve-viewer-source` — the captures above.
* Live measurement: 244 `/usage/<id>` responses fetched from `preview.coo.ee` across 19 catalogs, every produced URL HEAD-checked (220 unique, all 200).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EQDGHVyAHJsEfaQJ5netHb)_


yschimke:
Fix pr comments

PR #4354, opened 2026-08-20
fix(serve): keep the front door's compare action for provider-neutral references

yschimke:
Follow-up to #4349, addressing both review findings left on it after it merged.

## P1 — provider-neutral references lost the action

The card's compare action was gated on the design tool its references *name*, so a catalog publishing references whose `source.provider` is `png`, `svg`, `file` or any unmapped token lost the action entirely — even though its `compare?format=reference` route works, and the catalog landing has always offered exactly that under the neutral "compare to design references" label.

Availability and the vendor label are two questions, and the landing page already keeps them apart. Split them here the same way:

- `HomeSystem.hasReferenceComparison` is the **gate** — the same condition `comparisonPage` turns the `reference` format on with, so the action can never deep-link a format that page does not offer.
- `designToolLabel` is only ever the **label**, falling back to the landing's own neutral wording rather than suppressing the action.

`CatalogMeta` carries both, computed at snapshot time for the same reason as before (`peekHost` never resumes a suspended catalog).

## P2 — repeated links were announced identically

Several cards naming the same tool exposed several sibling links announced only as "compare to Figma". The tile that gives each one its context is a *sibling*, so nothing labelled the chip by it — a screen-reader link list or a voice command had nothing to tell them apart.

The accessible name now leads with the catalog (`Compose Material 3: compare to Figma`) while the visible text stays short. The visible string survives intact inside the accessible name, so WCAG 2.5.3 Label in Name holds and "click compare to Figma" still matches — a name like "compare Wear Material 3 to Figma" would read fine and break that.

## Visual evidence

The fixture's `homeassistant-remotecompose` now publishes provider-neutral references, so the golden pins the case that regressed. Crop of the `yschimke` section — the neutral chip renders, and `MeshCore` (no references) keeps its reserved empty row so both tiles still end level:

Front door, neutral label, light

Front door, neutral label, dark

The Design Systems row is unchanged from #4349 apart from the added `aria-label`, which has no pixels.

## Test plan

- [x] `./gradlew :cli:test ktfmtCheckAll` — green
- [x] `ServeWebTest`: the existing case gains a `png-kit` catalog that compares with no tool to name (asserts the neutral label and the working link); a new case asserts the two accessible names differ, lead with the catalog, and still contain the visible string verbatim
- [x] `ServeWebFixtureTest` goldens regenerated — `serve-home-index` and `serve-component-browser-home` only
- [x] `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot` — 147 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_013WZo8KqkvPbt8sq1GJBGGj)_


yschimke:
Can you move the link onto the card.also make sure all card are the same site.

Also can you increase the zoom of the hero component so on hover it slightly breaks out of the card, and with a transparent background.

PR #4366, opened 2026-08-20
feat(serve): add a GitHub-gated endpoint for uploading preview images

yschimke:
An agent preparing a PR has the pixels — a rendered PNG on disk — and no URL to embed them from. `share-preview` covers that with a gist or a pushed capture branch, both of which need credentials a hosted agent session often lacks (`gh`, or push rights). `--accept-images` is the third mechanism, and needs neither:

```bash
compose-preview serve --public --accept-images --image-upload-repo yschimke/compose-ai-tools

curl -sS -H "Authorization: Bearer $(gh auth token)" \
  --data-binary @build/renders/android/ButtonPreview.png \
  'https://preview.coo.ee/images?name=after.png'
# {"id":"K3f…","url":"https://preview.coo.ee/i/K3f….png","width":360,"height":128,
#  "markdown":"[after.png](https://preview.coo.ee/i/K3f….png)","uploadedBy":"yschimke","expiresIn":"7d", …}
```

`markdown` is handed back finished, so a caller never assembles it itself — and never trips the backticked-URL trap the PR-body rule calls out.

## Locked down: uploading is never anonymous

Unlike the document drop-box beside it, this lane hands out hosting on the operator's origin, so it is gated on a **GitHub account** — on a `--public` host too.

- The caller sends `Authorization: Bearer <github-token>`; the server asks GitHub who that is and whether they have real access to `--image-upload-repo` (falling back to `--github-auth-repo`). The bar is the playground's existing rule, unchanged and shared: write access on a public repo, any real grant on a private one. No access → `403`.
- **A bearer token, not the OAuth cookie**, because the whole audience is headless — a CI job or cloud session holding a `GITHUB_TOKEN`, not a browser that can round-trip a redirect. That also means the lane needs **no OAuth app**: a box with no `--github-auth-*` config can run it. Without a repository to gate on it **refuses to start**.
- The token is used for two GitHub reads and dropped — never stored, logged, or echoed back; the verification cache is keyed by its SHA-256 and bounded against a token spray.
- Per-account budget (60/min default, `--image-rate-limit`), which bounds both abuse and the GitHub API calls an uncached upload costs.

**Reads stay open on the unguessable 128-bit id, and that asymmetry is deliberate.** GitHub renders an embedded image by fetching it through its own proxy, *anonymously* — an authenticated read would mean the PR body never paints, so the feature would not exist. The document lane's trick of appending the host's browse token to the permalink is worse than useless here: the destination is a pull request, so it would publish that token to everyone who can read it. Nothing is listed or browseable; there is no index of what the box holds.

## The rest of the envelope

- **Content-sniffed**: PNG (APNG served as `image/apng`), GIF, WebP, JPEG. SVG deliberately excluded — it is an image to a human and a scriptable document to a browser. `?name=` is a label only, never a path and never the format decision.
- **Bounded**: 8 MB per image, 256 images, 128 MB heap, oldest-first eviction; 7-day links (`--image-ttl`), in memory, `nosniff` on the way out. The TTL caveat is documented: camo caching is a cache, not a guarantee — evidence that must live forever still belongs on a SHA-pinned capture branch.
- No `?url=` leg: the caller already holds the bytes, so a server-side fetcher would add an SSRF surface to buy nothing.
- `/status.json` reports the lane, its gating repo, and what it holds; `SERVE_ACCEPT_IMAGES` + `SERVE_IMAGE_UPLOAD_REPO` wire it on the deployed image.

## Visual evidence

The lane has no page of its own — two routes answering a script, and a browse view over other people's uploads is the one thing an unguessable-link store must not grow. Its entire visual surface is one row in `/status`'s Configuration block, captured from a real `compose-preview serve` run against this build ([evidence](https://github.com/yschimke/compose-ai-tools/tree/cccecd122686e8b4b540101a23d1a9937a020fe1/docs/design/evidence/serve-image-lane)):

| image lane off | image lane on |
| --- | --- |
| status with the image lane off | status with the image lane on |

That capture earned its keep twice: the first one had the gating repo in the cell, where it overran its own label, and rendered the default TTL as `168h` — so the value is now short and `ServeWeb.humanDuration` grew a days branch.

## Test plan

- `:cli:test` — **2771 tests, 1 failure**, and that failure (`ServeWebFixtureTest`, `serve-design-page.html` golden drift) also fails on a clean `origin/main`; verified by stashing this branch and re-running. It is untouched by this change.
- New: `ServeImageRoutingTest` (8) — 401 without a credential, 403 for an account without repo access, 201 + absolute URL + markdown for a collaborator, anonymous `GET` of the bytes with `nosniff`, 400 for a non-image, 404 on expiry and on a suffix that isn't the stored format's, 429 past the budget, token never echoed, routes absent without the opt-in. `ServeImageStoreTest` (7), `ServeImageFormatsTest` (4).
- Live smoke against the built CLI: startup refusal with no repo, `401` with no credential, and a real GitHub round-trip for a bogus token. The sandbox's egress proxy rewrites GitHub API auth, so the positive path could not be exercised end-to-end here — it is covered at the route level, and the access rule itself is the one the playground already ships.

## Follow-ups (not in this PR)

`share-preview` could grow a `--mechanism serve` that POSTs each image and rewrites a report's markdown to absolute URLs, so an agent gets this without hand-writing `curl`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wm56KASJ2s6kiHn82J1y2v)_


yschimke:
This merged while the review fixes were in flight, so they land in #4368 instead.

Taken there: installation tokens (`${{ github.token }}` has no user, so `GET /user` refused the lane's primary CI caller — now verified on the installation's write permission on the gating repo); verification metering (the per-login budget was only charged after identification, so a spray of unique invalid tokens bought one synchronous GitHub call each — now there's an address-keyed budget before verification); negative `--image-rate-limit`; and the misleading "nothing to serve" exit when `--image-upload-repo` is missing. The compose passthrough was already fixed in `ec3e80a` before this merged.

Not taken: publishing the flags in `yschimke/skills` — operator flags for running a serve host live in `docs/public-preview-server.md` alongside `--accept-docs`, `--accept-bundles` and the `--playground-*` set, and the skill tracks the *released* CLI, which this flag isn't in yet. Reasoning in #4368.

---

PR #4380, opened 2026-08-20
feat(serve): open the design-spec lane on the triptych

### LINKED ISSUE #4376: Make tryptych the default Figma view
### Base render — you were on the design spec (triptych)

render

### Server

| | |
| --- | --- |
| compose-preview | `1.25.0` |
| Mode | public (open) |
| Uptime | 49m |

### Page

| | |
| --- | --- |
| Page | [`/wear-m3-catalog/p/button-loading__ideal__outlined?bg=on&mode=spec&specView=triptych`](https://preview.coo.ee/wear-m3-catalog/p/button-loading__ideal__outlined?bg=on&mode=spec&specView=triptych) |
| Preview | `button-loading__ideal__outlined` |
| Catalog rendered by | compose-ai-tools 1.24.0 |
| Render lane | live daemon |
| View | design spec (triptych) |

### Browser

| | |
| --- | --- |
| User agent | `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36` |
| Viewport | 2560×1330 CSS px |
| Device pixel ratio | 1 |
| Page colour scheme | light |
| OS colour scheme | light |

### Recent failures

```
2026-08-20 12:56 UTC  meshcore-mobile: render failed after 3321ms — render failed: NoSuchMethodError: 'void androidx.compose.remote.player.compose.RemoteDocumentPlayerKt.RemoteDocumentPlayer(androidx.compose.remote.core.CoreDocument, int, int, androidx.compose.ui.Modifier, int, kotlin.jvm.functions.Function1, kotlin.jvm.functions.Function1, kotlin.jvm.functions.Func
```

yschimke:
Fixes #4376.

The design-spec lane's default view was `spec` — the imported reference alone on the stage. That answers *do these agree?* only by asking the eye to hold one frame while looking at the other, which is why the Figma chip had already grown a workaround: it forced `diff` on entry so the divergence it states ("Figma 96.3%") had somewhere to be seen. The triptych is the better answer one rung further out — it *is* the diff, with the two frames it was taken from either side of it — so it becomes the lane's own default and the chip's override goes away.

## Visual evidence

Both shots are the `pages-snapshot` harness's `spec-lane` state on the `serve-viewer-path` fixture: the lane is entered the way a visitor enters it, by clicking the `Figma` chip, and **no view button is pressed** — so the stage shows whatever the lane decides to open on.

| before — lands on Diff | after — lands on the triptych |
| --- | --- |
| Before: the magenta delta map alone on the stage, Diff pressed | After: Spec, Diff and Render side by side, Triptych pressed |

The dark pair, same state on the dark catalog:

| before | after |
| --- | --- |
| Before, dark | After, dark |

The fixture was already registered with the harness, so this state keeps being diffed by the visual-diff bot on every later PR without anyone remembering to ask; the pair above is committed under `renders/spec-lane-default-triptych/` with a README recording how it was captured.

## What changed

- **`spec/views.ts`** — `DEFAULT_VIEW` is now `triptych`. That constant was doing two jobs: *which view the lane opens on* (and therefore what `?specView=` may leave unsaid) and *which view paints nothing of its own*. Only the first moves; the second splits out as `PLAIN_VIEW = "spec"`.
- **`SpecCompare.ts`** — the checks that keep the comparison panel and the score off the stage key on `PLAIN_VIEW`, so pressing **Spec** still restores the lane's original behaviour exactly.
- **`viewer.ts`** — the Figma chip no longer requests an entry view, and `syncUrl` asks `viewParam()` what may be omitted instead of hardcoding `"spec"`. A link to the reference alone now carries `?specView=spec`; a link to the triptych carries nothing. A URL that names a view still wins over the default, as it won over the chip's request.
- **`ServeWeb.kt`** — one `SPEC_DEFAULT_VIEW` constant drives the served `aria-pressed` and `data-view`, so the page arrives pressing the button the script keeps pressed rather than flipping on first paint.
- **`ServeBugReport.kt`** — a spec-lane URL naming no view is not silent about which of four pictures the reporter was looking at, so the **View** row resolves the default (`design spec (triptych)`) instead of dropping it. The reference-alone view reads `design spec (reference only)` — "design spec (spec)" is a stutter.
- Page fixtures regenerated, docs updated.

`main` has been merged in; the only conflicts were the committed `serve-web` bundles, which are build output and were regenerated from the merged sources rather than hand-resolved (`check-committed-bundle.mjs` confirms they match a fresh build).

## Test plan

```
cd cli/serve-web && npm run typecheck && npm test
# 959 passing (post-merge)
node scripts/check-committed-bundle.mjs
# all six bundles up to date

./gradlew :cli:test --tests '*ServeWeb*' --tests '*ServeBugReport*' ktfmtCheckAll
# BUILD SUCCESSFUL

cd vscode-extension && HARNESS_FIXTURE=serve-viewer-path npx playwright test \
  -c preview-harness/playwright.config.mjs pages-snapshot.spec.mjs --grep 'snapshot ·'
# 2 passed — the shots above
```

New tests pin both halves of the split: that the lane opens comparing and scores the pair without being asked, that **Spec** still leaves the stage alone, and that `DEFAULT_VIEW`/`PLAIN_VIEW` do not collapse back into one constant.


yschimke:
fix merge conflicts

PR #4386, opened 2026-08-20
feat(cli): scope preview-server confirmations, and name this repo's host

yschimke:
Two halves of the same decision: tighten the trust gate from #4379, then use it here.

## Confirmations can be scoped to a repository

`COMPOSE_PREVIEW_SERVE_HOSTS=preview.coo.ee` is a standing grant — every checkout of every repository on that machine, forever. An entry may now be `owner/repo=host`:

```bash
export COMPOSE_PREVIEW_SERVE_HOSTS=yschimke/compose-ai-tools=preview.coo.ee
```

which confirms only while the checkout's `origin` is that repository. The identity comes from `.git/config`, which **no pull request can edit** — that's what makes it usable as an identity at all. Host comparison stays exact; a malformed entry (`=host`, `not/a/repo=host`, `owner/repo=`) grants nothing.

**The bare form still works, deliberately.** What it risks is much narrower than the gate itself: a branch of some other repo can name a host you already confirmed and cause an upload you didn't ask for — it cannot hand anyone your credential, because the destination is still a host you trust. That's worth narrowing, not worth breaking every CI image over. The refusal message offers the scoped form whenever it can identify the repository, so the tighter grant is the one that's easy to copy:

```
export COMPOSE_PREVIEW_SERVE_HOSTS=yschimke/compose-ai-tools=preview.coo.ee   (that host, for this repository)
```

## This repo names its preview server

```properties
composePreview.serveUrl=https://preview.coo.ee
```

**Nothing changes for anyone until a machine confirms the host** — the setting is a *value*; the decision to act on it comes from outside the checkout. Until then `share-preview` keeps making gists or capture-branch pushes and says why once, and `doctor` reports a warning. Where the confirmation belongs is now documented per context (developer machine → shell profile or `~/.gradle/gradle.properties`; Actions → a repo/org **variable**; agent sandbox → the environment's own variables), along with why it deliberately **isn't** an export in a checked-in script: that would be the checkout confirming itself.

## The rule is written down

`docs/AGENTS.md` gains the convention this whole episode was missing: **a setting read out of the checkout supplies a value, never trust.** It names `composePreview.serveUrl` as the worked example and the **version pin** as the deliberate exception — also checkout-controlled, but it only changes what a *build* injects, and running a build on a hostile checkout is already arbitrary code execution, so gating it would buy nothing. The next project-scoped setting should have to answer which side of that line it's on.

## Test plan

- Full `:cli:test` green.
- `ProjectPreviewServerTest` now 19: a scoped grant confirms its own repo, refuses a different one, and refuses a checkout with no identifiable repo; a bare entry still confirms anywhere; the refusal offers the narrower form; four malformed entry shapes grant nothing.
- Live in this checkout, all three states: unconfirmed → refused, offering `yschimke/compose-ai-tools=preview.coo.ee`; confirmed for this repo → passes the gate to the credential step; confirmed only for `someone/else` → still refused.
- That live check earned its keep: it caught that the `originRepo` wiring had landed in `doctor` but silently failed to apply in `share-preview`, so the two would have disagreed about whether a scoped confirmation applied.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wm56KASJ2s6kiHn82J1y2v)_


yschimke:
Fix pr comments

PR #4389, opened 2026-08-20
fix(harness): regenerate the capsule fixture so it names Roboto Flex

yschimke:
`WearScrollSvgGrowthTest` has been red on `main` since #4342, failing the `Renderer Android Unit Tests` and `Module Unit Tests` checks on every PR. It reads as a font regression. It is the opposite — the render got **better** and the committed fixture stayed behind.

## Reading the failure correctly

The drift guard is `assertEquals(message, expected, actual)` with the **fresh** render as `expected` and the **committed** file as `actual` — the reverse of how a failure report is usually read. So:

```
expected: font-family="'Roboto Flex', sans-serif"   ← freshly generated
actual:   font-family="sans-serif"                  ← the committed fixture
```

says the fresh SVG names Roboto Flex and the fixture doesn't. Nothing failed to resolve a font, and the seeding path (`PixelSystemFontAliases` / `SystemFontFamilies.unseeded`) is not involved — this test drives neither render lane that calls `seedSystemFonts()`.

## Why it drifted

The fixture was last regenerated in #4310. #4342 then added `deviceFontFamilyName` to the `compose/semantics` producer so a node drawn through `Font(DeviceFontFamilyName("roboto-flex"))` reports a typeface at all — as that PR puts it, before it "every text node in a family declared that way reported **no typeface at all**", and that shape "is how Wear Compose Material 3's own type ramp declares Roboto Flex, so it covered every Wear preview's text".

The committed fixture is a snapshot of that earlier era. Regenerating is the fix.

## The change

31 of the capsule's 32 text nodes gain the family; the remaining one inherits. **No renderer or producer code changes** — the diff is 32 lines, every one of them the `font-family` attribute.

The stitched Wear capsule, rendered from the committed fixture

**That screenshot does not show the change and is not offered as proof of it.** Roboto Flex isn't installed in the container that captured it, so the browser falls back to its own sans-serif for the old and new fixture alike — the two PNGs differ by 32 bytes and are indistinguishable by eye. It's here to show the capsule still renders correctly after regeneration. The textual diff is the evidence, and CI's `vscode-preview-diff` bot renders this fixture on every PR.

## Test plan

- `:renderer-android:testDebugUnitTest --tests '*WearScrollSvgGrowthTest*'` passes **without** `UPDATE_WEAR_SCROLL_FIXTURE`
- full `:renderer-android:testDebugUnitTest` — 269 tests, this was the only failure and it is now green
- write-up and capture commands in `docs/design/evidence/wear-capsule-typeface/`

> Note on this PR's own checks: `Renderer Android Unit Tests` and `Module Unit Tests` show as **skipped** here, because the affected-module resolver path-filters them and this PR touches only a fixture HTML plus docs. The fix takes effect for other PRs once this lands on `main`.

https://claude.ai/code/session_011pNZkQkWiGChd3GUUbfg2b


yschimke:
The one row in the visual-diff comment above is a **false positive** — this PR touches only `serve-wear-scroll-long-capsule.html` and `docs/`, nothing that could reach `serve-landing-tree-depth-component-open`.

That capture is non-deterministic. Rendering it repeatedly off a single unchanged commit:

| run | hash |
| --- | --- |
| 1 | `2c8a7383…` |
| 2 | `2c8a7383…` |
| 3 | `938a1969…` |
| 4 | `fc10b33c…` |

Three distinct hashes, no code change between them. A differing pair has 151 differing scanlines spanning y=35–760, and the visible difference is the **header block landing ~23px lower**: the breadcrumb row sits at y=54 in one capture and y=77 in the other, with the title overlapping it in the second. That reads as the screenshot occasionally being taken before the Lit chrome components (which are `display: contents`, so their upgrade reflows the header) have settled.

It flapped the same way on #4375, where it was equally unrelated to the diff. Worth its own fix — the bot puts a spurious "changed" row on unrelated PRs, which is how a visual-diff bot gets ignored. Not fixing it here: the settle would go in the shared `pages-snapshot` harness and affects all 330 captures, so it deserves its own change rather than riding in on a fixture regeneration.

Also worth noting the **absence** of a row for the capsule fixture this PR does change: that's expected and documented. CI runners have no Roboto Flex installed, so the browser falls back identically for the old and new SVG — the change is in the declared `font-family`, which the screenshot lane can't observe.

---

PR #4394, opened 2026-08-20
feat(serve): let an agent ask a human for temporary, scoped access

yschimke:
## The problem

An agent asked to fix something needs to *see* the preview server, and it can satisfy neither gate: a private box wants `--token`, a public one wants a GitHub cookie from an interactive OAuth redirect. So the practical answer today is "paste the operator token into the agent's context" — a permanent, unscoped, unrevocable, unattributable credential that then lives in a transcript.

## The shape

`--agent-grants` replaces that with a link you approve. It's [RFC 8628's device grant](https://datatracker.ietf.org/doc/html/rfc8628) narrowed to this server, which is the right shape for exactly the reason it was invented: the party that needs the credential can't render the authorization page, so it asks for a **link** that someone who can will open.

```
compose-preview auth request --server https://preview.coo.ee --scope live --ttl 2h \
    --label "fix wear-m3-catalog#68"
```

```
Ask a human with access to https://preview.coo.ee to open this and approve:

  https://preview.coo.ee/agent-access/9c2Qk1pTf0Xb7hLm4nRzQA
  verification code: KX7M-9QD4

Waiting for approval (this request expires in 10m)…
```

You open it, check the code, approve. The agent's poll returns a bearer it presents exactly where the operator token would go (`?token=`, `X-Compose-Preview-Token`, `Authorization: Bearer`). **No route changes shape** — the existing gates each learn one new way to say yes.

## Visual evidence

The consent page. Its one job is to make you suspicious of the link that brought you here, so the verification code is the loudest element on it and the agent-supplied purpose is rendered escaped, next to where it asked from:

The approval page — verification code in large type, the agent's stated purpose and origin, one radio per scope rung with plain-language descriptions and what each carries with it, a lifetime selector, and Approve/Deny

**Radios, not checkboxes** — that changed in review and the reason is the page's whole claim to being consent. The scopes are cumulative, so with independent boxes an approver could untick `live`, leave `playground` ticked, and get a grant containing the capability they had just unticked. A control that can describe an outcome other than the one it produces does not belong on this page. The rung is one choice; what it carries is spelled out beneath it. A scope the approver may not pass on is *named* rather than quietly dropped, so a request that comes back smaller than it was asked for is explainable.

Same page in dark:

The same page in dark mode

Approving lands here — the only place the lifetime and fingerprint are stated together, and the token is never on screen:

Access granted — the lifetime, the scopes, the grant fingerprint, and a link back to the status page

Both pages are added to the serve-web fixture set (`serve-agent-access.html`, `serve-agent-access-granted.html`), so the preview-harness screenshots and the visual-diff bot diff them on every future PR without anyone remembering to.

## Two properties carry the design

**The link is not the credential.** It holds a public request id; the token is delivered to whoever proves possession of a device secret that never left the agent. So a link that gets pasted into Slack, logged by a proxy, prefetched, or unfurled grants nothing — the worst an interceptor can do is approve a request whose token still goes to the agent that asked. That's also why approval is a `POST` from the page and never a `GET`: a chat client fetching the URL to build a preview card must not be able to grant access by looking at it.

**The verification code binds the page to the request.** Someone who can get a message in front of you could send *their* approval link dressed as the agent's; they cannot make your agent's terminal print their code.

## Scopes

| Scope | Unlocks |
|---|---|
| `preview` | browse catalogs, replay baked renders, `/status` |
| `live` | anything that puts the daemon to work — the viewer WebSocket, `/render` **with an override or a daemon-only suffix**, `/bundle.zip`, presence pings, the Storybook frame, the theme-render lease |
| `playground` | compile and run a snippet on the box |

Cumulative, and capped twice: by `--agent-grant-scopes` (default `preview,live` — `playground` runs caller-supplied Kotlin) and by the approver's own rights, so nobody delegates what they don't hold.

Rules the gates follow, each of which cost a review round to get right:

- **Scope is checked before, and independently of, whether GitHub auth exists.** Written the other way round, a private box with no OAuth let every grant through the live gates unread.
- **A route that commissions a render wants `live`, even where it looks like a read.** A bare replay of baked bytes stays `preview`; refusing that would break ordinary browsing for a grant given exactly that. There's a test for the floor as well as the ceiling.
- **The ingest lanes are outside the scope system entirely.** `POST /bundles/{name}` and `POST /docs` take the operator token and refuse grants outright — a grant labelled "browse this server's catalogs" must not be able to publish a document or replace a named runtime bundle.
- **Approval, expiry and purging are one atomic step.** Every poll and every `/status` runs a purge, so a lookup outside that lock is racing on the normal path — and losing it meant the page telling a human "granted" while the grant was unreachable forever.

## Who may approve

Both halves, and neither is satisfiable by a grant: on a non-`--public` box the server's own `--token`, plus a GitHub identity where one is configured. A session alone was not enough — any account the (by default empty) `--github-auth-users` allowlist admits could otherwise open a request and approve it into a server whose browse token it never had. On a private box the practical path is `/status`, whose waiting-request links already carry the token.

A `--public` server with no GitHub auth has neither half and is **refused at startup** rather than becoming a button the internet can press.

## Lifetime

Request TTL 10 min; grant TTL is the approver's choice up to `--agent-grant-max-ttl` (default 8h, hard ceiling 24h; a present-but-unparseable value is a startup error, not a silent fallback to the default). `/status` lists live grants by fingerprint with a Revoke button, plus requests still waiting on you. In-memory, so a redeploy ends every grant. Nothing ever renders a token — `/status`, the log and every error carry a 12-character SHA-256 fingerprint.

The request TTL bounds **how long a human has to decide**, and only that. A *pending* request dies on its deadline — nobody may approve a closed window — while an *approved* one is retained until its token is fetched or its grant expires, so a decision made in the last seconds still reaches the agent that asked. Both sides keep that rule: the CLI holds its device secret for the same span, which is the half I missed the first time.

## The bug this found on the way

Splitting `serverToken` from `linkToken()` in `ServeHttpServer` is not incidental. On a token-gated server **every generated link carries `?token=`** — that's how the UI stays navigable — so the moment a grant could load any HTML page, it would read the operator's permanent credential straight out of the markup. An agent handed twenty minutes of `preview` would have walked away with the keys to the box: precisely the trade this feature exists to abolish. A grant holder's pages are now wired with their own token; everyone else is byte-for-byte unchanged. There's a regression test, and a note in `docs/AGENTS.md` so the next page renderer doesn't reach for the wrong one.

## Client side

`compose-preview auth request|status|token|revoke|forget`, storing the grant per **server origin** in `~/.config/compose-preview/agent-access.json` (mode `0600`, exact-origin match only, written through a temp file and an atomic rename under an advisory lock). `share-preview --mechanism serve` picks it up without `--serve-token`. Transport rules match `ServeImageUploader`: HTTPS or loopback, no redirects, no credentials in the URL.

`--no-wait` persists the request so a later `auth status` or `auth token` collects the token once the human approves; opening a second request for a server keeps the first rather than discarding a secret that is still approvable. `auth status` asks each server whether the grant is still live rather than trusting the local file; an unreachable host reads `unverified`, never a deletion. A write that fails is reported as a failure everywhere, including the paths that used to read "nothing to revoke". `--json` is JSON Lines.

## Test plan

- **103 tests across five new suites**, all green:
  - `ServeAgentGrantStoreTest` (30) — the state machine: delivery against the device secret rather than the link, an unknown id indistinguishable from a wrong secret, approval narrowing but never widening, the cap not evicting the grant it just minted, a late approval still collectable while a pending one still expires on time, control characters kept out of the audit log, and a concurrency test racing approve against purge 40 times.
  - `ServeAgentGrantsTest` (14) — the scope lattice, the approver ceiling, the CSRF seal's three bindings, duration parsing.
  - `ServeAgentGrantRoutingTest` (28) — the whole flow over real HTTP, including the escalation cases: a grant cannot approve or revoke another, a forged seal mints nothing, a wrong secret never collects the token, no page leaks the operator token, a `preview` grant is refused every route that commissions work (and keeps every route that replays), and a raced denial reports a conflict instead of a false assurance.
  - `AgentAccessStoreTest` (22) — origin keying (no suffix match, no scheme/port confusion, IPv6 loopback, credentials-in-URL refused), retention past the approval window, `0600`, corrupt-store recovery, failed writes reported as failures.
  - `AgentAccessClientIntegrationTest` (9) — the CLI client against a real `ServeHttpServer`, so the two independently-declared wire shapes can't drift; plus the `--no-wait` → approve → collect round trip and the revoke-detection path.
- Full `:cli:test` suite green (**2947** tests, 0 failures).
- `ktfmtFormat` clean; the `/status` golden is byte-identical when the lane is off.

**Known non-blocking CI noise:** `CMP/Wasm Parity` flags `IndeterminateCircularProgressRemote`, a continuously rotating arc whose *baseline* also moves between runs. Across six heads of this branch — none touching the renderer, the RC player, wasm, or the parity harness — it has come back green four times and red twice, on the same 51 previews. Root cause and three proposed patches are in [this comment](https://github.com/yschimke/compose-ai-tools/pull/4394#issuecomment-5361425305); the fix belongs in `rc-settle.mjs`, not here.

Design and rationale: [`docs/design/AGENT_ACCESS_GRANTS.md`](https://github.com/yschimke/compose-ai-tools/blob/agent/temp-preview-token/docs/design/AGENT_ACCESS_GRANTS.md). Consumer guidance for agents ships in [yschimke/skills#55](https://github.com/yschimke/skills/pull/55).

Off by default — `preview.coo.ee` picks it up with `SERVE_AGENT_GRANTS=1`.


yschimke:
All twelve review findings are fixed in `46c3658`, each with a regression test. Summarising here rather than replying twelve times; I'm resolving the individual threads.

Three were privilege escalations, and they shared a shape worth naming: a rule that held on the deployment I actually tested and quietly did not hold on another.

| # | Finding | Fix |
|---|---|---|
| P1 | `preview` reached the ingest lanes | `POST /bundles/{name}` and `POST /docs` take `rejectBadTokenForIngest`, which no grant satisfies |
| P1 | Scope unenforced without GitHub OAuth | A presented grant is judged on its own scope **first**, independently of `githubAuth`, on both HTTP gates and the socket |
| P1 | A GitHub session approved without the browse token | Approving a non-`--public` box now requires `--token` **and** the identity |
| P2 | Overflow stranded uncollected tokens | `collected` flag; only denied/collected requests are shed |
| P2 | Control characters in audit labels | Flattened before storage |
| P2 | `--no-wait` discarded the device secret | Request persisted before anything is printed; `auth status` / `auth token` collect it |
| P2 | `auth status` trusted local bookkeeping | Verifies with `whoami`; drops dead grants, `unverified` when unreachable |
| P2 | `--json` emitted two documents | JSON Lines |
| P2 | IPv6 loopback refused | Host from `URI`, not a `:` split |
| P2 | `forget` claimed success on a failed write | Returns the write result |
| P2 | Unlocked read-modify-write | Advisory lock + temp file + atomic rename |
| P2 | Box-wide grants on site-scoped `/status` | Omitted on site hosts, HTML and JSON |

**Correction to the commit message:** its closing line says the store-serialization finding wasn't changed. That's wrong — it was fixed in the same commit (the bullet four lines above describes the lock and atomic rename). The commit body doesn't reach `main` under squash-merge, so I've left history alone rather than force-push.

### On the visual diff

The bot reports three *changed* captures alongside my four new ones. Those aren't from this branch. `serve-landing-tree-depth-component-open` and `serve-viewer-path-spec-typography` take their inputs from fixture HTML and `pages-snapshot.spec.mjs`, and this PR touches neither — its only fixture changes are the two new files. The visible delta is the Figma chip gaining a parity percentage (`Figma` → `Figma 71.5%`), which shifts the segmented control beside it; that landed on `main` after the `vscode-preview/main` baseline was last regenerated, and my branch merged it in. Nothing to do here, but worth knowing the baseline is behind.

New test coverage for the fixes: the live-lane refusal driven over a real WebSocket (the only place that gate is reachable on a box with no playground), the `--no-wait` → approve → collect round trip against a live server, and the revoke-detection path. Full `:cli:test` green at 2930.

---

yschimke:
Round two: all four fixed in `2f8a8b6`, each with a regression test. Resolving the threads.

| # | Finding | Fix |
|---|---|---|
| P1 | `/render` commissioned daemon work on a `preview` grant | Override query or daemon-only suffix now requires `live` — and so does `/bundle.zip` |
| P2 | A short-TTL grant evicted itself on a full map | The new grant is excluded from eviction outright |
| P2 | Device secret dropped before the token was safely stored | Save first, drop the pending record only on success — in **both** collection paths |
| P2 | `auth forget` claimed success over a failed clear | Reports the failure |

**The P1 is the same mistake as last round, one route further along.** The WebSocket and both GitHub gates learned to check scope; `/render` is reached through the token gate alone, so it never did. Worth stating plainly since it's twice now: the lesson isn't "check this route", it's that `rejectBadToken` authorises at `PREVIEW` and every route behind it that spends CPU needs to say so itself.

Two notes on how I read the P1, since both are judgement calls:

**I gated on what the caller asked for, not on whether bytes are baked.** The suggestion was to require `live` "whenever the request cannot replay baked bytes". I went with `requestCarriesOverrides() || wantsDaemonOnlyRenderProduct()` instead. `renderWouldReplayBakedBytes` short-circuits to `true` for any non-HEAD request by design, so it can't answer this question for a GET; and building a GET-shaped equivalent means a null `peekHost` (a suspended session) reads as "needs daemon", which would 403 ordinary browsing every time a session wasn't resident. A bare `/render/<id>.png` for a preview with no baked copy is the catalog serving its own content, at the cost any anonymous visitor imposes on a public box — that's what `preview` was given, so refusing it would withhold something nobody withheld.

**I also gated `/bundle.zip`, which you didn't name.** It renders every preview in the catalog — the same rule and the larger bill. Closing the named hole while leaving its bigger sibling open would have been theatre.

`AgentAccessStore.save` is now `open` so the failed-save path is testable. A wedged filesystem can't express it: a store that can't write can't hold the pending record the test is about. That's how the round-two P2 turned out to be **two** bugs — the waiting path and `collectPending` both dropped the record unconditionally, and only writing the test surfaced the second.

Full `:cli:test` green at 2934.

---

yschimke:
Round three: all seven fixed in `f148b49`, with 8 regression tests. Resolving the threads.

| # | Finding | Fix |
|---|---|---|
| P1 | `presence` and `/iframe.html` still did daemon work on `preview` | Both require `live` |
| P1 | A raced denial claimed nothing was granted | Reports 409 "already decided" |
| P2 | Private + OAuth looped on the agent's tokenless link | Missing front door → instructions, not OAuth |
| P2 | A GitHub bearer shadowed a grant in `?token=` | Each credential source resolved independently |
| P2 | Pending cap was a check-then-act | Checked and inserted under the existing lock |
| P2 | `auth --help` opened a real request | `--help`/`-h` short-circuit before dispatch |
| P2 | `auth revoke` claimed a failed local removal | Reports both halves, exits non-zero |

**The P1 is the third instance of one mistake, so the fix worth naming isn't the routes — it's the rule.** `rejectBadToken` authorises at `preview`; anything behind it that spends daemon CPU has to say so itself. Round one needed it stated on the WebSocket, round two on `/render` and `/bundle.zip`, round three on `presence` and the storybook frame. I've swept the remaining `rejectBadToken` callers rather than waiting for a round four.

Two of these were self-inflicted by earlier fixes, which is worth being explicit about:

- **The redirect loop** is a direct consequence of round one's "require the browse token *and* an identity". Correct as a security rule, and it made the agent's own printed link unusable on a private box with OAuth — the callback returns to the same tokenless URL and redirects again, forever. The missing-token case now gets instructions pointing at `/status`, whose waiting-request links already carry the token.
- **The shadowed grant** broke the `share-preview` integration this PR advertises. The `?:` chain took whatever the bearer held, and `share-preview --mechanism serve` legitimately puts a GitHub token there while the host credential rides in `?token=`.

One process note: this round I lost four edits to a scripted patch that asserted its way out before writing the file, and only caught it because two of the new tests failed. The tests were doing their job; I'd rather say that than let the green tick imply the first pass was right.

Full `:cli:test` green at 2938.

**On the CMP/Wasm parity failure** on `2f8a8b6`: `IndeterminateCircularProgressRemote` moved 0.13% → 0.64%. It's an *indeterminate* (animated) progress indicator, my diff at that commit was agent-grant Kotlin, tests and docs — nothing touching the renderer, the RC player, or wasm — and the immediately preceding run reported 0 regressions across the same 51 previews. The same run also produced a demonstrable capture flake in the harness (`serve-viewer-exploded-controls` screenshotted mid-`rendering…`, with `comparing…` still in the chip). I'm treating it as frame-timing noise; this push re-runs it, which is the confirmation.

---

yschimke:
### CMP/Wasm Parity is failing on this PR and it isn't this PR

Reporting rather than fixing, because the fix belongs in the parity harness and not in a branch about access grants.

**What's failing.** `IndeterminateCircularProgressRemote` exceeds the 0.25 pp movement limit. Across three runs of this branch:

| Head | Result | Baseline → current |
|---|---|---|
| `46c3658` | 0 regressions, 51 unchanged | — |
| `2f8a8b6` | 1 regression | 0.13% → 0.64% |
| `f148b49` | 1 regression | **0.10%** → 1.13% |

**The baseline moved too** — 0.13% to 0.10% — between two runs comparing against the same base. A deterministic renderer cannot do that. Neither side is stable, so this is not a regression measurement; it's two samples of a moving picture.

**Why.** The document is a `RemoteCircularProgressIndicator` — the sibling preview's own caption says the arc *"rotates and changes sweep continuously on the remote clock"*. `rc-settle.mjs` captures once pixels stop moving, and its header already documents this exact failure mode from #3558: *"the same commit scored `VariableWidthRemote` at 0.19% or 2.45% and the advisory check reported improvements and regressions on PRs that changed nothing a renderer can see."* The convergence loop plus the ink expectation fixed the *font-resolution* case. It cannot fix an indeterminate spinner, because that document never converges — the quiet window just accepts whichever phase it catches, in each player independently.

**Not this branch.** My diff touches no renderer, no RC player, no wasm, and nothing under `scripts/design-artifacts/` — across all four commits it is agent-grant Kotlin, its tests, and docs. The first run on this same branch reported zero regressions on the same 51 previews.

**Proposed patch** (happy to open it separately if you want it):
1. **Pin the clock** for RC playback in the parity lane, so both players score the same frame — closest to how `@AnimatedPreview` already handles this and the only option that actually measures parity for animated documents; or
2. **Exempt continuously-animated documents** from the pp threshold in `rc-compare-regression.mjs` via an explicit list, scoring them for presence rather than pixels — cheap, honest, and leaves a named gap; or
3. **Give the settle loop a phase-aware mode** that captures at a fixed offset from first paint rather than waiting for quiet.

(2) is the smallest change that stops this check crying wolf; (1) is the one that makes the number mean something. I'd not recommend simply raising the limit — that hides a real regression in the other 50 rows to accommodate one that was never measurable.

The harness's own visual-diff bot corroborates the timing story from the other direction: each run flags a *different* extra fixture, always caught mid-transition — `serve-viewer-exploded-controls` with `comparing…` and `rendering…` still on screen, then `serve-viewer-connecting` in its connecting state. Different fixture each run is what flake looks like; a real change reproduces.

The check is marked non-blocking, so I've left it. Say the word if you'd rather I take one of the three patches.

---

yschimke:
Round four: all five fixed in `8afd5f0`, with 5 regression tests. Resolving the threads.

| Finding | Fix |
|---|---|
| Cumulative scopes as independent checkboxes | Radios — one rung, with what it carries named beneath it |
| Late approval uncollectable after the request TTL | Approved-but-uncollected records retained until fetched or the grant dies |
| A second request discarded the first's device secret | Requests accumulate (bounded), dropped individually by id |
| Waiting after a failed local save | Aborts before printing the link; `--json` exempt (it prints the secret) |
| Misspelled `--agent-grant-max-ttl` → silent 8h | Present-but-unparseable is a startup error |

**The checkbox one is the best find of the four**, and worth spelling out: on the one page in this server whose entire job is to state accurately what is being agreed to, the control could describe an outcome different from the one it produced. Untick `live`, leave `playground`, and the page says live was withheld while the grant contains it. That's not a rendering nit — the page's whole claim to being consent is that it doesn't do that.

**Rendering the fix caught a second bug in the same change.** I screenshotted the new radios before pushing and found *nothing was preselected* — the default was the requested scope, which matches no radio when the approver's own ceiling capped the request, and with `required` the form couldn't be submitted at all. The default is now the highest rung this approver can actually give. I'd have shipped that if I'd trusted the diff instead of looking at the pixels.

Before / after, from the regenerated fixture:

| Before (checkboxes, `preview` fixed) | After (radios, "also includes") |
|---|---|
|  |  |

On the request-TTL one, the split I landed on: the TTL bounds how long a human has to *decide*, so a **pending** request still dies on the dot — nobody may approve a request whose window has closed — while an **approved** one is retained until its token is fetched or its grant expires, whichever comes first. Both directions have a test.

Full `:cli:test` green at 2944. `CMP/Wasm Parity` came back ✅ on the previous head (0 regressions, 51 unchanged) after failing twice on the two before it, which is the flake confirmation I said the next push would give.

---

yschimke:
Round five: all five fixed in `fac8c7f`, with 4 regression tests plus one existing test rewritten. Resolving the threads.

| Finding | Fix |
|---|---|
| CLI dropped a request the server still held | Records retained for polling to the server's own grant ceiling; the *window* still governs what a human is told |
| Approval and purge not atomic | Lookup, validation and transition under one lock — `deny` too |
| Superseded same-origin grant orphaned | The loser is revoked before it is discarded |
| `auth revoke` read a failed write as "nothing to revoke" | Asked before removing; fatal when removal fails |
| `Retry-After` ignored while polling | Honoured, and a throttle no longer spends an error |

**The first one is a pattern worth naming, because I caused it last round.** I changed a retention rule on the server (keep an approved request until its token is collected) and didn't change the matching rule on the client. The two halves of a two-party protocol then disagreed about how long a request exists, and the failure mode was the exact one the server-side fix existed to prevent — an approval landing in the last seconds and going nowhere. Worth remembering that "fix the retention bug" was only half the change.

**The atomicity one is the sharpest find of the round.** `approve` looked up outside the lock `purge` takes, and *every poll and every `/status` runs a purge* — so the race isn't exotic, it's on the normal path. The window is narrow but the outcome is bad in a specific way: the approval page reports success to a human while the grant is unreachable forever. There's a test that races approve against a purge 40 times and asserts that anything reported as minted is collectable.

One test **changed rather than added**, and I want to flag it rather than let it pass as a green tick: `a remembered request expires with its own deadline` asserted the record was gone at the deadline. That is precisely the behaviour this round changes, so the old assertion was wrong now, not the code. It's been rewritten to pin what the deadline still does govern — the window is closed, `secondsUntilExpiry` is zero, and `auth status` says so instead of counting down. Deleting it would have quietly dropped the coverage.

Full `:cli:test` green at 2947.

---

yschimke:
Round six: all four fixed in `026347d`. Resolving the threads.

| Finding | Fix |
|---|---|
| "Asked from" showed the proxy, not the agent | One `clientAddress()` shared with the rate limiter |
| Superseded hand-over only on one path | Shared helper used by both save paths |
| …and it swallowed the revoke result | Result is read; a failed hand-over warns and names the fingerprint |
| Pending cap evicted live requests | Only dead records are swept; otherwise the save is refused |

**Two of these are last round's fixes caught incomplete, which is the useful half.** I taught the `--no-wait` collector to revoke a grant it was about to orphan, and then (a) didn't do the same on the ordinary waiting path, which replaces the same entry, and (b) wrapped the call in `runCatching` when `revoke()` returns `Result.Err` instead of throwing — so every HTTP and network failure counted as a success and orphaned the credential exactly as before. The fix looked right and did nothing in the failure case it existed for. It's one helper now, used by both paths, that reads the result.

**The proxy one is the best find of the round** and I should have caught it myself. `preview.coo.ee` runs behind Caddy, so `request.origin.remoteHost` is Caddy on every request — meaning "Asked from" rendered the same string for the agent that asked and for anyone else, on the page whose entire purpose is helping someone decide whether a request is theirs. The rate limiter had resolved this correctly all along via `--trust-forwarded-for`; the display path just never asked it. They share one function now, so they can't diverge again.

Worth being explicit that this was **only ever a display bug** — no gate consulted the wrong address, and the security properties (device secret, verification code, POST-only approval) are unchanged. But a decision aid that reads identically in the good and bad cases isn't a decision aid.

One test **replaced rather than added**: `the remembered-request pile is bounded` asserted only that the pile stayed under the cap — which the evicting behaviour satisfied perfectly well. It's now two tests that pin the actual requirement: refuse rather than evict, every original secret survives, and a dead record is still swept so the pile keeps accepting.

Full `:cli:test` green at 2948.

---

yschimke:
Short addendum on the parity noise, because this run produced the cleanest evidence yet — and of a different kind.

`e65633f` reports **1 improvement**: `IndeterminateCircularProgressRemote` 0.84% → 0.06%.

Every reading that preview has produced across this branch, on both sides of the comparison:

| Head | Base reading | PR reading | Verdict |
|---|---|---|---|
| `46c3658` | — | — | ✅ under threshold |
| `2f8a8b6` | 0.13% | 0.64% | ❌ regression |
| `f148b49` | 0.10% | 1.13% | ❌ regression |
| `475fb4b` | — | — | ✅ |
| `8afd5f0` | — | — | ✅ |
| `026347d` | — | — | ✅ |
| `e65633f` | 0.84% | 0.06% | ✅ **improvement** |

Same preview, same 51-document corpus, and **no commit on this branch touches the renderer, the RC player, wasm, or the parity harness**. The readings span 0.06% to 1.13% — roughly 19× — against a 0.25pp threshold.

Two things follow that the earlier failures alone couldn't establish:

1. **The noise is bidirectional.** Until now I could only point at regressions, and "your PR sometimes makes it worse" is at least arguable. An *improvement* of the same magnitude, from the same unchanged code, isn't. The metric is unstable in both directions, which is the signature of sampling an animation at an arbitrary phase rather than of a rendering change.
2. **A green run here is not evidence of correctness either.** The 0.84% base reading in this very run would have been flagged as a regression had it landed on the PR side. Nobody should read "1 improvement" as this PR having improved rendering — it did nothing of the sort.

The recommendation is unchanged and this strengthens it: fix `rc-settle.mjs` (pin the clock for RC playback, or exempt animated documents explicitly) rather than widen the 0.25pp threshold, which would mask real regressions in the other 50 rows. Root cause and three proposed patches are in [the earlier comment](https://github.com/yschimke/compose-ai-tools/pull/4394#issuecomment-5361425305). Still out of scope for this branch.

---

yschimke:
Round eight: all three fixed in `bc168f1`. Resolving the threads.

| Finding | Fix |
|---|---|
| Poll could see a half-published approval | `poll` reads under the transition's lock; write order corrected too |
| `collected` was a deletion trigger | Approved records retained while the grant lives; `collected` only ranks eviction |
| `auth revoke` couldn't resolve a lone pending request | Wider resolver that includes pending origins |

**The first one is worth reporting properly, because my first fix was wrong and the test told me so.**

I did what the comment suggested — reordered the writes so `grantId` is published before `state = APPROVED` — added a racing regression test, and it passed. Then I re-ran that test against the **old, unfixed** code. It passed there too.

Which means it was worthless as evidence. And thinking about why exposes that the reorder was the wrong fix: the window here is two adjacent statements with `poll` reading unsynchronized, so a poller doesn't need any memory reordering to land in the middle — it just has to *interleave*. Ordering the writes narrows the window; it doesn't close it. A test sampling for that window is testing its own luck.

So `poll` now reads under the same lock `approve` holds, which makes the intermediate state unobservable by construction rather than improbable. The write ordering stays as belt-and-braces since it's the right shape anyway. The test stays, reduced from 200 iterations to 50 and renamed, with its scope written into the body:

> HONEST SCOPE: this is a smoke test, not the guarantee. […] A racing test like this one passed against the *broken* ordering too — the window is two adjacent statements, and hitting it by sampling is luck. Kept because it would catch a gross regression (a lock removed, an exception under contention) cheaply, and because the reader should be told which of the two it is.

I'd rather ship a test that says what it is than one that looks like proof and isn't.

On `collected`: it was carrying two meanings and got one wrong. It records that a poll *built* a response — not that the agent received one — so using it to authorise deletion meant a response lost in flight left the retry told `unknown` while the grant was live. Approved records are now kept while their grant lives, and `collected` keeps only its real job (choosing the safest thing to shed under overflow), documented as not a lifetime signal.

Full `:cli:test` green at 2952.

---

PR #4395, opened 2026-08-20
feat(deploy): turn on agent grants wherever there is someone to approve them

yschimke:
Follow-up to #4394, which merged while a further review round was in flight. Three things: switching the lane on for real, the findings that arrived too late for that PR, and the findings this branch's own fixes then attracted.

## Turning it on

#4394 shipped `--agent-grants` **off everywhere**, including on `preview.coo.ee` — the box it was built for. Enabling it needs care, because neither flat default is right, in opposite directions:

- Hardcoding `SERVE_AGENT_GRANTS=1` in the compose file **takes out every public box with no OAuth app configured**, on its next roll. The lane mints credentials, so it needs a human identity to approve against, and on the open profile the server deliberately refuses to start without one.
- Hardcoding `0` is what we have now: a feature nobody can use without first reading the deploy docs.

So the entrypoint derives it. Unset means **on where it can work** — on when this box has an approver, which (matching the server's own rule) is either all three `SERVE_GITHUB_AUTH_*` secrets, **or** a token-gated box (`SERVE_PUBLIC` not `1`, with `SERVE_TOKEN` set), where the operator token holder approves. Only a `--public` box with no GitHub auth has neither, and there the server refuses the lane outright. An explicit value still wins both ways: `0` opts out, and `1` insists on a box that can't and reaches the server's own refusal — the honest failure rather than a silent downgrade.

`preview.coo.ee` already sets those three secrets, so it picks the lane up on its next roll with no `.env` edit. **Set `SERVE_AGENT_GRANTS=0` before merging if you'd rather it waited.**

What *enables* the lane and what *approving* requires are different questions, and the docs now say so with a table: on a private box that also configures OAuth you need **both** the browse token and a signed-in account, because the server checks its front door first and a session is not a substitute for it.

## Why the derivation gets a test

`deploy/image/test-agent-grants-default.sh` extracts the real gate out of `entrypoint.sh` by its leading comment and runs it, so the test and the shipped logic can't drift. It's wired into CI beside the other deploy guards.

A derivation earns a test rather than a comment for the same reason the env-passthrough guard exists: it has no output of its own. Get it wrong and you either break unrelated deployments on their next roll, or quietly ship a dead feature — both invisible until someone goes looking. That is exactly how `SERVE_PLAYGROUND` spent its whole life inert on the prebuilt image.

## The findings that missed the merge

`bc168f1` merged at 22:40; two review rounds landed at 22:36 and after, so these are bugs on `main` rather than review comments.

**A label could reorder what the approval page appears to say.** `sanitizeLabel` stripped C0/C1 and DEL. `U+202E RIGHT-TO-LEFT OVERRIDE` is neither and survives HTML escaping untouched — so a requester, who picks their own label, could reverse the rendering of the text after it and change what the human appears to be agreeing to, **including the scope and lifetime**. The whole Unicode format category (Cf) now goes: exactly the set that renders as nothing while changing everything around it.

**Credentials could land in a checkout.** With no `XDG_CONFIG_HOME` and no `HOME`, the store fell back to `.` — bearer tokens in whatever directory the command ran in, which for an agent is a checkout.

**A malformed grant flag silently became its default — on all three.** The review named `--agent-grant-max-active`; an earlier round had fixed the same thing for `--agent-grant-max-ttl` and I hadn't swept for siblings. Sweeping found a third and worst: `--agent-grant-scopes preivew` is an operator narrowing the box to read-only, and the silent default is `preview,live` — a typo that **widens** what every grant may do. All three now refuse a present-but-invalid value at startup.

## What this branch's own fixes then attracted

Listed separately because that is what they are — each is a consequence of an earlier fix in this same PR, and two of them were the *same bug reached through its own mitigation*:

- **The working-directory leak was reachable three times.** The original `.` fallback, then a relative `user.home`, then a relative `XDG_CONFIG_HOME`/`HOME`. Each fix went to the branch that was pointed at rather than the question underneath. There is now one `absoluteHome` check every candidate passes through, with that history recorded next to it.
- **`auth request --json` died on the machine it exists for.** That mode prints the device secret so the caller can poll without local storage — which is what the no-credential-home refusal recommends. The refusal fired *before* printing, so it opened a server-side request and exited, leaving an approvable link nothing could redeem. Strictly worse than the stack trace it replaced. It also had to learn to hand back the token when it could not store it, rather than reporting `stored: false` and nothing else.
- **Two caps were fighting, then one leaked.** Retention correctly stopped shedding live approvals, so counting every row against the pending cap meant `--agent-grant-max-active` above it could never be reached. Counting only *pending* rows then let denials accumulate unbounded — an operator denying a hostile batch freed a slot with every click. The cap now charges what an anonymous caller can create (pending **and** denied) and excludes retained approvals, which their own limit already bounds.
- **The refusal reached only one of its two callers.** `share-preview --mechanism serve` opens the same store; handled at `Main`'s dispatch boundary now, so the next consumer inherits it. And it fired after the request was opened, leaving an uncollectable row holding a slot in the very map above — so it now runs before.
- **The gate test didn't control its own inputs**, so an ambient `SERVE_GITHUB_AUTH_*` turned its negative cases positive. `env -i` now, verified against the reviewer's reproduction.

## Test plan

- `test-agent-grants-default.sh` — 11 cases: full auth on, no/partial auth off, private+token on, private-by-default+token on, public+token off, private-without-token off, and explicit `0`/`1`/`true` winning in each direction.
- `test-compose-env-passthrough.sh` still passes — the variable stays forwarded.
- New CLI tests for: the no-home refusal naming its remedy, relative paths rejected from **all three** sources, a relative value falling through to an absolute one, the explicit override winning, a two-character credential filename, a bidi-laden label leaving no format character behind, a live grant surviving an anonymous attempt to fill the request map, denials still charged so denying a batch admits no new one, retained approvals not spending the pending budget, and `request --json` printing its secret with the store constructor throwing.
- **Three were verified against the unfixed code** — the storeless JSON test, the gate-isolation fix, and the concurrency ordering — because a test that passes either way is not evidence. The concurrency one is documented in its body as a smoke test rather than the guarantee, since the guarantee is a lock.
- Full `:cli:test` green (**2964**, 0 failures).

**Seven tests across this feature were rewritten rather than added.** Six pinned mechanism instead of promise. The seventh — `a resolved request makes room for a new one` — asserted that denying frees capacity, which *was* the unbounded-growth vector: the suite was defending the bug. Called out because a suite full of mechanism assertions makes every later refactor look like a regression, and because a green tick obtained by removing a test is worth nothing.

Text-only change to a deploy default, a sanitizer, and a credential path — no rendered surface, so no visual evidence. The consent pages this configures are unchanged and already in the fixture set from #4394.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1243,6 +1243,13 @@ jobs:
       # kept the public host's playground pinned to one catalog with the Android modes off.
       - name: Run deploy compose env-passthrough tests
         run: ./deploy/image/test-compose-env-passthrough.sh
+      # The agent-grant lane derives its own default (on where SERVE_GITHUB_AUTH_* gives it a human
+      # to approve against, off otherwise), because both flat answers are wrong in opposite
+      # directions: `1` fails startup on every box with no OAuth app, `0` ships the feature dead on
+      # the box it was built for. A derivation has no output of its own, so getting it wrong either
+      # breaks unrelated deployments on their next roll or quietly disables the lane — both silent.
+      - name: Run deploy agent-grant default tests
+        run: ./deploy/image/test-agent-grants-default.sh
       # The version-control config overlay is opt-in, so nothing exercises it on the way to a
       # release — a wrong source path or mount target is found on a box, after a restart, serving a
       # stale catalog set. Every failure mode is silent: a missing source makes Docker create an
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/AgentAccessStore.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/AgentAccessStore.kt
@@ -27,6 +27,17 @@ import kotlinx.serialization.json.Json
  * Expired entries are dropped on read rather than swept on a schedule: this is a CLI, it runs and
  * exits, and a stale row costs nothing until someone asks about it.
  */
+/**
+ * No safe place to keep credentials could be determined. Carries the remedy, because the only thing
+ * the user can do about it is name a location.
+ */
+internal class NoCredentialHomeException :
+  IllegalStateException(
+    "no user config directory could be determined (XDG_CONFIG_HOME, HOME and user.home are all " +
+      "unset), and credentials will not be written to the working directory. Set " +
+      "COMPOSE_PREVIEW_AGENT_ACCESS_FILE to a path you control."
+  )
+
 internal open class AgentAccessStore(
   private val file: File = defaultFile(),
   private val clock: () -> Long = System::currentTimeMillis,
@@ -342,10 +353,23 @@ internal open class AgentAccessStore(
 
     /**
      * `$COMPOSE_PREVIEW_AGENT_ACCESS_FILE`, else `$XDG_CONFIG_HOME/compose-preview/…`, else
-     * `~/.config/compose-preview/…`. The override exists for CI and for tests; the XDG path is
-     * where a user would look for it.
+     * `$HOME/.config/…`, else the JVM's `user.home`. The override exists for CI and for tests; the
+     * XDG path is where a user would look for it.
+     *
+     * Throws [NoCredentialHomeException] when none of those yields a location, rather than falling
+     * back to the working directory — see the body.
      */
-    fun defaultFile(env: (String) -> String? = System::getenv): File {
+    fun defaultFile(
+      /**
+       * Injected so a test can simulate a JVM with no `user.home`, which it otherwise always has.
+       *
+       * Deliberately **before** [env] rather than appended: several call sites pass the environment
+       * as a trailing lambda, and a new last parameter silently rebinds those to this one instead —
+       * they still compile, and they quietly test the wrong thing.
+       */
+      prop: (String) -> String? = System::getProperty,
+      env: (String) -> String? = System::getenv,
+    ): File {
       env("COMPOSE_PREVIEW_AGENT_ACCESS_FILE")
         ?.takeIf { it.isNotBlank() }
         ?.let {
@@ -353,7 +377,18 @@ internal open class AgentAccessStore(
         }
       val configHome =
         env("XDG_CONFIG_HOME")?.takeIf { it.isNotBlank() }
-          ?: (env("HOME")?.takeIf { it.isNotBlank() }?.let { "$it/.config" } ?: ".")
+          ?: env("HOME")?.takeIf { it.isNotBlank() }?.let { "$it/.config" }
+          // The JVM's own view of the user's home, which on Linux comes from the passwd entry
+          // rather than the environment — so it survives the minimal `env -i` service context that
+          // has neither variable set.
+          ?: prop("user.home")?.takeIf { it.isNotBlank() && it != "?" }?.let { "$it/.config" }
+          // …and if there is genuinely nowhere, REFUSE. The old fallback was `.`, which wrote
+          // bearer tokens and device secrets into whatever directory the command happened to run
+          // in — for an agent that is a checkout, so the credentials land somewhere that gets
+          // archived by CI or committed by the next `git add -A`. Worse, the target directory
+          // would then be attacker-influenced by anything that can add a `compose-preview/` path
+          // to the tree. Better to store nothing and say so.
+          ?: throw NoCredentialHomeException()
       return File("$configHome/compose-preview/agent-access.json")
     }
 
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/AgentAccessStore.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/AgentAccessStore.kt
@@ -375,27 +375,35 @@ internal open class AgentAccessStore(
         ?.let {
           return File(it)
         }
+      // ONE rule, applied to every candidate: a home must be an absolute path.
+      //
+      // This leak has now been reachable three times by three different routes — the original `.`
+      // fallback, then a relative `user.home`, then a relative `XDG_CONFIG_HOME`/`HOME` — because
+      // each fix was applied to the branch that was pointed at rather than to the question. The
+      // question is the same every time: is this a real home, or does it resolve under whatever
+      // directory the command happens to be run from? For an agent that directory is a checkout,
+      // so the credentials land somewhere CI archives or the next `git add -A` commits, and the
+      // target becomes influenceable by anything that can add a `compose-preview/` path to the
+      // tree. Anything that fails the question is treated as absent.
       val configHome =
-        env("XDG_CONFIG_HOME")?.takeIf { it.isNotBlank() }
-          ?: env("HOME")?.takeIf { it.isNotBlank() }?.let { "$it/.config" }
+        absoluteHome(env("XDG_CONFIG_HOME"))
+          ?: absoluteHome(env("HOME"))?.let { "$it/.config" }
           // The JVM's own view of the user's home, which on Linux comes from the passwd entry
           // rather than the environment — so it survives the minimal `env -i` service context that
           // has neither variable set.
-          // Absolute only. `-Duser.home=.` resolves to `./.config/…` — straight back under the
-          // working directory, which is the exact outcome this whole path exists to prevent. A
-          // relative home is not a home; treat it as absent and let the refusal below fire.
-          ?: prop("user.home")
-            ?.takeIf { it.isNotBlank() && it != "?" && File(it).isAbsolute }
-            ?.let { "$it/.config" }
-          // …and if there is genuinely nowhere, REFUSE. The old fallback was `.`, which wrote
-          // bearer tokens and device secrets into whatever directory the command happened to run
-          // in — for an agent that is a checkout, so the credentials land somewhere that gets
-          // archived by CI or committed by the next `git add -A`. Worse, the target directory
-          // would then be attacker-influenced by anything that can add a `compose-preview/` path
-          // to the tree. Better to store nothing and say so.
+          ?: absoluteHome(prop("user.home"))?.let { "$it/.config" }
+          // …and if there is genuinely nowhere, REFUSE rather than inventing one.
           ?: throw NoCredentialHomeException()
       return File("$configHome/compose-preview/agent-access.json")
     }
+
+    /**
+     * [candidate] if it is a usable absolute path, else null — the single test every
+     * credential-home candidate goes through. `?` is JVM shorthand for "unknown", and a relative
+     * path is not a home however it was supplied.
+     */
+    private fun absoluteHome(candidate: String?): String? =
+      candidate?.trim()?.takeIf { it.isNotEmpty() && it != "?" && File(it).isAbsolute }
 
     /**
      * `scheme://host[:port]`, lowercased, default ports dropped, path and query discarded.
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/AgentAccessStore.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/AgentAccessStore.kt
@@ -353,8 +353,8 @@ internal open class AgentAccessStore(
 
     /**
      * `$COMPOSE_PREVIEW_AGENT_ACCESS_FILE`, else `$XDG_CONFIG_HOME/compose-preview/…`, else
-     * `$HOME/.config/…`, else the JVM's `user.home`. The override exists for CI and for tests; the
-     * XDG path is where a user would look for it.
+     * `$HOME/.config/…`, else the JVM's `user.home` **when that is an absolute path**. The override
+     * exists for CI and for tests; the XDG path is where a user would look for it.
      *
      * Throws [NoCredentialHomeException] when none of those yields a location, rather than falling
      * back to the working directory — see the body.
@@ -381,7 +381,12 @@ internal open class AgentAccessStore(
           // The JVM's own view of the user's home, which on Linux comes from the passwd entry
           // rather than the environment — so it survives the minimal `env -i` service context that
           // has neither variable set.
-          ?: prop("user.home")?.takeIf { it.isNotBlank() && it != "?" }?.let { "$it/.config" }
+          // Absolute only. `-Duser.home=.` resolves to `./.config/…` — straight back under the
+          // working directory, which is the exact outcome this whole path exists to prevent. A
+          // relative home is not a home; treat it as absent and let the refusal below fire.
+          ?: prop("user.home")
+            ?.takeIf { it.isNotBlank() && it != "?" && File(it).isAbsolute }
+            ?.let { "$it/.config" }
           // …and if there is genuinely nowhere, REFUSE. The old fallback was `.`, which wrote
           // bearer tokens and device secrets into whatever directory the command happened to run
           // in — for an agent that is a checkout, so the credentials land somewhere that gets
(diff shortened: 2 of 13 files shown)


yschimke:
Pushed `cc4ee93`, adding two findings that landed on #4394 four minutes before it merged — so they're bugs on `main` now rather than review comments, and they belong here.

**A label could reorder what the approval page appears to say.** `sanitizeLabel` stripped C0/C1 and DEL. `U+202E RIGHT-TO-LEFT OVERRIDE` is neither, and it passes through HTML escaping untouched — so a requester, who picks their own label, could reverse the rendering of the text after it and change what the human appears to be agreeing to, **including the scope and the lifetime**. Isolates (`U+2066`–`U+2069`) do it more quietly, and the same trick reorders the audit line in the operator's console.

This is the one surface whose entire job is honest display, and escaping was never going to cover it: the characters that matter here survive escaping by design. The whole Unicode **format** category (Cf) now goes — overrides, embeddings, isolates, marks, zero-widths, the BOM. That's exactly the set that renders as nothing while changing everything around it, and nothing a purpose string legitimately needs.

**Overflow admission still treated `collected` as delivered.** Round eight taught `purge` that the flag means a poll *built* a response, not that the agent got one — and left the request-map overflow path shedding on it. So a full map plus one dropped packet still stranded a live grant, and filling that map is something an anonymous caller can attempt. That's the same mistake in the same file, found twice, three rounds apart; the rule is now stated once and applied in both places.

One existing test **rewritten rather than deleted**: `a collected request is shed to make room` asserted precisely the behaviour this changes. It now pins what may actually be shed — a request whose grant has expired — with a sibling for a denial.

Full `:cli:test` green at **2957**.

Worth noting for whoever reviews this: neither of these is reachable without the lane being enabled, and #4394 shipped it off everywhere — so nothing was exposed on `main` in the window between that merge and this PR. This branch is also what turns the lane on, so the two arriving together is the right order.

---

yschimke:
All three fixed in `a4a7ead`. Resolving the threads.

| Finding | Fix |
|---|---|
| Token-gated boxes excluded from the derivation | Private + `SERVE_TOKEN` counts as an approver |
| Gate test didn't control its own inputs | `env -i`, PATH only |
| Credential-home refusal reached only one caller | Handled at `Main`'s dispatch boundary |

**The token-gated one is a straight miss on my part.** The server names two ways to be an approver and refuses only when *both* are absent — its own message even says "drop `--public` (the `--token` holder then approves)". I keyed the default on GitHub auth alone, so the lane stayed off on exactly the configuration that message recommends. A default advertised as "on where it can work" has to actually mean it. Four new cases cover the axis: private+token on, private-by-default+token on, public+token off, private-without-token off.

**On the test-isolation one — I checked this the way I should have checked the race test two rounds ago.** Your reproduction fails against the old line and passes against the new one:

```
=== reviewer's repro against the NEW test:
PASS: the lane enables itself only where an approver exists

=== same repro against the OLD env line:
FAIL: unset + no GitHub auth => off: expected SERVE_AGENT_GRANTS=0, got '1'
```

Without that, a test that layers onto the ambient environment silently converts its negative cases into positive ones and passes for the wrong reason — which is worse than not having it, because it reports confidence.

**The `share-preview` one is a good catch on reach.** I added the refusal and handled it where I was looking. Fixed at the dispatch boundary rather than in the second caller, so the next consumer inherits it. Verified end to end on the real binary:

```
$ compose-preview share-preview n.md shot.png --mechanism serve …
compose-preview: no user config directory could be determined (XDG_CONFIG_HOME, HOME
and user.home are all unset), and credentials will not be written to the working
directory. Set COMPOSE_PREVIEW_AGENT_ACCESS_FILE to a path you control.
exit=1
```

Worth noting it took `-Duser.home=` to reach that at all — the `user.home` fallback added in the same round means a real machine essentially always has somewhere to write. The refusal is the floor, not the common path.

Both deploy guards pass; full `:cli:test` green at **2957**.

---

yschimke:
Both fixed in `1197c5e`. Resolving the threads.

**The cap one is the better find, and it's a consequence of my own earlier fix.** Retention (correctly) stopped shedding approvals whose grants are live — but the request map was bounded as a whole, so enough live approvals filled it and every subsequent request was refused. An operator setting `--agent-grant-max-active 100` could never reach it, with nothing in the output to say why.

The fix isn't to inflate a number. These are **two populations with two different reasons to exist**:

- *Pending* requests are what an anonymous caller can create at will — that's what the cap is for.
- A *retained approval* isn't that: it exists only while its grant does, and grants already have their own cap.

So the pending cap now counts pending requests only. Total memory stays bounded (pending + active), and both configured numbers mean what they say. Counting them together was conflating "what an attacker can inflate" with "what we owe someone", which is the same conflation the `collected` flag caused three rounds ago — worth noticing as a pattern in this file.

The docs one is straightforward drift: I widened the derivation to cover token-gated boxes last commit and left the prose saying "exactly when" the GitHub variables are set. Both approver configurations are now named, with the `--public`-and-no-auth case called out as the refusal.

**One existing test rewritten**, and the reason is worth stating: `overflow never strands a token the agent has not collected yet` asserted the *mechanism* — that a new ask got refused rather than evicting the uncollected approval. The mechanism changed (an approval no longer spends that budget, so there's nothing to refuse) while the guarantee didn't. Deleting it would have dropped real coverage; leaving it would have pinned an implementation detail. It now saturates the pending budget and asserts the uncollected token still arrives, which is what it was always for.

That's the fifth test this feature has needed rewritten rather than added. Each time the test was pinned to how something worked instead of what it promised — a habit worth naming, since a suite full of mechanism assertions makes every subsequent refactor look like a regression.

Deploy gate passes; full `:cli:test` green at **2959**.

---

yschimke:
Both fixed in `05584a8`. Resolving the threads.

**The JSON one is the sharpest find on this branch**, because the bug was *created by the fix for the previous finding* and made things strictly worse.

`auth request --json` prints the device secret so the caller can poll for itself — and that is precisely what the new no-credential-home refusal message recommends as the fallback. But the refusal fired while opening the store, *before* the JSON was printed. So on the exact machine the fallback was written for, the command opened a request on the server and exited: a human could still approve that link and mint a credential nothing could ever collect. The previous behaviour was an ugly stack trace; this was a silent orphaned credential, recommended by my own error text.

The store is optional at that one call site now. Every other path still demands somewhere to write; `--json` degrades to storeless, which is what it already promised.

**Verified against the unfixed code, which is the only way this kind of test earns its place.** Reverting the one-line change fails the new test:

```
--- confirming the new test fails without the fix:
Gradle Test Run :cli:test FAILED
```

That needed a narrow seam (`openStore`) — no test can make a real machine forget where its home directory is — documented as existing for that one purpose.

**On the docs finding: you're right, and the distinction is one I blurred.** The two approver configurations are alternatives for deciding *whether the lane turns on*. They are not alternatives for *approving*: `agentGrantApprover` checks the server's front door first and a GitHub session is not a substitute, so a private box that also configures OAuth needs both. An operator reading "either" could hold the documented credential and get bounced by the sign-in notice with nothing explaining why. Enabling and approving are now separate sections with a table:

| Box | To approve you need |
|---|---|
| public + GitHub auth | a signed-in GitHub account |
| private, no GitHub auth | the browse token |
| **private + GitHub auth** | **both** |

Deploy gate passes; full `:cli:test` green at **2960**.

---

yschimke:
Both fixed — `fc6c47c` here, plus [yschimke/skills#56](https://github.com/yschimke/skills/pull/56) for the P1. Resolving the threads.

**The cap hole is mine, opened one commit after I closed the opposite one.** Splitting the caps made the pending limit count only `PENDING`. But a denial stays in the map until its own deadline — it has to, so its owner learns they were *denied* rather than `unknown` — and it was no longer counted. So an operator working through hostile requests freed a slot with every click: submit a batch, have it denied, submit another, forever, in a map that was supposed to be bounded. The comment I wrote in the previous commit said "total memory stays bounded (pending + active)"; that had stopped being true by the time it was pushed.

Having now got this split wrong in **both** directions, the rule is written into the code rather than left to be re-derived:

- charged: **pending** rows, plus rows the caller **got denied** — everything an anonymous caller can create;
- not charged: a **retained approval**, which lives only as long as its grant and is bounded by `--agent-grant-max-active`;
- reclaimed first, unconditionally: rows nobody is owed anything for, so a dead approval can never be what pushes a legitimate caller over.

A denial is kept *and* charged. Only its own expiry frees the space, which is the honest accounting — the attacker caused the row, and the operator's decision shouldn't hand them a free one.

**One existing test inverted, and it's worth being blunt about why:** `a resolved request makes room for a new one` asserted that denying frees capacity. That test encoded the vulnerability. It passed happily for nine rounds. It now asserts the opposite, plus what's preserved (the denial stays readable by its owner; time still frees the space). A near-duplicate making the same claim was dropped rather than inverted twice.

That's the seventh test on this feature that had to be rewritten rather than added — and this one is a different, worse category than the other six. Those pinned mechanism instead of promise. This one asserted behaviour that was actually wrong, so the suite was actively defending the bug.

**On the P1:** agreed, and it's not just tidiness. The old skill text tells an agent that a 404 from `/agent-access/request` means a missing flag. After this PR, on the image, it more likely means the box has nobody who could approve — so an agent following it gives its human wrong advice about their own deployment. skills#56 distinguishes hand-run `serve` from the image and names the one configuration that has no approver. Worth merging close to this one; in between, the installed skill describes a default the image no longer has.

Deploy gate passes; full `:cli:test` green at **2961**.

---

yschimke:
Both fixed in `63baa36`. Resolving the threads.

**The first is the same bug, reached through the fix for it.** I added `user.home` as a fallback so a minimal service context would still have somewhere to write — and didn't check it was absolute. `-Duser.home=.` resolves to `./.config/compose-preview/…`: bearer tokens under the working directory, which for an agent is a checkout. That is precisely the outcome the refusal exists to prevent, arrived at through the mitigation for it.

Worth noting I used `-Duser.home=` myself two rounds ago to force the refusal in a manual check, so a relative value there was not hypothetical — I had the mechanism in my hands and still didn't validate the shape.

A relative home is now treated as absent, and the refusal fires.

**The second is a case where withholding the token stopped being the safe choice.** Normally `GrantedJson` deliberately omits it — the grant is on disk, `auth token` prints it for anything that needs the string, and putting a bearer in a machine-readable response would spread it through logs for no gain. With nowhere to store, that inverts: the caller waited for a grant that is live on the server, got `stored: false` and nothing else, and the credential stays minted until it expires. It now comes back, **only** when it could not be stored, with the reasoning written next to the field so nobody later "tidies" it into always-on or always-off.

Both directions have tests, including the storeless waited flow end to end against a real server.

Full `:cli:test` green at **2963**; deploy gate passes. [skills#56](https://github.com/yschimke/skills/pull/56) is clean and waiting alongside this.

---

yschimke:
Both fixed in `1790e38`. Resolving the threads.

**This is the third time the same leak has been reachable, and that's the finding — not the individual holes.** The original `.` fallback; then a relative `user.home`; now a relative `XDG_CONFIG_HOME` or `HOME`. Each time I patched the branch that was pointed at instead of the question underneath it, so the hole reopened immediately beside the patch for it.

The question was identical every time: *does this resolve under whatever directory the command happens to run from?* There's now one `absoluteHome` check every candidate goes through, and I've written the three-strike history next to it so the next person adding a source applies the rule rather than rediscovering it the hard way. A relative value now falls through to the next candidate instead of shadowing it.

The test covers all three sources × four shapes deliberately — not because a fourth route is likely, but because "I fixed the one you named" is exactly what produced rounds two and three of this.

**The second is a smaller version of the same habit.** I put the no-store abort where I was reading, which was after `client.open`. That left an uncollectable request occupying the server's bounded pending map for ten minutes — and a caller retrying the way anyone would could exhaust the slots for everyone else, which is the very map this branch spent two rounds bounding correctly. The check runs before the request is opened now. `--json` stays exempt (it prints the device secret, so it genuinely needs no store), and the after-the-fact check stays for a store that exists but can't be written, since that isn't knowable until the write is attempted.

Parity came back **green on `63baa36`**, so the failure on `fc6c47c` was the usual spinner noise; that preview has now read between 0.06% and 1.23% across eleven runs of this branch with nothing touching the renderer.

Deploy gate passes; full `:cli:test` green at **2964**.

---

yschimke:
## `wear-os-samples (ComposeStarter)` is failing, and it is not this PR

Unlike the parity noise, this one **blocks merge** — it's `pr_blocking: true` and a required status check — so flagging it rather than leaving it in the run log.

```
UnknownPluginException: Plugin [id: 'com.android.application', version: '9.1.0', apply: false] was not found
  could not resolve 'com.android.application:com.android.application.gradle.plugin:9.1.0'
  Searched: Gradle Central Plugin Repository, Google, MavenRepo, MavenLocal
```

**Why it isn't mine.** This repo pins its own AGP at `9.3.1` in `gradle/libs.versions.toml` and references `9.1.0` nowhere. That version comes from the external `android/wear-os-samples` checkout's own catalog. My diff touches `AgentAccessStore.kt`, `AuthCommand.kt`, `Main.kt`, `ServeAgentGrantStore.kt`, `entrypoint.sh`, one deploy test and docs — nothing that can reach an external repo's plugin resolution. `composePreviewDiscover` fails during *configuration*, before any of our code runs.

**Why it happened.** `integration.yml` checks that sample out at `ref: main` — unpinned upstream:

```yaml
- name: wear-os-samples (ComposeStarter)
  repo: android/wear-os-samples
  ref: main
```

So the moment upstream moved to an AGP our configured repositories can't serve, a required check on *every* PR in this repo started failing. Nothing here changed.

**The irony is the actual defect worth fixing:** this repo is careful about pins — its own AGP, the design-artifacts driver, the androidx snapshot build — and then takes third-party `main` unpinned on a check that gates merge. That's a supply-chain dependency on someone else's default branch sitting in the critical path.

### Proposed patch (not applied here — it would widen this PR)

Pin the cell to a known-good upstream commit and bump deliberately:

```yaml
- name: wear-os-samples (ComposeStarter)
  repo: android/wear-os-samples
  ref: <known-good SHA>   # was: main
```

That keeps the coverage this cell exists for while removing "upstream pushed something" from the set of things that can block an unrelated PR. A periodic bump (the nightly cron is the natural place) still catches real upstream drift, just not on the critical path. If the intent is genuinely to track upstream `main`, then the cell should not be `pr_blocking` — you can have one or the other, not both.

I have not applied it: it's unrelated to this branch and belongs in its own change where it can be reviewed on its merits. Happy to open that PR if you want it.

**Everything else on `1790e38` is green**, including the deploy gate, `ktfmt`, CodeQL, and the full `:cli:test` suite at 2964. Parity was green on `63baa36`.

---

yschimke:
**Correction to my previous comment: `wear-os-samples (ComposeStarter)` is green on `1790e38`** (00:57Z). It recovered without any change from me, so it is *not* blocking this PR, and my "blocks merge, needs your call" framing was wrong within about ten minutes of writing it.

What that changes, and what it doesn't:

- **Status:** not a blocker. Nothing here needs a decision from you to merge.
- **The mechanism:** unchanged and, if anything, better evidenced. The cell resolves `android/wear-os-samples` at `ref: main`, so upstream pushed something that briefly made AGP unresolvable and then pushed something that fixed it. A required check on every PR in this repo went red and green again on someone else's commits.

So the proposed patch stands as **hygiene, not urgency** — pin the cell to a known-good SHA and bump on the cron, or drop `pr_blocking` from it. I'd still open that as its own PR if you want it, but there's no longer anything to bypass or wait on.

I should have said "this is failing on something outside the diff, and I'll re-check whether it persists" rather than "this blocks merge". A single red run of an unpinned external dependency is exactly the case where the honest read is *wait one cycle*, and I had already established the mechanism that makes it volatile.

**Current state of `1790e38`:** every completed check green, including `CMP/Wasm Parity`, the whole downstream-consumer matrix, CodeQL, `ktfmt`, and the deploy gate. One job (`A11y + notification previews`) still running. All thirteen review threads resolved.

---

PR #4409, opened 2026-08-21
fix(serve): withhold override axes by lane AND family, and only for a real lane

yschimke:
Two gaps in the lane scoping that landed in #4407. Both were raised in review there and arrived before it auto-merged, so they are on `main` now. Both confirmed against the code before fixing.

## A browser lane forwards only half the controls

`wasmOverridePatch()` forwards `.cp-knob` and nothing else. The RC canvas (`applyRcOverrides`) and the CMP-Wasm player (`rcWasmNamedValues`) forward `.cp-rc-knob` and nothing else — and `syncServerControls()` disables whichever family its lane cannot reach.

Exempting the *whole* withheld set for any in-browser mode therefore hydrated a **disabled** control to a value the lane on screen never applies:

```
pinned / baked-fallback / replayed page, then
  ?mode=wasm&rc.x=…     → rc.x restored, Wasm never forwards it
  ?mode=rc&knob.x=…     → knob.x restored, the canvas never forwards it
```

The exemption is per family now: `wasm` releases `knob.*`, `rc` and `rc-wasm` release `rc.*`, and the other family stays withheld.

## `?mode=` is a request, not a fact

A stale or hand-shared mode naming a lane this session does not offer leaves the snapshot displayed — the bookmarked-mode guard returns early on a missing or disabled radio (`viewer.ts`, `if (!radio || radio.disabled) return`). Dropping the withholding for it let a control adopt an override the snapshot had not applied.

`laneIsEnterable` consults the same radio that guard does, so the two cannot disagree about whether a lane will draw the page.

## Shape

`effectiveUnseeded(all, mode, laneEnterable)` replaces `laneAppliesWithholding(mode)` — one function over the withheld set, the requested mode and whether that lane is reachable, because those are one decision and splitting them is what let the second gap through.

## Test plan

- `viewerOverrideSeeds.test.ts`: everything binds on each server-rendered lane and on the default; only the forwarded family is exempted per in-browser lane; withholding stands when the requested lane cannot be entered; an empty set is left alone.
- `cli/serve-web`: `npm run typecheck` clean, `npm test` 991 passing, `npm run verify` reports the committed bundle in sync.
- `./gradlew :cli:test --tests "ee.schimke.composeai.cli.serve.*"` and `ktfmtCheck` — green.

Text-only evidence: the change is which URL values a control opens on, per lane. No `@Preview` or fixture page draws it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XvUA1ogfsLK5wmCxAgah8r)_


yschimke:
## `CMP/Wasm Parity` red — not this PR's, and not pushing anything for it

`IndeterminateCircularProgressRemote_…_dpi_320` moved 0.23% → 0.63% (0.40 pp, limit 0.25 pp).

**The diff cannot reach it.** Four files, all viewer front-end:

```
cli/serve-web/src/viewer.ts
cli/serve-web/src/viewer/overrideSeeds.ts
cli/serve-web/test/viewerOverrideSeeds.test.ts
cli/src/main/resources/…/serve/assets/viewer.js   (the rebuilt bundle)
```

Nothing there is on the path the parity harness renders through — it draws `ir/<id>.rc` documents with the CMP and Wasm players offline; the viewer's URL-hydration code never runs.

**The same single preview has oscillated all session**, in both directions, across PRs whose diffs are equally unrelated (serve/viewer Kotlin + TypeScript). From this branch's own predecessors:

| PR | reported |
| --- | --- |
| #4398 | 1.02% → 0.08% ✓ |
| #4398 (re-run) | 0.96% → 0.66% ✓ |
| #4400 | 0.64% → 0.24% ✓ |
| #4401 | 51 unchanged |
| #4403 | 51 unchanged |
| #4407 | 1.10% → 0.64% ✓ |
| #4409 | 0.23% → 0.63% ✗ |

Same artifact observed at 0.08, 0.23, 0.24, 0.63, 0.64, 0.66, 0.96, 1.10 — no diff in that list plausibly moves an *indeterminate* animation's pixel mismatch. It is a time-sampled animation whose captured phase varies per run, and the ±0.25 pp gate is narrower than its natural spread, so it trips whichever direction the sample lands.

**So: no fix pushed, deliberately.** Pushing a change to move a number my diff didn't move would be papering over the check. The bot itself marks this one non-blocking ("does not block merging — read the rows below and decide"), and the other 50 rows are unchanged.

Worth separating from this PR: if that preview's spread really is wider than the gate, the gate or the capture is what needs attention — either pin the animation's phase for the parity capture, or widen the tolerance for indeterminate previews specifically. I have not touched either; say the word if you want that looked at as its own change.

---

PR #4424, opened 2026-08-22
feat(serve): find previews by the composables they call (`uses:`)

yschimke:
The landing filter box matches a card's label and id — names a catalog *chose*. `uses:Button` now narrows the same box to the previews whose declaration actually **calls** something matching `Button`.

```
uses:Button              every preview that calls a Button-ish composable
uses:Button tonal        …and whose name or id also contains "tonal"
```

## Visual evidence

| Resting — unchanged by this change | `uses:Button` |
| --- | --- |
| The landing filter box before anything is typed | Six previews narrowed to the two that call a Button |

**The right-hand shot is the argument for the feature.** The two cards left standing are *Card* and *Profile screen* — neither name contains "button", so the existing label-and-id filter could not have found either. Among the four it hid is `ButtonPreview`, which that filter *would* have matched. Asking what a preview is made of, rather than what it is called, is the question a name search structurally cannot answer.

A catalog that cannot be indexed says so, rather than showing an empty grid that reads as "nothing matched":

A catalog that cannot be indexed, reading 'call index unavailable'

Both are new `pages-snapshot.spec.mjs` fixture states, so every later change to this surface is diffed by the bot without anyone remembering to shoot it. Notes and reasoning in [`docs/design/evidence/serve-uses-filter/`](https://github.com/yschimke/compose-ai-tools/blob/agent/preview-uses-filter/docs/design/evidence/serve-uses-filter/README.md).

## How it works

- **`PreviewUsageIndex`** builds the answer from the catalog's own source, reusing what the viewer's Source panel already runs on: each preview's `repo` / `ref` / `sourceFile` / `bodyLine` from catalog metadata, and the `:usage-source-psi` parser in its isolated classloader. One fetch and one parse per distinct **file** (a catalog's previews come a section at a time), and the same declaration-bounds rule the seed slicer uses splits a file's calls among the previews declared in it. Cached per catalog with a short TTL, rebuilt when the preview list changes.
- **`GET /{system}/api/uses?q=`** serves it, **gated to Dev mode** — Catalog mode gets a 404, and the page emits no trace of the operator there (no `data-uses-id`, no branch of it in the script). The Catalog-mode page fixture is byte-identical to before.
- The operator **composes** with the ordinary text filter and rides the existing `?q=`, so a `uses:` search is as bookmarkable and shareable as any other filter. A readout under the search box names the active filter back to the reader.
- `PlaygroundSeedResolver.declarationLines` is extracted from `sliceDeclaration` so both callers share one rule about where a declaration ends, rather than keeping a near-miss copy.

### Stated limits

Documented in the new [server-docs section](https://github.com/yschimke/compose-ai-tools/blob/agent/preview-uses-filter/docs/public-preview-server.md) and on the class:

- **No resolution.** A parse reports the callee as written — "names called in this declaration", not "Compose symbols this preview binds to". Two `Button`s from two packages are one entry. Resolving properly needs a classpath and a frontend per catalog.
- **No expansion through delegation.** A catalog whose previews are one-line `Sticker("<slug>")` delegations indexes as calling `Sticker` — true, and not useful. Following `compose-usage.json`'s scaffold rules is a further change, deliberately not a silent behaviour of this one.
- **Unavailable is not empty**, which is why the endpoint carries an `available` flag rather than just a list.

## Test plan

- `PreviewUsageIndexTest` (new, 15 cases): a preview is credited with the calls in its own declaration and not its neighbour's; one file read once for the previews sharing it; TTL and preview-list-change rebuilds; CRLF source does not walk the offsets; and each of the four ways to be unavailable or truncated reports itself rather than returning a bare empty set.
- `ServeHttpRoutingTest`: `/api/uses` and `/{system}/api/uses` both 404 in Catalog mode and answer in Dev mode, and a host with no source fetcher answers `available: false`.
- `ServeWebFixtureTest` goldens regenerated — 15 Dev-mode landing pages moved, the Catalog-mode one did not.
- `./gradlew :cli:test` green; `./gradlew ktfmtCheckAll` green.
- Both new harness captures shot in-session (embedded above).

https://claude.ai/code/session_016sd9w15eKsxnedz8s4bdmY

---
_Generated by [Claude Code](https://claude.ai/code/session_016sd9w15eKsxnedz8s4bdmY)_


yschimke:
On the diff bot's **Changed** rows — `serve-viewer-exploded-controls` in both themes is **not** this PR. Recording what I measured so nobody re-chases it.

The "after" frame is the page caught mid-render: the match chip reads `comparing…` instead of `69.2% match`, there's a spinner over the sheet, and a `rendering…` line has pushed the export bar down. Those are states, not pixels — nothing about the exploded view changed.

That state is non-deterministic on the code already on `main`. Three local runs of `serve-viewer-exploded` with this branch's `pages-snapshot.spec.mjs` reverted to its committed form gave at least two distinct outcomes, roughly 1-in-3. Its wait is

```js
await page.waitForFunction(() => !document.querySelector('.cp-stage[aria-busy="true"]'));
```

which covers the SVG re-projection but not the two things the knob kicks off behind it: the viewer's re-render status line (`viewer.js` writes `rendering…` before each fetch) and the SVG-vs-render scorer, which re-scores against the new frame. When the stage's busy flag clears, neither is finished.

This PR's diff touches the catalog **landing** page, `serve.css`, and the CLI — not the viewer page, `viewer.js`, or the exploded lane. The plausible coupling is scheduling: two new fixture states on `serve-landing` shift what runs alongside what under `fullyParallel` with 2 workers.

I tried extending that wait to hold for both signals, and **it came out worse** — a blank stage with the chip still reading `comparing…`, because holding longer walks the page into a later frame rather than back to an earlier settled one. So I reverted it rather than push a speculative fix onto an unrelated page; the flake wants its own change, by someone who can decide whether the right fix is a settle or making the capture not depend on the scorer at all. Happy to open an issue if that's useful.

`Compare vscode-preview captures` is green, and the four **New** rows are the intended ones for this PR.

---

yschimke:
Thanks — all five verified against the code, none were false positives, and all five are fixed in `97585b3`. Each fix has a test that I checked fails without it.

**P1 · index build on request threads.** Fixed. `match` was `@Synchronized` across up to `maxFiles` network reads, on threads Ktor also serves renders on. Now per-catalog locks — which keep the property that actually mattered, two concurrent searches of one cold catalog doing a single build, and drop the process-wide one that was accidental — plus a double-checked read outside the lock, and `withContext(Dispatchers.IO)` in the handler.

**P2 · truncation at the fetcher's real cap.** Fixed, and this was the sharpest of the five. `httpFetch` reads `maxBytes + 1` precisely so a caller can tell a big file from the start of a bigger one, and my 512 KiB cap threw that signal away. `DEFAULT_MAX_BYTES` is now `PlaygroundSeedResolver.DEFAULT_MAX_BYTES` itself rather than a matching literal, so the two cannot drift apart again; the test asserts the equality as well as the behaviour.

**P2 · parser declaration spans.** Fixed as suggested. `:usage-source-psi` now reports top-level declaration spans and those decide attribution; the line scan remains only for an analyzer predating the field. The test is your exact case — `fun A() { Spacer() }` immediately followed by `fun B() { Icon() }` — and it fails on the old path.

**P2 · uses-only query and section spanning.** Fixed. Worth adding that `reflectTabs`'s own `searching` was already right, since it reads the raw input where the operator is still present — so the branches opened while the cards under them stayed hidden. Only `tabOk` disagreed.

**P2 · Pages pane.** Fixed. The pane filters on the raw query, so a uses-only query lands on its own empty state rather than listing every sheet; design sheets have no composables to call, and that is the honest answer.

Two notes on the tests, since one of them nearly slipped through. The section-spanning contract test **passed with the fix reverted** on its first draft: the sectioned fixture opens on the **All** row, where `current === "all"` satisfies the tab predicate whatever the query is. It now selects a real section and asserts the card is hidden before filtering, and it fails on the old expression. And `ServeWebFixtureTest` pinned `tabOk` as a literal string; its intent is the `current === "all"` tail, which this does not touch, so that assertion is now a regex over the tail with the leading conjunct left loose.

`:cli:test`, `:usage-source-psi:test`, `ktfmtCheckAll` and all 160 page-harness tests green locally.

---

PR #4450, opened 2026-08-22
fix(serve): keep a themed grid rendering after the on-screen batch drains

yschimke:
Picking a declared theme on m3.preview.coo.ee left the grid spinning on the previous theme's pixels and never recovering. Reproduced against the live box: `POST /api/theme-render-lease` grants a claim, the page releases it ~2ms later before a single render starts, and every themed render then comes back `429 theme render lease expired or saturated` — for the whole life of the page.

## What was wrong

Three faults compounding on one path:

1. **The All tab classified every card as off screen.** The visible/deferred partition ends with `themeVisible = themeSection.getAttribute("data-section") === current`, and on `?tab=all` no card's section is the current tab. The on-screen batch came out empty.
2. **The deferred cards rode the visible batch's claim.** Both queues were stamped with the same token, and that claim is handed back the moment the on-screen batch drains — instantly, when it is empty. Every card the visitor scrolled to then presented a claim the server had already reaped.
3. **The server answered a reaped token `429`.** No retry can make a reaped claim valid, so each card burned its three backoff retries and gave up.

On `?tab=all` nothing rendered at all; on a tabbed view only whichever cards made the first on-screen batch (one card, in the `?tab=actions` trace).

## The fix

- Each deferred batch takes a **claim of its own** when the viewport reaches it and hands it back when it drains. Claims for one catalog join one allocation, so this is bookkeeping, not extra width.
- The page tracks **every** claim it holds (`themeLeases`), so a theme change or a page exit releases all of them rather than just the last acquired.
- On the All tab every section is showing, so the partition is geometry alone.
- `ThemeRenderLeaseManager.admission()` now distinguishes `Saturated` (alive, full — still `429` + `Retry-After`, which a retry can satisfy) from `Unknown` (reaped, expired, or another catalog's). An `Unknown` token is treated exactly like a request carrying none: rendered on the serial unleased lane. That is the safety net for a claim that expires mid-scroll, which no client bookkeeping can avoid.

## Visual evidence

Both captures are the same page, the same mock render endpoint, the same interaction (pick **Brand Dark**, scroll the grid) — one against the page this branch replaces, one against the page it emits. The endpoint refuses a released token exactly as the server does, and answers a live one with dark pixels.

Before — every re-renderable card stuck spinning on its baked light pixels:

Themed catalog before the fix: every card spinning

After — all seven re-renderable cards carry the themed render (the two Themes swatches have no re-renderable base by design):

Themed catalog after the fix: every card themed

## Keeping the coverage automatic

`serve-landing-declared-tabbed-themes.html` is now written out as a preview-harness fixture — a sectioned catalog with an All tab, a declared theme and a burst capacity, big enough that its cards outrun one viewport. Fixtures are auto-discovered by `pages-snapshot.spec.mjs`, so it joins the snapshot captures and the CI visual-diff bot without further wiring, and the new contract test runs the deferred lane on it.

## Test plan

- [x] `./gradlew :cli:test` — full suite green
- [x] `./gradlew :cli:ktfmtCheckMain :cli:ktfmtCheckTest`
- [x] `npx playwright test -c preview-harness/playwright.config.mjs pages-snapshot.spec.mjs -g "contract"` — 14 passed
- [x] New contract test **fails against the pre-fix page** (0 cards render) and passes after
- [x] New unit coverage: `admission()` tri-state (saturated vs released vs expired vs never-issued), the All-tab partition, per-batch claims, release-all
- [x] New routing test: a render carrying a reaped `_themeLease` is served `200` by the unleased lane instead of `429`
- [x] Root cause reproduced against the live box first (lease granted → released 2ms later → `429` on every render)


yschimke:
On the preview-harness diff above — the two **new** captures are this PR's (the `serve-landing-declared-tabbed-themes` fixture it adds). The one **changed** capture is not.

`serve-viewer-exploded-controls` is a flaky *baseline*, not a change here. The baseline on `vscode-preview/main` was shot mid-computation: its match chip reads `comparing…`, the stage carries a spinner, and `rendering…` sits under the sidebar. This branch's capture is the settled state — `69.2% match`, no spinner. Nothing in the diff touches that fixture's HTML or the assets its capture routes, and it alternated theme between the two runs on this PR (`dark` on the first, `light` on the second), which only a race explains.

The first run also listed `serve-format-compare-reference-lane` (light + dark). That was base drift — this branch predated #4428 — and it disappeared once `main` was merged in.

Raising rather than fixing: making that capture wait on its scorer is someone else's fixture and would move its baselines, so it doesn't belong in this PR. Happy to do it as a fast-follow if you want it.

---

PR #4458, opened 2026-08-22
feat(catalog): point remote-m3 at the Wear reference catalog, not the harness sheet

yschimke:
Follow-up to #4447, which built the cross-repo + design-column capability. This uses it.

`remote-m3` compared against `wear-m3` — this repo's own **harness** sheet, which exists to exercise the preview pipeline rather than to reproduce a design system. Wrong yardstick, and it pinned by-component redundancy on that sheet purely so a published page wouldn't silently degrade. It now pairs against the Wear **reference** catalog in `yschimke/wear-m3-catalog`, and therefore also carries that catalog's kit references as a third column.

```jsonc
"compareWith": {
  "system": "wear-m3-catalog",
  "repo": "yschimke/wear-m3-catalog",
  "designTitle": "M3 Wear OS kit"
}
```

## Results, measured against the live published branches

| | before | after |
| --- | --- | --- |
| rows with a parallel | 42 | 40 |
| distinct parallels resolving in the sibling | 9 of 19 | **18 of 18** |
| rows resolving a kit reference | 20 | **38** |

## Visual evidence

The real page over the live branches, trimmed to eight rows — six whose parallel changed, plus `Button/Filled` and `Card` as unchanged controls. Full notes in [`renders/remote-m3-wear-catalog-repoint/`](https://github.com/yschimke/compose-ai-tools/tree/347a339296b5292db8fff5e2e7cc2c3b4894d608/renders/remote-m3-wear-catalog-repoint).

**Before** — paired against the harness sheet, no third column to arbitrate:

Component / Remote Compose Material 3 / Wear Compose Material 3 / Parallel

**After** — paired against the reference catalog, kit leading:

Component / M3 Wear OS kit / Remote Compose Material 3 / M3 Wear OS Apps Design Kit / Parallel

`Button/Compact` is the row that shows the point: kit and Wear catalog both draw **+ Primary label**; the Remote sticker draws a small pill reading "Compact". Against the harness sheet that was two components differing. Against the kit it's the Remote catalog using non-kit copy.

## The mapping

- **Re-pointed (18 rows).** `IconButton` splits by emphasis — the base `Button/Icon` uses `RemoteIconButtonDefaults.iconButtonColors()`, the stock *transparent* container, so it pairs with `IconButton/Standard`, while `-Filled` / `-Outlined` pair with their own variants. `CompactButton` → `Button/Compact`, `Progress/Circular*` → `CircularProgressIndicator`, `Template/TimeText/largeRound` → `Scaffold`, `Template/PageIndicator/largeRound` → `PageIndicator/Horizontal`, `Text/MaxLines-Truncated` → `Text/Body`.
- **Added (1).** `PageIndicator/Vertical` — the reference catalog publishes it, the harness sheet never did.
- **Dropped (3).** `Icon`, `Theme/Typography`, `Theme/ColorScheme` have no counterpart; listing them under "only in remote-m3" is truer than naming a component that doesn't exist.

Two paired rows read "no kit reference", both correctly: `Button/Group` → `ButtonGroup` and `Template/WatchScreen` → `Scaffold` are door-2 components whose `noReference` records that the kit publishes no such set.

## Doc changes

The "standing exception" note in `DESIGN_CATALOGS.md` is discharged — nothing on the `wear-m3` sheet is load-bearing for another catalog's page now, so it's free to be feature-scoped like `compose-m3`. **Not done here**, and it should confirm against a published `matches.html` first.

Added in its place: why the Remote catalog itself stays in this repo, with a stated trigger rather than a preference. It depends on `:third-party-rc-embedded-player`, which is deliberately unpublishable because it `api`-exposes `androidx.compose.remote:*` coordinates that only resolve from the build-id-pinned androidx.dev repo; all three remote groups are `1.0.0-SNAPSHOT` on a build id that ages out after weeks, and the published sibling modules keep those deps off their POMs on purpose. Moving to `wear-m3-catalog` today would import an expiring pin into the repo whose job is a stable kit reproduction, and drop the embedded-player lane. **Trigger: those groups reaching Google Maven on release coordinates** — then the move is cheap and `compareWith` drops to `spec: "../catalog.spec.json"` with no `repo`.

## Test plan

- `validate-catalog-spec.mjs --spec samples/design-catalog-remote-m3/catalog.spec.json` → **OK, 0 warnings, 0 errors**.
- Every one of the 18 distinct parallels resolved against the live `wear-m3-catalog` `catalog.json` (66 components) — **0 unresolved**. This is the check no single-spec validator can do, since the sibling inventory only exists on its branch.
- Kit-reference join run against the live `references/index.json`: 38 of 40 paired rows resolve, and the 2 that don't were confirmed to be `noReference` components rather than gaps.
- Full suite `node --test "scripts/design-artifacts/*.test.mjs"` → **854 pass / 7 fail**, identical to `main`. The 7 are pre-existing `ERR_MODULE_NOT_FOUND` for un-vendored `@design-parity/*` packages in this sandbox.
- No code changed — spec, doc and evidence only, so the capability's own tests from #4447 still cover the behaviour.

One capture note: `loading="lazy"` is stripped when screenshotting the evidence. A headless capture never scrolls, so below-fold lazy images photograph blank — that produced a misleading first capture here, which is why both are re-shot eagerly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CEZrWvRNvWRPw4c1D26pyy)_


yschimke:
The 18 changed renders above are **not from this PR** — they belong to #4455, which is one of the two commits the bot's own note says the baseline is behind.

`eb290a07` ("feat(rc): default to the embedded player, not the AOSP view player") changed `RemoteComposeIrReplay.kt` and `RemoteOverridablePreview.kt` — the connector every `remote-m3` sticker renders through — so a wholesale shift in this module's output is exactly what it should produce. It shipped its own before/after for the same components in `docs/design/evidence/rc-embedded-default/` (`buttongroup-before-java.png` → `buttongroup-after-embedded.png`), and `ButtonGroupRemote` is on the changed list here.

This PR touches five files: `catalog.spec.json`, one doc, and three evidence artifacts. No Kotlin, no theme, no dependency. Comparing the spec across `origin/main...HEAD`, every `preview` and `motionPreview` value is **identical** — those fields show in the line diff only because the `parallel` value on the same line changed:

```
render-input (preview/motionPreview) changes: NONE — every rendered function is identical
```

Nothing to do here. Worth a re-run of the preview diff once the baseline for `eb290a07` has published, which should come back empty for this PR.

### The a11y report is pre-existing too

Same reasoning applies to the 3 `SpeakableTextPresentCheck` errors on `WidgetContainerGradientRemote` / `WidgetContainerLargeRemote` / `WidgetContainerSmallRemote`. The flagged element is `androidx.compose.remote.player.view.RemoteComposePlayer` — an Android `View` hosting remote content, which ATF cannot see inside to find speakable text. That is structural to how those widget-container stickers render, and this PR changes no code on that path.

The reason the report appears here at all is scoping, not regression: the a11y job is change-scoped, and `catalog.spec.json` lives under `samples/design-catalog-remote-m3/`, so touching it pulled the whole module into the run. The `A11y + notification previews` check itself passed.

Worth a separate look at whether those three previews should carry a `contentDescription` on the player, or be exempted — but it is not this PR's to fix, and folding it in would widen a spec-only change.

---

PR #4480, opened 2026-08-24
feat(parity): score scoped acceptances, in both hosts of one engine

yschimke:
Batch 05's half that lives in this repo. Batch 04 settled every *gate* — what a mask is permitted to suppress — and deliberately computed no numbers, because §4's I1 says every gate resolves before any score does. This adds the score, and the two places it runs.

Companion PR: yschimke/design-parity#388 carries a repo's committed acceptances into the published catalog.

## The separated-plane score

`scripts/design-artifacts/known-difference-score.mjs`. Separation precedes every resample and applies to **both** inputs (I3/I4), one resample per region straight from the source into the score plane at the candidate box's dimensions (I10), region-aware edge and content masks, and masked coordinates excluded in **both roles and both directions** — so a regression within `EDGE_SEARCH_RADIUS` of a mask edge cannot find its match inside the accepted region and score as clean.

The seam to the gates is `survivingMasks`: the evaluator hands out the masks of the acceptances that reached `valid`, and no other status. That is where "a `resolved` mask suppresses nothing" actually lives, and it is pinned from both sides — `survivingMaskIds` on the gate cases, the `scoring/` group on what the survivors then suppress. Neither fixture asserts the other's half.

Six scoring cases, every expected number **declared as arithmetic over the geometry** rather than harvested from a run: the base case, a regression parked three pixels outside a mask edge, a mask edge that straddles the score-plane grid with equal and opposite deltas either side (a pre-averaged composite lands exactly on the reference and cancels two real findings), the all-masked denominator, the empty union where `unaccepted` is `raw` bit for bit, and an overlapping pair.

Two things D5's six answers did not cover, now settled in §4: the whole path is **binary64**, not the browser's `Float32Array` storage — that quantised a 768-pixel mean by ~4×10⁻⁷ and would have put a `Double` port somewhere else — and a coordinate is scored only where **both** sides are present, since separation can leave a footprint present on one side alone.

## The canonical plane, measured portably

`known-difference-plane.mjs`, with a `plane/` group. It was pinned by nothing: every gate case is *handed* its plane, deliberately, so it tests the gate. But the plane gate compares a **recomputed** plane against the recorded one, so a one-pixel disagreement about a content box is `plane-changed` on one engine and `valid` on the other, from identical bytes. `contentBox.ts`'s decisions are carried over unchanged; the one host-dependent step — the sampling downscale — becomes the portable area average.

## One engine, two hosts

§4 asked for two engines kept honest by fixtures. Where a stronger option was available it was taken: **the browser runs the same module**, because `png-lite.mjs` stopped needing `node:zlib` and `node:crypto` (`inflate-lite.mjs`, `sha256-lite.mjs`, both checked against the Node implementations over the committed corpus and randomised inputs at every block type and padding boundary). A guard test pins that nothing in the import graph reaches for `node:` or `Buffer`.

The alternative was decoding through an image element onto a canvas, and that is not merely slower: a canvas decode normalises **every** colour type to 8-bit RGBA, so the mask-encoding rules §4 spends paragraphs on — bit depth `8`, colour type `0`, the palette and 16-bit masks that sail through a sample-only check — are invisible to it. The browser would have accepted files the offline engine refuses, from identical bytes. `design-parity` remains the genuine second implementation and the fixtures' whole remaining job.

## The host carries; it does not decide

`ServeKnownDifferences` serves the document **verbatim as text**. `document-unreadable`, `document-too-large`, a duplicated id, a schema token from the future are all verdicts the engine must be able to reach, and it can only reach them if the bytes arrive intact — a host that pre-parsed would be a third implementation of the contract with no conformance suite behind it. What it does enforce is the three obligations §4 gives a *reader*: refuse by size before allocating, resolve containment into the acceptance's **own** directory (a symlink from `a/` into `b/` stays under the root while letting record `a` be hashed against `b`'s bytes), and open the exact spelling asked for. Each is a distinct status code so the browser can relay `path-not-contained` and `artifact-too-large` rather than collapsing both into "could not read".

`ServeTagIndex`'s KDoc said an acceptance that finds no tag entry "degrades to no element gate, which is the safe direction". Corrected: that is only safe if it also stops suppressing, or an element-scoped acceptance whose gate cannot run has silently become the ignore rectangle the non-goals rule out.

## Visual evidence

The band on a focused comparison, from the real bundle running the real engine over a catalog with **two** acceptances — one still matching, one that has stopped. Light and dark:

the acceptance band, light theme

the acceptance band, dark theme

Three numbers because none is the difference of the others: `raw` is the pair as though nothing had been accepted and is never hidden, `unaccepted` is what still needs looking at, `accepted` is the accepted region's own match on the same scale. A row per acceptance because a single aggregate cannot express a mixed-validity set — and the loud row is the point of the model: a known difference that has stopped matching suppresses **nothing** and says so.

`raw` and the result line above it disagree deliberately, and the band says so: acceptance is measured with the portable kernel while the result line still comes from `drawImage`. They converge when the versioned rebaseline (D3) lands, which is a change of its own with regenerated baselines and a release note.

Wired for future coverage per the repo rule: the richest reference-compare fixture now carries a `knownDifferences` scope, so the band's markup, payload and script tag are diffed by the preview harness on every PR. The fixture cannot show the band *filled* — a page opened as a file has nothing to fetch — which is the honest split, with `cli/serve-web/test/acceptance.test.ts` covering the engine end to end. `docs/design/evidence/parity-acceptance-band/` carries the two screenshots and the scripts that make them.

## What this does not do

- **The offline engine in `design-parity` (#3808) is not written.** It is the second implementation the fixtures exist for, and a batch's worth of work on its own — the whole ladder in another language, plus a decision about how the committed fixture tree reaches another repository.
- **The host-level catalog walk is written and not mounted.** `walkCatalog` in `src/parity/acceptance.ts` is what makes `orphaned-target` reachable from the browser; nothing on the dashboard calls it yet, so that status is still browser-invisible.
- **The D3 rebaseline is not here**, deliberately: it moves every catalog's published number and must not share a diff with a changed verdict.
- **No catalog publishes acceptances yet**, so the band is reachable today only on a hand-placed `parity/known-differences.json`. The publisher is the companion PR.

## Test plan

- `node --test` in `scripts/design-artifacts/` — 224 pass across the gate, scoring, plane, inflate and sha256 suites, including the test that regenerates the whole fixture tree into a scratch directory and compares digests.
- `npm run verify` in `cli/serve-web/` — typecheck, 1104 mocha tests, rebuild, committed-bundle freshness.
- `./gradlew :cli:test` — green, including the 14 new `ServeKnownDifferencesTest` cases and the regenerated `serve-reference-compare` golden (whose diff is exactly the four lines the band adds; the three sibling goldens are byte-identical).

Refs #3808, #3810, #3809, #3680.


yschimke:
**On the two `serve-viewer-exploded-controls` rows above: they are a capture race, not this diff — and they are *not* the phantom row already on the loose-ends list.** Measured rather than assumed, because that list exists so real rows stop being discounted.

The two captures genuinely differ — 4.05% of pixels dark, 4.02% light, max channel delta 231, bounding box `28,149 968×752`. Nothing like resampling noise. But the change is entirely this:

| | Before | After |
| --- | --- | --- |
| SVG comparison chip | `comparing…` | `69.2% match` · `view diff →` |
| Below the panel | `rendering…` | *(gone)* |

That is the viewer's client-side SVG-vs-PNG score finishing before the screenshot in one run and not the other; the footer then shifts up by the height of the status line, which is what the large bounding box is measuring.

It cannot be this branch, and the checks are cheap:

- `viewer.js`, `format-compare.js`, `serve-components.js` and `serve-chrome.js` are **byte-identical to `main`** on this branch — `git diff --quiet origin/main..HEAD` on all four.
- The only page golden that changed is `serve-reference-compare.html`. `serve-viewer-exploded-controls.html` is untouched.
- `serve.css` gains 19 lines, every one scoped to a `.cp-acceptance*` class that does not exist on the viewer page.

Worth distinguishing from the instability [`00-decisions.md`'s loose ends](https://github.com/yschimke/compose-ai-tools/blob/main/docs/design/parity-batches/00-decisions.md) already records. That one is a fixture that "captures byte-identically at base and head across repeated runs" — a stored-baseline drift. This one is not byte-identical and the difference is legible: an async score that had not settled when the shutter fired. Two different problems, and only the second is fixable by making the harness wait for the score before capturing `serve-viewer-exploded-controls`.

Not fixed here — it is a harness race in a fixture this PR does not touch, and widening the diff to chase it would make the acceptance change harder to review, not easier.

---

yschimke:
Four of the five are fixed; the fifth is a real gap in `v1` that cannot be closed in the adapter without the two engines disagreeing. Each was checked against the code first.

**Transform tag bounds into the canonical plane** — correct, and it is the exact mistake §4's own table warns about: *"an engine that expects canonical bounds from the index either compares raw render coordinates or transforms an already-transformed box, and both report `element-moved` for an element that never moved."* I passed the raw index straight through.

Fixed in the **shared** module rather than the adapter, so `design-parity` inherits it: `projectTagIndex` / `projectRenderBox` in `known-difference-plane.mjs` subtract the candidate box's origin, scale x and y independently, round outward at both ends and clip to the plane. A box that clips away keeps its `count` and loses its `bounds` — the tag is still in the tree, and "vanished" is a different verdict about a different fact. Six new fixtures in a `tag-projection/` group, pure geometry so every expectation is derivable by hand; the first is deliberately the identity case that *hides* the bug (zero origin, 1:1 scale), because an engine skipping the transform passes that one and fails the other five. The adapter test also now pins it end to end on a cropped plane, where skipping the projection gives a displacement of 9 against a tolerance of 0.1.

**Disable current acceptances on pinned comparisons** — correct, and embarrassingly my own code comment had already worked out why the pin must not reach the document URL and then failed to draw the conclusion. Both panels take the pin, so the pixels are historical while the document and `referenceSha256` are current: the fingerprint gate can pass on today's metadata against yesterday's bytes, and an acceptance authored long after that revision would be reported `valid` and suppress pixels on a comparison it was never published for. The payload is now gated off `pinned`, like the annotation and tag layers beside it.

**Render document-level validation failures** — correct. A document rejected wholesale returns no `statuses`, so the band showed scores above an empty list, which reads as "this catalog accepts nothing here" rather than "this catalog's acceptance document was refused" — the same picture with opposite meanings. The document-level rows (those with no `id`) now render as their own line.

**Distinguish document fetch errors from an unpublished catalog** — correct. `absent`, `unavailable` and `evaluated` are three states now, not two booleans: a 404 is the only absence, and a 401/500/network drop says so on the band rather than hiding it. Hiding it was the worse failure, because the page only carries this evaluator when the *server* already found a document.

---

**Not fixed: buffering every artifact.** The measurement is right — `prefetch` fetches both artifacts for every record and holds them until evaluation returns, and `v1` permits 256 records with two 8 MiB artifacts each. Neither remedy suggested works, and the reasons are worth recording:

- **Scoping the fetch to in-scope records changes verdicts.** The engine's preflight deliberately runs over *every* record before scope is considered, because the pixel and count caps are **document**-level verdicts. A browser that fetched only in-scope artifacts would report a different document verdict from the offline engine for the same bytes — which is the one outcome this whole contract exists to prevent, traded for a memory bound.
- **Streaming needs an async `readArtifact`,** and it is synchronous by design: the ladder is a sequence of ordering requirements (preflight strictly before decode, gates strictly before scoring) and threading a promise through it turns each into a race.

The real fix is in the schema, not here: `v1` caps count, decoded pixels, per-axis size and *per-artifact* encoded bytes, and has no **aggregate encoded-byte** cap. Adding one — a `document-too-large` sibling checked in the same bounded preflight that already reads each header, with boundary fixtures on each side — makes both engines refuse the same documents and gives the browser a bound it can rely on. That is batch-04 schema work with fixtures and two engines to update, so it belongs in its own change rather than widened into this one.

Worth sizing the exposure honestly: it takes a deliberately-constructed document with hundreds of maximal artifacts. The measured pilot population is four acceptances across six sites, the host already refuses any single artifact over 8 MiB with a 413, and a catalog is third-party data — which is exactly the threat model the caps exist for, so I am not arguing it away, only saying where the fix belongs.

---

PR #4514, opened 2026-08-26
fix(parity): take the locator's system from the session id, not the URL path

yschimke:
Follow-up to #4508, which merged while this was being written. Found by review there, on the last head before the merge.

`basePath.trim('/')` is the session id put through `urlEncodeSegment`, and the two differ for a catalog whose name carries a character a URL segment escapes — `@` is legal in a session name (`ServeBundleStore.sanitizeName` permits `[A-Za-z0-9._@-]`) and encodes to `%40`.

The locator's identity then changed spelling with the route form: `/<system>/parity` and `/<system>/compare/…` recorded `foo%40bar`, while `?session=foo@bar` and a top-level site recorded `foo@bar`. An acceptance matches on **every** recorded field, so the same `known-differences.json` is healthy through one route and `orphaned-target` through the other, from identical bytes — and an issue filed through one route carries a locator the index joins differently from one filed through the other.

`sessionId` is the resolved catalog id in every route form — the path parameter, the site host and `?session=` all resolve to it — so the four locator sites read it directly. Two other call sites in the same file already spelled it that way. For every catalog whose name needs no escaping the emitted bytes are identical, so no page or report changes.

The URLs stay escaped, which is what a path segment is for; only the identity is raw. The new case asserts both against a session named `acc@ept`: `"system":"acc@ept"` in the audit payload, `/acc%40ept/parity/known-differences.json` as its document URL.

## Test plan

- `./gradlew :cli:test` green in full.
- `ServeParityAcceptanceRouteTest` gains the `@`-named catalog case above.
- `./gradlew ktfmtFormat`.

---


yschimke:
**CMP Remote Compose Player** is red on `61fcc64`, and it is not this PR's:

```
Execution failed for task ':slot-preview-runtime:ktfmtCheckScripts'
> Could not read path '…/runtimes/slots/build/tmp/ktfmtCheckKmpCommonMain/5463ca3c-1893-4470-becb-13da7fd74966'
```

A ktfmt task failing to read its own Gradle temp file, in `slot-preview-runtime` — a module this diff does not touch. The diff is two files: `ServeHttpServer.kt` (four locator sites now read `sessionId`) and `ServeParityAcceptanceRouteTest.kt`. Nothing in it reaches the runtimes tree, and the build died in a formatting task's scratch directory rather than in any test body, which is the signature of the tmp-dir race between parallel ktfmt tasks rather than a real check failure. `./gradlew ktfmtFormat` and the full `:cli:test` pass locally on this commit.

No fix exists to port — there is nothing to fix here yet beyond the race itself. Re-run requested and refused with *"This workflow is already running"* while the rest of the run is in flight; I will re-run this job once the run settles, and if it fails a second time I will treat it as real.

---

Issue #4522, opened 2026-08-26
serve: a refreshing catalog serves one host against another generation's files


yschimke:
From review of #4515. The document half is fixed there (c1c9483); this is the general case, which belongs to `ServeCatalogStore`'s storage model rather than to any one writer.

## The window

`ServeCatalogStore.load` stages a catalog into a temporary directory, then **renames it over the live `dir`** — and only *after* that does it do its post-swap work (the Wasm app, vectors, themes, live bundles) before finally calling `register(safe, host)` with a freshly-built host.

Between the swap and the registration, the **old host is still the registered one**, and it is reading from a directory that now holds the **new** generation's files. Every accessor that reads lazily from `bundleDir` is exposed:

- `knownDifferenceArtifact(relativePath)` — reads bytes per call, and must, since holding every artifact is what `maxArtifactBytes` and the streaming fetch exist to prevent.
- `raster(referenceId)` in `ServeDesignReferenceStore`, preview PNGs, design-page crops, and every other byte-serving path.

The accessors that read **once at construction** — `previews`, `parityIssues`, `parityActivity`, annotations, the reference manifest, and now the known-difference document — are consistent with each other and with the generation the host was built for. So the host is internally coherent and its *files* are not.

## What it actually costs

Worth being precise, because the severity varies by artifact and it is not uniform:

- **Hash-guarded content degrades to a refusal.** A known-difference mask or accepted candidate is admitted only if its bytes hash to the record's recorded `sha256`. Changed bytes fail the gate (`plane-changed` / `candidate-changed`); removed bytes are `artifact-unreadable`. The acceptance is not applied, so the raw difference stays reported. Late, not wrong — and it self-heals on the next tick.
- **Un-hashed content can be served mismatched.** A preview PNG or a design-reference raster fetched in this window is the new generation's pixels beside the old generation's metadata, with nothing to notice. This is the half that can be quietly wrong rather than visibly refused.

The window is not microseconds. The rename is, but the post-swap work between it and the registration includes network fetches.

## Options

1. **Generation-scoped directories.** Stage into `dir-<generation>`, register the new host against it, and delete the previous generation once no request is using it. The host and its bytes then change together, by construction, and the swap stops being a mutation of shared state. This is the real fix and the largest: it touches how every catalog's storage is named, cleaned up, and referenced.
[...]

Issue #4521, opened 2026-08-26
serve: a catalog asset the transport refuses by size is indistinguishable from one that isn't published


yschimke:
**Update: the availability half is fixed in #4525.** A load now carries `incomplete` when a read ended in a "right now" failure, the refresher withholds the recorded head so the next tick re-reads, and `seedInitialHeads` excludes an incomplete startup load. That closes option (2) below — the silent, permanent loss.

**What remains is the verdict half**, retitled accordingly.

---

## The property that is left

`fetchCatalogAsset` is `cachedBranchRead(url).bytesOrNull` — it keeps the bytes and discards the `BranchFetch` outcome. Two events still collapse into one answer:

| what happened | what the writer sees | what the catalog then serves |
| --- | --- | --- |
| the producer publishes no such file (a real 404) | `null` | nothing — correct |
| the body exceeds `MAX_FETCH_BYTES` (25 MiB), so `readCapped` throws | `null` | nothing — **a lost verdict** |

`compose-preview-known-differences/v1` says a document past `maxDocumentBytes` is `document-too-large`/413 and an artifact past `maxArtifactBytes` is `artifact-too-large`/413. #4515 stages over-sized files precisely so `ServeKnownDifferences` can answer `TooLarge` from their length — but only up to the transport's own 25 MiB ceiling. Above it the file never arrives, the reader sees nothing, and the engine records `artifact-unreadable`/404: a different verdict, and one that hides why.

It applies to every writer that reaches the network through `fetchCatalogAsset`, but only known differences has a contract that distinguishes "too large" from "absent", so only there does it cost anything today.

## Why it wasn't done in #4525

#4525 is deliberately outcome-agnostic: it asks one question ("did anything fail transiently?") from a counter, and needs no per-writer judgement. This half is the opposite — each writer has to decide what a size refusal *means* for its own artifact, and for most of them the honest answer is "nothing, treat it as absent".

## Options

1. **Return the outcome, not just the bytes.** A `fetchCatalogAssetOutcome` beside the existing helper, and only the writers that care consult it. For known differences, a size refusal stages a marker whose *length alone* is past the ceiling — the reader measures the file and answers `TooLarge` without ever reading it.
2. **Raise `MAX_FETCH_BYTES` for known differences.** Cheapest, and wrong: it moves the hole one size up rather than closing it.

[...]

PR #4523, opened 2026-08-26
fix(parity): mirror JavaScript's blank-id test, not the JVM's

yschimke:
Follow-up to #4515, which merged while this was being written. Found by Codex on that PR's last review round.

`rejectsWholeDocument` mirrors the engine's document-level rejections so a fetch planner knows when the engine will read nothing. Its KDoc states the one direction it must never be wrong in:

> claiming a rejection the engine would not make starves legal records of their artifacts and turns them into `artifact-unreadable`

It was wrong in exactly that direction.

## The mismatch

`String.isBlank()` delegates to `Character.isWhitespace`, which counts the four separators **U+001C..U+001F** and **U+0085** as whitespace. ECMAScript's `TrimString` does not:

| char | `trim()` empties it | `Character.isWhitespace` |
| --- | --- | --- |
| U+001C..U+001F | **no** | **yes** |
| U+0085 (NEL) | **no** | **yes** |
| U+00A0, U+2007, U+202F | yes | **no** |
| U+FEFF | yes | **no** |

So a record with an id of a single U+001C is **keyable to the engine**: `identityFailures` lets it through, `preflightRecord` refuses that one record as `id-not-safe`, and every other record in the document is read normally. The stager called it unkeyable, rejected the whole document, and skipped the artifacts of every legal record beside it — turning them into `artifact-unreadable`. A changed verdict on legal records, from one control character in one id.

The disagreement runs the other way too — `trim()` removes the non-breaking U+00A0, U+2007, U+202F and U+FEFF that `Character.isWhitespace` does not — but that direction only costs a fetch, so it was never the dangerous half.

## The fix

The set is spelled out from the specification rather than borrowed from either language: `WhiteSpace` ∪ `LineTerminator` — the three format controls (U+0009, U+000B, U+000C), the four line terminators (U+000A, U+000D, U+2028, U+2029), the Unicode `Zs` category via `Character.getType`, and the byte-order mark.

Borrowing a host's definition is what caused this; the KDoc says so in place, so the next reader doesn't reach for `isBlank()` again.

## Test plan

Two tests, one per direction, both in `ServeCatalogStoreTest`:

- `an id blank only to the JVM does not cost the other records their artifacts` — a document pairing a U+001C id with a legal `glyph` record asserts `glyph`'s mask **is** fetched and reaches the host. **Verified this fails on the pre-fix code**, so it pins the bug rather than merely passing.
- `an id blank to JavaScript still rejects the document` — a U+00A0 id must reject, so nothing is fetched.

`./gradlew :cli:test` green in full on `main` at `b2163c6`; `./gradlew ktfmtFormat`.

Text-only evidence: this changes which bytes a serving host fetches, not how anything is drawn.

## Also on that review round

Two other findings, neither needing code here:

- **Transient artifact-fetch failures collapse to a missing file**, and `lastHead` then advances so the revision is never retried. Same root as the document case — tracked in #4521, which covers every writer in the store rather than this one.
- **The artifact accessor stays directory-backed** across a refresh's swap window. Bounded to refusals (the hash gates turn changed bytes into `plane-changed` / `candidate-changed`, missing bytes into `artifact-unreadable`), never a false accept. The real fix is generation-scoped directories — #4522.

---

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogStore.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogStore.kt
@@ -2112,9 +2112,47 @@ class ServeCatalogStore(
     return false
   }
 
-  /** A record's `id` as the engine keys it: a non-blank string, or nothing. */
+  /**
+   * A record's `id` as the engine keys it: a string that is not blank *to JavaScript*, or nothing.
+   */
   private fun recordId(fields: JsonObject): String? =
-    (fields["id"] as? JsonPrimitive)?.takeIf { it.isString }?.content?.takeIf { it.isNotBlank() }
+    (fields["id"] as? JsonPrimitive)
+      ?.takeIf { it.isString }
+      ?.content
+      ?.takeIf { !isEcmaScriptBlank(it) }
+
+  /**
+   * Whether `String.prototype.trim()` would empty this string — the engine's own test for an
+   * unkeyable id, spelled out rather than borrowed from the JVM.
+   *
+   * `String.isBlank()` is the obvious Kotlin answer and it is the wrong one, in the direction this
+   * mirror must never be wrong in. It delegates to `Character.isWhitespace`, which counts the four
+   * separators U+001C..U+001F and U+0085 as whitespace; ECMAScript's `TrimString` does not. So an
+   * id of a single U+001C is *keyable* to the engine — the record fails on its own as `id-not-safe`
+   * while every other record in the document is read normally — and would have rejected the whole
+   * document here, skipping the artifacts of every legal record beside it and turning them into
+   * `artifact-unreadable`. That is a changed verdict on legal records, which is the one failure
+   * direction [rejectsWholeDocument] exists to avoid.
+   *
+   * (It differs the other way too — `Character.isWhitespace` excludes the non-breaking U+00A0,
+   * U+2007, U+202F and U+FEFF that `trim()` removes — but that direction only costs a fetch.)
+   *
+   * The set is `WhiteSpace` ∪ `LineTerminator` from the specification: the three format controls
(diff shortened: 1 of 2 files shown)


yschimke:
**`Playground compiles + redeems a snippet end to end` failed, and it isn't this PR's.** Recording it once rather than leaving it unexplained.

The build died resolving dependencies, before any test body ran:

```
Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException:
  Could not GET 'https://dl.google.com/dl/android/maven2/org/jetbrains/compose/runtime/
                 runtime-iossimulatorarm64/1.11…'
    at DefaultExternalResourceArtifactResolver.downloadByCoords(…:149)
    at DefaultCacheAwareExternalResourceAccessor.copyToCache(…:188)
```

surfaced through Kotlin's `KmpPartiallyResolvedDependenciesChecker`. `BUILD FAILED in 3m 11s`, and `playground-lane.log` was never created — the subsequent steps report `No such file or directory` and the artifact upload found nothing to upload. So the playground lane never started; the failure is an artifact download from `dl.google.com`.

This diff is two files — `ServeCatalogStore.kt` and `ServeCatalogStoreTest.kt` — changing how one function decides whether a string is blank. It adds no dependency, touches no Kotlin Multiplatform target, and goes nowhere near the playground or an iOS-simulator artifact. There is no version of this change that makes a Maven GET fail.

**Re-run:** I tried `rerun-failed-jobs` and got `403 This workflow is already running` — the run still has jobs in flight, so its failed job can't be re-run yet. I'll re-run it once the run settles; the allowance is unspent and this is the case it exists for (a job that died before any test body ran). I'm keeping the PR watched until it's green.

Locally on `19886de`, against `main` at `b2163c6`: `./gradlew :cli:test` green in full, including the two new cases, and `ktfmtFormat` clean.

---

PR #4531, opened 2026-08-27
fix(parity): keep a prototype-named tag an own key in the projected index

yschimke:
Found while porting #4529 into `design-parity`, from the far side: `design-parity`'s vendored copy of `known-difference-plane.mjs` had been hand-patched with `Object.create(null)` and the fix never came back here. Re-vendoring from `main` reverted that hardening, and `design-parity`'s own test caught it.

## The defect

`projectTagIndex` built its result as a plain `{}` and keyed it by producer-controlled `testTag` names. `__proto__` is an ordinary string in a semantics tree and a catastrophic object key:

```
own keys of result: [ 'glyph' ]
__proto__ tag present as own property? false
result prototype polluted? true
the tag the producer published: { count: 1, bounds: { x: 1, y: 2, width: 3, height: 4 } }
```

The assignment **replaces the prototype** instead of creating an own property. So the tag vanishes from `Object.keys` and every iteration built on it, while a direct `index["__proto__"]` still answers through the prototype chain. A consumer that iterates and a consumer that reads by name disagree about whether the producer published that tag at all, and an element acceptance targeting it resolves differently depending on which the reader used — the silent, reader-dependent divergence §4 exists to prevent.

The plain object has a second edge the same change closes: a tag the producer **never** published — `constructor`, `toString`, `valueOf` — reads as present through the prototype, so a consumer testing `if (index[tag])` gets a false positive.

This repo already treats the class as important. `known-differences.mjs` refuses `__proto__` as a record `id` (`id-not-safe-proto`, whose `why` says "the browser builds `statuses` as a `Map` or a null-prototype object"), and `joinedIssues` at `known-differences.mjs:485` already builds its map with `Object.create(null)`. Tag names never pass through the id rules, and this was the one `= {}` left in the module.

## The fix

`Object.create(null)`, matching `joinedIssues` and the copy `design-parity` had already arrived at independently.

**The runner needed one change with it.** `known-difference-score.test.mjs` compared with a strict `deepEqual` against a `JSON.parse`d expectation, which compares prototypes — so a null-prototype result failed all seven tag-projection cases even where the keys and values matched exactly. That comparison is an artifact of how this runner happens to read the fixture, not something the tree pins: a runtime without JavaScript prototypes has no such distinction to make. It now compares own enumerable entries, which is what the fixtures actually specify. Spread is used rather than a manual copy because spread *defines* own properties, so `__proto__` survives the normalisation.

## The fixture taught the lesson twice

`a-prototype-named-tag-stays-an-own-key`, in the `tag-projection` group.

My first version wrote the case the obvious way — `tagIndex: { __proto__: {...}, glyph: {...} }` — and produced a `case.json` whose `tagIndex` had **no `__proto__` key at all**, because that literal syntax *is* the prototype setter. The fixture would have passed against the buggy code by testing nothing. It is now assigned onto a null-prototype object, and the committed `case.json` and `expected.json` both carry `["__proto__", "glyph"]`.

## Test plan

- `known-difference-score.test.mjs` 23/23 (22 before), `known-differences.test.mjs` 225/225, `cli/serve-web` 1134 passing, `tsc --noEmit` clean, `check-serve-seam: OK — 151 crossing symbol(s)`.
- The new fixture verified to pin the bug: with `Object.create(null)` reverted to `{}`, **exactly** `a-prototype-named-tag-stays-an-own-key` fails and nothing else. With the fix and the old prototype-comparing assertion, all seven tag-projection cases fail — which is how the runner change was found rather than assumed.
- `npm run build` in `cli/serve-web` — `known-differences.js` embeds the engine, so the committed bundle is regenerated.

Text-only evidence: this changes which keys a projected index carries, not how anything is drawn.

## Follow-on

`design-parity`'s vendor is being brought forward in its own PR; with this merged its copy is a verbatim mirror again rather than a locally-patched one. That port also revealed the fixture tree moved to root-relative content-addressed `rasters/`, which its conformance runner still resolved against the case directory — carried there, not here.

---


yschimke:
The preview bot's one changed capture (`SwitchButtonOn`, wear-m3) is **not this PR's** — recording the check so a reviewer doesn't have to repeat it, since the body claims this changes nothing that draws.

**This PR cannot reach a renderer.** Its whole diff is the parity engine, its fixtures, its own test, and the browser bundle that embeds the engine:

```
cli/src/main/resources/.../assets/known-differences.js
scripts/design-artifacts/build-known-difference-fixtures.mjs
scripts/design-artifacts/fixtures/known-differences/…
scripts/design-artifacts/known-difference-plane.mjs
scripts/design-artifacts/known-difference-score.test.mjs
```

Nothing in `renderers/`, `gradle-plugin/`, or `samples/`. `projectTagIndex` builds the semantics tag index the acceptance gates read; it is downstream of rendering and no part of it.

**The bot says so itself**, in its own note: the baseline is rendered from `5ad68765`, **two commits behind** this PR's base `2551bd2e`, and "any preview those commits changed is attributed to this PR below." Those two are #4529 and #4528.

**Checked rather than assumed**, because an APNG is exactly where I would expect capture nondeterminism and I would rather rule it out than wave at it:

| | before | after |
| --- | --- | --- |
| frames | 114 | 114 |
| `acTL` num_frames / plays | 114 / 0 | 114 / 0 |
| `IDAT` / `fdAT` chunks | 1 / 113 | 1 / 113 |
| frame delays | all `1/60` | all `1/60`, **0 differing** |
| bytes | 532,202 | 545,282 |

Identical structure and identical timing across all 114 frames, so it is not a capture-rate or frame-count wobble; the compressed size moved because frame *content* did. That is a real difference in the render, produced by something in the baseline gap rather than by this diff.

Re-running the preview diff once the baseline for `2551bd2e` has published will show this PR's own changes, which should be none.

---

PR #4540, opened 2026-08-27
chore: drop the sticky release-as key now that v1.36.0 has published

yschimke:
The cleanup #4537 promised. One line removed, exactly the inverse of what `6a80699c` added.

```diff
       "package-name": "compose-ai-tools",
-      "release-as": "1.36.0",
       "include-component-in-tag": false,
```

## It did its job, and now it has to go

`release-as` is **not consumed by the release it forces**. Left in place, the next release PR proposes 1.36.0 again, the tag already exists, and the release wedges. `docs/RELEASING.md` records this having happened twice — once via a `1.0.0` pin, once via a `0.7.0` override that outlived its release. This PR is what stops it being a third.

Verified the override actually worked before removing it, rather than assuming:

| | |
| --- | --- |
| #4500 final title | `chore(main): release 1.36.0` (was `chore(main): release 2.0.0`) |
| merged | 2026-08-27T08:06:58Z |
| tag | `v1.36.0` present |
| release | `v1.36.0`, draft — same as `v1.35.0` / `v1.34.0`, i.e. the configured `"draft": true` |

`jq .` clean.

## Why it took three PRs to force one version

Recorded here because the mechanism `docs/RELEASING.md` recommends cannot work in this repo, and the two failures were both **silent** — merged, green, no effect:

| PR | Payload location | Outcome |
| --- | --- | --- |
| #4532 | PR body | Ruleset allows only `squash`; repo sets `squash_merge_commit_message: BLANK`, so the body is discarded. Commit `0a073692` landed with an empty body. |
| #4533 | Commit message, **empty commit** | Squashing an empty diff produces no commit at all — GitHub recorded `merge_commit_sha` as the pre-existing `0a073692`. Enabling rebase would not have helped: git drops empty commits by default. |
| #4537 | **File content** | Worked. Squash preserves the diff. |

The lesson is narrow and worth keeping: under `squash` + `BLANK`, the only thing that reliably reaches `main` is a **file change**. Anything carried in a commit message or PR description is discarded.

## Follow-ups this exposed

Both docs are actively misleading and cost three round-trips:

- **`CLAUDE.md`** says PRs are "squash-merged from the **PR title + PR body**". The body is discarded. Its instruction to scrub `Co-authored-by:` from PR *descriptions* is therefore moot — a description cannot reach `main`.
- **`docs/RELEASING.md`** recommends the `Release-As:` footer as the preferred mechanism, which this repo's squash-only ruleset makes unreachable. It should recommend the config key *plus a mandatory cleanup PR*, and say why the footer does not work here.

Happy to fix both in a follow-up if you want them.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YQJKxzuDLpRUepa8RYeaU5)_

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -6,7 +6,6 @@
   "packages": {
     ".": {
       "package-name": "compose-ai-tools",
-      "release-as": "1.36.0",
       "include-component-in-tag": false,
       "draft": true,
       "extra-files": [


yschimke:
`Build CLI` failed on `abd36744`, and it is not this PR's.

```
ServeStatusLiveFramesTest > a live socket's frames are reported on status_json() FAILED
    java.lang.IllegalStateException at ServeStatusLiveFramesTest.kt:127
3159 tests completed, 1 failed
```

**Why it cannot be this change.** The diff is one deleted line in `release-please-config.json`. That file is read by release-please, never by `:cli:test` — there is no path by which it reaches a serve HTTP socket. This is structural rather than a judgement call: the PR touches no Kotlin, no build logic, and no test fixture.

**What the failure actually is.** Line 127 is the `error(...)` at the end of `awaitLiveFrames`:

```kotlin
repeat(100) {
  val (code, body) = get(server, "/status.json")
  …
  Thread.sleep(50)
}
error("liveFrames never reported the expected frames: $last")
```

A 5-second bound (100 × 50 ms) on frames that cross threads to reach `/status.json`, over a live socket, on a shared runner executing the rest of the matrix beside it. Nothing in the test asserts frames arrive *quickly* — only that they arrive — so the deadline is a liveness bound, and a runner with no CPU to spare exhausts it. Same shape as the `pollUntil` timeout fixed in #4528 for `:daemon:core`, and it is the identical fix if it recurs: scale the bound on CI rather than assert latency.

I have not pushed that fix here. It would widen a one-line release-config PR into `cli/`'s test suite, and one failure is not yet evidence of a pattern.

Re-running the job once — my single re-run for this PR, per the flake rules. A second failure is real and I will treat it as such.

If this test starts showing up on unrelated PRs, it is worth making robust on its own terms rather than re-run each time; happy to open that separately.

---

yschimke:
`Module Unit Tests` also failed, and this one is settled rather than suspected: **it is red on `main` too.**

I checked the base rather than inferring from the diff. `main` at `b78dfef6` — the `chore(main): release 1.36.0` commit — fails `Module Unit Tests` with the same three tests:

| Test | Where | Fails on `main` @ `b78dfef6` | Fails here @ `abd36744` |
| --- | --- | --- | --- |
| `AndroidRippleFrameTest > pressedRippleKeepsAnimatingAcrossHeldFrames` | `:daemon:android` (`:121`) | ✅ | ✅ |
| `LivePressRippleTest > aLiveClickPaintsPressFeedback` | `:daemon:android` (`:92`) | ✅ | ✅ |
| `ServeStatusLiveFramesTest > a live socket's frames…` | `:cli` (`:127`) | ✅ | ✅ |

Same tests, same line numbers, same counts (`327 tests completed, 2 failed`; `3159 tests completed, 1 failed`). This PR deletes one line from `release-please-config.json` and cannot reach any of them.

**Correcting my earlier comment on this thread.** I attributed the `Build CLI` failure to a loaded runner and treated it as a one-off flake. That was incomplete: `ServeStatusLiveFramesTest` fails on `main` as well, inside `main`'s own `Module Unit Tests` job. `main`'s separate `Build CLI` check did pass, so the test is *intermittent* rather than reliably broken — but it is failing on the base branch, not just here. My re-run has not run yet (`rerun_failed_jobs` returned `403 This workflow is already running`), and on this evidence it is not worth spending: a re-run cannot make a base-branch failure into a pass for the right reasons.

**No fix exists to port.** I looked for one; there is no open PR addressing these. Per the rules I am not widening a one-line release-config PR to fix three timing-sensitive tests across `:daemon:android` and `:cli` — that is its own change, with its own review.

**What this means beyond this PR:** `main` is currently red. All three are timing assertions — ripple animation frames, and a 5s poll for socket frames crossing threads — which is the same failure class as the `pollUntil` bound fixed in #4528 for `:daemon:core`, and the same fix shape applies: make the deadline a liveness bound scaled for CI, rather than an implicit latency assertion. Worth doing deliberately rather than as a drive-by.

Flagging it to the repo owner rather than acting unilaterally, since a red `main` affects every PR in flight, not just this one.

---

PR #4619, opened 2026-08-28
test(renderer): fail the build when the forward Compose pin stops being forward

yschimke:
## Summary

A P2 left on #4503 after it merged — with a correction to its premise, which changes what is worth doing about it.

**The reported state does not exist on `main` today.** #4527 holds `compose-multiplatform` at `1.11.1` against `compose-multiplatform-forward = "1.12.0-rc01"`, so `composeForwardTestRuntime` genuinely is the newer line and `forwardComposeSystemThemeTest` covers what it claims to. Verified in the catalog rather than assumed. So there is no broken pin to bump.

**What is missing is anything that keeps it that way.** The two are separate catalog entries moved by separate Renovate rules, and the relationship between them lived only in a reader's head — the gap was visible on #4503 only because someone read both lines. The task passes whether the forward runtime is ahead, level, or behind, so the coverage can lapse silently. That is the durable version of the finding, and it is what this fixes.

The build now refuses it at configuration time. A pre-release sorts **below** the release it precedes, so a production bump to `1.12.0` final fails against a forward pin of `1.12.0-rc01` rather than quietly levelling — which is the exact shape #4503 described.

Two smaller drift points in the same file, both of which would go stale the same way: the task description hardcoded `1.12.0-rc01`, and the `strictly` version re-read the catalog independently. Both now derive from the one value the check reads.

## Test plan

- Guard fires on the reported scenario. Setting `compose-multiplatform = "1.12.0"`:

  ```
  * What went wrong:
  compose-multiplatform-forward (1.12.0-rc01) must be strictly newer than compose-multiplatform
  (1.12.0). forwardComposeSystemThemeTest exists to run the renderer invariant on the next Compose
  line; at or below the production pin it runs on an identical or downgraded graph and passes while
  covering nothing. Bump the forward pin in gradle/libs.versions.toml.
  ```

  That case is the one a naive comparison gets wrong — the cores are equal and only the pre-release ordering separates them.
- Catalog restored: `:renderer-desktop:tasks` configures clean and prints `forwardComposeSystemThemeTest - Runs the light/dark renderer invariant on Compose Multiplatform 1.12.0-rc01`, i.e. the description is now read from the catalog.
- `:renderer-desktop:ktfmtFormatScripts` clean. (Worth noting: `ktfmtFormatMain`/`Test` do **not** cover `*.gradle.kts` — the `pre-push` hook caught it, and the `steward` skill's pre-push list could say so.)
- No visual surface.

https://claude.ai/code/session_01LJVTdPE3RwJ1gLMbLpThZP

---


yschimke:
Check the failing build

yschimke:
Fix build issues

yschimke:
Checked the failing build. Both red checks — **Actions Script Tests** and **Build CLI** — were failing on the same thing, and it was not this PR's diff:

```
test/serveInternalsUsedByCli: refusing to write 1 NEW crossing(s) into the allowlist.
FAIL: test_repo_is_green (__main__.RealTree.test_repo_is_green)
AssertionError: 1 != 0
```

This branch only touches `renderers/desktop/build.gradle.kts`. The unregistered crossing came from #4616, which branched here before #4620 landed the `serve.ServeBackgroundWork` registration in `scripts/serve-seam-allowlist.json`. The fix already exists on `main`, so I ported it the only way that works — merged `origin/main` (now at `ebf561561`) into the branch rather than waiting for a rebase.

Validated locally on the merged head before pushing:

- `python3 scripts/test_check_serve_seam.py` — 45 tests, OK
- `./gradlew :cli:checkServeSeam` — `OK — 123 crossing symbol(s), all on the allowlist`
- `./gradlew ktfmtCheckAll` — clean
- `./gradlew :renderer-desktop:tasks` — configures, and `forwardComposeSystemThemeTest` still resolves against Compose Multiplatform 1.12.0-rc01

One note for anyone reproducing the pre-push hook locally: `vscode-extension/node_modules` is absent in a fresh sandbox, so the hook's `npx prettier` resolves to whatever npm serves rather than the lockfile's 3.9.6, and reports ~32 files unformatted that are byte-identical to `main`. `npm --prefix vscode-extension install --no-save prettier@3.9.6` makes the hook agree with CI; nothing in the tree needed reformatting.

---

Issue #2957, opened 2026-07-29
Capture a still frame for AndroidView-hosted and scrolling-container previews


yschimke:
Follow-up to #2946, which is closed by #2956 — that PR added the *declarative* half (`"capture": "animated"` keeps a non-static component in the spec without sinking the publish gate). This issue tracks the renderer half the original issue asked for first.

## What's missing

Two shapes of ordinary preview render no `previews/<id>.png`, even though a still frame is well-defined for both:

- **`AndroidView`-hosted content** — the hosted platform `View` can be drawn to a `Canvas` (`view.draw(canvas)` after measure/layout), so the composition's pixels are obtainable.
- **Scrolling containers** (horologist `ScalingLazyColumn` and friends) — a scrolling screen has a well-defined initial frame; today these come out as a scrolling GIF (or nothing) rather than a still.

Both currently land in the "no static PNG" bucket, so the export drops them from the candidate join. With #2956 a consumer can declare that and publish, but the sticker sheet still carries no pixels for surfaces that are often the most important in the app (on Wear, the watch list *is* the home screen).

## Done when

- A plain `@Preview` whose body is an `AndroidView` emits a static PNG.
- A plain `@Preview` of a `ScalingLazyColumn`-based screen emits a static PNG of its initial frame.
- The catalog for such a component joins and exports a sticker with no `capture` declaration needed — i.e. `"capture": "animated"` becomes the exception (genuine animations) rather than the workaround.

Repro material: the run and previews listed in #2946 (`yschimke/pocket-casts-android@feffc63`, jobs 90641670656 / 90641670663) — `HtmlText` for the `AndroidView` case, `WatchListScreen` / `EpisodeListScreen` for the horologist case. The two `Summary*` previews in that report have no established cause and should be diagnosed separately.

yschimke:
I read the render logs from run 30471225420 (jobs 90641670656 / 90641670663) before touching code, and the framing in this issue doesn't survive them. **All eight previews failed to render** — none of them rendered-then-got-mis-bucketed — for three unrelated reasons:

**Wear (all four), `IllegalStateException`:**
```
Given component holder class androidx.activity.ComponentActivity does not implement
interface dagger.hilt.internal.GeneratedComponent or interface dagger.hilt.internal.GeneratedComponentManager
  at dagger.hilt.EntryPoints.get(EntryPoints.java:62)
[...]

PR #4646, opened 2026-08-28
fix(renderer): resolve previews whose parameters are all defaulted

yschimke:
Triage of the 71 failed renders on [preview.coo.ee/status](https://preview.coo.ee/status), and the two root causes that turned out to be ours.

## The triage

Every failure was pulled from each catalog's `catalog.json` `failures[]` on its `design-artifacts/<system>` branch and grouped by root cause. All 71 land in twelve clusters:

| # | Cluster | n | Catalogs | Owner |
|---|---|---|---|---|
| A | Activity tour — Hilt app-level DI | 36 | home-assistant-android ×17, pocketcasts-wear ×10, home-assistant-wear ×8, jetcaster-wear ×1 | renderer (open question, below) |
| B | Activity tour — Koin app-level DI | 2 | confetti-mobile, confetti-wear | same as A |
| C | Activity tour — DI-constructed Activity (Metro `AppComponentFactory`) | 1 | cadence | same as A |
| **D** | **Activity tour — Firebase SDK trampoline** | **4** | confetti-mobile ×2, confetti-wear ×2 | **fixed here** |
| E | Activity tour — needs Intent extras | 1 | confetti-mobile | consumer |
| F | Activity tour — nothing drawn | 1 | home-assistant-android | renderer, unclustered |
| **G** | **Preview — defaulted parameters not resolved** | **11** | jetlagged | **fixed here** |
| H | Preview — `hiltViewModel()` in the tree | 4 | pocketcasts-wear | consumer (fails in Studio too) |
| I | Preview — `AndroidView` needs the app theme | 5 | pocketcasts | consumer (`composePreview.hostTheme`) |
| J | Preview — Glance body under a Compose `@Preview` | 2 | jetchat | consumer / follow-up |
| K | Preview — never reaches idle | 2 | thunderbird | consumer |
| L | Preview — Compose ABI mismatch | 2 | confetti-mobile | consumer |

Two of those are bugs in this repo. They are what this PR fixes; 15 renders.

## G — previews whose parameters are all defaulted (11)

`ComposePreviewStrategy` resolved a no-argument preview through
`clazz.getDeclaredComposableMethod(functionName)`. That builds the JVM signature it looks for out of the argument types the caller passes, so with none it can only ever match `name(Composer, changed[, default])` — a preview whose own parameter list is empty.

It therefore misses the shape Studio renders happily and that samples reach for constantly: a component that *is* its own preview, every parameter defaulted.

```kotlin
@Preview
@Preview(widthDp = 500, name = "larger screen")
@Composable
fun AverageTimeInBedCard(modifier: Modifier = Modifier) { … }
```

Its compiled signature carries the real parameters ahead of the synthetic tail, the exact-signature lookup misses, and the render fails with a bare `NoSuchMethodException` naming a function that is plainly there:

```
java.lang.NoSuchMethodException: com.example.jetlagged.HomeScreenCardsKt.AverageTimeInBedCard
	at androidx.compose.runtime.reflect.ComposableMethodKt.getDeclaredComposableMethod(ComposableMethod.jvmAndAndroid.kt:202)
	at ee.schimke.composeai.renderer.ComposePreviewStrategy.Render(PreviewRenderStrategy.kt:133)
```

All 11 JetLagged failures are this one shape, `JetLaggedScreen`'s mangled `-uVTuf4s` name (a `WindowWidthSizeClass` value-class parameter) included.

**Fix.** Try the officially supported lookup first; only when it misses, scan `declaredMethods` for the name and take the composable overload that carries a defaults mask. `ComposableMethod.invoke` already knows how to call one — it fills every unsupplied parameter with the type's zero value and sets that parameter's bit in the mask, so the callee substitutes its own default expression. An overload *without* a mask is deliberately skipped: those zero values would go through as real arguments (a null `Modifier`, straight into an NPE), which is worse than reporting the miss.

## D — Firebase's headless activities (4)

`AppTourDiscovery` already drops library-injected activities by class-name prefix, but the list stopped at `com.google.android.`. Firebase Auth's `GenericIdpActivity` and `RecaptchaActivity` sit under `com.google.firebase.`, so manifest-merger's copies reached both Confetti catalogs as tour cards — and NPE'd inside `onResume` the moment they launched off a real device:

```
java.lang.NullPointerException: Cannot read the array length because "bytes" is null
	at com.google.android.gms.common.util.Hex.bytesToStringUppercase(play-services-basement@@18.9.0:2)
	at com.google.firebase.auth.internal.GenericIdpActivity.onResume(firebase-auth@@24.2.0:242)
```

Widening the prefix to `com.google.` subsumes the old entry and covers both. The list stays evidence-led rather than pre-emptive — every prefix is also a way to silently drop an app's own screen, so an SDK earns one once it has been seen producing a dead card.

## Not fixed here, and why

**A/B/C — 39 activity tours blocked by app-level DI.** Not one activity tour renders anywhere in the fleet except meshcore-mobile's; every DI-using app fails identically. The cause is the one `docs/APP_TOURS.md` already lists as a v1 limitation: the generated `robolectric.properties` pins `application=android.app.Application` for the whole render run, so an Activity that needs its `@HiltAndroidApp` Application never gets one. `composePreview.useConsumerApplication = true` restores it, but it is a *per-module* switch — flipping it would also push every composable preview in that module through the consumer's `Application.onCreate()`, which is the exact platform-init minefield the stub default exists to dodge. Doing this properly means giving ACTIVITY/APP_TOUR previews their own Robolectric config (a sibling package with its own properties file and test class), and that is a feature, not a triage fix. It also cannot be validated from here — Home Assistant's real `Application` under Robolectric may simply fail differently. Worth a decision before anyone builds it.

**I — 5 pocketcasts renders** are a one-line consumer fix, not a bug: `:modules:services:compose` is a library module, so its merged manifest has no `<application android:theme>` and `HtmlText`'s `setTextAppearance(UR.style.H50)` can't resolve `?attr/primary_text_01`. That is precisely what `composePreview.hostTheme` (#2957) exists for — `hostTheme = "@style/ThemeLight"` on that module should clear all five.

**H, J, K, L** are upstream preview defects (a `hiltViewModel()` call in the tree, Glance composables annotated with the Compose `@Preview`, a composition that never idles, a material3 ABI skew) that fail in Studio too.

## Test plan

- `./gradlew :renderer-android:testDebugUnitTest` — 301 tests, 0 failures, including the new `DefaultedComposableLookupTest`: six cases reflecting over *compiled* `@Composable` fixtures, since the compiled signature is the whole point and a hand-written `Method` stub could only approximate it. Covers a single defaulted parameter, several defaulted parameters, a required parameter (must be skipped), a plain function that merely has defaults (must be skipped), a parameterless composable the primary lookup already handles (fallback stays out of the way), and an absent name.
- `./gradlew :gradle-plugin:preview-discovery:test` — 199 tests, 0 failures, including a new `AppTourDiscoveryTest` case that feeds a realistic merged manifest (app activities + ui-tooling's `PreviewActivity` + `GoogleApiActivity` + both Firebase trampolines) and asserts only the app's own two survive.
- `./gradlew :renderer-android:ktfmtFormat* :gradle-plugin:preview-discovery:ktfmtFormat*` — clean.

No visual evidence: neither change alters how anything is painted. G turns a hard failure into the render the preview always described, and D removes cards that never had pixels. The proof is the JetLagged and Confetti catalogs regenerating without those failures.

https://claude.ai/code/session_01HsnMASEm93V7HBUe89PhqQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01HsnMASEm93V7HBUe89PhqQ)_

### DIFF
--- a/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/AppTourDiscovery.kt
+++ b/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/AppTourDiscovery.kt
@@ -32,11 +32,20 @@ object AppTourDiscovery {
   /**
    * Class-name prefixes of activities that manifest-merger injects from libraries (Glance's action
    * trampoline, ui-tooling's `PreviewActivity`, ui-test-manifest's `ComponentActivity`, Play
-   * Services dialogs, …). They're not part of the *app* — not previewable entry points, not tour
-   * material — so they're dropped at parse time.
+   * Services dialogs, Firebase Auth's reCAPTCHA / federated-IdP trampolines, …). They're not part
+   * of the *app* — not previewable entry points, not tour material — so they're dropped at parse
+   * time.
+   *
+   * `com.google.` subsumes the narrower `com.google.android.` this list used to carry, so that
+   * Firebase Auth's `GenericIdpActivity` and `RecaptchaActivity` — headless trampolines that NPE
+   * inside `onResume` the moment they are launched off a real device — stop reaching the catalog as
+   * broken cards, as they did in both Confetti catalogs.
+   *
+   * The list stays evidence-led rather than pre-emptive: every prefix is also a way to silently
+   * drop an app's own screen, so an SDK earns one once it has been seen producing a dead card.
    */
   private val LIBRARY_ACTIVITY_PREFIXES =
-    listOf("android.", "androidx.", "com.android.", "com.google.android.")
+    listOf("android.", "androidx.", "com.android.", "com.google.")
 
   private val xmlFactory: XMLInputFactory =
     XMLInputFactory.newFactory().apply {
@@ -267,10 +276,9 @@ object AppTourDiscovery {
   )
 
   /** Parses one tour spec file. Returns `null` (best-effort) when unreadable or not a tour. */
-  fun parseTourSpec(file: File): TourSpec? =
-    runCatching { TOUR_JSON.decodeFromString(TourSpec.serializer(), file.readText()) }
-      .getOrNull()
-      ?.takeIf { it.steps.isNotEmpty() || it.start != null }
+  fun parseTourSpec(file: File): TourSpec? = runCatching {
+    TOUR_JSON.decodeFromString(TourSpec.serializer(), file.readText())
+  }.getOrNull()?.takeIf { it.steps.isNotEmpty() || it.start != null }
 
   /**
    * Turns committed tour spec files into one synthetic [PreviewKind.APP_TOUR] preview each. Every
--- a/gradle-plugin/preview-discovery/src/test/kotlin/ee/schimke/composeai/discovery/AppTourDiscoveryTest.kt
+++ b/gradle-plugin/preview-discovery/src/test/kotlin/ee/schimke/composeai/discovery/AppTourDiscoveryTest.kt
@@ -144,4 +144,39 @@ class AppTourDiscoveryTest {
     assertTrue(previews.isEmpty())
     assertEquals(1, warnings.size)
   }
+
+  @Test
+  fun `library-injected activities are dropped`() {
+    // A merged manifest carries every library's activities alongside the app's own. None of them
+    // are app screens; Firebase Auth's two headless trampolines in particular NPE on launch and
+    // used to land in the Confetti catalogs as permanently-broken cards.
+    val merged =
+      """
+      <?xml version="1.0" encoding="utf-8"?>
+      <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="com.example.app">
+          <application>
+              <activity android:name=".MainActivity" android:exported="true">
+                  <intent-filter>
+                      <action android:name="android.intent.action.MAIN" />
+                      <category android:name="android.intent.category.LAUNCHER" />
+                  </intent-filter>
+              </activity>
+              <activity android:name="androidx.compose.ui.tooling.PreviewActivity" />
+              <activity android:name="com.google.android.gms.common.api.GoogleApiActivity" />
+              <activity android:name="com.google.firebase.auth.internal.GenericIdpActivity" />
+              <activity android:name="com.google.firebase.auth.internal.RecaptchaActivity" />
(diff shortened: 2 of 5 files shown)


yschimke:
Codex's review landed after this merged, so the follow-ups are on #4662 (commit b545a5d).

**"Preserve app-owned activities under com.google" (P2) — valid, fixed.** `com.google.` was far too wide for a prefix match: Now in Android really does ship as `com.google.samples.apps.nowinandroid`, and the short prefix would have classified every one of its screens as library-injected, leaving its catalog with no activities at all. Narrowed to the two library namespaces the evidence names — `com.google.android.` (Play Services, restored) and `com.google.firebase.` (the Auth trampolines that reached both Confetti catalogs). A new `AppTourDiscoveryTest` case feeds a manifest from an app owning the vendor namespace and asserts it keeps its own activity while both SDK trampolines are still dropped.

**"Resolve the preview overload instead of the shortest sibling" (P2) — valid, fixed differently.** Picking by arity could render a real component, plausibly, and the wrong one. Rather than guess, ambiguity is now refused and diagnosed with the competing signatures — a catalog quietly showing the wrong card is worse than one showing an error.

The suggested fix (preserve the discovered JVM descriptor) is the right long-term shape but needs a manifest wire-format change, and reading the annotation instead is not an option: `@Preview` is `AnnotationRetention.BINARY`, so it lands in `RuntimeInvisibleAnnotations` where reflection cannot see it. Held off because the shape is nearly unusable in Kotlin source — with both overloads fully defaulted, every call site is ambiguous — so a diagnosed refusal buys the correctness without the wire change. Worth revisiting if it is ever hit in practice.

**"Re-author the commit with a human identity" (P1) — does not hold.** The reviewed commit d948e84 is authored *and* committed by `Yuri Schimke <yuri@schimke.ee>`:

```
$ git log -1 --format='author=%an <%ae> committer=%cn <%ce>' d948e84
author=Yuri Schimke <yuri@schimke.ee> committer=Yuri Schimke <yuri@schimke.ee>
```

The `No Agent Attribution` gate passed on this PR before it merged. No `codex@openai.com` identity appears anywhere in this branch's history.

---

Issue #4624, opened 2026-08-28
Can we sort figma diffs and use default diffs from the delivery branches


yschimke:
### What went wrong

Ideally the page is sorted correctly, and uses presomputed results. But the diffs can be updated after that.

Also show a column with known bugs, and a link for opening a bug?

### Server

| | |
| --- | --- |
| compose-preview | `1.40.0` |
| Mode | public (open) |
| Uptime | 1h 24m |
| Server JVM | `21.0.12 (Eclipse Adoptium)` |
| Server OS | `Linux 7.0.0-28-generic (amd64)` |

### Page

| | |
| --- | --- |
| Page | [`/m3-catalog/compare?format=reference&theme=light`](https://preview.coo.ee/m3-catalog/compare?format=reference&theme=light) |
| Design system | `m3-catalog` |
| Catalog | `yschimke/m3-catalog@design-artifacts/m3-catalog` |
| Catalog rendered by | compose-ai-tools 1.36.0 |
| Trust | branch:yschimke/m3-catalog@design-artifacts/m3-catalog |
| Render lane | live daemon |

### Browser

| | |
| --- | --- |
| User agent | `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36` |
| Viewport | 2560×1330 CSS px |
| Device pixel ratio | 1 |
| Page colour scheme | light |
| OS colour scheme | light |

### Recent failures

```
2026-08-28 07:54 UTC  meshcore-mobile: render failed after 5442ms — render rejected: kind=APP_TOUR is not renderable by the daemon yet; the Gradle composePreviewRender pipeline produces its captures
2026-08-28 07:54 UTC  meshcore-mobile: render failed after 5361ms — render rejected: kind=APP_TOUR is not renderable by the daemon yet; the Gradle composePreviewRender pipeline produces its captures
2026-08-28 07:54 UTC  meshcore-mobile: render failed after 0ms — render rejected: kind=APP_TOUR is not renderable by the daemon yet; the Gradle composePreviewRender pipeline produces its captures
2026-08-28 07:54 UTC  meshcore-mobile: render failed after 1ms — render rejected: kind=APP_TOUR is not renderable by the daemon yet; the Gradle composePreviewRender pipeline produces its captures
```

PR #4660, opened 2026-08-28
fix(serve): dress the compare wall's rows only when they are shown

yschimke:
Closes the unaddressed review finding left on #4643 after it merged.

## The cost

`dressRow` assigns both of a row's image `src` values, and the run pass called it for **every** catalog row before `applySearch` hid the nonmatches:

```ts
const dressed = new Set(this.rows.filter((row) => this.dressRow(row)));
this.applySearch();
for (const row of this.rows) if (!dressed.has(row)) row.hidden = true;
```

So opening the wall through a narrowed `?q=` or `?preview=` link had the browser fetch and decode the whole wall — on the large catalogs this page exists for, hundreds of full-resolution pairs for a single visible comparison. The bandwidth is one thing; the decode is the part that stalls a laptop, which is the same reason #4624 moved the scoring chain to one row at a time.

## The fix, and the trap in it

Dressing now happens inside `applySearch`, on the rows it is about to leave visible.

The naive version of that is wrong, and the wrong version is invisible in a first-load test: a row hidden on the first pass was never dressed, so a later **widening** — the search box clearing, a Back to a broader query, the `applySearch` after the scoring chain — would reveal a row with no pictures in it. `ensureDressed` handles both directions, and the invariant is simply **a visible row is a dressed row**.

The dressed set resets per `run()`, because a format or theme switch changes which pair each row shows; a row dressed for the previous lane has to be dressed again for this one. A dressing *failure* is deliberately not remembered — it costs one repeated `querySelector` sweep on a row that will not be shown either way, and remembering it would mean a second set whose only purpose is to skip work nobody waits on.

## One behaviour change, and it is a correction

The count now equals the number of rows actually on screen. It was computed from the filter alone and then rows that failed to pair were hidden *underneath* it, so a wall containing an unpairable row over-reported by one. The reorder forces the question either way, and matching what the reader can see is the honest answer. The existing `narrows on the search box, and says how many are left` and `says so when the filter leaves nothing` tests both still pass unchanged.

## Test plan

Two tests added to `compareWallElement.test.ts`, **both verified to fail against the pre-fix component** (restored from `origin/main`, suite re-run: `2 failing`):

- **`does not load the pictures of a row the filter hides`** — narrows to one row, blanks what the previous pass drew, then switches theme to force a full re-run. The visible row gets a `src`; the filtered-out row gets none.
- **`dresses a row a widened filter reveals, rather than showing it empty`** — the same setup, then clears the search and asserts the revealed row carries its picture and the count follows.

Both use a fixture carrying light *and* dark variants, so the theme switch is a real re-run rather than an empty wall — the first draft used the default single-theme rows and passed for the wrong reason (`0 comparisons`), which is why the fixture is explicit about it.

`cd cli/serve-web && npm run verify` — typecheck, **1164 tests passing**, build, and the committed-bundle check all green. `serve-components.js` is rebuilt and committed, as that check requires. `npm run format:check` clean.

## Not in this diff

#4644's two findings on the same page (anchor slug collisions, and the fixed 44px scroll-margin against a wrapping sticky header) are a different surface — `scripts/design-artifacts/render-cross-system-html.mjs`, a generator whose `c-<slug>` scheme is shared with `render-index-html.mjs` and cross-links between the two pages. Kept separate rather than mixed into a serve-web component change.

---


yschimke:
Both open items on this PR are pushed.

## The review finding is right, and it was a real hole

Dressing a revealed row gives it its pictures, but nothing gave it a **number**. The row was never in the run's `visible` snapshot, so no `scoreRow` was ever scheduled for it — an SVG row sat at the server's `waiting…` for the rest of the run, a reference row was never remeasured, and neither took part in the score-based reordering.

One correction to the framing: the gap **predates dressing lazily**. A row the initial filter hid was left out of `visible` before this PR too, so it went unmeasured then as well. What this PR changed is that reveal now has one place to be handled from — which is why the fix belongs beside `ensureDressed` rather than nowhere.

The run's chain is kept on the instance and a revealed row is appended to the **back** of it, not raced beside it: each row decodes two full frames, and a widening that reveals twenty at once is exactly the stampede the serial chain exists to avoid. `scoredRows` keeps that to once per row per run rather than once per keystroke.

`applySearch` gained a `schedule` flag, because `run` calls it *before* the sorted loop that does its own queueing. Without the flag that pass read the previous run's set and chain and queued every visible row twice — and the order it would impose is DOM order, not the baked-score order the sort exists to produce. I found that by two existing tests going red (`bounds the map it keeps to what the column can draw`, `fixes the RC player's theme and fonts BEFORE it measures`), not by reading the diff.

Two tests added:

- **`measures a row a widened filter reveals, not just its pictures`** — verified to fail against `8cc115cc` (the reviewed head): the revealed row is never handed to the scorer.
- **`schedules a revealed row once, not on every keystroke`** — a guard on the new queueing rather than a regression test; it passes on both heads, and is there so a future change cannot turn a filter into a measurement storm.

## The CI failure was not this PR's

`Module Unit Tests` on `8cc115cc` was `ServeWebTest > the spec lane offers the compareWith sibling as a second source`, which asserted `under that catalog's theme and knobs` unescaped while the page serves `catalog&#39;s`. Nothing to do with this diff — this branch was cut before the fix for it landed on `main`. Merged `origin/main`, and that test passes on the new head.

Verified on the pushed head: `npm run verify` in `cli/serve-web` (typecheck, **1166 tests**, build, committed-bundle check) and `./gradlew :cli:serve:test ktfmtCheckAll` both green. `serve-components.js` is rebuilt and committed.

---

PR #4662, opened 2026-08-28
feat(renderer): render activity tours against the manifest Application

yschimke:
Follow-up to #4646 (merged), which fixed 15 of the 71 failed renders on [preview.coo.ee/status](https://preview.coo.ee/status). This takes the largest remaining cluster: **39 activity tours blocked by app-level DI** — 36 Hilt, 2 Koin, 1 `AppComponentFactory` — plus the review findings that followed (last section).

## The failure

Not one activity tour rendered in any app that uses app-level DI. Across the fleet that was 39 of 45 activity-tour renders, all failing before a pixel was drawn:

```
java.lang.IllegalStateException: Hilt Activity must be attached to an @HiltAndroidApp
Application. Did you forget to specify your Application's class name in your manifest's
<application />'s android:name attribute?
	at dagger.hilt.android.internal.managers.ActivityComponentManager.createComponent(…:86)
	at io.homeassistant.companion.android.matter.Hilt_MatterCommissioningActivity.inject(…)
	at ee.schimke.composeai.renderer.AppTourRenderer$TourSession.launchIntent(AppTourRenderer.kt:194)
```

and the Koin (`KoinApplication has not been started`) and Metro (`Couldn't call constructor`) variants of the same thing. Only meshcore-mobile's single tour rendered anywhere — the one app in the fleet whose activities need nothing from its Application.

## Why one flag couldn't fix it

The generated `robolectric.properties` pins `application=android.app.Application` for the whole render run. That is **right for an isolated composable** — it has no business running the consumer's `Application.onCreate()`, and DI graphs, `BridgingManager`, Firebase and WorkManager routinely fail inside the sandbox — and **wrong for an Activity, which *is* the app**.

`composePreview.useConsumerApplication = true` already restores the manifest Application, but per *module*. Flipping it on `home-assistant-android:app` to rescue 18 activity tours would also push that module's 26 working composable previews through Home Assistant's real `onCreate()` — the exact platform-init minefield the stub default exists to dodge. One knob, two incompatible needs.

Robolectric settles the Application **per test class**, from the `robolectric.properties` files merged down that class's package hierarchy — never per test method. So two previews in one manifest cannot have different Applications unless two classes render them.

## The split

A second render lane: `ee.schimke.composeai.apptour.AppTourRobolectricRenderTest`, with a generated properties file of its own that omits `application=` and lets Robolectric fall back to the merged manifest.

The package is a **sibling** of `…renderer`, not a child, and that is load-bearing: Robolectric merges parent-package properties into a child's, so a nested package would inherit the `application=android.app.Application` line and have to override it. With a sibling, omitting the line means exactly what it says.

`PreviewManifestLoader.Lane` splits the manifest so the two classes' selections are **disjoint and jointly complete**. The two-argument `loadShard(N, M)` still means the composable lane, so the `RobolectricRenderTest_ShardN` subclasses the plugin generates into consumer builds are unaffected. `allEntries` stays the full manifest in both lanes, so the stale-fan-out sweep still expects the other lane's outputs and neither class deletes the other's PNGs.

**Both entry points run both lanes** — the Gradle `composePreviewRender` task by include pattern, and `AndroidRendererMainKt` (what `BundleRenderer.spawnAndroidRenderer` launches for a detached bundle render) by running both classes. Those are the only two; the daemon's `RenderEngine` has never rendered app-level kinds.

Everything except the Application is copied rather than shared — `sdk`, `graphicsMode`, `shadows`, `instrumentedPackages` — so a module can't render its activities at a different API level, or without the font/coil/paused-clock shadows, than its composables. Tests pin that on both the Gradle and bundle paths.

**The failure mode is now the consumer's `Application.onCreate()` failing under Robolectric, and it is contained** to this class rather than taking the module's composable previews with it. `composePreview.appTourUseConsumerApplication = false` puts activities back on the stub for a module whose app cannot survive the sandbox — the escape hatch matters because I can't validate 39 third-party apps from here, and some of their Applications may simply fail differently.

**Never sharded.** A module has a handful of app-level previews, and the cost that matters is the Application init paid once per sandbox — sharding would multiply that, not divide it. `ShardTuning.perPreviewRowCosts` therefore stops counting ACTIVITY/APP_TOUR rows: sizing the shards for rows they will never be given would spin up idle forks on an activity-heavy module (Home Assistant declares 18 activities against 26 composable previews).

## Test plan

**End-to-end, on `:samples:android`** — `composePreviewRenderAll` green, and the split does what it says:

| class | tests | rendered |
|---|---|---|
| `apptour.AppTourRobolectricRenderTest` | 3 | `activity__MainActivity`, `activity__NowPlayingActivity`, the 4-step `apptour__getting-started` |
| `renderer.RobolectricRenderTest` | 169 | everything else, **no** activity or tour |

Nothing rendered twice, nothing dropped. The generated `…/apptour/robolectric.properties` came out with no `application=` line and the same `sdk=35` / `graphicsMode=NATIVE` / shadows as the composable lane's.

**Per-lane Application proven directly.** The claim that Robolectric honours a sibling-package properties file is the load-bearing assumption here, so I checked it rather than assuming: a temporary `ProbeApp : Application` declared in the sample's manifest wrote a marker file from `onCreate` (reverted before commit). Running **only** the app-tour class produced the marker; running **only** a composable shard did not. Both directions, as intended — the tour lane gets the manifest Application, the composable lane keeps the stub.

**Unit tests** — 1008 across `:renderer-android`, `:gradle-plugin`, `:gradle-plugin:preview-discovery` and `:bundle-format`, 0 failures:

- `PreviewManifestLoaderLaneTest` — the two lanes partition the manifest (disjoint + complete over every kind the Android renderer handles), and the two-argument `loadShard` still defaults to the composable lane so generated shard subclasses are unaffected.
- `GenerateRobolectricPropertiesTaskTest` ×4 — the app-tour file's Application line under each flag combination (including that module-wide `useConsumerApplication = true` is not undone by `appTourUseConsumerApplication = false`), and that it pins the same sdk/graphics/shadows as the composable lane.
- `AndroidBundleLaunchTest` ×2 — the bundle path writes the app-tour package's file too, and its body omits `application=` while otherwise matching the composable lane's line for line.
- `ShardTuningTest` — app-level rows excluded from shard sizing, anchored on the two kind names so an entry's nested `dataProducts` (which carry a `kind` of their own, like `render/scroll/gif`) can't be mistaken for one.

`ktfmtCheck` clean.

No before/after images: this changes *which Application an Activity is constructed under*, not how anything is painted. For the sample — whose manifest declares no custom Application — the tour PNGs are unchanged. The real evidence is the catalogs: Home Assistant, Pocket Casts Wear, Confetti, JetCaster Wear and Cadence regenerating with tours that render instead of a stack trace.

## Review findings folded in

Codex reviewed #4646 after it had merged, so its findings there have no other home — and this branch is the one this task develops on. Each is small and pinned by a new test.

**The `com.google.` activity prefix was too wide** (b545a5d). The exclusion list matches by prefix, and an app is free to own that namespace — Now in Android ships as `com.google.samples.apps.nowinandroid`. The short prefix would have classified every one of its screens as library-injected and left its catalog with no activities at all. Narrowed to the two library namespaces the evidence actually names: `com.google.android.` (Play Services, restored) and `com.google.firebase.` (the Auth trampolines that reached both Confetti catalogs). The new test feeds a manifest from an app that owns the vendor namespace and asserts it keeps its own activity while both SDK trampolines are still dropped.

**`findDefaultedComposableMethod` picked the shortest sibling** (b545a5d). With several fully-defaulted composable overloads of one name, it took the fewest-parameter one — and a manifest entry carries the function *name* only, so nothing at render time can tell which carried the `@Preview`. Picking by arity renders a real component, plausibly, and the wrong one; a catalog quietly showing the wrong card is worse than one showing an error. Ambiguity is now refused and diagnosed with the competing signatures.

Reading the annotation would settle it and cannot: `@Preview` is `AnnotationRetention.BINARY`, so it lands in `RuntimeInvisibleAnnotations` where reflection cannot see it. Settling it properly means threading the JVM descriptor from discovery through the manifest — worth doing if this is ever hit, but the shape is nearly unusable in Kotlin source (with both overloads fully defaulted, every call site is ambiguous), so it stays a diagnosed refusal rather than a wire-format change.

**The standalone entry point ran only one lane** (072bbca) — a real regression in the first commit here, and the one worth reading twice. `AndroidRendererMainKt` ran `JUnitCore.runClasses(RobolectricRenderTest::class.java)` and nothing else, so with the lane filter in place a detached bundle render would have rendered the manifest *minus* its app-level previews, **exited 0**, and left reconciliation reporting every activity and tour as failed with no diagnostic. Silent, and worse than the failure this PR exists to fix. Both classes now run there, and `AndroidBundleLaunch.writeRobolectricConfig` writes the sibling-package properties file to match — without it that lane would fall back to whatever the shipped jar carries rather than the synthesized config the caller prepends.

**Not actioned: the repeated attribution P1.** Three iterations named three different commits (`d948e84`, `493d44f`, `2be14a8`); the last two are not objects in this repository at all, and `d948e84` merged with the gate green. Every commit on this branch is authored and committed by `Yuri Schimke <yuri@schimke.ee>`, the repo's own detector exits 0 over the range, and the `Reject agent attribution` check has now **passed on this head**. Evidence in [this comment](https://github.com/yschimke/compose-ai-tools/pull/4662#issuecomment-5451077201).

## Remaining after this

Of the original 71: 15 fixed in #4646, up to 39 addressed here, 5 in yschimke/pocket-casts-android#28 (`hostTheme` on a library module). The rest — 4 `hiltViewModel()` previews, 2 Glance composables annotated with the Compose `@Preview`, 2 that never reach idle, 2 on a material3 ABI skew, 1 needing Intent extras, 1 tour that drew nothing — are upstream preview defects that fail in Studio too, or single cases worth their own look.

https://claude.ai/code/session_01HsnMASEm93V7HBUe89PhqQ


yschimke:
On the Codex P1 ("Recreate the commit with a human identity", citing `493d44fd3a13b5e749da97927efdc7c5587c01ff`) — **this does not hold, and the cited commit is not in this repository.**

```
$ git cat-file -t 493d44fd3a13b5e749da97927efdc7c5587c01ff
fatal: git cat-file: could not get object info

$ git log --format='%H %an <%ae> | committer %cn <%ce>' origin/main..origin/agent/failed-renders-triage-3d0q6l
b545a5daeba7e1e50137005760808e549ed1e33d Yuri Schimke <yuri@schimke.ee> | committer Yuri Schimke <yuri@schimke.ee>
f3dc1d5a07112f92f32a1a972ea84b32e7f216af Yuri Schimke <yuri@schimke.ee> | committer Yuri Schimke <yuri@schimke.ee>
```

Both commits on this branch are authored *and* committed by the human identity. Run against the repo's own detector — the same script the `commit-msg` hook, the `pre-push` hook and the CI gate share:

```
$ bash .github/scripts/agent-attribution-scan.sh --range origin/main..origin/agent/failed-renders-triage-3d0q6l
$ echo $?
0
```

This is the second time the same finding has come back on a commit that isn't the one under review — it made the identical claim about `d948e84` on #4646, which merged with the `No Agent Attribution` gate green. It reads like Codex reporting on a commit in its own sandbox rather than the reviewed branch. Flagging rather than acting on it; no change made.

The review's other two findings on #4646 were valid and are fixed in b545a5d on this branch — details in the PR description and in [this comment](https://github.com/yschimke/compose-ai-tools/pull/4646#issuecomment-5451064934).

---

yschimke:
**"Run the app-tour lane from the standalone renderer" (P2) — correct, and a good catch. Fixed in 072bbca.**

`AndroidRendererMainKt` ran `JUnitCore.runClasses(RobolectricRenderTest::class.java)` and nothing else. With the lane filter in place that class drops every `ACTIVITY` / `APP_TOUR` entry, and nothing ran `AppTourRobolectricRenderTest` — so a detached bundle render (`BundleRenderer.spawnAndroidRenderer`) would have rendered the manifest *minus* its app-level previews, exited 0, and left reconciliation reporting every activity and tour as failed with no diagnostic. Silent, and worse than the failure the split exists to fix.

Both classes now run there. The Gradle `composePreviewRender` task already included both; these are the only two entry points — the daemon's `RenderEngine` has never rendered app-level kinds (`docs/APP_TOURS.md` lists it as a standing limitation), and every other mention of `RobolectricRenderTest` in the tree is KDoc.

The review's second half was right too: `AndroidBundleLaunch.writeRobolectricConfig` now writes the sibling-package file at `ee/schimke/composeai/apptour/` alongside the renderer one. Robolectric resolves each class's Application from its own package's file, so without it the app-tour lane would fall back to whatever the shipped jar carries rather than the synthesized config the caller prepends. Two new `AndroidBundleLaunchTest` cases pin that the file is written and that its body omits `application=` while otherwise matching the composable lane's line for line.

1008 tests green across `:renderer-android`, `:gradle-plugin`, `:gradle-plugin:preview-discovery` and `:bundle-format`; ktfmtCheck clean.

---

On the P1 attribution finding, now citing `2be14a8980cafc974697e2f820b54979fa77eeb3`: that commit is not in this repository either.

```
$ git cat-file -t 2be14a8980cafc974697e2f820b54979fa77eeb3
fatal: git cat-file: could not get object info
```

This is the third iteration of the same finding against a third SHA that isn't the reviewed commit (`d948e84` on #4646, then `493d44f`, now `2be14a8`). The branch's actual commits are all authored and committed by `Yuri Schimke <yuri@schimke.ee>`, and the repo's own detector exits 0 over the range — evidence in [my earlier comment](https://github.com/yschimke/compose-ai-tools/pull/4662#issuecomment-5451077201). Not acting on it; flagging that this particular check appears to be reporting on Codex's own sandbox rather than the branch.

---

yschimke:
**"Put the packed manifest on the standalone renderer classpath" (P2) — the facts are right. Scoped rather than fixed, in 0ff0b4b.**

Verified: `BundleRenderer.renderAndroid` builds its classpath from `configRoot + classesDir + libJars + resolvedJars + rendererJars + androidJar` and never extracts `android/AndroidManifest.xml` or calls `AndroidBundleResources.writeTestConfig`. Those are reachable only from `BundleDaemonCommand` and `ServeBundleDaemon` — the daemon lane. So on the one-shot bundle path Robolectric has no packed merged manifest and falls back to its own default, whose `<application>` names no class. A bundle whose app declares `android:name` does not get that Application there.

My KDoc claimed otherwise ("leaves it to the merged manifest"), which is true on the Gradle path and wrong here. Corrected: the comment now states plainly that this resolves to the platform default today, and names the follow-up — pack the manifest for this lane, and keep the Application class in `BundlePreviewTask`'s minimized `app.jar`, which is seeded from preview class names rather than from the manifest.

**Not implementing it here.** Android bundle resource packing is Phase 2 by design — `AndroidBundleLaunch`'s own header scopes it that way, and it exists for the daemon lane only. Building it would mean changing what `BundlePreviewTask` packs, extraction and `writeTestConfig` in `renderAndroid`, and the `app.jar` minimization seed: a feature, and one this PR would be a poor place to review it in.

Worth being precise about what this PR does and doesn't change for that lane, since it's easy to read as a regression and isn't one:

- **Before:** the bundle path ran only `RobolectricRenderTest`, and activity tours rendered under the stub Application — so they failed, exactly as everywhere else.
- **After 072bbca:** both lanes run, so tours are attempted rather than silently dropped, and the properties file is in place.
- **Still open (this finding):** with no packed manifest, that lane's Application is the platform default, so a DI-using app's tours still won't render *via `bundle render`*. The Gradle path — where AGP supplies the merged manifest, and which is what produces the catalogs on preview.coo.ee — is unaffected and does get the consumer Application (proven with the `ProbeApp` marker in the PR description).

I left the `application=` line absent rather than pinning the stub on purpose: pinning would make the gap permanent, whereas once the manifest is packed this lane starts honouring it with no further change.

---

The P1 attribution finding has now cited four commits (`d948e84`, `493d44f`, `2be14a8`, `d2eb4c2`); the last three are not objects in this repository. The `Reject agent attribution` check has since **passed on 072bbca** — the gate this finding says will reject the change is green on the reviewed head. Not commenting on it further.

---

PR #4665, opened 2026-08-28
fix: close three review findings left on merged PRs

yschimke:
Three unaddressed review findings, from #4617, #4631 and #4615. Independent, all small, and each one is a guard that was not guarding.

## #4617 — the contract probe named an artifact it never used

`:common-image-crop` was added to `CONTRACT_PROJECTS`, but `ContractSurface` referenced nothing from it. Adding the artifact to the probe's `contracts` only proves that a JAR with that coordinate resolves — so removing or renaming the exact crop surface serve calls would have left `checkContractSurface` **green**, which is the one failure that file exists to make impossible.

`ContentCrop` alone. It is serve's whole import from the module (`ServeWeb` and `ServeHttpServer` name it and nothing else); the sibling `ContentBox` is published but unused here, and naming it would have the probe assert a dependency floor *higher* than the real one — against this file's own stated rule that every reference is drawn from serve's real import set.

> My first draft did include `ContentBox`, justified with "`ContentCrop.box` returns it". It doesn't — they are unrelated types, and serve imports neither the type nor any accessor returning it. Caught by checking the `.api` dump rather than trusting the sentence.

## #4631 — a wholly deferred component lost its declared parallel

`applyParallels` walks `manifest.components`. An entry marked `priority: "deferred"` short-circuits into `deferred[]` *before* it ever gets there, so its `parallel` was dropped — on a catalog that publishes `compareWith` and whose whole point is the cross-system comparison. A server reconstructing the deferred card could not resolve its sibling.

The deferred records are attached **after** `applyParallels` runs (`generate-design-catalog.mjs` builds them further down), so they cannot be reached from inside it. `parallelIndex` is now exported and both stampers read it, rather than growing a second hand-rolled reader that would eventually disagree about the blank-means-none rule — the only subtle part of reading a spec entry. The deferred stamp also refuses to clobber a record that already carries one, matching `applyParallels`.

## #4615 — a regression test that could not see the code it guarded

`--id shatters a comma-bearing id into its fragments` rebuilt `args.drop(2).flatMap { it.split(',') }` inside the test, so it kept passing if the real `--id` parsing stopped splitting, changed its trimming, or lost the `--id-file` precedence entirely.

`PackPreviewIdExclusions.selectedIds` extracts the selection `bundle pack` actually makes (`BundleCommand` now calls it), and the test goes through it. Three more cover what the hand-rolled version structurally could not reach:

- the file keeps a comma-bearing id **whole** — the regression `--id-file` exists for
- the file **outranks** `--id` rather than merging with it
- repeated `--id` flags accumulate, and blank values are dropped

## Test plan

- `./gradlew :cli:test :cli:serve:test` — BUILD SUCCESSFUL. `PackPreviewIdExclusionsTest`: **30 tests, 0 failures.**
- `node --test scripts/design-artifacts/apply-parallels.test.mjs` — **13 tests, 0 failures** (9 existing untouched, 4 new on `parallelIndex`, one of which asserts `applyParallels` stamps exactly what the index holds — the property that makes exporting it worth anything).
- `node --check scripts/design-artifacts/generate-design-catalog.mjs` — parses.
- **`scripts/check-preview-server-contracts.sh`** — the 16 contract modules publish, and `preview-server` then compiles against the *published* artifacts with the new import resolving. The script's second half needed `--no-daemon` in this sandbox (the separate build wants a JDK named by `org.gradle.java.home` and the running daemon's JVM did not match) — an environment mismatch, not a build one; the compile itself is clean.
- `./gradlew :cli:checkServeSeam` — `OK — 28 crossing symbol(s), all on the allowlist`. `./gradlew ktfmtCheckAll` — clean. `python3 .github/ci/test_path_scope.py` — OK.

Seven `scripts/design-artifacts/*.test.mjs` suites fail in this sandbox on `main` as well as on this branch (missing npm workspace deps — `@design-parity/candidate`); `apply-parallels` is not among them and runs clean.

Non-visual: a probe reference, a manifest field, and a test seam.

---

### DIFF
--- a/cli/serve/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogStore.kt
+++ b/cli/serve/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeCatalogStore.kt
@@ -934,17 +934,30 @@ class ServeCatalogStore(
           // straight off `catalog.json`; a catalog that declares neither gets null + empty and the
           // viewer's second comparison source is simply never offered.
           compareWithSystem = catalog.compareWith?.system?.takeIf { it.isNotBlank() },
+          // Read from BOTH lists, components last so they win — the same shape as
+          // [captionByComponentId] and for the same reason. A wholly deferred component never
+          // reaches `components[]`, and the export writes its pairing onto the deferred record
+          // instead; reading only `components` discarded it and left the deferred card unable to
+          // offer the sibling source. That also covers its deferred *variant* records, which
+          // inherit the entry's pairing.
           parallelByComponentId =
-            catalog.components
-              .mapNotNull { component ->
-                val id = component.componentId?.takeIf { it.isNotBlank() } ?: return@mapNotNull null
-                val parallel =
-                  component.parallel?.takeIf { it.isNotBlank() } ?: return@mapNotNull null
-                id to parallel
+            buildMap {
+              catalog.deferred.forEach { deferred ->
+                val id = deferred.componentId?.takeIf { it.isNotBlank() } ?: return@forEach
+                val parallel = deferred.parallel?.takeIf { it.isNotBlank() } ?: return@forEach
+                // First wins among the deferred records themselves, like [previewAliasFor]: a
+                // catalog that somehow listed a component id twice gets the producer's own order
+                // rather than a throw.
+                putIfAbsent(id, parallel)
               }
-              // First wins, like [previewAliasFor]: a catalog that somehow published a component id
-              // twice gets the producer's own order rather than a throw.
-              .toMap(),
+              catalog.components.forEach { component ->
+                val id = component.componentId?.takeIf { it.isNotBlank() } ?: return@forEach
+                val parallel = component.parallel?.takeIf { it.isNotBlank() } ?: return@forEach
+                // `put`, not `putIfAbsent`: a component that IS in `components[]` states its own
+                // pairing, and a deferred *variant* record sharing its id only inherits one.
+                put(id, parallel)
+              }
+            },
           degradations = degradations,
           liveOnly = liveOnly,
           // Kept in step with [scheduleRcCompareFetch]'s own guard through the shared helper: a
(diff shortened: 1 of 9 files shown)


yschimke:
The deferred-parallels finding is right, and it is the more important half of the two. Fixed in `4d2c3a7b98`.

I shipped the generator half without the server half, so the `parallel` this PR started writing onto a deferred record went into a field nothing read. Both legs of the finding check out against the code:

- `ServeCatalogStore.Deferred` had no `parallel` property, and the decoder ignores unknown keys — so the value was dropped at parse.
- `parallelByComponentId` was built from `catalog.components` alone, which a wholly deferred component never reaches by construction.

Net effect: the deferred card stayed the one card that cannot offer the sibling comparison source, on a catalog that publishes `compareWith` precisely to have it — the exact condition the generator change was meant to remove. A fix that writes a field nobody reads is not a fix.

The model gains the field, and the lookup now reads both lists in the same shape `captionByComponentId` already uses for the identical problem (deferred first, then components). Precedence only decides the **mirror** case, since a wholly deferred component is never in both: a deferred *variant* record sharing an id with a component that IS published. There the component states its own pairing and the variant merely inherits one, so the component wins — `put`, not `putIfAbsent`.

> My first pass wrote `putIfAbsent` for both while the comment beside it said components win. Caught re-reading my own diff; the two now agree, and the second test below pins which way round it goes.

Two tests in `ServeCatalogStoreTest`, the first **verified to fail against current `main`'s store** (`a wholly deferred component's pairing reaches the server's parallel lookup` → FAILED):

- a wholly deferred component's pairing reaches `parallelByComponentId`, alongside a published component's
- an authored component pairing outranks one inherited by its deferred variant record

Green on the pushed head: `./gradlew :cli:test :cli:serve:test :cli:checkServeSeam ktfmtCheckAll` (`ServeCatalogStoreTest`: 94 tests, 0 failures; seam `OK — 28 crossing symbol(s)`), and `node --test scripts/design-artifacts/apply-parallels.test.mjs` — 13 pass.

---

PR #4714, opened 2026-08-28
fix(serve): give the page, its verdict and the rendered frame one cache generation

### LINKED ISSUE #4695: serve: give the page, its verdict and the rendered frame one cache generation
Raised by the review bot on #4672 and deliberately deferred there; filing so it is tracked rather than living in a merged PR body.

## The gap

A comparison page's HTML is served `STATIC_PAGE_CACHE_CONTROL` (`max-age=60, stale-while-revalidate=300`); the render it points at is served `STATIC_RESOURCE_CACHE_CONTROL` (`max-age=300, stale-while-revalidate=3600`). The two lifetimes are independent, so after a catalog refresh a browser or intermediary can pair one generation's HTML with another generation's pixels.

That has always been true of the annotation redline, whose boxes are bounds in the render's pixel space. It matters more now that the same page carries a **parity verdict**: a redline that is one generation stale is a reading aid slightly out of date, whereas "padding is 24 where the spec says 16" over the wrong pixels is a false claim with a highlight pointing at it.

## Already named in the code

`ServeHttpServer`'s `frameIsReplayedBaked` comment states it:

> Necessary but NOT sufficient, and the gap is caching: an override-free baked `/render/<id>.png` is served `STATIC_RESOURCE_CACHE_CONTROL` while this index is `no-store`, so a client can pair previous-generation pixels with a freshly-fetched index. Closing it needs the image and the index to carry a shared generation — batch 05's coupling, which also moves the render lane's URL and caching contract.

## Why it is its own change

Every surface that builds a render URL moves with it: the viewer, the compare page, the compare wall, the acceptance engine's `candidateUrl` and `referenceUrl`, the tag-index gate, and the issue reporter's embedded `renderUrl`. A partial fix is worse than none — generation-scoping only the `` on the compare page would put the acceptance engine and the displayed pixels on different URLs, which is a fresh inconsistency in exchange for closing an old one.

## Mitigations already in place

The verdict is withheld on a pinned revision and under any render override, so the window is narrowed to an unpinned, override-free browse across a catalog refresh. It is not closed.

## Acceptance

- The page, the verdict payload and the rendered frame share one cache identity, so a client cannot hold a mismatched pair.
- The published render URL contract changes in one place, with every builder of that URL moved with it.
- Acceptance scoping keeps referring to the same bytes the page displays.

yschimke:
Closes #4695.

## The gap

A comparison page's HTML was served `public, max-age=60, stale-while-revalidate=300` and the render it points at `public, max-age=300, stale-while-revalidate=3600` — two independent lifetimes over content that is regenerated **in place**, so after a catalog refresh a browser or a CDN could hold one generation's HTML beside another generation's pixels.

That was tolerable while the only thing riding on the pairing was the annotation redline (boxes a publish out of date are a reading aid a publish out of date). It stopped being tolerable once the same page carries a parity **verdict** — "padding is 24 where the spec says 16" drawn over the wrong frame is a false claim with a highlight pointing at it — and it was never tolerable for a **tag selection**, which persists the index's bounds as an acceptance baseline, so bounds measured on another frame survive into a record that later reports an unchanged element as moved. `ServeHttpServer`'s `frameIsReplayedBaked` comment named this gap and deferred it.

## The coupling

Every published frame URL a catalog page writes now names the publish that page was assembled from — `…/render/<id>.png?gen=<sha>` — so the pairing is decided inside the asset lane rather than by a race between two `max-age`s:

- **the generation on disk** → the ordinary browse, and the URL is now content-addressed (a republish moves the page's generation and therefore the URL), so the response takes the `immutable` lifetime `/hero/` and `?thumb=` already take instead of a short public one;
- **any other generation** → the page asking is a publish behind, and the frame it needs is that publish's. The delivery branch still has it, so this resolves through the **existing pinned lane** — the same fetch, the same bounds, the same rules about what a sha may name. A generation the branch cannot answer is a `404`, never today's pixels.

`gen=` and `at=` are deliberately two parameters over one mechanism because they claim different things. `at=` is a *request*: a person asked for an older publish, the page announces the pin, and the redline and verdict are withheld because they describe today's render. `gen=` is a *coherence note* the server made to itself, invisible to the reader, withholding nothing — and written only onto asset URLs, never onto a `/p/` or `/compare/` link somebody might copy. A pinned page writes the pin and no generation; a second sha beside it is only something to disagree with.

Where a pin **refuses** a URL that also asks for a product made to order (`&fontScale=1.5`, `.svg` / `.slots` / `.a11y` / `.annotations` / `.rc`), a generation **steps aside** for one. Those responses are `no-store` and reflect no published bytes, so there is no pair for a cache to hold wrongly — and refusing would cost the one interaction this coupling must not: turning a knob on a page a refresh overtook. A `gen=` that is not a sha is a `400`, like `?at=main`.

## Everything that builds a frame URL moved with it

A partial fix would be worse than none, so the whole set moved in one change: the comparison page (both panels), the comparison wall, the parity dashboard's scored `candidateUrl`/`referenceUrl` pair, the viewer's spec raster and its unfurl `og:image`, the issue reporter's embedded `renderUrl` on both the viewer and the comparison, and the tag-index gate — whose comment now records that the condition is sufficient rather than merely necessary. The viewer builds its own frame URL in the browser, so the generation reaches it as `data-generation` and `viewer.ts` applies the same rule the server does.

Sessions with no delivery branch — an uploaded bundle, a local project, a daemon-backed module — write no generation and are byte-identical to before.

## Test plan

- New `ServeCacheGenerationTest` over the delivery-branch stub whose tip and previous publish serve **different** bytes for one preview id: an unpinned comparison scopes its frames without pinning the page or leaking `gen=` into a page link; a stale `gen=` returns the *older* render and the *older* reference while the unscoped URL returns today's; an unanswerable generation is a `404` and not today's pixels; a scoped frame is `immutable` and an unscoped one is not; a ref-shaped `gen=` is a `400`; a stale `gen=` plus an override renders while the same URL with `at=` is refused; a pinned page writes the pin and no `gen=`; the viewer is handed `data-generation` and its unfurl card points at the scoped frame.
- New `renderQueryGeneration.test.ts` for the client-side rule.
- `./gradlew :cli:serve:test ktfmtCheckAll` — green.
- `npm --prefix cli/serve-web run verify` — typecheck, tests, and the committed bundle back in sync.

Non-visual: this changes URL query strings and `Cache-Control` headers, not pixels. The rendered frames are byte-identical to what the same catalog served before.

---
_Generated by [Claude Code](https://claude.ai/code/session_01C7yW8gVFW5qkYubwEewA7E)_


yschimke:
Checked these five against `main` before starting on them, and **all five are already fixed** — the threads are just still open. Recording where, so nobody re-does the work:

| Finding | Fixed at |
| --- | --- |
| Route generation-scoped thumbnails before the fast path | `ServeHttpServer.kt:7715` — `staleGeneration(renderHost) == null` now guards the fast path |
| Do not treat viewer mode as a generated PNG | `ServeHttpServer.kt:7797` — a stale generation on a non-raster product refuses |
| Version selection metadata with the displayed frame (P1) | `ServeWeb.kt:10188` — `tagIndexUrl` carries `assetQuery(…, revisions)` |
| Preserve generation when hiding revision controls | `ServeWeb.kt:12186` — `CatalogRevisions(generation = revisions.generation)` instead of `NONE` |
| Thread generation through the published RC lanes | `ServeWeb.kt:9756` — `rcLanesSection` takes `generation` and scopes its assets |

Each carries a `(#4714 review)` citation at the fix, which is what made them findable.

The remaining thread is the agent-commit-identity P1, which is the known false positive — the `Reject agent attribution` gate is the authority and it passed.

I came here from a sweep of post-merge reviews across #4680, #4688, #4702, #4706 and #4713, where the same pattern left findings unanswered because the review landed after the merge. This one was handled; the others are in #4728, #4729 and the merged #4724.

---

PR #4769, opened 2026-08-29
feat: name what is filed on the compare wall, and ask where a report belongs

yschimke:
Three changes to `/<system>/compare?format=reference` and the report it files, all about the same gap: the wall could say that *something* was filed, and could file one more thing, but neither half said **what**.

## 1. The Bugs pill carries the issue title

It was the number alone, with the title on the tooltip — a width decision that cost the column its whole purpose. "Does someone already know about this difference?" is not answered by `#77`, so every row with a number on it had to be hovered, or opened, before the reader learned whether the filed issue was even about the difference in front of them.

The pill is now `#40 Glyph colour is da…`: capped at the column width, ellipsised on one line, and dropped entirely below 1100px so the three picture panels keep the room they need. The tooltip still carries state, number and the untruncated title. Titles join the wall's filter haystack too, because a filter over a table has to match what the reader can see.

## 2. The report form asks where the difference belongs

A third question beside Summary — **Where does it belong?** — with three answers:

| Answer | Label filed |
| --- | --- |
| Upstream — not this catalog | `parity:upstream` |
| A catalog bug — this repository | `parity:catalog` |
| Needs investigating — not sure yet (default) | `parity:verification-needed` |

It is a `<select name="labels">`, so GitHub applies the label straight from the query with or without JavaScript; `<cp-report-classification>` restates it in the body's `**Where it belongs:**` line for a repository that has no such label to apply. `upstream` and `catalog` join the `parity:` vocabulary at **both** ends of the round trip — `parity-issues.mjs` and `ServeParityIssuesStore` — since a value only one end knows is a classification the index drops without saying so.

Not knowing is the default on purpose: a report filed without a thought about this is an unclassified one, and saying so beats a confident wrong label.

## 3. Rows can be reported together

The wall's own report is page-scoped — it names the page, because a wall singles out no preview — which is right for "this whole lane scores zero" and wrong for a reader who has spotted four components drifting the same way. Filed page-scoped, that issue carries no locator, so the index skips it and it reaches the Bugs column of **none** of the rows it is about.

Each row on the reference lane now carries a picker beside its name, and the wall writes one `compose-parity-locator/v1` block per ticked row into the report it already had. A line above the report says what it will name.

**One row per component, enforced at the tick.** The producer refuses a body naming a component, a preview or a reference twice — and refuses the *whole body*, so such a report is filed and then indexed nowhere, with nothing anywhere to notice. The second variant's checkbox goes disabled the moment the first is ticked, with a title saying why; unticking gives it back. The same rule is applied again where the body is written, because a rule enforced only in the UI is one a re-render can lose.

The block writer is ported to the browser (`report/locator.ts`) and pinned to the **same shared fixture** as the Kotlin writer and the JavaScript producer, over every case the fixture says a writer emits — including the astral-key case that catches a plain `.sort()`, which orders by UTF-16 code unit rather than by code point.

## Visual evidence

Full write-up, capture commands and both pairs: [`docs/design/evidence/compare-wall-bug-titles-and-picking/`](https://github.com/yschimke/compose-ai-tools/tree/490bbb6011247db5115e59bc71b7a5a8d52b9f60/docs/design/evidence/compare-wall-bug-titles-and-picking).

The wall at 1280px, before and after:

The wall before: the Bugs column shows #40 and #41 as bare numbers, and rows carry no picker

The wall after: the Bugs pills read "#40 Glyph colour is da…" and "#41 Verify the disable…", the row carries a ticked checkbox, and a line above the report reads "1 comparison will be named in the report below."

The catalog report form, before and after:

The report form before: a Summary field and the submit button

The report form after: "Where does it belong?" sits between Summary and the button, defaulting to "Needs investigating — not sure yet"

### Why the shots are 1280 wide

At the harness's default 1024px the three picture panels, the preview column, `Match` and the pills already come to within ~10px of the content width, so the titles are a wider-than-1024 affordance and the default capture legitimately shows the number-only fallback — unchanged here, which is the other thing worth diffing. Rather than leave the titled column diffed by nothing, a new `serve-format-compare-picked` harness state shoots the same fixture at 1280 with a row ticked, so **both** new surfaces are under the visual-diff bot from now on.

## Test plan

- `./gradlew :cli:serve:test` — green, including the regenerated page fixtures.
- `npm --prefix cli/serve-web run verify` — 1214 passing (23 new), typecheck, build, committed bundle in sync.
- `node --test parity-issues.test.mjs` in `scripts/design-artifacts` — 32 passing (1 new, covering the extended `parity:` vocabulary).
- `npm --prefix preview-server/preview-harness run harness:pages` for `serve-format-compare` and `serve-viewer` — the captures above.

New coverage worth calling out:

- `test/reportLocator.test.ts` runs the browser block writer against every `writer` case in `scripts/design-artifacts/fixtures/parity-locators.json`, byte for byte — the same fixture the Kotlin writer and the JS parser are pinned to.
- `test/comparePicks.test.ts` pins the failure with no symptom: two variants of one component, a row the lane cannot pair, and a lane switch dropping the picks.
- `ServeWebTest` pins that deleting the `{{locators}}` line reproduces the page-scoped body byte for byte — which is what a report filed with nothing ticked, and what a visitor with no script, has to be.

## One thing this does not fix

19 of the 567 rows on the live Wear wall have no design reference in this lane, so their "+ file" falls back to the viewer's own report — which carries no locator by construction, and so never reaches the Bugs column. Those rows cannot be picked either. Worth a follow-up; out of scope here.

https://claude.ai/code/session_019tGmLMhnGkRaw4oiC5YFBc

---
_Generated by [Claude Code](https://claude.ai/code/session_019tGmLMhnGkRaw4oiC5YFBc)_


yschimke:
All six review findings were real and are fixed in `5784ece`. Verified each against the code rather than taken on trust; one summary here rather than six replies.

**The classification default no longer applies a label.** It carried `parity:verification-needed`, which was wrong twice over: that value already meant "a fix landed; the comparison has not confirmed it" — a triage state somebody reaches, not the state every report starts in — and attaching it to every report made `emit-parity-issues.mjs` warn "labelled as parity work but carries no locator block" for reports filed exactly as designed, which is the one place that warning has to keep meaning something. The absence of a classification *is* the unclassified state, so the default now says so in the body and leaves the label list alone. GitHub ignores an empty `labels=`.

**The control is offered only where a `parity:` answer is about something** — the viewer, the focused comparison and the wall. A design page, the pages index and the motion browser file a report about a *page*: there is no comparison there for "upstream or ours" to be a question about, and — the practical half the review found — those pages do not all load `serve-components.js`, so the element would never upgrade and the body would keep the generic line while the select showed an answer. Both reasons point the same way, so the same change closes both findings.

**`viewer.ts`'s `refreshReportLink` now goes through `report/body.ts`.** It assigned `body.value` directly, which made it a second writer of a field whose whole design is that it has one. Harmless while the render URL was the only thing in the body; not harmless once the classification control writes a line, since every knob, device or theme change reverted the reporter's answer with the select still showing it. `ServeWebFixtureTest`'s spelling assertion moved with it and still pins "input value, never an href".

**A row's locator is read from its own `data-reference-detail-<variant>`**, not from the `.cp-compare-bug-new` anchor. That anchor lives in the Bugs cell, which `showBugs` omits entirely for a catalog whose index publishes no issues — so picking was dead on exactly the catalogs filing their first report — and it is only re-pointed when a row is *dressed*, so a row picked, filtered away, then viewed in the other theme kept the previous theme's pair. The attributes are on every row, always, and carry every variant, so neither depends on anything having been rendered. One change, both findings.

**Picks are synchronised before the RC lane's early return**, which skipped the path that normally recomputes them and left ticked locators sitting in a form the reader could still submit.

New regression tests for the two that had none: `comparePicks.test.ts` now files a pick on a wall with no Bugs column at all, and moves a picked-then-filtered row between themes and asserts the body names the *dark* pair. `ServeWebTest` pins that a page-scoped report is not asked where it belongs, and that the default option's value is empty.

Full suite green before pushing: `:cli:serve:test`, `ktfmtCheckAll`, `npm run verify` in `cli/serve-web` (1216 passing), and the 32 producer tests. The evidence capture was retaken for the reworded control.

---

PR #4775, opened 2026-08-29
feat!: take back the five modules that were never wire contracts

yschimke:
The consuming half of [compose-preview-contracts#23](https://github.com/yschimke/compose-preview-contracts/pull/23) (contracts 2.1.0).

> [!IMPORTANT]
> **Do not merge until 2.1.0 is on Central.** This build cannot resolve a coordinate that does not exist. Same ordering as the cutover: contracts releases, then this adopts.

## Why

The contracts repository published nine coordinates. Five were not wire contracts, and `:daemon-protocol` `api`-exported four of them — so a client deserialising **one** daemon message resolved 9,111 lines of published ABI across five coordinates to reach seven types. Those types now live in `:daemon-protocol` itself, and the five modules come back here, where their **81 dependency edges are `project(...)` deps again** rather than a cross-repo release apiece.

`common-io` comes back for a different reason: it was never a contract at all. Two files imported one symbol from it — an alias for `FileSystem.SYSTEM` — and it is entirely behaviour: filesystem access, XDG cache directories, classpath-argument files.

## How

Restored from `5ccfa7fd^` — byte-identical to the pre-cutover tree, confirmed by `git log 611bd2b..HEAD` over those paths in the contracts repo being empty — then the 23 declarations that moved to the protocol were cut with the *same script* that added them there.

| | |
| --- | --- |
| imports repointed to `…daemon.protocol` | 179 across 130 files |
| build references `libs.composeai.*` → `project(...)` | 81 across 53 files |
| `externalContracts` in the contract probe | 9 → 4 |
| `CONTRACT_PROJECTS` | +5 (built here again) |
| `ci-paths.json` | their globs restored to both job groups |
| documentation links | 9 pointed back here from the contracts repo |

## The part that would have quietly broken Central

All five came back still carrying `composeai.maven-publishing`, so a 1.49.0 release would have published them. **`ee.schimke.composeai:common-io` and the four `data-*-core` coordinates already carry 2.0.0 from the contracts repo.** Stamping 1.49.0 onto them would make Gradle and Renovate rank the *new* artifact below the old one, and `latest` would resolve backwards to the stale contracts build — the same ranking trap that forced that repository's re-base *up* rather than down.

They keep the plugin, because the contract probe still publishes them to Maven Local at the probe version; only the Central fan-out excludes them, following the `:rc-player-*` precedent already in `release.yml`. The reasoning is written into the workflow beside the exclusions.

## Test plan

Validated against compose-preview-contracts published to Maven Local at `2.1.0-SNAPSHOT` (behind a temporary `mavenLocal()` + snapshot-pin shim, reverted before commit and verified absent from the diff):

- `./gradlew projects` — the whole build configures with the five modules back in tree.
- All five compile: `:common-io`, `:data-render-core`, `:data-layoutinspector-core`, `:data-theme-core`, `:data-preview-overrides-core`.
- `./gradlew ktfmtFormat` applied.
- `.github/ci/test_path_scope.py` passes against the restored globs.

**The first compile attempt failed, and the failure is worth recording**: 62 unresolved references in files that referenced the moved types from the *same package* and therefore had no import for the rewrite to find — `SemanticsDiff` sat beside `SemanticsDelta` and never needed one. Those 18 files now import them explicitly. A rewrite driven only by existing import statements cannot see that class of reference; the compiler can.

CI here is the real check, and it needs 2.1.0 published first.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Th8Vytwmpg6kQJLdSJdKUd)_


yschimke:
Closing — this was the wrong shape, and it would have broken the preview-server split.

Bringing `common-io` and the four `data-*-core` modules back in-tree meant un-publishing their coordinates. But `preview-server/` is a **separate Gradle build, deliberately not `includeBuild`-ed**, and its own `settings.gradle.kts` explains exactly why that matters:

> everything here resolves the way it will resolve after the split — as a published artifact, by coordinate, from a repository. That is why this build is NOT included in the root `settings.gradle.kts`. Wiring it in with `includeBuild` … would defeat the point: the contracts would resolve from the workspace and **the missing ones would never be missed**.

`cli/serve` depends on all five, and all five are in the probe's `contracts` list. Un-publishing them would have left `check-preview-server-contracts.sh` **green** — it publishes contracts to Maven Local at a fixed probe version — while the real coordinates froze at 2.0.0. A gate that stays green while the thing it guards rots is the precise failure that build exists to prevent.

I had also counted the 81 cross-repo dependency edges as a cost to eliminate. For the split, those edges being published coordinates is the goal, not the cost.

The narrowing itself is unaffected and lands entirely in [compose-preview-contracts#23](https://github.com/yschimke/compose-preview-contracts/pull/23): the 21 wire types move into `:daemon-protocol`, the four `api()` edges go away, and a protocol consumer drops from 9,111 lines across five coordinates to 5,695 in one. All nine coordinates keep publishing, so `serve` still resolves them.

The follow-up that repo needs is much smaller than this PR: bump the `composeai-contracts` pin to 2.1.0 and repoint imports for the moved types. No module moves, no publishing changes. I will open that once 2.1.0 is released.

---

Issue #4821, opened 2026-08-29
Native catalog rendering diverges from the published snapshots it replaces (unaddressed review on #4803)


yschimke:
#4803 made native Wasm rendering the default lane for `compose-m3` and merged with six review threads unanswered, five of them P1. They share a theme worth stating plainly: **the native lane substitutes different pixels for the snapshots it replaces**, which is a problem for a tool whose entire job is parity.

## Fonts — the one I'd check first

[Thread](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770917) — `NativeCatalog.kt:69`

Every native text-bearing preview calls `CatalogApp` without `fontFamily`, `genericFamilies` or `namedFamilies`, falling back to the CMP bundled font rather than the fonts the published snapshots used. The standalone Wasm catalog explicitly loads the `fonts.json` families before composing (`samples/cmp-wasm-catalog/.../Main.kt:65-75`), and `wasmFrontendDist` does not package `fonts/**`.

Consequence: text metrics, wrapping and every typography-themed variant differ in the default lane. For a parity tool this is not a cosmetic gap — it is the measurement drifting.

## Harness-dependent variants rendered without their harness

[Thread](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770920) — `NativeCatalog.kt:38`

The pressed and keyboard-focus captures in `CatalogStates.kt` compose plain buttons and rely on `@FocusedPreview` to dispatch real focus/pointer input. Mapping them to `button-filled-pressed` / `-focused` natively displays an ordinary unpressed, unfocused button. Same shape for `card-slots__…__slot-mode`, which falls through to `card-slots` without providing `LocalSlotMode`.

So these variants don't just differ — they silently show the wrong state under the right label.

## Version-gating

[Thread](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770919) — `NativeCatalog.kt:30`

Native selection keys on the session slug alone. When a host refreshes `design-artifacts/compose-m3` in place (which `deploy/image/entrypoint.sh:556-560` explicitly enables), the slug is unchanged, so the frontend keeps substituting its own compiled composables for newly published snapshots until redeployment.

## Also unaddressed

[...]

Issue #4819, opened 2026-08-29
The serve-wasm surfaces have no automated visual coverage (flagged on three PRs in one afternoon)


yschimke:
The same finding was raised on three separate PRs on 29 Aug, each citing [`AGENTS.md:L122-123`](https://github.com/yschimke/compose-ai-tools/blob/main/AGENTS.md) — "wire new visual surfaces into the preview workflow so the next change to them is diffed without anyone remembering to do it". All three merged unaddressed:

| PR | Surface | Thread |
| --- | --- | --- |
| #4803 | native CMP catalog cards/detail | [link](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770913) |
| #4807 | `UiComposer` (three-pane composer) | [link](https://github.com/yschimke/compose-ai-tools/pull/4807#discussion_r3886838834) |
| #4811 | slot-target, nested-content, inspector states | [link](https://github.com/yschimke/compose-ai-tools/pull/4811#discussion_r3886956249) |

## Verified

```
$ grep -rln "serve-wasm\|UiComposer\|NativeCatalog" .github/workflows/
$
```

No workflow mentions any of it. Each PR shipped a one-time manually captured PNG, so the evidence in those PR bodies is already frozen at the commit that produced it and nothing will diff these surfaces on the next change.

## Why it's worth doing once rather than three times

Three findings in one afternoon on one package is a pattern, not three oversights — the composer and native catalog arrived fast and the preview wiring didn't keep up. A single pass covering the `serve-wasm` surfaces closes all three, and makes the fourth PR into this package cheap instead of a fourth finding.

## Suggested scope

- The native catalog render path (`NativeCatalog.kt`) at the states the published snapshots cover.
- `UiComposer` in its load-bearing states: empty canvas, populated slot, nested slot host, inspector open, preview mode.
- Whatever breakpoint states matter, given #4807 also flagged the three-pane layout collapsing at 760–990dp.

Worth deciding as part of this whether these render through the existing preview lane or need their own capture step, since the surface is a Wasm frontend rather than a `@Preview`.

yschimke:
The gap is wider than visual coverage, and there is a concrete root cause worth recording here.

**`:cli:serve-wasm` has no unit-test coverage in CI either.** The module has a `wasmJsTest` source set with real tests in it (`NativeCatalogTest`, `UiComposerTest`), and nothing runs them:

[...]

PR #4824, opened 2026-08-29
fix(serve-wasm): read a colour override's argb, not a value it never has

yschimke:
An unaddressed P2 from the review on #4803 ([thread](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770923)). Triaged in #4821.

## The problem

`PreviewOverrideValue` serialises each case under its own property name. Every kind except colour spells its payload `value`; colour spells it `argb`:

```json
{"kind":"color","argb":"#FF7DE2FF"}     // ServeBundleStoreTest fixes this shape
{"kind":"bool","value":true}            // every other kind
```

`Main.kt` read `value` alone, so every colour declaration was dropped from `knobSeeds` and the native preview rendered the composable's **author default** instead of the catalog's current colour.

Quiet by construction: nothing errors, and a preview still appears — it is simply the wrong colour. On a tool whose whole job is parity, that is the failure that matters most and the one least likely to be spotted.

## The change

The parse now reads `value` and falls back to `argb`.

`overrideSeeds` becomes the top-level `internal overrideSeedsOf` so the parsing can be tested without a browser — it is the half of this client that has rules, and the half where a missing case looks like a working render.

## ⚠️ Verification status — corrected

**The five tests in this PR do not execute anywhere.** My first version of this description said "CI is the first place these run". That was wrong, and the correction matters more than the original claim.

Locally, `:cli:serve-wasm:wasmJsTest` cannot run in the sandbox — the Kotlin/Wasm test runner provisions Node tooling through yarn and the fetch fails behind the proxy:

```
Execution failed for task ':kotlinWasmToolingSetup'
> Process 'Setup of tooling dependencies' returns 1
```

**And CI does not run them either.** `:cli:serve-wasm` has no `test` task — only `wasmJsTest` — so the affected-module resolver finds nothing to invoke, `tasks` resolves to `none`, and the `Module Unit Tests` job skips. That is observable on this PR's own check runs: `Module Unit Tests — skipped`, on a PR whose only changes are in that module. `grep -rn "wasmJsTest\|serve-wasm" .github/workflows/` returns nothing.

What I *can* evidence:

- `:cli:serve-wasm:compileKotlinWasmJs` — clean
- `:cli:serve-wasm:compileTestKotlinWasmJs` — clean (so the tests are at least valid Kotlin against the real API)
- `:cli:serve-wasm:ktfmtCheck` — clean

So the fix rests on reading, and the tests are provision for when the module is wired up. By construction the colour case returns an empty map without the fix — but I have not watched it fail, and nothing else will until `wasmJsTest` runs somewhere.

I have not wired `wasmJsTest` into CI in this PR: I cannot validate that the Wasm test runner comes up on a CI runner (it does not here), and speculatively adding a job that might go red for everyone is not a change to make blind. Filed as a concrete root cause on #4819 instead.

## Not fixed here

The other findings on that review — [fonts not loaded natively](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770917), [harness-dependent variants rendered without their harness](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770920), and [version-gating against the catalog artifact](https://github.com/yschimke/compose-ai-tools/pull/4803#discussion_r3886770919) — are all larger than a parse fix and point at one decision: whether the native lane should fall back to the server-backed snapshot for anything it cannot faithfully reproduce. Left on #4821 for that call.

## Visual evidence

This changes which colour a native preview renders — a real visual effect I cannot capture, since the surface has no automated preview coverage at all (#4819). The rendered result needs either that coverage or a human opening the composer against a colour-seeded catalog.

### DIFF
--- a/cli/serve-wasm/src/wasmJsMain/kotlin/ee/schimke/composeai/servewasm/Main.kt
+++ b/cli/serve-wasm/src/wasmJsMain/kotlin/ee/schimke/composeai/servewasm/Main.kt
@@ -233,23 +233,37 @@ class BrowserPreviewClient(private val config: ClientConfig) {
   private fun JsonObject.int(key: String): Int? =
     this[key]?.jsonPrimitive?.contentOrNull?.toIntOrNull()
 
-  private fun JsonObject.overrideSeeds(): Map<String, String> =
-    this["overrides"]
-      ?.jsonArray
-      .orEmpty()
-      .mapNotNull { value ->
-        val declaration = value as? JsonObject ?: return@mapNotNull null
-        val key = declaration.string("key")?.takeIf { it.isNotBlank() } ?: return@mapNotNull null
-        val index = declaration.int("index")
-        val encoded =
-          ((declaration["current"] ?: declaration["default"]) as? JsonObject)
-            ?.get("value")
-            ?.jsonPrimitive
-            ?.contentOrNull ?: return@mapNotNull null
-        (if (index == null) key else "$key[$index]") to encoded
-      }
-      .toMap()
+  private fun JsonObject.overrideSeeds(): Map<String, String> = overrideSeedsOf(this)
 }
+
+/**
+ * The knob seeds a published preview's override declarations carry.
+ *
+ * A declaration's value is not always spelled `value`. `PreviewOverrideValue` serialises each case
+ * under its own property name, and the colour case carries `argb` — so reading only `value` dropped
+ * every colour override on the floor and rendered the composable's author default instead of the
+ * catalog's current colour, silently and only for colours.
+ *
+ * Top-level and `internal` so the parsing can be tested without a browser: it is the half of this
+ * client that has rules, and the half where a missing case looks like a working render.
+ */
+internal fun overrideSeedsOf(root: JsonObject): Map<String, String> =
+  root["overrides"]
+    ?.jsonArray
+    .orEmpty()
+    .mapNotNull { value ->
+      val declaration = value as? JsonObject ?: return@mapNotNull null
+      val key =
+        (declaration["key"]?.jsonPrimitive?.contentOrNull)?.takeIf { it.isNotBlank() }
+          ?: return@mapNotNull null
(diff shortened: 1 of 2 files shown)


yschimke:
Fix ci workflows

yschimke:
**`ktfmt (Google style)` failed, and it is not this PR's.** Re-run triggered; diagnosis below so it does not have to be redone.

It is not a formatting violation — no file is reported as unformatted. It is the plugin failing on its own temp file, in a module this diff does not touch:

```
Execution failed for task ':rc-player-trace:ktfmtCheckScripts'
  (registered by plugin 'com.ncorti.ktfmt.gradle').
> Could not read path '…/rc-player/trace/build/tmp/ktfmtCheckKmpWasmJsMain/<uuid>/<uuid>.txt'.
```

Note the path: `ktfmtCheckScripts` is reading a temp file under **`ktfmtCheckKmpWasmJsMain`'s** directory. One ktfmt task holding another's temp path is plugin-internal state, not anything a diff can cause.

Three independent reasons it is not this PR's:

1. **`./gradlew ktfmtCheckAll` passes locally on this exact SHA** (`926b588`).
2. **The failing module is `:rc-player-trace`.** This diff touches only `cli/serve-wasm`.
3. **It is recurring across unrelated branches.** In the last hour of `format.yml` runs: #11033 (this PR) failed, #11024 (`release-please--…`) failed, #11009 (`agent/refresh-driver-pin`) failed — and #11030 is **attempt 2 of that same `agent/refresh-driver-pin` commit and it passed**. Same tree, same code, opposite results, so it is nondeterministic by demonstration rather than by assumption.

Worth noting for whoever fixes it properly: `com.ncorti.ktfmt.gradle` is at **0.27.0, which is already the latest published version**, so there is no upgrade to take. The build runs `org.gradle.parallel=true` with the configuration cache on, and the symptom is two ktfmt tasks in one project disagreeing about which temp directory is theirs.

A real formatting violation is deterministic and would fail both attempts, so the re-run distinguishes the two cleanly rather than papering over anything.

---

PR #4862, opened 2026-08-30
fix(ci): an import renders composables, not the third-party app's activities

yschimke:
## What the diagnostics found

With [#4860](https://github.com/yschimke/compose-ai-tools/pull/4860) in place, the first PeopleInSpace import finally said why it failed ([run 33312180392](https://github.com/yschimke/compose-preview-imports/actions/runs/33312180392)):

```
##[group]Failing previews
--- app/build/test-results/composePreviewRender/TEST-…apptour.AppTourRobolectricRenderTest.xml
FAILED: renderPreview[activity__MainActivity]
  org.koin.core.error.KoinApplicationAlreadyStartedException: A Koin Application has already been started
    at dev.johnoreilly.common.di.KoinKt.initKoin(Koin.kt:31)
    at dev.johnoreilly.peopleinspace.PeopleInSpaceApplication.onCreate(PeopleInSpaceApplication.kt:30)
--- app/build/test-results/composePreviewRender/TEST-…renderer.RobolectricRenderTest.xml
```

The second file has **no failures**. Both of the module's authored `@Preview` functions rendered clean. The import failed entirely on the **app-tour lane** — a lane it never asked for.

## The policy

`activity__*` / `apptour__*` previews are synthesised from the merged manifest by `AppTourDiscovery`, not authored by anyone. Rendering one **launches the app**: Robolectric instantiates the manifest `Application` and runs its `onCreate`, so the app's whole DI graph — and whatever that constructs — executes.

For a first-party catalog that is exactly the point; the launcher activity's capture is the app's hero image. For an imported third-party project it is wrong twice over: it runs far more of an unvouched-for build than rendering a composable does, cutting against the reason imports build in a disposable job with no credential; and it captures the app's own screen, not the components an imported catalog exists to show.

So: **an import renders composables, never the app.**

`appTourUseConsumerApplication = false` is the wrong lever — it swaps the real `Application` for a stub, relocating the failure into an activity that then has no Koin at all. `allow-incomplete` doesn't help either: it applies at catalog **generation**, while this dies in the render step.

## What changed

A workflow-level `IMPORT_EXCLUDE_PREVIEW_IDS` — `activity__*,apptour__*` when `upstream-repository` is set, empty otherwise — enforced at every point that can reach a synthetic preview:

| Path | Treatment |
| --- | --- |
| Module **selection** (`modules: all`) | `discovered-previews.json` is thinned with `jq` before module selection, so a synthetic-only module never reaches the render |
| Primary module | import-wide ids **+** the per-mode ids |
| Later modules of a `modules: all` sweep | import-wide ids |
| `extra-module` | import-wide ids |
| Sharded render | **refused** |

The two id scopes are kept apart deliberately: the per-mode ids belong to the primary module alone, for the same reason its render filter does, while the import-wide ids belong everywhere, because "render in full" for a fallback module means every *authored* preview.

First-party runs are unaffected throughout: `IMPORT_EXCLUDE_PREVIEW_IDS` is empty, the JSON is not thinned, and every path composes to exactly what it passed before.

**Sharding an import is refused** rather than half-supported. The planner partitions every discovered id, synthetic ones included, and the render then drops them; both ways of disagreeing fail — a shard handed only synthetic ids excludes its whole set (which `excludePreviewIds` throws on), and any other shard leaves a planned preview unrendered for `verify-shard-renders.mjs` to reject at merge. Teaching the planner is the proper fix and more than this PR should carry, so the driver fails fast and says why. No import needs it: `import.yml` doesn't expose `render-shards`.

## Review rounds

Three rounds of Codex review found five real gaps, every one of them in a path the PeopleInSpace import doesn't take — it names `:app` explicitly, so no green run could have revealed them. Two are worth reading:

> **A `modules: all` import would have thrown on any app module whose previews are all synthetic.**
> `excludePreviewIds` throws when an exclusion removes a module's entire set — by design, so a render with no outputs can't pack a bundle of missing stickers — and `previewModuleRecords` selects modules with no kind filter. An Android application module with activities but no authored `@Preview` is the *normal* shape when previews live in feature or design modules, and `modules: all` is what `import.yml` passes whenever an `import.json` omits `modules`.

> **The first fix for that used a new `preview-modules.mjs` flag, which would not exist where imports run.** `scripts/` and this workflow reach an external caller by different routes: the caller executes the export driver at the commit in [`design-artifacts-driver-pin.txt`](.github/design-artifacts-driver-pin.txt) (`12847f31`, v1.54.0) while the workflow YAML comes from `@main`. Confirmed by running it — the pinned script aborts with `ERR_PARSE_ARGS_UNKNOWN_OPTION` before rendering anything.

So the rule is enforced by **thinning the input rather than extending the tool**. That works on any pin with no release or pin bump, leaves the pinned script's contract untouched, and keeps `--sources-output` in lockstep for free, since both lists derive from the same records. The `preview-modules.mjs` option from the previous revision is reverted: two mechanisms for one rule invites drift, and this one has to live in the workflow to reach external callers at all.

The others: `extra-module` never received the exclusions; the sharded path never received them (now refused); and the first revision applied them only to the primary module of a sweep.

One finding was declined, and it is a genuine limitation worth stating plainly: **`activity__*` is not a protected namespace.** An authored `fun activity__Header()` would be caught by the glob. Fixing that needs a kind predicate `--exclude-preview-id` doesn't have.

That is a symptom of the real structural point: **render-time id exclusion is the wrong layer.** The right fix is for discovery not to synthesise these previews for an import at all — but `PreviewDiscovery` gates only on a merged manifest being present, with no flag, so that means a plugin change and a release rather than a workflow change. This PR makes the proven path correct and refuses what it cannot do correctly; the plugin-level fix is the follow-up.

## Test plan

- `yaml.safe_load` parses the workflow.
- Extracted the shell composition into a harness and ran all seven exclusion cases: first-party ± mode filter, import ± mode filter, shard first-party, shard import ± partition. The three first-party cases produce byte-for-byte today's arguments.
- Exercised the module-selection filter **against the pinned driver script**, which is what an external caller actually runs:
  ```
  $ node pinned.mjs --input filtered.json --output m.txt --sources-output s.txt
  modules: :feature :legacy :ui        # :app was activity-only, correctly dropped
  sources: /w/feature …/legacy /w/ui   # in lockstep, including the projectDirectory fallback
  $ node pinned.mjs --input unfiltered.json --output m2.txt
  modules: :app :feature :legacy :ui   # first-party keeps :app — its capture is the point
  ```
  A preview with no `id` (an older CLI's `list --json`) survives the filter rather than silently vanishing.
- Reproduced the pin incompatibility before fixing it: the pinned script throws `ERR_PARSE_ARGS_UNKNOWN_OPTION` on `--ignore-id-prefix`.
- Verified the load-bearing claims in source rather than from failure text: `excludePreviewIds` throws on a fully-excluded set; `previewModuleRecords` groups with no kind filter; `AppTourDiscovery.kt:236/:330` build the two id prefixes; `PackPreviewIdExclusions` splits on `,`.
- The run logs `N authored preview(s) of M discovered` when it thins, so a narrowed selection is visible rather than silent.

**End-to-end: proven for the path it takes.** [Import run 33313037741](https://github.com/yschimke/compose-preview-imports/actions/runs/33313037741), pinned to this branch, completed `success` — the first third-party project to render through the import pipeline. That run used the first revision and an explicit `module: :app`, so it exercised the primary-module path only; the `modules: all`, extra-module and shard-refusal paths are covered by the verification above rather than by a live run.

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -1431,7 +1431,24 @@ jobs:
           # render and leave that pass at full width, which is half the axis saving. The flag does
           # both — it forwards `-PcomposePreview.idExclude` to the render and skips the same ids in
           # the daemon capture. Empty ⇒ the flag is omitted ⇒ render + capture everything.
-          EXCLUDE_PREVIEW_IDS: ${{ steps.mode-filter.outputs.exclude-ids }}
+          #
+          # An IMPORT additionally drops the activity and app-tour lanes. Those previews are
+          # synthesised from the merged manifest rather than authored (`activity__<Activity>`,
+          # `apptour__<tour>`), and rendering one LAUNCHES the app: Robolectric instantiates the
+          # manifest `Application` and runs its `onCreate`, so the app's whole DI graph, and
+          # whatever that constructs, executes. For a first-party catalog that is the point — the
+          # launcher activity's capture is the app's hero image. For a third-party project it is
+          # wrong twice over: it runs far more of an unvouched-for build than rendering a
+          # composable does, and it is the app's own screen rather than the components an imported
+          # catalog exists to show.
+          #
+          # It also does not work. PeopleInSpace's `Application.onCreate` calls `initKoin()`, and
+          # the tour died on `KoinApplicationAlreadyStartedException` while BOTH of its authored
+          # @Preview functions rendered clean — an import failing entirely on a lane it never
+          # wanted. `appTourUseConsumerApplication = false` is the wrong lever: it swaps the real
+          # Application for a stub, which just moves the failure into an activity that then has no
+          # Koin at all.
+          EXCLUDE_PREVIEW_IDS: ${{ inputs.upstream-repository != '' && format('activity__*,apptour__*{0}{1}', steps.mode-filter.outputs.exclude-ids != '' && ',' || '', steps.mode-filter.outputs.exclude-ids) || steps.mode-filter.outputs.exclude-ids }}
           EXCLUDE_PREVIEW_ROWS: ${{ steps.mode-filter.outputs.exclude-rows }}
         run: |
           set -euo pipefail
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -430,6 +430,23 @@ env:
   # `upstream/`, with `working-directory` applied INSIDE it. Computed once here because seven steps
   # ask the same question and answering it seven times is how they drift apart.
   RENDER_DIR: ${{ inputs.upstream-repository != '' && (inputs.working-directory != '' && format('upstream/{0}', inputs.working-directory) || 'upstream') || (inputs.working-directory || '.') }}
+  # An IMPORT renders composables, never the app. Activity and app-tour previews are synthesised
+  # from the merged manifest (`activity__<Activity>`, `apptour__<tour>`) rather than authored, and
+  # rendering one LAUNCHES the app: Robolectric instantiates the manifest `Application` and runs its
+  # `onCreate`, so the project's whole DI graph, and whatever that constructs, executes. For a
+  # first-party catalog that is the point — the launcher activity's capture is the app's hero image.
+  # For a third-party project it runs far more of an unvouched-for build than drawing a composable
+  # does, and it captures the app's own screen rather than the components the catalog exists to show.
+  #
+  # PeopleInSpace is the worked example: both authored @Preview functions rendered clean while the
+  # activity tour died in `initKoin()` with KoinApplicationAlreadyStartedException.
+  #
+  # Empty for a first-party run, so those keep rendering activities exactly as before.
+  #
+  # This is workflow-level, not step-level, because it must reach EVERY render: the primary module,
+  # every later module of a `modules: all` sweep, and each shard of a sharded render. It is
+  # deliberately separate from the per-mode exclusions, which belong to the primary module alone.
+  IMPORT_EXCLUDE_PREVIEW_IDS: ${{ inputs.upstream-repository != '' && 'activity__*,apptour__*' || '' }}
   INPUT_LIVE_RERENDER_SOURCE: ${{ inputs.live-rerender-source }}
   INPUT_ALLOW_INCOMPLETE: ${{ inputs.allow-incomplete }}
   INPUT_PUBLISH_LIVE_BUNDLE: ${{ inputs.publish-live-bundle }}
@@ -1010,10 +1027,17 @@ jobs:
           set -euo pipefail
           run() { if [ "$INPUT_DESKTOP_RENDER" = "true" ]; then xvfb-run -a -s "-screen 0 2000x1400x24" "$@"; else "$@"; fi; }
           embed=(); if [ "$INPUT_EMBED_DEPS" = "true" ]; then embed=(--embed-deps); fi
-          exclude=()
+          # The shard's own partition, plus the import-wide exclusions. Without the latter the
+          # shard holding an `activity__*` id would launch the imported app — the same failure the
+          # unsharded path guards, reached by the other route.
+          shard_ids=""
           if [ -s "$GITHUB_WORKSPACE/shard-exclude.txt" ]; then
-            exclude=(--exclude-preview-id "$(cat "$GITHUB_WORKSPACE/shard-exclude.txt")")
+            shard_ids="$(cat "$GITHUB_WORKSPACE/shard-exclude.txt")"
           fi
+          if [ -n "${IMPORT_EXCLUDE_PREVIEW_IDS:-}" ]; then
+            shard_ids="${shard_ids:+${shard_ids},}${IMPORT_EXCLUDE_PREVIEW_IDS}"
+          fi
+          exclude=(); if [ -n "$shard_ids" ]; then exclude=(--exclude-preview-id "$shard_ids"); fi
           if [ -n "${EXCLUDE_PREVIEW_ROWS:-}" ]; then exclude+=(--exclude-preview-row "$EXCLUDE_PREVIEW_ROWS"); fi
           run compose-preview bundle pack --module "$INPUT_MODULE" --with-semantics --progress \
             "${embed[@]}" "${exclude[@]}" -o "$GITHUB_WORKSPACE/bundle.png" \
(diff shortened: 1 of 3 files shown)


yschimke:
Thanks — two of these were real, and one was the failure mode this PR exists to prevent, reintroduced by a path I hadn't traced. Both fixed in 799c863f5. Taking them in order.

**P2 "Preserve import exclusions for every discovered module" — correct, and it was the default import path.** In a `modules: all` sweep only index 0 received `${exclude[@]}`; every later module rendered with none. An import that omits `modules` from its `import.json` takes exactly that branch, so a project with a second Android application module would have launched the app from module two onwards. Fixed by splitting the two scopes, which is what the review asked for: `IMPORT_EXCLUDE_PREVIEW_IDS` is import-wide and reaches every module, while the per-mode ids stay primary-only for the reason the surrounding comment already gives. "Render in full" for a fallback module now means every *authored* preview, not the synthetic ones.

**P2 "Apply the import exclusions to sharded renders" — correct, fixed.** Unreachable through `import.yml` today (it exposes neither `render-shards` nor `cli-version`), but this workflow is the API rather than that caller, so the shard path now composes the partition with the import-wide ids. Confirmed the separator first: `shard-preview-ids.mjs:528` writes `mine.exclude.join(",")` with no trailing newline, so appending `,…` is right.

**P2 "Restrict exclusions to synthetic preview kinds" — real, and I'm not fixing it here.** You're right that these are not a protected namespace: `fun activity__Header()` is legal Kotlin and its id would be caught by the glob. But `--exclude-preview-id` has no kind predicate, so "derive exclusions from discovery records whose kind is ACTIVITY or APP_TOUR" needs new CLI surface — a flag that doesn't exist. That is a larger change than this fix warrants, it only ever affects imports, and it trades a remote collision for a live failure. Worth a follow-up issue rather than widening this PR; flagging it rather than closing it as a non-issue.

**P2 "Require a CLI version that supports import exclusions" — not reachable, so not guarded.** `cli-version` defaults to `latest` and `import.yml` doesn't expose it, so no import can select a 0.17.20–0.19.8 CLI. If it ever does, the right shape is the capability probe this file already uses for `bundle merge` and `=<id>` matching, not a version comparison.

**P1 "Remove the agent identity from the commit metadata" — this is incorrect.** The branch has one commit, `7f37d5b`, authored and committed by `Yuri Schimke <yuri@schimke.ee>`. The cited `7411be` is not an object in this repository (`git cat-file -t 7411be` → `Not a valid object name`). The repo's own detector agrees:

```
$ bash .github/scripts/agent-attribution-scan.sh --range origin/main..agent/import-skips-activity-lanes
No agent co-author trailers or agent commit identities found.   # exit 0
```

This is the fourth time this finding has been raised against a branch whose commits carry the human identity, each time citing a SHA that isn't in the repo. Recording it here so the pattern is visible; there's nothing to recreate.

---

yschimke:
Second round: all three acted on, in `4f312846a`. One of them was a real bug on the **default** import path, and I'd missed it twice.

**"Skip synthetic-only modules before invoking pack" — correct, reachable, and the important one.** I verified both halves rather than taking them on trust:

- `excludePreviewIds` (`RenderPreviewsTask.kt`) does throw when an exclusion removes the whole set — deliberately, so a render with no outputs can't pack a bundle of missing stickers.
- `previewModuleRecords` groups by `preview.module` with no kind filter, so a module whose only previews are synthetic **is** selected.

And that shape isn't a corner case — an Android application module with activities but no authored `@Preview` is what you get whenever the previews live in feature or design modules. A `modules: all` import of any such project would have thrown on the app module and sunk the sweep, taking the modules that *did* have composables with it. `modules: all` is what `import.yml` passes whenever an `import.json` omits `modules`, so this was live.

Fixed at the module-selection step rather than the render: `preview-modules.mjs` takes `--ignore-id-prefix`, so a synthetic-only module never reaches `bundle pack`. Prefixes, not globs — this is our code rather than the CLI's flag, so it matches `AppTourDiscovery`'s `"activity__" + simpleName` / `"apptour__" + name` exactly. Three tests added, including one asserting first-party runs pass no prefixes and keep activity-only modules, and one for a preview with no `id` (an older CLI's `list --json`) surviving rather than silently vanishing. 8/8 pass.

**"Pass import exclusions to the extra-module render" — correct, fixed.** One line; that path had only `embed`.

**"Exclude import-only IDs before planning shards" — correct, and I'm refusing the combination rather than half-fixing it.** You're right that the planner partitions ids the render then drops, and that it fails both ways: a synthetic-only shard excludes its entire set, and any other shard leaves a planned preview unrendered for `verify-shard-renders.mjs` to reject. Teaching the planner the exclusions is the proper fix, and it's more than this PR should carry. So the driver now fails fast on `upstream-repository` + `render-shards > 1` with that reasoning, instead of producing a wrong catalog later somewhere less legible. No import needs it: `import.yml` doesn't expose `render-shards`, and these projects are nowhere near large enough to shard. I kept the shard-side composition behind the refusal so that teaching the planner is the only thing that combination ever needs.

On process: this is the second round where fixing one path surfaced the next, and the underlying reason is structural. Render-time id exclusion is the wrong layer — the right fix is for discovery not to synthesise these previews for an import at all, which needs a plugin flag and a release rather than a workflow change. I've filed that thinking into the PR body. I'd rather land this, which makes the proven path correct and refuses what it can't do, than keep widening it.

I have not re-litigated the P1; it remains false, and `agent-attribution-scan.sh` still exits 0 on all three commits.

---

yschimke:
**The driver-pin finding is correct, and it would have broken exactly the thing this PR exists to enable.** Fixed in `ac4f8e3e3`.

I had missed that `scripts/` and the workflow reach an external caller by different routes. `.github/design-artifacts-driver-pin.txt` spells it out — an external caller executes the export driver at the pinned commit (`12847f31`, v1.54.0) while the workflow YAML comes from `@main` — and compose-preview-imports is an external caller. So a new `preview-modules.mjs` option would not exist where it runs.

Reproduced rather than reasoned about:

```
$ git show 12847f31:scripts/design-artifacts/preview-modules.mjs | grep ignore-id-prefix
(no output)

$ node pinned.mjs --input d.json --output m.txt --ignore-id-prefix activity__
node:internal/util/parse_args/parse_args:102
      throw new ERR_PARSE_ARGS_UNKNOWN_OPTION(
```

Every `modules: all` import would have died there before rendering anything.

**The fix inverts it: thin the input instead of extending the tool.** The workflow now filters `discovered-previews.json` with `jq` and hands the pinned script the authored-only list. Three things fall out of that which the flag approach didn't have:

- it works on *any* pin, so it needs no release and no pin bump;
- the pinned script's contract is untouched, so there's nothing to keep in sync;
- `--sources-output` stays in lockstep for free, since it's derived from the same records — the flag version would have needed the module list and the sources list filtered together.

Verified against the **pinned** script, not the branch's:

```
$ jq '.previews |= map(select((.id // "") | (startswith("activity__") or startswith("apptour__")) | not))' d.json > f.json
$ node pinned.mjs --input f.json --output m.txt --sources-output s.txt
modules: :feature :legacy :ui          # :app was activity-only, correctly dropped
sources: /w/feature …/legacy /w/ui     # in lockstep, including the projectDirectory fallback
$ node pinned.mjs --input d.json --output m2.txt
modules: :app :feature :legacy :ui     # first-party unfiltered, :app kept — its capture is the point
```

I reverted the `preview-modules.mjs` option and its tests: two mechanisms for one rule invites drift, and this one has to live in the workflow to reach external callers at all. The run also now logs `N authored preview(s) of M discovered`, so a thinned selection is visible rather than silent.

**P1, fifth time:** `bfc77115324d3fae2d21c5d9be644f84fd52301d` is not an object in this repository (`git cat-file` → `could not get object info`). All three commits on this branch are `Yuri Schimke <yuri@schimke.ee>` as both author and committer, and `agent-attribution-scan.sh --range origin/main..HEAD` reports `No agent co-author trailers or agent commit identities found`. There is nothing to recreate. I won't re-answer this one again.

---

yschimke:
Correction to my previous comment: I said the `preview-modules.mjs` option was reverted in `ac4f8e3e3`. **It wasn't.** I used `git checkout <paths>`, which restores from the index — and since those files were already committed in `4f312846a`, it was a silent no-op, reporting `Updated 0 paths from the index`. The branch carried both mechanisms for two commits: the unused `--ignore-id-prefix` option *and* the `jq` filter that replaced it. Precisely the drift I claimed to be avoiding.

Actually reverted now in `7c311a83a`, from `origin/main` rather than the index. The diff is one file again:

```
$ git diff --stat origin/main..HEAD
 .github/workflows/design-artifacts-reusable.yml | 105 ++++++++++++++++---
 1 file changed, 99 insertions(+), 6 deletions(-)

$ node --test scripts/design-artifacts/preview-modules.test.mjs
# pass 5   # main's suite, unmodified
```

The PR body already described the end state correctly, so it needed no change — but it was describing something the branch didn't yet match, which is worse than a stale body. Flagging rather than quietly fixing, since a reviewer reading that comment would have believed the tree was in a state it wasn't.

---

Issue #4853, opened 2026-08-30
RotateToLookAtUserPreview's XR composite is nondeterministic, so every PR gets flagged for it


yschimke:
`renders/samples:xr-spatial/composite.png` (`RotateToLookAtUserPreview`) renders differently on PRs that do not touch `samples/xr-spatial`. The visual-diff bot attributes the change to whichever PR happens to be open, so unrelated PRs carry a spurious "Changed" entry.

Surfaced on #4852 (a version-catalog + workflow change) and triaged per [`.claude/skills/flake-triage`](.claude/skills/flake-triage/SKILL.md).

## Evidence

Blob hashes of `renders/samples:xr-spatial/composite.png`:

| blob | render commit | that change touched |
| --- | --- | --- |
| `3150fd98` | `1a52da22` — baseline from `cab7d8fc` | — |
| `3150fd98` | `4c73f1eb` — baseline from `ea76cd1a` | — |
| `246ad1f6` | `68b0905d` — PR #4851 | catalog annotations, `design-map.mjs`, docs |
| `df837b84` | `b467a624` — PR #4852 | `gradle/libs.versions.toml`, `.github/workflows/ci.yml` |

Two independent main baselines are **byte-identical**. Two PRs with disjoint diffs, neither touching `samples/xr-spatial`, each produce a different render — and differ from each other.

Pixel differences (pixelmatch, threshold 0.1, 1280x800):

```
base   vs #4852:  397065/1024000 px  (38.78%)
base   vs #4851:  461512/1024000 px  (45.07%)
#4851  vs #4852:  364840/1024000 px  (35.63%)
```

35–45% of the frame is not a subtle antialiasing wobble; it reads as a wholly different camera or pose.

## Where to look

The preview source is static — three panels with `.rotateToLookAtUser()`, no clock, no randomness ([`XrModifierPreviews.kt:160`](samples/xr-spatial/src/main/kotlin/com/example/samplexrspatial/XrModifierPreviews.kt#L160)). So the nondeterminism is downstream, in the bake:

- the seeded head pose (`:renderer-xr`'s `FakeXrHeadPose`) — if it is not pinned identically on every run, `rotateToLookAtUser` turns the panels to a different angle, which would move most of the frame exactly as observed;
- software-GL rasterisation under Xvfb/Mesa in the pinned `xr-composite` binary;
- panel-texture compositing order.

[...]

PR #4865, opened 2026-08-30
fix(ci): pin download-artifact to a SHA that exists

yschimke:
## This is a live break on `main`, and I introduced it

`publish-catalog` pins `actions/download-artifact` to `018cc2cf5baa6db3ef3c5f28c025e4dfd1ca53b3`, which is **not a commit in `actions/download-artifact`**. GitHub cannot resolve it, so the job dies in *Set up job* — before any step runs:

```
##[error]Unable to resolve action `actions/download-artifact@018cc2cf5baa6db3ef3c5f28c025e4dfd1ca53b3`,
unable to find version `018cc2cf5baa6db3ef3c5f28c025e4dfd1ca53b3`
```

**Any catalog publish fails**, first-party included — the render completes, uploads its bundle, and then the job that would push it cannot start.

I added it in [#4856](https://github.com/yschimke/compose-ai-tools/pull/4856), the generate/publish split. It has been on `main` since, latent only because no publish has run:

```
$ git log -S018cc2cf5baa6db3ef3c5f28c025e4dfd1ca53b3 -- .github/workflows/design-artifacts-reusable.yml
83de8c485 refactor(ci): publish from a job that runs none of the rendered code (#4856)
```

Found because it broke the first PeopleInSpace publish ([run 33318777406](https://github.com/yschimke/compose-preview-imports/actions/runs/33318777406)) — the render succeeded and the `publish` job failed at setup with the identical error, from the identical bad pin copied into `import.yml`. That repo gets its own fix; this one is the more serious of the two.

## The fix

Use `3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1` — the pin the other two `download-artifact` steps in this same file already use, in place since [#4771](https://github.com/yschimke/compose-ai-tools/pull/4771) and exercised on every sharded render. All three uses now agree.

## Test plan

- `yaml.safe_load` parses.
- Audited **every** action pin my recent PRs (#4846, #4856, #4859, #4860, #4862) added to this workflow, against `4e961c6a0`:

  | pin | uses in repo | existed before my work |
  | --- | --- | --- |
  | `actions/checkout@3d3c42e5aa` | 85 | yes |
  | `actions/upload-artifact@043fb46d1a` | 48 | yes |
  | `actions/download-artifact@018cc2cf5b` | **1** | **no** |

  The bad pin is the only one I introduced from nowhere, and the only one used exactly once. Every other pin I added was already load-bearing elsewhere.
- The replacement is verified by use rather than by assertion: `018cc2cf…` is used once and never resolved; `3e5f45b2…` is used by "Download shard renders" and "Download section bundle to fold in", both of which run in production. I could not check either SHA against `actions/download-artifact` directly — this sandbox's proxy returns 403 for repositories outside the session's scope — so "exercised in this repo's own green runs" is the strongest evidence available here, and it is stronger than an API lookup would have been anyway.

**A pin nothing else in the repo uses is worth a second look.** That is the lesson I'd take from this: the audit above would have caught it at review time in #4856, and it is a cheap check to run on any PR that adds a pinned action.

---
_Generated by [Claude Code](https://claude.ai/code/session_0134yHmD5U7amNDUz34oRnwE)_

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -2509,7 +2509,7 @@ jobs:
           persist-credentials: false
 
       - name: Download the rendered catalog
-        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f28c025e4dfd1ca53b3 # v6.0.0
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: ${{ inputs.system }}-catalog-out
           path: ${{ github.workspace }}/out


yschimke:
The preview diff flags `RotateToLookAtUserPreview` (`samples:xr-spatial`) as changed. **It is not this PR's**, and I want to say why rather than let a "Changed" section sit unexplained under a body that claims no visual impact.

The entire diff is one line:

```diff
-        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f28c025e4dfd1ca53b3 # v6.0.0
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
```

A SHA in a `uses:` field, in a job that runs after every render. There is no path by which it changes a pixel.

This is the known instability in [#4853](https://github.com/yschimke/compose-ai-tools/issues/4853) — that preview's composite is nondeterministic and gets attributed to whichever PR is open, with 35–45% of the frame differing between runs of untouched code. The bot's own note adds a second reason here: its baseline is `83de8c48`, three commits behind this PR's base `ce62c4a1`, so anything those commits changed lands in this PR's column too.

Not re-running the diff and not touching the render. #4853 says the fix is to pin the nondeterminism at source with a pixel test, not to loosen a threshold or re-roll until it comes up clean — and it also warns that flagging unrelated PRs "trains reviewers to wave through a Changed section", which is the reason I'm writing this out instead of ignoring it.

Everything else on this head is green: CI, No Agent Attribution, CodeQL, Daemon Harness, Integration, Actions Script Tests, Format Check and the rest. Compose Preview is the only workflow still running.

---

Issue #4838, opened 2026-08-30
serve: the cross-catalog pairing resolves one canonical sticker per component, and carries no typography or theme diff


yschimke:
**Correction to the first version of this issue, which claimed no cross-catalog comparison existed at all. It does** — `compareWith` + `parallel`, added in #4621 — and I had missed it. What follows is the narrower, real gap.

## What exists

A catalog names its sibling system in `catalog.spec.json`:

```json
"compareWith": { "system": "wear-m3-catalog", "spec": "../catalog.spec.json" }
```

and a component names its counterpart with `@CatalogComponent(parallel = …)`; `@CatalogVariant.parallel` carries the same field, explicitly so a variant is "compared in its own right, not through its parent". `parallelSpecSource` (`ServeHttpServer.kt` ~4497) walks system → component → sibling preview and offers the counterpart's render as a **second source for the spec lane**, failing soft at every step. `remote-m3` already declares `compareWith: wear-m3-catalog` and authors `parallel` on 31 of its 48 components.

## The two gaps

**1. Component granularity, not cell granularity.** The walk ends at:

```kotlin
val siblingPreview =
  siblingHost.previews.firstOrNull { it.componentId == parallelId } ?: return null
// "a component with several previews has one canonical sticker and the manifest's order is the producer's"
```

So `Button/Child` on one sheet pairs with whichever `Button/Child` preview the sibling's manifest happens to list first — its `default`, in practice. The `disabled`, `icon` and `left` cells have no counterpart even though the sibling draws every one of them, and `CatalogVariant.parallel`'s own docs promise the opposite ("a variant is compared in its own right"). The natural key is the **kit cell**: both sides carry `reference` + `kitAxis`/`kitValue` seeds in `design-map.json`, so `Disabled=Yes` on one sheet can find `Disabled=Yes` on the other without either catalog naming its previews the same. Falling back to the canonical sticker when a cell has no counterpart keeps today's behaviour as the floor.

**2. No compare wall, and no typography or theme diff.** The pairing feeds the spec lane only. The `/compare` surfaces stay single-system — `ServeWeb.kt` ~9204–9211 hardcodes `svg` / `rc` / `reference`, all resolved inside one catalog, and `defaultFormat` (~9493) picks among the same three. An unknown `?format=` falls back silently rather than erroring, which is a small separate papercut.

[...]

Issue #4732, opened 2026-08-28
Plan: split `compose-preview serve` and the extension into separate repos


yschimke:
Successor to #3824, which asked only for the prep work and a way to measure. The prep is done; this issue carries the plan and the residual work.

## Where we actually stand

**Structurally, serve is separable.** `:cli:serve` is a module with an enforced boundary (`checkServeModuleBoundary`), the seam ratchet (`checkServeSeam`) is wired into `check`, the Playwright harness lives with the server rather than the extension, and serve imports nothing from `daemon.core` — 171 imports land on `daemon.protocol`, 2 on `daemon.devices`, 2 on `daemon.bta`, all three of which are published contracts. Twenty contracts are registered in all four places the registration gate checks. `checkKotlinAbi` now runs for every module that has it, both in the affected-scope resolver and on the `full` branch.

**By the ready-to-split test, it is not.** Measured on the unshallowed history (numbers and caveats in [#3824's closing comment](https://github.com/yschimke/compose-ai-tools/issues/3824)):

| condition | target | today |
| --- | --- | --- |
| 1. Structural | seam green, no allowlist growth | ✅ holds |
| 2. Volume — % of all PRs | ≤ 5% | 10.3% – 12.5% |
| 2. Volume — % of serve-touching PRs | ≤ 15% | 25.1% – 27.3% |
| 3. Depth — deep crossings/week | ≤ 2 | **9.0 – 10.7** |
| 4. Protocol | no serve-motivated bump | ✅ holds |

## What #3824 got wrong, and what to do about it

The assumption behind the daemon split was that deep crossings were daemon-protocol traffic. They are not. Over a clean two-week window the deep traffic was `daemon/android` ×71, `renderers/android` ×68, `gradle-plugin/src` ×31, `daemon/desktop` ×23, `daemon/core` ×14. Serve co-changes with the **render path**, not the daemon path — because a change to how a preview renders is usually also a change to how serve presents it. That is product cohesion, not a missing boundary, and no further module extraction will move it.

Three ways forward, in the order I'd try them:

1. **Split the extension first, not serve.** The extension already passes the volume-of-all-PRs condition (2.7%) and touches the render path far less. Its failing ratio is over only 10 touching PRs — noise from a component in maintenance (see #3856). It is the cheap, low-risk rehearsal for the mechanics: publishing, versioning, cross-repo CI, the contract-drift gate. Learn the migration on the component where being wrong is survivable.
[...]

PR #4870, opened 2026-08-30
fix(serve): pair the spec lane's sibling on the kit cell, not the component

yschimke:
Gap 1 of #4838. The compare wall and the typography/theme layers in that issue are untouched — see **Not in this PR** below.

## The bug

`parallelSpecSource` resolved the counterpart by taking the sibling's first preview for the component:

```kotlin
val siblingPreview =
  siblingHost.previews.firstOrNull { it.componentId == parallelId } ?: return null
```

So every **cell** of a component was compared against one canonical sticker. `Disabled=Yes` on one sheet was scored against the other sheet's **default** render — while both sheets drew the disabled cell — and the difference reported was mostly the difference between an enabled button and a disabled one. `@CatalogVariant.parallel` documents the opposite: *"a variant is compared in its own right, not through its parent."*

## The key

Both sheets reproduce one design kit, so the **kit node** is a key that means the same thing on both sides. It is already published — a Figma-backed reference carries `figma:<fileKey>/<nodeId>` — so this only projects it out:

```kotlin
fun kitCell(reference: DesignReference): String?   // "<fileKey>/<nodeId>"
fun kitCells(references: List<DesignReference>): Set<String>
```

and the pairing prefers a sibling preview sharing one. Neither catalog has to name its previews, variants or components the way the other does — which is what makes this survive the renames that defeat id-matching (`iconbutton-filled` vs `button-icon-filled`; a breakpoint suffix on one side only). Node ids are normalised to the `:` form, so a producer emitting the `-` URL form still pairs with one emitting `:`. Parsing is the same strict validation `of()` already does, for the same reason: a catalog is third-party data, so a malformed handle yields no key and therefore no pairing, rather than a key that matches something by accident.

## The floor is unchanged

The canonical sticker remains the fallback, so **nothing that paired before stops pairing**: a preview naming no kit node, a sibling publishing no Figma references, and a cell the other sheet genuinely does not draw all keep exactly what they had.

## The provenance now says which pairing you got

The two are not equally strong, and conflating them would relocate the bug rather than remove it:

- cell match → *"Paired on the design-kit cell both sheets name."*
- fallback → *"Paired at component level: that catalog's canonical sticker, which may draw a different cell of this component than the render beside it."*

A difference in a cell match is a real finding. A difference in the fallback may be nothing but the fact that it is a different cell.

## Tests

Two, against one fixture whose sibling draws two cells:

- the disabled cell pairs with the **disabled** cell, and its provenance says so;
- a preview whose reference is a **PNG** rather than Figma still pairs, and declares itself component-level.

The fallback's *identity* is deliberately not asserted. It follows the manifest's order — the producer's — and in this fixture that happens to be alphabetical, which puts `chip-disabled` ahead of `chip-filled` and makes the "canonical sticker" the disabled cell. Pinning that would pin an accident of naming, and it is precisely the fragility the cell key routes around.

Two things the fixture work turned up, both recorded in comments so the next person does not rediscover them: the spec lane only exists for a preview that has **some** reference, so a preview with none never reaches this code at all; and the existing tests in this file keep exercising the fallback path, because their reference declares `provider: "figma"` with no URI.

## Verification

`:cli:serve:test` green in full (not just the touched class), `ktfmtFormat` clean.

## Not in this PR

Gap 2 of #4838 — the compare wall, and the typography / theme / layout diff layers. Those are additive surfaces rather than a correction to an existing one, and this pairing is the thing they would be built on.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SxnGrAmCM3y1AbrFixVcqK)_

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -1578,6 +1578,15 @@ jobs:
         run: |
           set -euo pipefail
           echo "Projecting design-map.json for '$INPUT_SYSTEM'."
+          # DELETE FIRST, so the check below proves the command RECREATED the map rather than that
+          # a map exists. In the two-catalog repo this input is for, the committed one at this path
+          # belongs to the OTHER system — so a command that succeeds but writes somewhere else (the
+          # `--out-dir` mistake, which `scripts/design-map.sh` makes easy) would leave it standing,
+          # pass a mere existence check, and publish the other system's mappings under this one's
+          # name. That is the exact failure this step exists to stop, and it would sail through.
+          # Safe because the checkout is disposable: nothing later in the job wants the committed
+          # map back, and a command that writes none fails the step anyway.
+          rm -f design-map.json
           bash -c "$DESIGN_MAP_COMMAND"
           if [ ! -f design-map.json ]; then
             echo "::error title=design-map-command wrote no map::The command ran but left no design-map.json at the repo root, which is the only path the lanes below read. Check that it projects to the root rather than to an --out-dir."
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -265,6 +265,19 @@ on:
         required: false
         type: string
         default: ''
+      design-map-command:
+        description: >-
+          Shell command that projects THIS system's `design-map.json` into the caller repo root,
+          run once before every step that reads it. Same input name and semantics as design-parity's
+          reusable workflow, and for the same reason: a repo publishing TWO catalogs from two modules
+          has one map PATH and two systems needing different maps, so without this the second system
+          is scored against the first one's map — every entry dangles and coverage publishes as 0%.
+          `system`, `spec` and `module` already vary per call; the map is the fourth thing that does.
+          Empty ⇒ the committed map is used as-is, which is right for the single-catalog repos that
+          are the common case.
+        required: false
+        type: string
+        default: ''
       preview-annotations:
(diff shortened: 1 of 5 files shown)


yschimke:
Closing — this is both **misplaced and redundant**, and I should have caught either before opening it.

**Misplaced.** I branched from a checkout at `27df77b`, which is not an ancestor of `main`, and never checked. `main` no longer owns `cli/serve` at all — its own version catalog says so:

> The preview server, published from yschimke/compose-preview-server since its 2.0.0 — the extraction #4732 planned. This repository no longer owns `cli/serve`.

So every file this PR edits is absent from `main`, which is why the merge reported them as "deleted in origin/main". I verified that against GitHub rather than trusting the shallow clone: both paths 404 on `main`, and `main`'s HEAD is exactly `2519718`.

**Redundant.** Both gaps of #4838 are already fixed in `compose-preview-server`, and the pairing there is better than what I wrote. It has three tiers where mine had two:

```kotlin
enum class Basis {
  /** Both renders are specified by the same design-kit node. The strongest pairing there is. */
  KIT_CELL,
  /** Both renders carry the same `state` / `props` / `size` coordinates. */
  VARIANT_CELL,
  /** No counterpart for this cell; the component's first published render, as before. */
  CANONICAL,
}
```

`VARIANT_CELL` is the case I missed entirely: two catalogs that draw the same cell but where one of them names no kit node still pair on their state/props/size coordinates, instead of dropping to the canonical sticker as mine would. The provenance is also more specific than mine — it names the cell ("paired on the `disabled` cell", "its default render, because that catalog publishes no `disabled` cell for this component") where I only distinguished cell-level from component-level.

Gap 2 landed there too, as `feat(serve): diff the derived layers across a cross-catalog pair` (#40).

Nothing here is worth porting. Leaving #4838 open or closed is a call for its author, but as far as I can tell both halves of it are now implemented in the repo that owns the code.

---

PR #5037, opened 2026-09-02
feat(import): let a caller name the Android variant to render

yschimke:
The preview plugin attaches its `composePreview*` tasks to **one** variant, and its convention is `debug`. A module with product flavors has no plain `debug`, so the variant that gets picked may not be the one that builds — and there was no way to say which.

home-assistant/android is the case in point. `app/src/screenshotTest/` is flavor-agnostic but references `SettingsWearOnboardingViewContent`, which exists only in `app/src/full/`. The plugin compiles screenshotTest sources deliberately — it renders `@Preview` functions declared there — so the render dies before a preview is produced:

```
e: …/SettingsWearOnboardingViewPreviewsTest.kt:16:13
   Unresolved reference 'SettingsWearOnboardingViewContent'.
Execution failed for task ':app:compileMinimalDebugScreenshotTestKotlin'
```

`fullDebug` compiles. Nothing is wrong with the import, the module, or the plugin — the run was simply building the flavor upstream never builds those sources against. (Confirmed against the upstream tree: `app/src/full/…/SettingsWearOnboardingView.kt` exists, `app/src/main/…` and `app/src/minimal/…` do not; the flavors come from `AndroidFullMinimalFlavorConventionPlugin`.)

## Changes

The CLI already has `--variant` for exactly this — *"a project with only flavored variants and no plain `debug`"* — so this exposes it as a workflow input.

It reaches **discovery** as well as the render, and that is not optional: `compose-preview list` enumerates a module by its `composePreview*` task presence, so a list run on a different variant than the render enumerates a different task set — on a flavored module, none at all.

Threaded through:

- all five `bundle pack` call sites — sharded, repository-wide import, repository-wide non-import, single-module, extra-module
- all three `compose-preview list --json` sites

## Compatibility

Empty by default, and an empty `INPUT_VARIANT` expands to no argument at all, so every existing caller's command line is unchanged byte for byte. Verified under `set -euo pipefail` for unset, set, and empty:

```
unset  → [compose-preview] [bundle] [pack] [--module] [:app] [-o] [out.png]
set    → [compose-preview] [bundle] [pack] [--module] [:app] [--variant] [fullDebug] [-o] [out.png]
empty  → [compose-preview] [bundle] [pack] [--module] [:app] [-o] [out.png]
```

---
_Generated by [Claude Code](https://claude.ai/code/session_01UXFgrNVx7S1jsk1ridhagp)_

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -50,6 +50,21 @@ on:
         required: false
         type: string
         default: ''
+      variant:
+        description: >
+          Android variant the preview plugin attaches its tasks to, forwarded to the CLI as
+          `--variant` (which becomes `-PcomposePreview.variant=<name>` on every Gradle
+          invocation). Omit for the plugin's own default of `debug`, which is right for any
+          module without product flavors.
+
+          Needed when a module HAS flavors, because there is then no plain `debug` variant and
+          the flavor picked for you may not be the one that builds. home-assistant/android is
+          the case in point: its `app/src/screenshotTest/` source set is flavor-agnostic but
+          references a symbol that exists only in the `full` flavor, so `minimalDebug` fails to
+          compile before a preview is rendered, while `fullDebug` is fine.
+        required: false
+        type: string
+        default: ''
       modules:
         description: 'Set to "all" to discover and publish every preview-enabled Gradle module. Equivalent to omitting module; other values are rejected.'
         required: false
@@ -445,6 +460,7 @@ env:
   INPUT_CATALOG_KEY: ${{ inputs.catalog-key }}
   INPUT_CATALOG_PATH: ${{ inputs.catalog-path }}
   INPUT_MODULE: ${{ inputs.module }}
+  INPUT_VARIANT: ${{ inputs.variant }}
   INPUT_MODULES: ${{ inputs.modules }}
   INPUT_EXTRA_MODULE: ${{ inputs.extra-module }}
   INPUT_RENDER_SHARDS: ${{ inputs.render-shards }}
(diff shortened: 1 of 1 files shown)


yschimke:
Following up after merge on the three review findings.

**The two P1s about the input itself were right, and are fixed in #5040:**

1. *Reject a global variant for repository-wide discovery* — confirmed against `AndroidPreviewSupport.variantMatchesTarget`, whose own KDoc says a flavored target "does NOT match a flavorless `debug` variant … the module is silently skipped". `modules: all` + a flavored variant would have published a smaller catalog with nothing in the log. Now refused.
2. *Preserve the variant in live source metadata* — confirmed: `source_args` carries repo/ref/module only, and `generate-design-catalog.mjs` accepts no `--source-variant`, so a source rebuild would run the default `debug` — for this input's motivating case, the variant that does not compile. #5040 refuses the combination where source rebuild is the only live path, rather than publishing a catalog that looks complete and re-renders wrong. Carrying it properly (generator flag, catalog source block, serve applying it) is noted in the comment for whoever picks it up.

**The P1 about commit attribution does not hold.** It cites `8aeebc9eccf3ec4ae1351131214bf2fb97b99d09` with author and committer `Codex <codex@openai.com>`; that object does not exist in this repository:

```
$ git cat-file -t 8aeebc9
fatal: Not a valid object name 8aeebc9

$ git log --pretty="%h %an <%ae>" -3 origin/main
60c45e280 Yuri Schimke <yuri@schimke.ee>
02a435245 Yuri Schimke <yuri@schimke.ee>
92efd7ea1 github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
```

`agent-attribution-scan.sh --range origin/main..HEAD` — the script the gate runs — exited 0 on this branch before it was pushed, and the same finding with a different non-resolving SHA (`c32c9c9`) appeared on #5035. That is the ninth such report in this session, always naming a SHA that does not resolve. Worth looking at how those ranges are constructed.

---

PR #5052, opened 2026-09-03
fix(cli): resolve the build root by settings file, and name the plugin publication race

### LINKED ISSUE #5031: CLI resolves the wrong Gradle build for a nested build with no wrapper of its own
`findGradleProjectRoot` (`cli/src/main/kotlin/ee/schimke/composeai/cli/VersionPin.kt`) answers "which directory is the project" by walking **up** from the cwd until it finds a `gradlew`:

```kotlin
internal fun findGradleProjectRoot(start: File = File(".").absoluteFile): File? {
  var dir: File? = start
  while (dir != null) {
    if (File(dir, "gradlew").exists()) return dir
    dir = dir.parentFile
  }
  return null
}
```

A wrapper is not what makes a directory a Gradle build — a `settings.gradle(.kts)` is. A nested build that borrows its parent repository's wrapper is therefore walked straight past, and the CLI silently drives the **enclosing** build instead.

## How it shows up

thunderbird-android's `components/` is a complete build: its own `settings.gradle.kts`, `rootProject.name = "components"`, three projects, and no `gradlew` (it uses the repository root's). Rendering `:ui:bolt` from `components/` ran the repository root build instead:

```
Module ':ui:bolt' not found or does not apply the compose-ai-tools plugin.
206 project(s) failed to configure during discovery and were skipped
  :: InvalidUserCodeException: Project ':' cannot access 'Project.buildscript' …
  :app-k9mail: … :app-thunderbird: … :backend:imap: …
```

206 is the root build's project count, and the `InvalidUserCodeException` is the root build's `org.gradle.unsafe.isolated-projects=true` — neither of which is a property of the build the user asked for. `components/` itself configures clean under Gradle 9.7 with the same `allprojects { buildscript { … } }` injection auto-inject performs, with Isolated Projects probing as inactive.

The failure mode is bad in a specific way: nothing says "I picked a different build than the directory you were standing in". The diagnostics that do surface are all true statements about a build the user never named, which sends the reader off diagnosing Isolated Projects.

## Suggested shape

- Resolve the **build root** as the nearest ancestor holding `settings.gradle` / `settings.gradle.kts`, falling back to the current wrapper search when there is none.
- Keep the wrapper search for what it is actually for: `GradleConnection` passes the root to `GradleConnector.forProjectDirectory(...)`, which takes the distribution from that directory's `gradle/wrapper/gradle-wrapper.properties`. When the build root has none, take the distribution from the nearest ancestor that does (`useDistribution`), rather than falling back to the Tooling API default.
- `PinCommand` shares this helper, so consider whether "where the pin is written" should follow the build root or stay with the wrapper; they are the same directory for every project that has both.

Worth a diagnostic either way: when the resolved root is not the directory the CLI was invoked from, say so, so a mis-resolution is legible instead of arriving as a wall of unrelated configuration failures.

## Workaround in place

yschimke/compose-ai-tools#5030 copies the nearest ancestor's wrapper into the throwaway checkout for imports that set `working-directory`, which stops the walk at the right build. That unblocks the import pipeline immediately (it does not wait on a CLI release), but it papers over this; anyone driving the CLI by hand at a nested build still gets the wrong build.

### LINKED ISSUE #5034: `cli-version: latest` races the Gradle plugin's publication, and every import in that window dies
## What happens

The install action resolves `cli-version: latest` from the **GitHub release**, but the render then asks Gradle to resolve `ee.schimke.composeai.preview` from **plugins.gradle.org**. Those two become available minutes apart, and every import dispatched in the gap dies during configuration:

```
:: ProjectConfigurationException: A problem occurred configuring root project '…'.
   -> Could not resolve all dependencies for configuration 'classpath'.
   -> Could not find ee.schimke.composeai.preview:ee.schimke.composeai.preview.gradle.plugin:1.66.1
```

The job's own log says both halves in the same run — `Resolved compose-preview version: 1.66.1` as a notice, and then a build that cannot resolve `1.66.1`.

## Frequency

Three rounds of import runs so far, one per release cut:

| release | dispatched | outcome |
| --- | --- | --- |
| 1.64.0 | ~20:35 | every import in flight failed |
| 1.65.0 | ~21:58 | every import in flight failed |
| 1.66.1 | ~22:38 | home-assistant-android, home-assistant-wear, pocketcasts all failed |

Each round costs a full re-dispatch of everything that was in flight, and the failure is indistinguishable at a glance from a real import problem — it presents as a configuration failure in the imported project, which is where the time goes. In the 1.66.1 round I confirmed the race directly: `plugins.gradle.org/m2/…/1.66.1.pom` returned 200 by the time I looked, minutes after the runs that could not resolve it.

## Why it is worth fixing rather than retrying

The window is invisible from the caller's side. Nothing the import declares is wrong, and nothing about the failure says "come back in five minutes" — so the honest options today are to re-dispatch blindly or to check the POM by hand before every batch.

## Possible shapes

- **Resolve `latest` from the artifact the render actually needs.** Have the install action confirm `ee.schimke.composeai.preview.gradle.plugin:<version>.pom` is resolvable from plugins.gradle.org before accepting a resolved version, and fall back to the newest version that is. This puts the check where the knowledge is, and makes `latest` mean "latest usable" rather than "latest released".
- **Gate the release itself** so the GitHub release is only published after the plugin portal has the marker — removes the window rather than tolerating it.
- **At minimum, name the race in the failure.** When plugin resolution fails on exactly the version the run just resolved, say so ("the CLI release is newer than the published plugin — this version is not on plugins.gradle.org yet"), so it is not diagnosed as the imported project's problem.

The first two are real fixes; the third is worth doing regardless, since the misdiagnosis is most of the cost.

yschimke:
Fixes #5031, fixes #5034 — two diagnosis bugs with the same shape: the failure text describes the consumer's project, and the cause is ours.

## Which build the CLI drives (#5031)

`findGradleProjectRoot` answered "which build am I in" with the nearest `gradlew`. A wrapper does not make a directory a build — a `settings.gradle(.kts)` does — so a nested build borrowing its parent repository's wrapper (thunderbird-android's `components/`) was walked straight past and the **enclosing** build was driven instead, reporting the root build's 206-project count and the root build's Isolated Projects failures for a build the user never named.

- The build root is now the nearest ancestor holding a settings file, with the wrapper search kept as the fallback for a build that has no settings file at all (`findGradleWrapperRoot`).
- The wrapper stays what it always was — where the Gradle **distribution** comes from. `GradleConnector.forProjectDirectory` reads it from that directory alone and silently falls back to the Tooling API default, so `GradleConnection` now inherits the nearest ancestor's `distributionUrl` when the build root has no wrapper: exactly the Gradle `../gradlew` would have used.
- The CLI says which build it resolved whenever that is not the directory it was invoked from, once per process, so a mis-resolution is legible instead of arriving as a wall of unrelated configuration failures.

`PinCommand` follows the same root, as the issue suggested — the two are the same directory for every project that has both.

## Naming the publication race (#5034)

Per the issue's follow-up comment, the readiness gate is #5029's business; what remained was the failure text. `Could not find ee.schimke.composeai.preview:…gradle.plugin:1.66.1` arrives as a configuration failure in the *imported project*, and that is where three release rounds of diagnosis time went.

A new `PluginResolutionDiagnosis.kt` recognises our own marker failing to resolve — narrowly: the marker coordinate, a resolution verb, and the version — and prints what the text never said: this is the version's availability, not your project; a release is downloadable as a CLI before the plugin of the same version is resolvable from plugins.gradle.org and Maven Central; wait, pin the previous release, or check the portal URL (given). The version is attributed to the pin that chose it when the failing version is the one this run injected.

Wired into all three places the failure can surface: discovery-failure reporting (it pre-empts the AGP guidance and the per-project list), the Gradle model-query failure path, and build failures via a new optional `failureAdvice` hook on `GradleConnection` — the driver stays ignorant of *why* a string is worth explaining; the CLI supplies the knowledge.

## Test plan

New unit coverage, all passing locally:

- `GradleBuildRootTest` — a nested build with settings and no wrapper is its own root (and the repository's wrapper still supplies the distribution); a subproject resolves to its containing build; Groovy settings count; wrapper-only fallback; null when neither exists.
- `InheritedWrapperDistributionTest` — inherits the ancestor's `distributionUrl` (including the properties-escaped `https\:` form), leaves a directory with its own wrapper alone, null for no ancestor and for an unusable URL.
- `PluginResolutionDiagnosisTest` — the verbatim failure from the issue is recognised at the injected version and not at another; unrelated dependency failures and "failed to apply plugin" messages are not; the version is read out of the coordinate; the guidance names the version, the race, the pin route and the portal URL; attribution is present for a pinned version and absent for one this run did not inject; discovery reporting leads with the explanation and is unchanged for ordinary failures.

Also ran `:cli:test` in full (green) and `ktfmtCheckAll`. Non-visual change, so no render evidence. One pre-existing failure in this sandbox — `DiscoverPreviewModulesIntegrationTest.discovery completes while GradleProject blows up on the poison module` — reproduces identically on a clean `origin/main` checkout here and is unrelated to the diff.

https://claude.ai/code/session_01U2JaJbzzWWJMRnLMwXdWkE

---
_Generated by [Claude Code](https://claude.ai/code/session_01U2JaJbzzWWJMRnLMwXdWkE)_

### DIFF
--- a/api/gradle-preview-driver/src/main/kotlin/ee/schimke/composeai/previewdriver/GradleConnector.kt
+++ b/api/gradle-preview-driver/src/main/kotlin/ee/schimke/composeai/previewdriver/GradleConnector.kt
@@ -545,6 +545,7 @@ class GradleConnection(
 internal fun inheritedWrapperDistribution(
   projectDir: File,
   warn: (String) -> Unit = System.err::println,
+  gradleUserHome: File? = defaultGradleUserHome(),
 ): URI? {
   if (wrapperProperties(projectDir) != null) return null
   var dir: File? = projectDir.parentFile
@@ -554,17 +555,22 @@ internal fun inheritedWrapperDistribution(
         val loaded = Properties().apply { props.inputStream().use { load(it) } }
         val url = loaded.getProperty("distributionUrl")?.trim()?.takeIf { it.isNotEmpty() }
         val resolved = url?.let { wrapperDistributionUri(it, props) } ?: return@runCatching null
-        // The Tooling API's `useDistribution(URI)` carries a URL and nothing else — there is no
-        // way to hand it the `distributionSha256Sum` the wrapper would have verified. Say so
-        // rather than silently dropping a repository's integrity pin: the distribution is taken
-        // from the shared wrapper cache, which is where `../gradlew` put the copy it verified.
-        if (loaded.getProperty("distributionSha256Sum")?.isNotBlank() == true) {
+        val pinned = loaded.getProperty("distributionSha256Sum")?.trim()?.takeIf { it.isNotEmpty() }
+        if (pinned != null && !distributionAlreadyInstalled(resolved, gradleUserHome)) {
+          // The Tooling API's `useDistribution(URI)` carries a URL and nothing else — there is
+          // no way to hand it the `distributionSha256Sum` the wrapper would have verified. So
+          // when the distribution is not already in the wrapper's cache, inheriting the URL
+          // would mean downloading and executing it with the repository's integrity pin
+          // dropped. Refuse instead: the Tooling API falls back to its own distribution, which
+          // is at least not a checksum this build asked for and did not get.
           warn(
-            "compose-preview: ${props.path} pins distributionSha256Sum, which the Gradle " +
-              "Tooling API cannot be given. Using the wrapper's distribution URL as-is " +
-              "($resolved); run ./gradlew once from ${props.parentFile?.parentFile?.parent} to " +
-              "have the wrapper download and verify it first."
+            "compose-preview: not inheriting the Gradle distribution from ${props.path} — it " +
+              "pins distributionSha256Sum, the Tooling API cannot be given a checksum, and " +
+              "${redactedDistribution(resolved)} is not in the wrapper cache yet. Run " +
+              "./gradlew once from ${props.parentFile?.parentFile?.path} (which verifies and " +
+              "caches it), or give this build its own gradle/wrapper/gradle-wrapper.properties."
           )
+          return@runCatching null
         }
         resolved
       }
@@ -585,6 +591,39 @@ private fun wrapperDistributionUri(url: String, props: File): URI? = runCatching
   if (uri.isAbsolute) uri else props.parentFile.toURI().resolve(url)
 }
   .getOrNull()
+
+/**
+ * True when [distribution] is already unpacked in the wrapper's own cache, which means the wrapper
+ * downloaded it and — where a `distributionSha256Sum` was pinned — verified it. Reusing that copy
+ * involves no download, so no unverified bytes reach this build.
+ *
+ * Matched by the cache's directory layout (`wrapper/dists/<name>/<hash>/…` with the `.ok` marker
+ * Gradle writes after a successful unpack), not by recomputing Gradle's internal URL hash — the
+ * layout is stable and public, the hash function is neither.
+ */
+private fun distributionAlreadyInstalled(distribution: URI, gradleUserHome: File?): Boolean {
+  val home = gradleUserHome ?: return false
+  val name = distribution.path?.substringAfterLast('/')?.removeSuffix(".zip") ?: return false
+  val dists = File(home, "wrapper/dists/$name")
+  val versions = dists.listFiles()?.filter { it.isDirectory } ?: return false
+  return versions.any { dir -> dir.listFiles()?.any { it.name.endsWith(".ok") } == true }
+}
+
+internal fun defaultGradleUserHome(): File? =
+  System.getenv("GRADLE_USER_HOME")?.takeIf { it.isNotBlank() }?.let(::File)
+    ?: System.getProperty("user.home")?.takeIf { it.isNotBlank() }?.let { File(it, ".gradle") }
+
+/**
+ * A distribution URL safe to print: userinfo and query string removed.
+ *
+ * A private distribution can carry credentials in either — `https://user:token@host/…` or a signed
+ * `?X-Amz-Signature=…` — and a warning goes to stderr, which in CI means the build log.
+ */
+internal fun redactedDistribution(uri: URI): String = runCatching {
+  URI(uri.scheme, null, uri.host, uri.port, uri.path, null, null).toString() +
+    if (uri.rawQuery != null) "?…" else ""
+}
+  .getOrElse { "(distribution URL withheld)" }
 
 private fun wrapperProperties(dir: File): File? =
   File(dir, "gradle/wrapper/gradle-wrapper.properties").takeIf { it.isFile }
--- a/api/gradle-preview-driver/src/main/kotlin/ee/schimke/composeai/previewdriver/GradleConnector.kt
+++ b/api/gradle-preview-driver/src/main/kotlin/ee/schimke/composeai/previewdriver/GradleConnector.kt
@@ -52,6 +52,7 @@ class GradleConnection(
    * manually wired it in their `build.gradle.kts`. See [autoInjectInitScriptArgs] for the source.
    */
   private val extraArguments: List<String> = emptyList(),
+) : AutoCloseable {
   /**
    * Optional second opinion on a build failure: given everything this connection saw of it
    * (Gradle's captured stderr plus the exception chain), returns one message to print after the
@@ -61,9 +62,15 @@ class GradleConnection(
    * CLI supplies the knowledge, because that is where it lives. Today the CLI uses it to name the
    * publication race behind an unresolvable plugin marker (issue #5034), which otherwise arrives as
    * a configuration failure in the consumer's own project and is diagnosed as one.
+   *
+   * A settable property rather than a constructor parameter **on purpose**: this module is a
+   * published library, and adding a defaulted parameter would change the primary constructor's JVM
+   * descriptor and its synthetic defaults constructor — every consumer compiled against the
+   * previous release would get a `NoSuchMethodError` on `GradleConnection(…)`, whether or not they
+   * ever wanted failure advice. Adding a property only adds methods.
    */
-  private val failureAdvice: (String) -> String? = { null },
-) : AutoCloseable {
+  var failureAdvice: ((String) -> String?)? = null
+
   companion object {
     /**
      * Wall-clock budget a Gradle invocation gets before it is cancelled.
@@ -335,7 +342,7 @@ class GradleConnection(
       }
     }
 
-    failureAdvice((captured + "\n" + messages.joinToString("\n")).trim())?.let {
+    failureAdvice?.invoke((captured + "\n" + messages.joinToString("\n")).trim())?.let {
       System.err.println()
       System.err.println(it)
     }
@@ -535,7 +542,10 @@ class GradleConnection(
  * Only the ancestor case is interesting: a build root with its own wrapper is already handled by
  * `forProjectDirectory`, and overriding it here would change which Gradle drives it.
  */
-internal fun inheritedWrapperDistribution(projectDir: File): URI? {
+internal fun inheritedWrapperDistribution(
+  projectDir: File,
+  warn: (String) -> Unit = System.err::println,
+): URI? {
   if (wrapperProperties(projectDir) != null) return null
   var dir: File? = projectDir.parentFile
   while (dir != null) {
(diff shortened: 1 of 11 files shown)


yschimke:
Round on the Codex review, plus the one red check. Pushed as f5e2ca9 (the review-threads API is rate-limited for this account right now, so this is one comment rather than per-thread replies).

## Fixed

- **P1 · Preserve the published constructor's JVM signature.** Correct — `:gradle-preview-driver` is published, and a defaulted parameter changes both the primary constructor descriptor and the synthetic defaults constructor. `failureAdvice` is now a settable `var` on the class instead, so the constructor is byte-for-byte what it was and the change only adds methods. The CLI sets it with `.apply { failureAdvice = ::buildFailureAdvice }`.
- **P2 · Require marker failures to dominate before hiding the list.** Fixed. The marker guidance still leads, but it only *replaces* the per-project list when marker failures account for at least half the collection — the same bar `agpClassloaderGuidance` uses. Below that bar the guidance prints and the full list follows. New test: `a lone marker failure among unrelated ones does not hide them`.
- **P2 · Narrow the publication-race matcher.** Fixed. It now requires a missing-artifact verb — `Could not find`, or the plugins DSL's `could not resolve plugin artifact` — and rejects a message naming a transport or repository-configuration cause (`no repositories are defined`, connection/TLS/proxy/auth failures). The generic `Could not resolve` / `Cannot resolve external dependency` forms no longer match on their own. Two new tests cover the no-repositories and 407-proxy shapes.
- **P2 · Parse the complete plugin version.** Fixed. The version now runs to the first character that can't be part of a coordinate rather than a numeric-prefixed subset, so `dev-SNAPSHOT` and `1.2.3+build` are named correctly in both the message and the availability URL. The only trim left is a trailing `.`, which is Gradle's sentence punctuation.
- **P2 · Resolve relative wrapper distribution URLs.** Fixed — a relative `distributionUrl` resolves against the directory holding `gradle-wrapper.properties`, exactly as the wrapper does, instead of being discarded and leaving the Tooling API on its own default Gradle.
- **P2 · Preserve the inherited wrapper's checksum.** Half-fixed, and worth saying which half. The Tooling API's `useDistribution(URI)` carries a URL and nothing else — there is no API to hand it a `distributionSha256Sum`, and `useInstallation` / `useGradleVersion` don't help either. Dropping the pin silently was the real problem, so an inherited wrapper that pins one now warns, naming the file and the URL being used, and pointing at running `./gradlew` once so the wrapper downloads and verifies the distribution into the shared cache the Tooling API then reuses. If you'd rather not inherit at all in that case, say so and I'll make it fall back instead — but that reinstates #5031 for exactly the repositories that care most about their toolchain.

## Not a finding

- **P1 · Recreate this commit with the human Git identity.** This is the false positive `AGENTS.md` warns about, and the commit it cites (`78c29833…`) is not in this PR's range at all:

  ```
  f5e2ca9 Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>
  a72019e Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>

  ```

  The `Reject agent attribution` check is green on the head accordingly.

## Red check, not this PR's

**Daemon Harness (android, real renderer)** fails with `S3_5RecompileSaveLoopRealModeTest > recompiled bytecode flows through to the next render`, `IllegalStateException` at `S3_5RecompileSaveLoopRealModeTest.kt:217` (a `renderFinished` with no `pngPath`). The identical failure, same test and same line, is on this PR's base commit `761a7e2` — [run 33722792383](https://github.com/yschimke/compose-ai-tools/actions/runs/33722792383) — and on `61b544d` after it. This diff touches the CLI's Gradle Tooling-API driver and never the daemon's render path, so there is nothing here to root-cause, and no fix exists anywhere yet to port in. I have not re-run it: two independent `main` runs failing the same assertion is not a flake, and a third would only spend the budget. Flagging it rather than sitting on it — if it turns out to be a real daemon regression it wants its own issue.

---

yschimke:
Second Codex round, on `f5e2ca9`. Pushed as 0b55870.

## Fixed — and it was a real hole this PR opened

**P1 · Preserve the checkout boundary for server trust.** Correct, and the reasoning holds exactly as written. `confirmProjectServeHost` measures "is this `GRADLE_USER_HOME` outside the checkout?" against the Gradle root it is handed, and `SharePreviewCommand` hands it `findGradleProjectRoot()`. Narrowing that to the nearest settings file moved the boundary *below* the checkout: for a repository driven at a nested build (`components/`), the repository's own committed `.gradle/gradle.properties` stopped being "inside", so one branch writing the same `composePreview.serveUrl` into both files would have confirmed itself — and `share-preview` would upload renders and a repository-scoped GitHub token to that host. Before this PR the wrapper search put the root at the checkout, so this is a regression the PR introduced, not a pre-existing gap.

The boundary is now the **VCS checkout root** — `findVcsCheckoutRoot`, the nearest ancestor holding a `.git` entry, matched as a file as well as a directory so worktrees and submodules count — falling back to the build root when there is no checkout. It is a defaulted `checkoutRoot` parameter on `confirmProjectServeHost`, so both call sites keep working unchanged and any future caller gets the safe boundary without having to know about this. New test: `a nested build root does not put the checkout's own gradle home outside it`, which is the attack in the finding, asserting `NeedsConfirmation`.

Worth saying plainly: this is the second boundary the root change touched, after the wrapper/distribution one. If there is a third consumer of `findGradleProjectRoot` where "the build" and "the checkout" being the same directory is load-bearing, it's worth a look — I checked the other four call sites (`InitScriptCommand`, `PinCommand`, `Command.findProjectRoot`, and `resolveProjectServeUrl`'s own read) and none of them makes a trust decision, but that is my reading, not a proof.

## Not a finding (third time)

**P1 · Recreate this commit with a human Git identity.** The cited commit `a1499218…` is not in this PR, and the range in the comment ends at it rather than at the head under review (`f5e2ca9`). On the actual range:

```
0b55870 Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>
f5e2ca9 Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>
a72019e Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>

```

`Reject agent attribution` is green on the head. This is the finding `AGENTS.md` singles out as the most-reported and least-real on this repository; it has now been raised on three separate heads, each time naming a commit that is not in the PR.

## Test plan

`:cli:test` in full and `ktfmtCheckAll`, both green. The Daemon Harness stand-down from my earlier comment still stands unchanged.

---

yschimke:
Third Codex round, on `0b55870`. Pushed as f9129da.

## Fixed

- **P2 · Refuse inherited distributions whose checksum cannot be verified.** Right, and it corrects my previous answer: a warning does nothing for the invocation that is happening *now*. The inheritance is now **refused** when the ancestor wrapper pins `distributionSha256Sum` — with one exception that is not a loophole: when the distribution is already unpacked in the wrapper's own cache. There the wrapper downloaded and verified it, reusing it fetches nothing, and refusing would cost the correct Gradle for no security gain. Detected by the cache's public directory layout (`wrapper/dists/<name>/<hash>/…` plus the `.ok` marker), not by recomputing Gradle's internal URL hash, which is neither public nor stable. Tests cover refuse-when-not-cached and inherit-when-cached.

  The cost is real and worth naming: a repository that pins a checksum, has a nested wrapper-less build, and no warm cache goes back to the Tooling API's default Gradle — i.e. #5031's symptom, but loudly and with instructions, rather than an unverified download. The message names the two ways out (run `./gradlew` once at the ancestor, or give the nested build its own wrapper properties).

- **P2 · Redact credentials from the inherited-wrapper warning.** Fixed — userinfo and query string are stripped before the URL is printed, keeping the host so the message stays useful. A private distribution can carry a token in either, and a warning on stderr is a warning in the CI log. Tests assert a `user:token@host/…?sig=…` URL leaks neither part.

- **P2 · Wire the publication diagnosis into `doctor`.** Correct and a genuine gap: `DoctorCommand` builds its own `GradleConnection` and formats `model.failures` itself, so it kept emitting the generic wrapper/cache remediation during the window — for the one command people run *because* something is wrong. It now installs the same `failureAdvice` on its connection and applies the diagnosis to both the model-access failure (whose check message names the unresolvable version, with `compose-preview pin <previous-version>` as the command) and the per-project discovery failures.

## Not a finding (fourth time)

**P1 · Recreate this commit with the human Git identity**, now citing `a4e95f6b…` — again a commit that is not in this PR, and again a range that does not end at the head under review. On the real range every commit is `Yuri Schimke <yuri@schimke.ee>` and `agent-attribution-scan.sh --range 'origin/main..HEAD'` prints `No agent co-author trailers or agent commit identities found`; `Reject agent attribution` is green on the head. Four heads, four different phantom commits — `AGENTS.md` asks reviewers to run the detector on the real range before reporting this, and it is worth the reviewer configuration doing so.

## Test plan

`:cli:test` in full, `:gradle-preview-driver:test --tests '*InheritedWrapperDistributionTest'`, and `ktfmtCheckAll` — all green. Daemon Harness stand-down unchanged.

---

PR #5057, opened 2026-09-03
docs(parity): record that the viewer now emits a parity locator

### LINKED ISSUE #5000: The viewer's report form emits no parity locator, so issues filed from a preview page never reach parity/issues.json
> **Corrected.** The first version of this issue claimed #3802 had not landed or had regressed. That was wrong, and the correction narrows the bug rather than removing it: the locator writer exists and works — the **focused comparison** emits it. The **viewer** does not, and the viewer is the form reachable from every preview page and every catalog grid card.

## The gap

`/{system}/compare/{previewId}` prefills a complete, valid locator. Verified live on `preview.coo.ee` (serve `2.16.0`):

```compose-parity-locator/v1
repository: yschimke/m3-catalog
system: m3-catalog
component: DatePicker/Modal
preview: datepicker-modal__ideal__default__light__compact
reference: datepicker-modal__ideal__default__light__compact
variant: ideal/default/light/compact
overrides: {}
revision: yschimke/m3-catalog@design-artifacts/m3-catalog
```

`/{system}/p/{id}` prefills the `What's wrong` / `Screenshot` / `Which preview` sections and stops at the `[Open this preview]` link. The string `locator` does not occur anywhere in that page — nor on the catalog landing page, whose cards carry the same form. `parity/issues.json` is built from the locator, so an issue filed from either of those never reaches the index and the component's card never badges.

This matches D2 in `docs/design/parity-batches/01-locator-and-report.md` — *"move the form to the focused comparison (D2, recommended)"* — landing on the comparison. What the batch does not appear to have settled is what the viewer's form should then do. It is still there, still files issues, and still tells the reporter:

> Applied as a `parity:` label, so the catalog's issue index can tell a difference that is ours from one that is not.

which is not true of a report filed from that surface.

## Evidence it bites

- [yschimke/m3-catalog#256](https://github.com/yschimke/m3-catalog/issues/256) — filed from the viewer form on 2026-09-01, open, `parity:upstream`, table but no locator.
- [yschimke/m3-catalog#269](https://github.com/yschimke/m3-catalog/issues/269) — same.
- `parity/issues.json` on `yschimke/m3-catalog@design-artifacts/m3-catalog`, `generatedAt 2026-09-02T20:02:38.049Z`: 17 entries across issues 40, 41, 42×3, 54, 87, 89, 91×5, 93×2, 121, 122. Neither #256 nor #269 is in it. Every entry that is present comes from a hand-written locator.

The failure is silent at every step: the form gives no hint, the filed issue looks complete, and `buildIssueIndex` skips a body with no locator by design (`NO_LOCATOR`), which is right for ordinary issues and indistinguishable from this.

## Options

The batch's own constraints rule out the naive fix — the viewer's overrides are live, substituted by `refreshReportLink()`, so a server-rendered locator would record the default variant while the embedded screenshot shows whatever the reporter dialled in. But the two ends of that are both fine:

1. **Emit the locator when the viewer is at the served defaults**, and substitute the override-dependent fields client-side (or omit the block) when it is not. Both #256 and #269 were filed from an unmodified frame, so this alone would have indexed them.
2. **Route the reporter to the comparison** when the preview has a reference, so the surface that carries the locator is the one they file from.
3. If neither is wanted, **stop promising the index on that surface** — reword the classification note so a reporter is not told their label feeds an index it cannot reach.

## Repo

`settings.gradle.kts` records that the server moved out (*"The preview server is no longer built here. It lives in yschimke/compose-preview-server"*, extraction #4732), so the form code is presumably there now and this may want refiling. Keeping it here for the moment because the consumer half — `scripts/design-artifacts/parity-issues.mjs` — and the batch design docs are in this repo. Move it if that is the wrong call.

## ~~Secondary question: may `preview` and `reference` differ?~~ Answered

Yes. I asked this because the 17 indexed entries all have them equal, and I had written a locator where they differ ([m3-catalog#271](https://github.com/yschimke/m3-catalog/issues/271): the affected cells are `@OverrideVariant`s with no kit node, so they are published previews absent from `references/index.json`). Running that body through `parseLocators` accepts it, and nothing in `parity-issues.mjs` couples the two fields — the duplicate checks are per-field, and `issuesForPreview` matches by component as well as preview id. No change needed; noting it since the locator contract does not spell it out.

yschimke:
The docs-and-evidence half of #5000. The code fix is
[yschimke/compose-preview-server#220](https://github.com/yschimke/compose-preview-server/pull/220),
which owns the server since extraction #4732.

## The bug

`parity/issues.json` is built from the `compose-parity-locator/v1` fence, and only the focused
comparison emitted one. The **viewer** — the report form on every preview page and every catalog
grid card — filed a complete-looking issue with a `parity:` label and no fence, so `buildIssueIndex`
skipped it as `NO_LOCATOR` and the component's card never badged. Silent at both ends, and the
form's own note went on telling the reporter that their label fed that index.
[m3-catalog#256](https://github.com/yschimke/m3-catalog/issues/256) and
[#269](https://github.com/yschimke/m3-catalog/issues/269) are the two that were lost.

## What this PR changes

Nothing executable — the contract and its errata live here, so the outcome is recorded here.

- **`docs/design/parity-batches/01-locator-and-report.md`** gains an erratum beside the existing
  one: D2 landed the form on the comparison and never settled what the viewer's should do, so it
  stayed, kept filing, and kept promising the index.
- **`docs/design/COMPONENT_PARITY_WORKFLOW.md` §2** now states that both reporting surfaces emit the
  block and that they fill `overrides` differently — the comparison writes the value, the viewer
  leaves an `{{overrides}}` placeholder its own script fills from live control state on the same
  pass that fills `{{render}}`.
- **`docs/design/evidence/viewer-report-parity-locator/`** carries the before/after of the report
  panel and the body it now prefills.

## Visual evidence

The committed page fixture (`serve-viewer.html`), production CSS and JS, with the report disclosure
opened. The locator is what puts **Show this issue on** in the panel — the scope control the
comparison has had since batch 02.

| Before | After |
| --- | --- |
| The viewer's report panel before: Summary, then Where does it belong?, then the submit button | The viewer's report panel after: Summary, then Show this issue on, then Where does it belong?, then the submit button |

## Test plan

Docs and PNGs only; no code paths touched in this repository. The behaviour they describe is tested
in the server PR (`ServeViewerIssueReportRouteTest`, `ServeIssueReportTest`,
`serve-web/test/reportOverrides.test.ts`, and the regenerated `ServeWebFixtureTest` golden the shots
above are taken from).

Fixes #5000

https://claude.ai/code/session_011piAXfYfYUhf6g3Nyq5mZR


yschimke:
Worked all five Codex findings. Three led to a change (pushed as `f33a49d`); two do not reproduce. Consolidated here rather than on five threads because the API is rate-limited right now.

**P2 — "Qualify the fixture command with the server repository": correct, fixed.** `:server` is `compose-preview-server`'s project and this build has no such module. The README now clones that repository first and says why.

**P2 — "Use a longer outer fence": correct, fixed.** The nested locator's closing fence was terminating the outer block, so the rendered README dropped it and set the closing paragraph in code. Four backticks outside now.

**P1 — "Adopt the server release before declaring the locator shipped": the mechanism is right, the remedy is not available.** `composeai-preview-serve` is pinned at `2.15.0` here and the fix is not in any release yet — the server repository's latest tag is 2.18.0, so this pin already lags three minors independently of this change, and bumping to an unreleased version is not a thing this PR can do. What was wrong was the wording, and that is fixed: the README and the batch-01 erratum now say the change is the *server's*, that it reaches a deployed catalog with that repository's next release and this repository's bundled `serve` only when `composeai-preview-serve` is bumped past 2.15.0 (Renovate's job), and that the stale fixture copy under `preview-server/preview-harness/fixtures/pages/` moves with the version rather than with this commit. `Fixes #5000` stays: the issue is about the server's behaviour and the server is where it is fixed — flagging the judgement rather than burying it.

**P1 — "Track the viewer's selected reference in the locator": does not reproduce.** The premise is a viewer control that selects a different *design reference*; there is none. The spec lane's picker chooses between `kit` — the imported reference — and `parallel`, which is another **catalog's render**, not a design reference of this preview (`serve-web/src/spec/sources.ts`, `KIT_SOURCE`; `ServeHttpServer.parallelSpecSource`). The viewer's stage, its `referenceAnnotations` and `ServeFigmaSpec` all already resolve `designReferencesFor(preview.id).firstOrNull()`, documented in place as *"a preview with several has one canonical spec, and the manifest's order is the producer's own"* — so the locator names exactly the reference the page is showing, by the same rule the page shows it. Where the preview has no reference of its own and the lane borrows a paired sibling's spec, `designReferencesFor` is empty, `ServeIssueReport.locator` returns null, and no fence is emitted at all — so there is no case where a report is indexed against a reference the reporter could not have been viewing.

The `public-preview-server.md:1718-1724` citation is about `referenceUrl` — embedding the reference **image**, where a server guess would be a claim about pixels the page may not be drawing. That hazard is untouched: `referenceUrl` stays null on the viewer, and its report is still a single render rather than a pair. The locator's `reference:` is identity, not pixels.

**P1 — "Recreate the commit with the human Git identity": does not reproduce.** The commit it names, `82552a5`, is not a valid object in this repository, let alone in this branch. On the real range:

```
97317ca Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>

(exit 0)
```

The `Reject agent attribution` check is green on `97317ca`. This is the pattern `AGENTS.md` documents under *"Reviewers: run the detector before reporting an attribution finding"* — 14 PRs drew this same P1 over 2026-08-31/09-01 and none was real.

---

PR #5058, opened 2026-09-03
fix(actions): make install's `latest` mean latest usable, and name the publication window

### LINKED ISSUE #5051: Imports render against @main, so a release breaks every import until the plugin publishes
`yschimke/compose-preview-imports/.github/workflows/import.yml` calls the pipeline at `@main`:

```yaml
uses: yschimke/compose-ai-tools/.github/workflows/design-artifacts-reusable.yml@main
```

That is deliberate — imports get pipeline fixes the moment they land, without waiting for a release. But it also means the *plugin version the render asks for* bumps the instant a release-please PR merges into main, while the artifact itself is not on the Plugin Portal or Central for some minutes afterwards. Every import dispatched in that window dies during configuration:

```
Could not find ee.schimke.composeai.preview:ee.schimke.composeai.preview.gradle.plugin:1.68.0
Searched in the following locations:
  - https://plugins.gradle.org/m2/.../ee.schimke.composeai.preview.gradle.plugin-1.68.0.pom
  ...
Required by:
    buildscript of root project 'conference-app-2026'
```

The failure is indistinguishable, in the log, from an import that is genuinely broken: it names the upstream project and 30+ of its subprojects, and says nothing about a race. Diagnosing it costs a full log read and a manual POM fetch every time.

`maven-readiness.yml` already knows how to wait for this — it retries a real Gradle resolve up to 100 times (~55 minutes) before attaching the release marker. But it gates only the *release marker*; nothing consults it on the render path, so the imports pipeline has no idea the version it is about to demand does not exist yet.

**This has now happened four times**: 1.64.0, 1.65.0, 1.66.1 and 1.68.0. The 1.68.0 instance was `compose-preview-imports` run 78 (droidkaigi), which had to be re-dispatched once the POM appeared.

Some options, roughly in increasing order of cost:

1. Have the render step wait for the plugin coordinate to resolve before configuring the upstream, reusing the retry that `maven-readiness.yml` already implements. Turns a hard failure into a delay, and needs no change on the caller.
2. Have the pipeline resolve the plugin version from the newest release that carries the readiness marker, rather than from the driver checkout's own version. Keeps `@main` for workflow fixes while pinning the plugin to something known-published.
3. Fail fast and legibly: probe the coordinate in the driver job and stop with an explicit "plugin `<v>` is not published yet, retry in a few minutes" rather than letting Gradle report it as an upstream configuration error.

(3) is worth doing regardless of which of (1)/(2) is chosen, because the current log actively misleads whoever reads it.

yschimke:
Fixes #5051 — the CI half of the publication race (#5052 covers the CLI's own failure text for #5034).

## What changes

**`latest` gates on the Maven-readiness marker.** `maven-readiness.yml` already resolves a release's plugin classpath from public Maven Central for real before attaching `compose-preview-maven-ready-<version>.json`, and the bootstrap installer withholds the CLI until that marker is downloadable. `resolve-version.py` now requires the same marker alongside the CLI tarball, so `latest` means **latest usable** rather than latest released. A release still inside its publication window is skipped with a warning that names it, and the previous release is installed. This is option (2) from the issue, and it needs no change on the caller — an import dispatched during a release cut renders against the last known-published plugin instead of dying.

When *no* release in the search window carries a marker — pre-marker releases, or a readiness job that has been failing across all of them — it falls back to the newest complete release and says so loudly. Refusing to install anything would be worse than the behaviour this gate replaces.

**A legible failure for every other mode.** `pin`, `catalog` and a literal version install exactly what the consumer named, and no marker can help there. After resolving, the action probes the plugin marker POM (portal, then Central) and waits up to `plugin-readiness-timeout` seconds — default 300, `0` disables for a job that never drives Gradle — then fails with an error that names the publication window and what to do about it, instead of letting Gradle report it as a configuration failure in the consumer's project. This is option (1) plus option (3) from the issue: the wait turns a hard failure into a delay, and the failure at the end of the budget at least says what happened.

A registry answering `429` or `5xx` is treated as *no verdict* — both registries are probed before that call, so a rate-limited portal cannot hide a Central `200` — because a Central rate limit must never look like an unpublished release.

## Note on the reusable pipeline

`design-artifacts-reusable.yml` installs through a **SHA-pinned** ref of this action (a deliberate security boundary — the workflow must not execute a movable ref). Imports pick this up when that pin is bumped to a commit carrying this change; the pin bump is not in this PR because the SHA does not exist until it merges.

## Test plan

- `python3 .github/actions/install/test_resolve_version.py` — 23 tests green, including four new `LatestReleaseTests` (skips a release with no marker and takes the previous one; takes the newest ready release silently; still skips a release with no CLI tarball; falls back loudly when nothing carries a marker) and two new `ReleaseAssetPredicateTests` (a partially-uploaded marker doesn't count; a marker for another version doesn't count). This file already runs in CI's Actions Script Tests job.
- The probe step was syntax-checked (`bash -n`) and run against a real published version (`1.69.0` → resolvable, exit 0) and a nonexistent one.
- No visual surface, so no render evidence.

https://claude.ai/code/session_01U2JaJbzzWWJMRnLMwXdWkE

---
_Generated by [Claude Code](https://claude.ai/code/session_01U2JaJbzzWWJMRnLMwXdWkE)_


yschimke:
Round on the Codex review, plus the one red check. Pushed as 14c840d (the review-threads API is rate-limited for this account, so this is one comment rather than per-thread replies).

## Fixed

- **P2 · Validate the timeout before arithmetic expansion.** Real, and the sharpest of the three. `plugin-readiness-timeout` is now rejected unless it is digits-only, before the value reaches any `$(( … ))` — Bash evaluates array subscripts recursively there, so `x[$(cmd)]` would have executed `cmd` on the runner, and this is a consumer-facing input a caller could wire to a dispatch or pull-request value. Verified: `plugin-readiness-timeout: 'x[$(touch /tmp/PWNED)]'` now exits 1 with a clear message and no file is created.
- **P2 · Bound probes by the remaining readiness budget.** Fixed. Each `curl --max-time` and each retry sleep is capped by what is left of the budget, so the step can no longer overrun by a probe cycle plus a fixed sleep. Probes floor at 5s because curl reads `--max-time 0` as *no limit*, which would invert the intent.
- **P2 · Verify the plugin classpath before declaring readiness.** Right about the gap: the marker is a `pom`-packaged stub whose only content is a dependency on `ee.schimke.composeai:compose-preview-plugin`, in a different group, so a 200 on it is not proof the buildscript classpath resolves. The step now requires **both** the marker POM and the implementation jar, on either mirror, before it passes.

  I did not go as far as a clean Gradle resolve. `maven-readiness.yml` can afford one because it runs in this repository with a wrapper and a JDK already set up; this action is consumer-facing, runs before the CLI is even downloaded, and would have to synthesise a build and require a JDK to do the same — a much larger contract for a guard whose job is to convert a confusing failure into a legible one. Both artifacts answering is the propagation state that actually bit here, and `latest` no longer depends on this path at all now that it gates on the readiness marker, which *is* backed by a real resolve. Happy to revisit if a partial propagation slips through the two-artifact check.

## Not a finding

- **P1 · Replace the agent-attributed commit identity.** This is the false positive `AGENTS.md` documents, and the commit it names (`0c4c9ec`) is not in this PR:

  ```
  14c840d Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>
  36ead79 Yuri Schimke <yuri@schimke.ee> | Yuri Schimke <yuri@schimke.ee>

  ```

## Red check, not this PR's

**Daemon Harness (android, real renderer)** fails with `S3_5RecompileSaveLoopRealModeTest > recompiled bytecode flows through to the next render`, `IllegalStateException` at `S3_5RecompileSaveLoopRealModeTest.kt:217`. This PR changes `.github/actions/install/**` and two docs — nothing that a daemon render can see — and the identical failure, same test and same line, is on the base commit `761a7e2` ([run 33722792383](https://github.com/yschimke/compose-ai-tools/actions/runs/33722792383)) and on `61b544d` after it. Same stand-down as yschimke/compose-ai-tools#5052: no fix exists anywhere yet to port in, and I have not spent a re-run, because two independent `main` runs failing the same assertion is not a flake. It looks like it wants its own issue.

## Test plan for this round

`python3 .github/actions/install/test_resolve_version.py` — 23 tests green. The probe step was syntax-checked and exercised against a real published version (`1.69.0` → marker **and** implementation jar resolvable, exit 0) and against the injection payload above.

---

yschimke:
Second Codex round, on `14c840d`. All three were right; pushed as accf7cd.

The guard moves out of inline shell into **`plugin-readiness.py`**, unit-tested in `test_plugin_readiness.py` (10 tests, wired into CI's Actions Script Tests). `python3` is already a hard requirement of this action — the resolve step runs `resolve-version.py` — and none of the three fixes was expressible in shell without the result becoming unreadable.

- **Resolve the whole plugin classpath before succeeding.** Fair, and my previous answer was too narrow. Rather than hardcode `preview-discovery` / `daemon-launch-builder` / `gradle-plugin-config` — a list that goes stale the moment the plugin's dependencies change — the guard now **derives** the set from the POMs: marker → implementation → the implementation's *same-version, same-group* dependencies, requiring every POM and jar in that closure. Third-party dependencies and anything pinned to another version are skipped deliberately: OkHttp and friends have been on Central for months, and it is the artifacts published by *this* release, together, that propagate independently. Tests cover a lagging sibling jar and prove `okhttp` and an older-versioned `preview-data-api` are not walked.

  Still not a Gradle resolution, and now with a reason I am happier defending: that needs a JDK and a Gradle distribution, and this action installs neither by design (consumers compose `actions/setup-java` themselves) and runs before the CLI is downloaded. The POM walk covers the same coordinate set for the propagation states that actually strand a consumer.

- **Do not let an inconclusive artifact mask an absent one.** Real bug, fixed. The state machine is per-artifact now: while anything is conclusively 404 on both mirrors, the run keeps waiting and then fails, whatever a neighbour answered. `429`/`5xx` still means "no verdict" rather than "unpublished". Test: marker 404/404 with the implementation at 503 → exit 1, not the warning-pass.

- **Keep every probe within the remaining budget.** Also real — the 5s floor could stack across four probes. Now: every request past the first sweep is bounded by what is actually left, and running out of time raises a distinct `BudgetExhausted` rather than returning the no-verdict pass, so "we ran out of time" can never come back as success. One sweep always happens, with a 5s per-request minimum, because a guard that never checks is worse than a slightly late answer — a `plugin-readiness-timeout` of 1 buys one attempt, not none. That is the one place the wall clock can exceed the input, and it is now documented in the README as such rather than promised away.

## Test plan

- `python3 .github/actions/install/test_plugin_readiness.py` — 10 green, and `test_resolve_version.py` still 23 green. Both run in CI now.
- Live against the real registries: `VER=1.69.0` walks marker → implementation → `preview-discovery` and exits 0; `VER=99.99.99` reports the 404 on both mirrors and exits 1 with the publication-window message.

The Daemon Harness stand-down from my earlier comment is unchanged — same base-branch failure, now also on this head.

---

yschimke:
**CodeQL is red on `accf7cd` and I need one thing to act on it.**

The aggregate `CodeQL` check reports `1 high` under "New alerts in code changed by this pull request". I can't see which alert: the code-scanning alerts are behind the repository's Security tab, which this session's access doesn't reach, and the check run itself carries no annotations I can read.

What I could establish:

- All three analyses **succeeded** — `Analyze (actions)`, `Analyze (java-kotlin)`, `Analyze (javascript-typescript)` — so the red is the alert, not an analysis failure. The check title is the unrelated standing warning about two configurations not found on `main`.
- This PR changes no Kotlin and no JS/TS, and the CodeQL matrix here is java-kotlin / javascript-typescript / actions only — Python is not analysed — so the alert is almost certainly from the **actions** pack, over `.github/actions/install/action.yml` or the one step added to `.github/workflows/ci.yml`.
- Reading the `Analyze (actions)` log gives the query list but not the results, and the previous head `14c840d` did not trip it, so whatever it is arrived with accf7cd.

I'd rather not guess: the plausible candidates in that pack (code injection, env-var/PATH injection, untrusted checkout, missing permissions) call for quite different fixes, and a speculative one would just add noise to a PR that is already three review rounds deep.

**What would unblock me:** the rule id, file and line from [the PR's code-scanning alerts](https://github.com/yschimke/compose-ai-tools/security/code-scanning?query=pr%3A5058+tool%3ACodeQL+is%3Aopen) — pasting the alert text here is enough. I'll keep the PR watched and fix it as soon as I can see it. If it turns out to be one of the pack's known false positives on this layout, I'll say so with the evidence instead.

The Daemon Harness stand-down from the earlier comment is unchanged.

---

Issue #5067, opened 2026-09-03
feat(discovery): let a call site construct a no-arg-constructible required parameter


yschimke:
## What

`ComponentSnippets.callSite` refuses a required parameter whose type has no literal it can write down. Over the 28-component Material 3 surface the compile gate measures, that is now the *only* remaining cause of refusal: **26 of 28 emit, and the 2 that don't are `TextField` / `OutlinedTextField`, both blocked on `state: TextFieldState`.**

`TextFieldState` is in fact constructible with no arguments. From `foundation-desktop-1.11.1.jar`:

```
private   TextFieldState(String, long)
public    TextFieldState(String, long, int, DefaultConstructorMarker)
```

The `DefaultConstructorMarker` bridge is how Kotlin emits a constructor whose parameters all carry defaults, so `TextFieldState()` compiles from source and `TextField(state = TextFieldState())` would too.

So the refusal is not "this cannot be written", it is "the record does not carry enough to know it can be".

## Why it is not a one-liner

`TargetParameter.type` is a **rendered, simple-name** spelling — `"TextFieldState"` — deliberately: it is a scaffolding hint, not a resolvable reference. It carries no package and nothing about constructibility. And `ComponentSnippets` currently guarantees, by construction, that a snippet needs no import beyond the callable:

> **No import beyond the callable.** Every placeholder this object writes is a literal or an empty lambda, so the emitted snippet never needs a second import to resolve. That is a property of the placeholder table, not a coincidence: widening the table means widening the emitted imports too.

Emitting `TextFieldState()` breaks that invariant on purpose, and needs the type's FQN to emit the import.

## Two designs, and the one to avoid

**A hand-authored table** mapping a type name to an expression plus an import. Cheap, and wrong for this codebase twice over: it is exactly the authored table `docs/design/COMPONENT_RECORD.md` keeps trying to delete, and keying on a simple name collides the moment two libraries define one.

**Resolve it at discovery time.** ClassGraph already has the classpath and `ComposableSignature` already reads `@kotlin.Metadata`. For a required parameter with no literal, look up the type and decide whether it is:

- public and non-abstract (and not an interface, enum entry, or `object`);
- callable with **zero Kotlin arguments** — a genuine no-arg constructor, or a primary constructor whose parameters all default, recognised by the `DefaultConstructorMarker` bridge rather than by counting JVM parameters;
- not itself a value class (mangling applies to constructors too).

[...]

PR #5068, opened 2026-09-03
feat(discovery): persist each component's call site in components.json

yschimke:
## Summary

`ComponentSnippets` can print a call site, but only in-process — so the consumer that needs one could not reach it. This puts the answer on the record.

`ComponentRecord.code` now carries either a `call` with its `imports`, or a `refusedReason`:

```json
"code": {
  "call": "Button(onClick = {}, content = {})",
  "imports": ["androidx.compose.material3.Button"]
}
```

```json
"code": { "refusedReason": "no placeholder can be written for required parameter `imageVector: ImageVector`" }
```

## Why persist rather than let the consumer compute it

I checked before assuming. compose-preview-server consumes ~10 published compose-ai-tools artifacts and **`preview-discovery` is not one of them**, so a consumer-side call site meant one of two bad options: take a dependency on a module that drags ClassGraph, ASM and kotlin-metadata in to produce one string, or reimplement the three rules that make a refusal sound — `signatureKnown`, `symbol.receiver`, and the `…Kt`-facade evidence in `symbol.callable`.

A second implementation of a rule that exacting is how two sides of a contract start disagreeing, and [compose-preview-server#208](https://github.com/yschimke/compose-preview-server/pull/208) already says this side should not re-derive it. `codeFor` is a **projection** of `callSite`, never a parallel path, so the persisted answer and the in-process one cannot differ — there is one decision.

It also makes `components.json` answerable on its own: a reader gets the call site without linking the library that produced it.

## The gate now tests what consumers receive

`ComponentCallSiteCompileFunctionalTest` reads `record.code` off the written `components.json` instead of calling `ComponentSnippets` itself. So the Kotlin compiler accepts the exact bytes a consumer gets. A generator that worked in-process while the record persisted something else would have passed the old test and failed every consumer — that gap is now closed rather than assumed shut.

A record carrying no `code` at all is reported as a producer bug, not counted among the legitimate refusals.

## What this unblocks

The consumer side, in two places, neither in this PR:

- **Playground seeding.** `components.json` already travels inside every bundle (`BundlePreviewTask.kt:1571`), and nothing on the server reads it today — I grepped, and the only hit is an unrelated wasm protocol marker. With `code` present, seeding the playground from a component is: read the record, wrap `call` in the `@Preview` the adapter already adds, submit to `PlaygroundCompileService`. No unsolved problem in the path.
- **The ui-builder's `CodeCapability(symbol, imports)`**, which is the same shape this emits.

And the refusal is a **mechanical tier signal**: a component whose call site cannot be printed cannot reach a Compose exporter. That is the gate the exporter currently spells as `EMITTER_IDS`, a hand-written allowlist of 29 component ids that someone has to remember to extend.

What it does **not** unblock is argument binding. A printed call site has placeholders; the exporter needs the user's values. That is still Phase 2, and it is the same wall as [#5067](https://github.com/yschimke/compose-ai-tools/issues/5067).

## Test plan

- `./gradlew :gradle-plugin:preview-discovery:test :gradle-plugin:test --rerun-tasks` — **BUILD SUCCESSFUL**.
- `./gradlew :gradle-plugin:functionalTest` — the **whole** suite, **BUILD SUCCESSFUL in 5m16s**, run alone from a cleared results dir (two runs sharing `build/test-results` is how I broke it on a previous PR).
- New `ComponentRecordsTest` cases: a record carries the printed call site and its import; a component with no writable literal records *why* rather than falling silent; an unread signature refuses instead of printing `Card()`.
- `ktfmtCheck` on all three touched source sets — clean.
- Wire compatibility: `code` defaults to null, so a `components.json` written by an older plugin still deserialises. Null means "not recorded", never "no call site" — that case is `refusedReason`.
- Non-visual change: no render evidence applies.

---

### DIFF
--- a/docs/design/COMPONENT_RECORD.md
+++ b/docs/design/COMPONENT_RECORD.md
@@ -790,6 +790,27 @@ gate over the corpora; retire the cleaner for migrated catalogs.
 > with you. Both fixtures I picked by hand happened to avoid the one construct that
 > breaks the path, and only compiling something real found it.
 >
+> **The call site is persisted, not recomputed.** `ComponentRecord.code` carries what
+> `ComponentSnippets` printed — either a `call` plus its `imports`, or a `refusedReason`.
+> That is a deliberate choice about where the rule lives. The consumer this record
+> exists for (compose-preview-server's Compose exporter and playground) depends on
+> published compose-ai-tools artifacts but **not** on `preview-discovery`, so a
+> consumer-side call site would mean either a new dependency on a module that drags
+> ClassGraph, ASM and kotlin-metadata in to produce one string, or a second
+> implementation of the three rules that make a refusal sound — `signatureKnown`,
+> `symbol.receiver`, and the `…Kt`-facade evidence in `symbol.callable`. A second
+> implementation of a rule this exacting is how two sides of a contract start
+> disagreeing, which is the drift §1.4 already records once.
+>
+> `codeFor` is a projection of `callSite`, never a parallel path, so the persisted answer
+> and the in-process one cannot differ: there is one decision.
+>
+> The compile gate was moved onto the persisted field for the same reason. It now reads
+> `record.code` off the written `components.json` instead of calling `ComponentSnippets`
+> itself, so what the Kotlin compiler accepts is the exact bytes a consumer receives — a
+> generator that worked in-process while the record persisted something else would have
+> passed the old test and failed every consumer.
+>
 > **Measured reach.** Over a 28-component Material 3 surface (buttons, cards, chips,
 > toggles, progress, scaffolding, list and navigation items, dialogs), **26 emit a
 > call site and 2 refuse** — the two being `TextField` / `OutlinedTextField`, whose
--- a/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/ComponentRecord.kt
+++ b/gradle-plugin/preview-discovery/src/main/kotlin/ee/schimke/composeai/discovery/ComponentRecord.kt
@@ -68,6 +68,21 @@ data class ComponentRecord(
   val slots: List<ComponentSlot> = emptyList(),
   val bindings: List<ComponentBinding> = emptyList(),
   /**
+   * The Kotlin call site for this component, printed by `ComponentSnippets` — or the reason there
+   * isn't one.
+   *
+   * Persisted rather than left for a consumer to compute, because the three things that make a
+   * refusal sound ([signatureKnown], [ComponentSymbol.receiver], and the `…Kt`-facade evidence in
+   * [ComponentSymbol.callable]) are producer-side knowledge. A consumer re-deriving the call site
+   * would have to rediscover all three, and a second implementation of a rule this exacting is how
+   * two sides of a contract start disagreeing. It also makes this file answerable on its own: a
+   * reader gets the call site without linking the discovery library that produced it.
+   *
+   * Null only in a record written before this field existed — never "no call site", which
+   * [ComponentCode.refusedReason] says explicitly.
+   */
+  val code: ComponentCode? = null,
+  /**
    * Whether [parameters], [slots] and [ComponentSymbol.receiver] were read from `@kotlin.Metadata`
    * rather than defaulted away.
    *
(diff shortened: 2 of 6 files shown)


yschimke:
@codex review

The first review ended `⚠️ Failed` on `9284ca1` after 40 seconds (08:22:34 → 08:23:14), which reads as the reviewer erroring rather than reporting findings. Retriggering once so the review actually happens — this PR changes a published wire format (`components.json` gains `ComponentRecord.code`), and Codex caught two real bugs in this same generator on #5036, so it is worth having.

---

yschimke:
## `Daemon Harness (android, real renderer)` is red, and it is not this PR's

```
S3_5RecompileSaveLoopRealModeTest > recompiled bytecode flows through to the next render FAILED
    java.lang.IllegalStateException at S3_5RecompileSaveLoopRealModeTest.kt:218
45 tests completed, 1 failed, 22 skipped
```

**The same job fails on `main` at `08eab8144` — this PR's exact base commit** ([run 33731766825](https://github.com/yschimke/compose-ai-tools/actions/runs/33731766825), 08:08 UTC), with the same test, the same line and the same 45/1/22 counts. Every completed `Daemon Harness` run on `main` today failed the same way:

| main commit | | |
| --- | --- | --- |
| `08eab8144` | release 1.70.0 (#5055) | failure ← this PR's base |
| `7b562055` | #5062 | failure |
| `aee0ec49` | #5061 | failure |
| `865eec84` | #5052 | failure |
| `61b544d7` | #5053 | failure |
| `761a7e2b` | release 1.69.0 (#5048) | failure |
| `00d3dfd0` | #5046 | failure |
| `152835c0` | release 1.68.0 (#5038) | failure |
| `60c45e28` | #5035 | failure |
| `92efd7ea` | release 1.67.0 (#5032) | failure |

(The gaps are `cancelled` runs superseded by the next push, not passes.)

**No fix exists to port.** [#5060](https://github.com/yschimke/compose-ai-tools/pull/5060) — "keep the S3.5 clone out of the fixture's static initializer" — was the candidate, and `git merge-base --is-ancestor 312befe1d 08eab8144` confirms it is **already in this PR's base**. The failure survives it unchanged, so whatever #5060 addressed was not this. This branch is level with `origin/main`, so there is nothing to merge in either.

**I have not spent the permitted re-run.** Its purpose is to confirm a failure is not this PR's; a run of the same job failing on this PR's own base commit is stronger evidence than a re-run would produce, so spending it would tell us nothing new.

**Why the diff cannot plausibly cause it.** This PR adds a `code` field to `ComponentRecord` (populated by a pure string projection of an existing function), and changes a functional test to read that field instead of calling the generator. It touches nothing in `:daemon:*`, `:renderer-android`, or the recompile/save loop the failing test exercises. Every other check on this head is green or still running.

Nothing here is actionable inside this PR without widening it into a daemon-harness fix, which would be the wrong place for it. Flagging separately that this is now red on `main` across ~10 hours and a dozen commits, and that the fix intended for it has landed without fixing it.

---

yschimke:
All six review findings were verified as real and are fixed in **[#5073](https://github.com/yschimke/compose-ai-tools/pull/5073)**, opened against `main` because this PR merged before the fixes were pushed. Replying once here rather than on each thread, since they are moot on a merged PR.

| Finding | Verdict | Fix in #5073 |
| --- | --- | --- |
| Escape Kotlin keywords in persisted calls | real | hard keywords backtick-escaped in the call and in **every import segment**; soft keywords (`value`, `data`, `by`) deliberately left alone, with a test |
| Refuse ambiguous merged overloads | real | `ComponentRecord.overloadsCollided` refuses. The disagreement was already being detected in `ComponentRecords` and discarded — it only needed keeping |
| Preserve required opt-ins with emitted calls | real | `code.requiredOptIns` travels with the call; the compile gate now generates `@OptIn(...)` wrappers, so the contract is compiler-checked |
| Refuse inaccessible private targets | real | visibility read from `@kotlin.Metadata`; non-public refuses |
| Refuse calls with uninferred type parameters | real | a declaration with type parameters refuses |
| Distinguish Kotlin scalars from same-named domain types | real | placeholders match `TargetParameter.typeFqn`, never the rendered spelling |

Two notes on the choices, since neither is the obvious reading of the finding.

**Opt-in is the one not answered by refusing.** Refusing would drop much of Material 3 over something a caller fixes with one annotation on the wrapper it already has to write, so the call is emitted and the markers travel beside it. A marker is identified by its own class carrying `@RequiresOptIn`, resolved one level up — not by matching names like `Experimental…`, which plenty of annotations use without gating anything.

**Two defaults are permissive on purpose.** `callableFromAnotherFile` defaults to true and `typeFqn` falls back to the rendered spelling, because a record written before those fields carries neither, and reading "not recorded" as "private" or "not a Kotlin scalar" would silently retract call sites already published.

The common thread is worth recording: every one of these came from deciding off `TargetParameter.type`, which is a deliberately lossy human-readable rendering. That is the same trap `nullable` was added for in #5036, and it recurred — so the qualified classifier is now what placeholder selection reads.

---

PR #5073, opened 2026-09-03
fix(discovery): close six ways an emitted call site could fail to compile

yschimke:
## Summary

Follow-up to [#5068](https://github.com/yschimke/compose-ai-tools/pull/5068), which merged before these fixes were pushed. That PR started publishing `ComponentRecord.code` — a call site labelled as compiling — and the review pass it drew found **six ways that label could be wrong**. All six were real; `main` currently publishes them.

They share one root cause worth naming: `callSite` decided from a **rendered** view of the signature. That is the same trap `TargetParameter.nullable` was introduced for one level down, and it recurred, so placeholders now match the qualified classifier and never the spelling.

| What broke | Now |
| --- | --- |
| ``fun `when`(`is`: String)`` printed `when(is = "")` and an import ending `.when` | hard keywords backtick-escaped in the call and in **every import segment** |
| merged overloads emitted `Chip()`, which resolves to neither | `overloadsCollided` refuses — the signal was already computed in `ComponentRecords` and thrown away |
| a `@RequiresOptIn` component compiled in its own preview, not in a generated wrapper | `code.requiredOptIns` travels with the call; the wrapper applies them |
| a `private` composable was advertised as importable | visibility read from `@kotlin.Metadata`; non-public refuses |
| `fun <T> Picker(items: List<T> = emptyList())` emitted `Picker()` | a declaration with type parameters refuses — omitting every default leaves nothing to infer from |
| `com.example.String` got `""`, because it renders exactly like `kotlin.String` | placeholders match `TargetParameter.typeFqn` |

## The one that is not answered by refusing

Opt-in. Refusing would drop much of Material 3 over something a caller fixes with one annotation on the wrapper it already has to write, so the call is emitted and the markers travel beside it. `code.call` was always an expression needing a `@Composable` wrapper the caller supplies; when `requiredOptIns` is non-empty that wrapper also needs `@OptIn(Marker::class)`.

**The compile gate now generates those wrappers**, so this half of the contract is compiler-checked rather than asserted — `ComponentCallSiteCompileFunctionalTest` imports each marker and emits `@OptIn(...)` above the function.

An opt-in marker is identified by its own class carrying `@RequiresOptIn`, resolved one level up — not by pattern-matching names like `Experimental…`, which plenty of annotations use without gating anything. A marker class outside the scan resolves to null and is skipped: unrecognised, never assumed.

## Two deliberate permissive defaults

`callableFromAnotherFile` defaults to **true** and `typeFqn` falls back to the rendered spelling. A record written before these fields existed carries neither, and reading "not recorded" as "private" or "not a Kotlin scalar" would silently retract call sites already being published — the same backward-compatibility trap the `nullable` field hit in #5036.

## Test plan

- `./gradlew :gradle-plugin:functionalTest --rerun` — **BUILD SUCCESSFUL in 3m42s**, run alone. The `--rerun` matters: a plain invocation reported success in 933ms because Gradle held the task UP-TO-DATE, which is not a test result.
- `./gradlew :gradle-plugin:preview-discovery:test :gradle-plugin:test` — **BUILD SUCCESSFUL**.
- `ktfmtCheck` on all three touched source sets — clean.
- Eight new `ComponentSnippetsTest` cases, one per finding plus the two that must *not* change: a soft keyword (`value`) is left unescaped, and a real `kotlin.Int` still resolves to `0`.
- The 26-of-28 reach is unchanged: no expected component is private, generic or collided, and the gate's vacuity guard still passes.

## Note on sequencing

The fixes were committed before #5068 merged but validated after, so they could not be pushed to that branch. This is a fresh branch from `origin/main` with the commit cherry-picked; the source tree is byte-identical to the one the 5m01s suite passed on, re-validated on the new base.

---

### DIFF
--- a/docs/design/COMPONENT_RECORD.md
+++ b/docs/design/COMPONENT_RECORD.md
@@ -811,6 +811,33 @@ gate over the corpora; retire the cleaner for migrated catalogs.
 > generator that worked in-process while the record persisted something else would have
 > passed the old test and failed every consumer.
 >
+> **Six ways the compile claim was wider than the record could justify.** Persisting
+> the call site drew a review pass over the guarantee itself, and every one of these
+> was real. They share a root cause worth naming: `callSite` was deciding from a
+> *rendered* view of the signature, which is the same trap `nullable` was introduced
+> for one level down.
+>
+> | What broke | Now |
+> | --- | --- |
+> | ``fun `when`(`is`: String)`` printed `when(is = "")` and an import ending `.when` | hard keywords backtick-escaped in the call and every import segment |
+> | merged overloads emitted `Chip()`, which resolves to neither | `overloadsCollided` refuses — the signal was already computed and discarded |
+> | a `@RequiresOptIn` component compiled in its preview and not in a generated wrapper | `code.requiredOptIns` travels with the call; the wrapper applies them |
+> | a `private` composable was advertised as importable | visibility read from metadata; non-public refuses |
+> | `fun <T> Picker(items: List<T> = emptyList())` emitted `Picker()` | a declaration with type parameters refuses |
+> | `com.example.String` got `""`, because it renders exactly like `kotlin.String` | placeholders match `TargetParameter.typeFqn`, never the spelling |
+>
+> The opt-in one is the only one not answered by refusing. Refusing would drop most of
+> Material 3 over something a caller fixes with one annotation on the wrapper it
+> already has to write, so the call is emitted and the markers travel beside it — and
+> the compile gate generates `@OptIn(…)` wrappers, so that half of the contract is
+> compiler-checked rather than asserted.
+>
+> Two of these defaulted permissively on purpose. `callableFromAnotherFile` defaults
+> to true and `typeFqn` falls back to the rendered spelling, because a record written
+> before those fields existed carries neither, and reading "not recorded" as "private"
+> or "not a Kotlin scalar" would silently retract call sites that were already being
+> published.
+>
 > **Measured reach.** Over a 28-component Material 3 surface (buttons, cards, chips,
 > toggles, progress, scaffolding, list and navigation items, dialogs), **26 emit a
 > call site and 2 refuse** — the two being `TextField` / `OutlinedTextField`, whose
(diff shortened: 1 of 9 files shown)


yschimke:
## `Daemon Harness (android, real renderer)` is red, and it is not this PR's

```
S3_5RecompileSaveLoopRealModeTest > recompiled bytecode flows through to the next render FAILED
    java.lang.TypeNotPresentException at S3_5RecompileSaveLoopRealModeTest.kt:402
        Caused by: java.lang.ClassNotFoundException at S3_5RecompileSaveLoopRealModeTest.kt:402
45 tests completed, 1 failed, 22 skipped
```

**`main` at `bb2bff694` — this PR's exact base — fails byte-identically** ([run 33735052685](https://github.com/yschimke/compose-ai-tools/actions/runs/33735052685), 08:45 UTC): same test, same exception, same cause, same line 402, same 45/1/22 counts. This PR's diff touches `:gradle-plugin:preview-discovery` only and nothing in `:daemon:*`.

### The part that is new, and worth someone's attention

**The fix attempt changed the failure mode without closing it.** Earlier today the same test failed differently:

| base | failure |
| --- | --- |
| `08eab8144` (before [#5069](https://github.com/yschimke/compose-ai-tools/pull/5069)) | `IllegalStateException` at **line 218** |
| `bb2bff694` (after #5069 + [#5072](https://github.com/yschimke/compose-ai-tools/pull/5072)) | `TypeNotPresentException` → `ClassNotFoundException` at **line 402** |

So "clone S3.5's `MutableSquare` from the target's own fixture" got past the original assertion and the test now dies later, loading a class that is not on the harness classpath. That reads like the clone is now sourced correctly but the recompiled bytecode is not reaching the classloader the next render uses — which is the behaviour the test exists to prove. Whoever owns #5069 will want to know it moved rather than landed; I have not touched it, since a daemon-harness fix does not belong in a PR about the discovery record.

This is the second distinct failure mode for this test today and `main` has been red on it since at least `92efd7ea` (release 1.67.0, 22:54 yesterday).

### Re-run

Not spent, deliberately. Its purpose is to confirm a failure is not this PR's, and a byte-identical failure on this PR's own base commit is stronger evidence than a re-run could produce.

### Everything else

Green or still running on this head. Local evidence for the diff itself: `:gradle-plugin:functionalTest --rerun` **BUILD SUCCESSFUL in 3m42s** on this exact tree and base, run alone.

---

PR #5081, opened 2026-09-04
feat(discovery): generate a screen that holds state and handlers that write it

yschimke:
## Why

`ScreenGenerator` could express a screen's shape and none of its behaviour. Its own KDoc named the gap — "the project-specific state adapter" — and the consumer in yschimke/compose-preview-server refuses **every document carrying state** for exactly this reason. A builder could export the static half of a screen and nothing else, which is a limit you notice the moment you try to design a UX rather than a picture of one.

The checked-in fixtures already carry state: Jetcaster declares `searchQuery` and `selectedCategory`, Confetti declares `selectedTrack` and `selectedDay`. None of it could be generated.

## What

`ScreenDocument.state` declares variables, `ScreenValue.StateRead` reads one, and `ScreenNode.handlers` binds a parameter to actions that write them.

```kotlin
@Composable
fun Stateful() {
    val expanded = androidx.compose.runtime.remember {
      androidx.compose.runtime.mutableStateOf<kotlin.Boolean>(false) }
    androidx.compose.material3.Button(
      onClick = { expanded.value = !expanded.value },
      content = { androidx.compose.material3.Text(text = caption.value) })
}
```

### Three spelling decisions

**`remember`, not a bare `mutableStateOf`.** Without it the value is rebuilt on every recomposition that touches it — which does not present as a bug in the preamble, it presents as *the state never changing*, the hardest kind of generated-code defect to trace back to its cause.

**`.value`, not a `by` delegate.** `by` needs `getValue` and `setValue` imported, and this generator's rule is that nothing it emits can be shadowed by the package it lands in. Two more imports are two more chances of the conflict the import check already refuses, bought for a spelling nobody reads twice in generated code.

**The type is declared, not inferred from the initial value.** A `String?` initialised to null is indistinguishable from any other nullable, and inferring `Boolean` from `false` would silently forbid a variable meant to widen later. It is also what every read and write is checked against — and, since it is written into `mutableStateOf<…>`, it is checked as a **name** before it is written as source. See the review round below.

### Actions are a closed set, not an expression language

`Toggle` and `Set` each write one declared variable. A generated lambda that can compute is a generated lambda nobody can review, and the whole claim of this generator is that its output is the screen somebody designed.

Handlers are keyed by **source parameter name** — `onClick`, not a wire event name — the same way arguments are, and for the same reason: the generator writes a call site against a recovered signature, so the only name it can bind is the one the function declares. Mapping a catalog's event vocabulary onto parameter names stays the projection's job.

### Ten refusals, each covering source that would compile and be wrong

| refused | because |
| --- | --- |
| a read or write naming an undeclared variable | the file would not compile, and the error would name a symbol the designer never typed — so the message lists what the screen *does* declare |
| a read whose claimed type disagrees with the declaration | |
| a duplicate state name | reported once against the declaration, not once per node that reads it |
| **a state type that is not a qualified Kotlin name** | it is interpolated into `mutableStateOf<…>` from wire data, so a malformed one does not compile and a crafted one closes the call and splices statements into the composable |
| **a state name colliding with a component imported by simple name** | the property captures the component's call site inside the function body, so `Text(...)` resolves to the state — source that compiles and is not the screen anyone designed |
| a handler on a parameter the component does not declare | dropping it silently ships a screen whose button does nothing, and compiles perfectly |
| **a handler on a callback that takes an argument** | `onValueChange: (String) -> Unit` exists to deliver the new value; a generated body that ignores it compiles by accident and silently drops what the control reported |
| **an empty action list** | an empty lambda is a button that looks live and is not — the one reading that hides the mistake instead of showing it |
| toggling a non-boolean | |
| setting a variable to the wrong type | literals are checked by *rendering against the declaration*, since they carry no type of their own — the same rule `argument` already uses |

## The review round

Two of the rows above are fixes to this branch rather than things it shipped with, both found by the Codex review and both worth stating plainly:

**The state type was source and was not checked as one.** This file already carries a `qualifiedName` whose comment names its own trust boundary — *"Everything else in this file asks whether a name can be written; this asks whether it may be called, and the two are not the same question once a document arrives over the wire."* — and `typeFqn` bypassed both checks on its way into `mutableStateOf<…>`.

**The zero-argument refusal was claimed and not implemented.** An earlier version of this description listed it. `acceptsBareLambda("(String) -> Unit")` is true, because `emptyLambda` accepts one top-level parameter — right for a **slot**, where children placed in `content: (RowScope) -> Unit` may ignore the receiver, and wrong for a handler. `acceptsZeroArgLambda` asks the handler's own question and leaves slots alone.

Four further findings are open and untouched, listed in [a comment below](https://github.com/yschimke/compose-ai-tools/pull/5081#issuecomment-5535339539). Two of them are the same missing thing — the declared type drives assignments but not initializers or nullability — and are one coherent follow-up rather than four patches.

## Test plan

`./gradlew :gradle-plugin:preview-discovery:test` — **352 tests, 0 failures**, fourteen of them new.

**The emitted source was compiled against real Compose, not asserted as a string.** Probe files carrying the exact generated spellings — qualified `remember`, qualified `mutableStateOf<kotlin.Boolean>`, `.value` reads, and the multi-statement `onClick` lambda — were built by `:daemon:desktop:compileKotlin` and then removed. That is the check that matters here: a string assertion only proves the generator emits what I expected it to emit, and I have been wrong about what Kotlin accepts more than once in this file's history.

Text-only evidence is correct: this is a code generator's data model and printing rules, with no rendered surface.

## What this does and does not unblock

It makes the generator able to express state. It does **not** by itself let anyone author state in the UI builder — that needs `setStateVariable` / `setEventBinding` on the wire, which is yschimke/compose-preview-contracts#40, plus projection work in the server once both release. The two halves are independent and this is the one that lives here.


yschimke:
Worked the Codex review. Two findings verified and fixed at cda7c9e; 352 tests, three new.

**"Validate state types before interpolating them into source" — correct, and it is the one I should not have missed.** This file already has a `qualifiedName` with a comment naming its own trust boundary — *"Everything else in this file asks whether a name can be written; this asks whether it may be called, and the two are not the same question once a document arrives over the wire."* — and `ScreenState.typeFqn` went straight into `mutableStateOf<…>` without either check. A malformed type produced source that does not compile; a crafted one closes the call and splices statements into the composable, past the `expressionPackages` vocabulary that exists to stop exactly that. Shape-checked now, with a test using a spliced payload and one for the plain unqualified case.

**"Reject handlers for callbacks that take arguments" — correct, and my description claimed otherwise.** The PR body says a parameter that is not a zero-argument function type is refused. It was not: `acceptsBareLambda("(String) -> Unit")` is true because `emptyLambda` accepts one top-level parameter. That is right for a **slot** — children placed in `content: (RowScope) -> Unit` may ignore the receiver — and wrong for a handler, where `onValueChange` exists to deliver the value and a generated body that ignores it compiles by accident while silently dropping what the control reported. `acceptsZeroArgLambda` asks the handler's own question; slots are untouched.

**The commit-identity finding is the one I am not acting on.** `Reject agent attribution` is green on this PR's range, and this repository's `AGENTS.md` says in as many words that this is the most-reported and least-real finding here — fourteen P1s over two days, none real — and that a reviewer must run the detector before reporting one.

**Not yet acted on**, unverified and not disputed:

- state initializers rendered by the untyped path, so `kotlin.Float` with `0.5` emits `0.5` rather than `0.5f`
- nullable state having no `typeFqn` spelling that satisfies both the state-read check and the parameter check
- the shadow check scanning the whole catalog rather than the components this document actually emits
- `handlers` silently taking precedence over `arguments`/`slots` for the same parameter instead of refusing the conflict

The first two look real to me on reading and both are about the same missing thing — the declared type is used for assignments but not for initializers or nullability. That is one coherent follow-up rather than four patches, and it is the shape of change I would rather make against a reviewed base than stack onto this one.

---

yschimke:
Six findings worked across b06a2af, 6f2ec2a and 0f4bdaa. 362 tests, each new check verified to fail without its fix.

**"Restore serialization for Chain values" — right, and the most serious thing on this branch.** Adding `StateRead` directly after `@Serializable` put that annotation on the new subtype and left `ScreenValue.Chain` without one. A sealed subtype missing it is **not a compile error**: it throws when a document carrying that shape is encoded, which is to say in a producer this module's tests never run. So the wire format was broken for every modifier expression and every check here stayed green.

The annotation is back, and so is the check that would have caught it. `ScreenDocumentSerializationTest` round-trips one document holding every `ScreenValue` shape and both `ScreenAction`s; without the annotation both tests fail with `SerializationException`. `ScreenDocument`'s KDoc calls it wire data throughout and nothing exercised it as such — that gap was the real defect and the missing annotation was its first symptom.

**"Track roots of components qualified in receiver scopes" — right, and it is a hole in the check I added two commits earlier.** I reserved the roots of components that *cannot* claim a simple name, but `node` writes a simply-importable one qualified too when it sits in a receiver scope, where an import would not reach it. Every component's root is reserved now.

**"Escape accepted state type names before emission" — right.** `isQualifiedName` allows a hard keyword and a segment holding a space; both are legal Kotlin once backticked and neither is legal as written, and the type goes straight into `mutableStateOf<…>`. It now goes through `escapeCallableIfKeyword`, the same escaping every other qualified name here already gets. A segment containing a literal backtick stays refused, since that is not escapable.

Earlier in the same batch: a handler bound to a composable slot (`@Composable` lives in `composableSlot`, not `type`, so it satisfied the zero-argument gate and would have written state during composition — a screen recomposing forever); a state name shadowing a package root the file writes in full; and an initializer reading state declared after it, or itself.

**Not acted on: the commit-identity finding.** `Reject agent attribution` is green on this PR's head, and this repository's own guidance names that gate the authority and this finding the one to skip. I have not re-run the scanner on the range and will not.

**Still open from your earlier review**, both the same shape and both real — a composable expression reaching a non-composable position: an initializer evaluated inside `remember`'s `@DisallowComposableCalls` lambda, and a `Set` value rendered inside an event callback. They want the same treatment (hoist while composing, or refuse), and I would rather do them together than half now.

---

yschimke:
Both remaining findings are done, at 074e1210 and 13da07e1. **366 tests.** Two corrections to the description above, which is now out of date:

- it says **352 tests**; it is 366, twenty-eight of them new
- it says *"Four further findings are open and untouched"* — none are. All eight from the review rounds have been worked, and every check added was run against the unfixed code first and observed to fail there.

**"Evaluate composable initializers outside remember" — right.** `remember`'s calculation lambda is `@DisallowComposableCalls`, and this vocabulary can name a composable read; `MaterialTheme.colorScheme.primary` is legal one line above the lambda and rejected inside it. An expression naming an external API — reference, construct, chain — is now bound to a local first and the lambda closes over the binding:

```kotlin
val tintInitial = androidx.compose.material3.MaterialTheme.colorScheme.primary
val tint = androidx.compose.runtime.remember {
  androidx.compose.runtime.mutableStateOf<androidx.compose.ui.graphics.Color>(tintInitial) }
```

A literal and a state read stay where they are — one is a constant, the other reads a local `MutableState`, neither can be a composable call, and hoisting every `""` would double an ordinary preamble to guard against nothing. The binding takes the state's own name plus `Initial`, bumped with underscores until it collides with nothing, because a screen may legitimately declare both `tint` and `tintInitial`.

**"Hoist composable action values out of event callbacks" — right, and I refused rather than hoisted.** The preamble can hoist because it *has* a composable scope to hoist into. A handler is emitted inside the tree with nowhere to put a binding, and inventing a mechanism to lift a value out of an arbitrary node into the enclosing composable is a much larger change than the defect warrants. `Set` from a reference, construct or chain is refused with a reason; a literal and a state read still work, which is the whole vocabulary a handler had business writing. Nothing that worked is lost — the alternative was a screen that looked generated and did not compile.

If you would rather have the hoist than the refusal, say so and I will do it as its own change; it wants a way for `node()` to contribute preamble lines, which is worth designing once rather than bolting on here.

One note on the shape of this round. Two of today's eight findings were defects that **compiled cleanly and no test caught** — the missing `@Serializable`, and a stale generated fixture in the sibling repo. Both are now covered by a check rather than by my attention, which is the part I would want a reviewer to weigh: the fixes matter less than the fact that neither would have been caught again.

---

PR #5093, opened 2026-09-04
feat(discovery): recover a parameter knob's literal default from the class file

yschimke:
The half of the declaration gap that looked hardest, and turned out to be tractable.

A parameter knob has never carried its **default value**, and that is what stops a migrated preview from being editable: `PreviewOverrideDeclaration` holds `default` and `current`, which is how a viewer shows what a field contains before anyone touches it and how it offers "reset". The `previewOverride*` surface gets both free because the author passes the default at the call site. A parameter knob's is compiled away — Kotlin metadata records only *that* a default exists.

## It isn't gone, it's inlined

The Compose compiler emits no `name$default` bridge whose constant pool could be read. It inlines each default expression into the function body, behind a bit of the synthetic `$default` mask. Straight from `samples/cmp`'s reference preview:

```
307: iload         8              // the $default mask
309: iconst_1                     // 1 shl 0
310: iand
311: ifeq          317
314: ldc           #53            // String Shopping list
316: astore_0                     // title
317: iload         8
319: iconst_2                     // 1 shl 1
320: iand
321: ifeq          328
324: ldc2_w        #54            // long 4281558783l
327: lstore_1                     // accentArgb
```

`PreviewKnobDefaults` matches exactly that six-instruction shape, checking **both** the guard bit and the store slot against the parameter's own position. `previews.json` now carries the value:

```json
{ "name": "title",      "index": 0, "type": "STRING", "default": "Shopping list" },
{ "name": "accentArgb", "index": 1, "type": "LONG",   "default": "4281558783" },
{ "name": "itemCount",  "index": 2, "type": "INT",    "default": "3" }
```

## What it refuses, and why that's the point

**Only a lone constant push counts.** `stringResource(...)`, `Color(0xFF3366FF)`, `Modifier` and `itemCount + 1` all come back as *no default* rather than a guess — a viewer showing an absent default knows to say nothing, while one showing an invented value tells the reader the preview does something it does not. The refusal is structural: the same check that recognises the shape rejects everything else, so it isn't a list of expressions someone has to remember to extend.

Two decoys it has to survive, both pinned by tests:

* **The dirty-bits block.** Compose reads the same mask under the same guard at the top of every defaulted composable (`iload mask; push bit; iand; ifeq …; iload dirty; push 6; ior; istore dirty`). A matcher keyed only on the guard would report `6` as the parameter's default. Requiring a *constant* immediately after the branch is what excludes it.
* **Two defaulted overloads of one name.** Discovery records the function name only, so nothing here can tell which one carries the `@Preview`. Refused outright rather than guessed between — the same posture `renderer-android`'s resolver already takes for the same reason.

`iconst_1` is `"true"` for a `Boolean` and `"1"` for an `Int`; the instruction is identical and the declared type is the only thing that says which the author wrote, so the constant is checked against the parameter's type before it's rendered.

## Test plan

Non-visual — discovery output only, no rendered surface changes.

* `:gradle-plugin:preview-discovery:test` and `:gradle-plugin:test` — **869 tests, 0 failures**.
* `PreviewKnobDefaultsTest` (7 cases) assembles its fixtures with ASM rather than writing them in Kotlin. This module has no Compose plugin, so a Kotlin fixture would compile to the ordinary `name$default` bridge and prove nothing about the pattern — and hand assembly is what lets the dirty-bits decoy be byte-exact. Covers every seedable kind, the decoy, expression defaults, a literal alongside an expression, a bit/slot mismatch, a non-Compose tail, and the ambiguous overload.
* `DiscoveryFunctionalTest.composePreviewDiscover records a parameter knob's literal default` compiles a **real** `@Preview` with the Compose plugin CI resolves and reads the manifest back, including a `computed: Int = itemCount + 1` that must come back null. That is the guard against the unit fixtures being precise about a shape the compiler has stopped emitting.
* Verified end to end against the real sample: `:samples:cmp:composePreviewDiscover --rerun` recovers all five of `ParameterKnobListPreview`'s defaults, matching source exactly.
* `ktfmtFormat` on every touched module.

## What's left of gap 1

The publish itself: turning these knobs into `PreviewOverrideDeclaration`s on the render, so the bundle sidecar, `data/fetch?kind=compose/overrides` and the viewer's controls pick them up through the channel `previewOverride*` already uses. That's the next PR, and it is the last thing between the samples and a real migration. `docs/design/PARAMETER_KNOB_MIGRATION.md` is updated to say so.

---

### DIFF
--- a/daemon/core/api/core.api
+++ b/daemon/core/api/core.api
@@ -931,10 +931,11 @@ public final class ee/schimke/composeai/daemon/PreviewIndexKt {
 
 public final class ee/schimke/composeai/daemon/PreviewInfoDto {
 	public static final field Companion Lee/schimke/composeai/daemon/PreviewInfoDto$Companion;
-	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lee/schimke/composeai/daemon/PreviewParamsDto;Ljava/util/List;Ljava/util/List;Lee/schimke/composeai/data/overrides/OverrideVariantSpec;)V
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lee/schimke/composeai/daemon/PreviewParamsDto;Ljava/util/List;Ljava/util/List;Lee/schimke/composeai/data/overrides/OverrideVariantSpec;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lee/schimke/composeai/daemon/PreviewParamsDto;Ljava/util/List;Ljava/util/List;Ljava/util/List;Lee/schimke/composeai/data/overrides/OverrideVariantSpec;)V
+	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lee/schimke/composeai/daemon/PreviewParamsDto;Ljava/util/List;Ljava/util/List;Ljava/util/List;Lee/schimke/composeai/data/overrides/OverrideVariantSpec;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()Ljava/lang/String;
-	public final fun component10 ()Lee/schimke/composeai/data/overrides/OverrideVariantSpec;
+	public final fun component10 ()Ljava/util/List;
+	public final fun component11 ()Lee/schimke/composeai/data/overrides/OverrideVariantSpec;
 	public final fun component2 ()Ljava/lang/String;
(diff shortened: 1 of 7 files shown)


yschimke:
`Module Unit Tests` failed here on `:daemon:core:checkKotlinAbi`, and it is **not this PR's** — this PR touches `preview-discovery` and the plugin, no daemon code at all.

It is red on `main` too. Confirmed rather than inferred:

```
$ git checkout --detach origin/main
$ ./gradlew :daemon:core:checkLegacyAbi
> Task :daemon:core:checkKotlinAbi FAILED
  +public final class ee/schimke/composeai/daemon/PreviewKnobDto { …
```

The cause is mine: #5089 added public API to `:daemon:core` (`PreviewKnobDto`, `PreviewKnobSeeds`, `PreviewKnobToken`, a `knobs` component on `PreviewInfoDto`) without regenerating the reference dump, and #5091 widened the same surface without regenerating it either. Both merged because `Module Unit Tests` is not a required check for auto-merge, so the failure landed on `main` instead of blocking the PR that caused it.

Fix opened as #5094, and **ported into this PR** (`61aac1be7`) rather than waiting on it — it no-ops once `main` carries the same regeneration. `:daemon:core:checkLegacyAbi` passes on this head, and `checkLegacyAbi` across the whole build shows `:daemon:core` was the only stale dump.

No re-run spent; the push above is the fix.

---

PR #5137, opened 2026-09-05
feat(render-host): bring the render host home from the preview server

yschimke:
Step 2 of [compose-preview-server#180](https://github.com/yschimke/compose-preview-server/issues/180), and the first half of a two-repository handover. Follows [#5129](https://github.com/yschimke/compose-ai-tools/pull/5129) (the layer rule) and [#5135](https://github.com/yschimke/compose-ai-tools/pull/5135) (the gate).

## Why

The render host, the bundle daemon and the git-backed preview history were written inside the `serve` package, so when the server was extracted they went with it. That left `:cli` reaching across a repository boundary for them: `bundle render`, `render matrix`, `history manifest` and the missing-render report are **offline** commands that open no socket, yet depended on an artifact published by the repository that owns the web server.

That is one of the two edges of the cycle, and the half with no reason to exist. The evidence is mechanical: **the module has zero project dependencies inside the server** — its entire dependency block is compose-ai-tools and contracts coordinates. `docs/design/REPOSITORY_LAYERS.md` settles it: layer 1 is behaviour that opens no socket, and this renders.

## What moved

81 Kotlin files (48 main / 17.8k lines, 32 test, 1 test fixture), one package, unchanged.

**Sources are unchanged, package included** (`ee.schimke.composeai.cli.serve`), so the move is source-compatible — call sites don't change, they resolve from a module in the right repository. The package rename stays a separately reviewed change in both repos.

**Dependencies are not unchanged, and that is the point.** Seven published coordinates become project dependencies:

| was | is |
| --- | --- |
| `libs.composeai.preview.data.api` | `project(":preview-data-api")` |
| `libs.composeai.bundle.format` | `project(":bundle-format")` |
| `libs.composeai.bundle.coordinates` | `project(":bundle-coordinates")` |
| `libs.composeai.daemon.core` | `project(":daemon:core")` |
| `libs.composeai.render.session.api` | `project(":render-session-api")` |
| `libs.composeai.render.session.subprocess` | `project(":render-session-subprocess")` |
| `libs.composeai.data.remotecompose.core` | `project(":data-remotecompose-core")` |

The contracts coordinates (layer 0) stay published in both repositories — that direction is what the layer rule allows. `parity-issues-protocol` is added to the catalog; it's the one alias this repo didn't already have.

The render host is now compiled and tested against the source tree it ships beside, instead of against whichever compose-ai-tools release the server happened to pin. **The skew that made #180 worth filing — built against 1.62.0, run against `main` — cannot recur from this side.**

## The coordinate changes: `render-host`, not `compose-preview-render-host`

Keeping the old coordinate would put one artifact on two version lines from two repositories, and a 1.x release sorts **below** the 2.x the server already shipped — a downgrade to every resolver and to Renovate. The new name matches this repository's own convention (`bundle-format`, `daemon-core`, `render-session-api`), and the old coordinate stays resolvable at its final 2.x for anyone pinned to it.

## Two things the move earns

- **The test-fixtures capability workaround is gone.** In the server the project was named `render-host` but published as `compose-preview-render-host`, so `java-test-fixtures` derived a capability that didn't match what `testFixtures(...)` asks for — a block of variant configuration plus a `checkTestFixturesCapabilities` guard to reconcile them. Here the project name and artifactId are the same string. Verified in the published module metadata rather than assumed:
  ```
  testFixturesApiElements      -> ee.schimke.composeai:render-host-test-fixtures
  testFixturesRuntimeElements  -> ee.schimke.composeai:render-host-test-fixtures
  testFixturesSourcesElements  -> ee.schimke.composeai:render-host-test-fixtures
  ```
- **`checkRenderHostIsServerFree` comes along**, because "no web server" is now this repository's *layer-1 test* rather than the server's internal boundary. It is narrower than `checkLayerBoundary`: that one asks where a coordinate comes from, this asks what kind of thing it is.

## What this PR deliberately does *not* do

`:cli` is **not** switched to the new module. It still depends on `compose-preview-serve`, whose POM names the published `compose-preview-render-host`, so wiring `project(":render-host")` in as well would put the same classes from `ee.schimke.composeai.cli.serve` on one classpath twice.

The order is: this lands and releases → the server adopts `ee.schimke.composeai:render-host` and deletes its copy → `:cli` swaps over and two entries leave `CheckLayerBoundary.knownLayerTwoEdges`. Until then the two copies coexist, which is why the window should be short.

## Test plan

- `./gradlew :render-host:compileKotlin` — green. Worth stating plainly: the moved sources compile against compose-ai-tools `main` unmodified, so the pinned-version skew had not yet become a source incompatibility.
- `./gradlew :render-host:test --rerun-tasks` — **452 tests across 33 classes, 0 failures.**
- `./gradlew :render-host:check` — green, including `checkRenderHostIsServerFree` and the new `checkLayerBoundary`.
- `./gradlew :render-host:publishToMavenLocal` — publishes `ee.schimke.composeai:render-host:1.75.1-SNAPSHOT` with jar, sources, javadoc and test-fixtures; capability metadata quoted above.
- `./gradlew :render-host:ktfmtCheck` — green.

Non-visual change; no render evidence applies.

https://claude.ai/code/session_014NtzoUCr4pou3y1qRDxcpZ

---
_Generated by [Claude Code](https://claude.ai/code/session_014NtzoUCr4pou3y1qRDxcpZ)_

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -410,6 +410,9 @@ composeai-data-layoutinspector-core = { module = "ee.schimke.composeai:data-layo
 composeai-data-theme-core = { module = "ee.schimke.composeai:data-theme-core", version.ref = "composeai-contracts" }
 composeai-data-preview-overrides-core = { module = "ee.schimke.composeai:data-preview-overrides-core", version.ref = "composeai-contracts" }
 composeai-common-io = { module = "ee.schimke.composeai:common-io", version.ref = "composeai-contracts" }
+# `ServeHost.parityIssues()` — the shape a catalog publishes for component-parity issues. Contracts
+# (layer 0) owns the wire shape; `:render-host` owns validation and storage over it.
+composeai-parity-issues-protocol = { module = "ee.schimke.composeai:parity-issues-protocol", version.ref = "composeai-contracts" }
 
 # The preview server (see `composeai-preview-serve`). `-test-fixtures` rides the same coordinates
 # as a Gradle feature variant, so `testFixtures(libs.composeai.preview.serve)` picks it up without
--- a/render-host/build.gradle.kts
+++ b/render-host/build.gradle.kts
@@ -1,0 +1,186 @@
+// The render host, the bundle daemon and the git-backed preview history — what renders and reads
+// history, with no web server underneath it.
+//
+// **This module came home.** It was written inside the `serve` package, so when the server was
+// extracted to yschimke/compose-preview-server it went along, and `:cli` had to reach back across a
+// repository boundary for it: `bundle render`, `render matrix`, `history manifest` and the
+// missing-render report are OFFLINE commands that opened no socket, yet depended on a published
+// artifact from the repository that owns the web server. That is one of the two edges of the
+// dependency cycle in yschimke/compose-preview-server#180, and the half with no reason to exist —
+// the module has zero project dependencies inside the server, and its entire dependency block is
+// compose-ai-tools and contracts coordinates.
+//
+// `docs/design/REPOSITORY_LAYERS.md` settles where it belongs and the answer is here: layer 1 is
+// behaviour that opens no socket, and this renders. Nothing about the sources changed in the move.
+//
+// Coordinate change, deliberate: it published as `compose-preview-render-host` from the server and
+// publishes as `render-host` from here. Keeping the old coordinate would mean two repositories
+// publishing one artifact on two version lines — a 1.x release that sorts BELOW the 2.x the server
+// already shipped, which is a downgrade to every resolver and to Renovate. A new coordinate in this
+// repository's own naming (`bundle-format`, `daemon-core`, `render-session-api`) has neither
+// problem, and the old one stays resolvable at its final 2.x for anyone pinned to it.
+//
+// Package note: the sources keep `ee.schimke.composeai.cli.serve`, exactly as they did in the
+// server. The rename is a separately reviewed change in both repositories, and keeping it is what
+// makes this move source-compatible — `:cli`'s call sites do not change at all, they just resolve
+// from a module in the right repository.
+//
+// One simplification the move earns for free. In the server this project was named `render-host`
+// but published as `compose-preview-render-host`, so `java-test-fixtures` derived a capability
+// (`…:render-host-test-fixtures`) that did not match what a consumer's `testFixtures(...)` asks for
+// (`…:compose-preview-render-host-test-fixtures`), and the build file carried a block of
+// configuration plus a `checkTestFixturesCapabilities` guard to reconcile them. Here the project
+// name and the artifactId are the same string, so the derived capability is already the right one
+// and all of that goes away.
+plugins {
+  id("composeai.base-conventions")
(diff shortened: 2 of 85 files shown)


yschimke:
`Actions Script Tests` went red on the first push, and it was right to. Pushed a fix in `796b385`; the failure is worth recording because it is not what it looks like.

`check-daemon-launch-schema.py` refuses a descriptor construction site whose `schemaVersion` stamp does not resolve to a registered constant, and `ServeBundleDaemon` stamps `DAEMON_LAUNCH_SCHEMA_VERSION`:

```
AssertionError: 'DAEMON_LAUNCH_SCHEMA_VERSION' not found in
{'DAEMON_DESCRIPTOR_SCHEMA_VERSION', 'schemaVersion', 'EXPECTED_DESCRIPTOR_SCHEMA_VERSION'}
  : render-host/.../ServeBundleDaemon.kt stamps unregistered DAEMON_LAUNCH_SCHEMA_VERSION
```

**This is not a new violation — it is an old one coming back into view.** The allowlist's own `writerEncoders` note already records that serve's writer *was* registered here until the preview server left in [#4732](https://github.com/yschimke/compose-ai-tools/issues/4732), and that its departure left a gap:

> Serve's writer (`ServeBundleDaemon`, …) was registered here until the server left the repository in #4732; it is now a cross-repo mirror this check cannot see, which is residual item 1 on that issue — a `daemon-launch.json` shape change here no longer fails the server's build, and nothing yet makes it.

Bringing the render host home puts that mirror back in a tree the check can see. So the fix is to **gate it again**, not to exempt it — the move closes part of a hole #4732 recorded and could not close from here.

Registered two things:

- **The schema-version site.** Worth noting the spelling differs from the other two — `DAEMON_LAUNCH_SCHEMA_VERSION`, not `*_DESCRIPTOR_SCHEMA_VERSION` — so the name-based mirror scan never finds it and only discovery-by-use does. Its value is 2, agreeing with the writer, which the check now enforces on every future bump.
- **The writer encoder.** Serve's `Json { encodeDefaults = true }` is deliberately *not* identical to the two Gradle-plugin writers: no pretty-print, and it leans on kotlinx's `explicitNulls = true` default rather than stating it. Recorded rather than normalised, since that block exists to make such differences visible.

The recorded v2 wire fingerprint is unchanged, so registering the encoder did not alter the shape as emitted.

Verified against what CI actually runs, with a `compose-preview-contracts` checkout so nothing skips:

```
$ COMPOSE_PREVIEW_CONTRACTS_ROOT=… python3 scripts/check-daemon-launch-schema.py
check-daemon-launch-schema: OK — 12 writer field(s), 1 reader(s) agree (jvm), 4 schema-version site(s) at v2

$ COMPOSE_PREVIEW_CONTRACTS_ROOT=… python3 scripts/test_check_daemon_launch_schema.py
Ran 43 tests — OK
```

(The earlier local run had 5 skips for the missing contracts checkout, which is why it went green here and red in CI. Worth knowing for anyone else running this script locally.)

One consequence for the plan: residual item 1 on #4732 is **narrowed, not closed**. The writer is gated again, but the server's own deployment image still passes descriptor keys this check cannot see from here.

---

Issue #1546, opened 2026-05-27
Auto-inject only targets flavorless application modules, ignoring library modules and flavored apps


yschimke:
### Summary
When running `compose-preview` on a multi-module project (like Google's `nowinandroid` sample) with no static plugin declarations (relying on the CLI's auto-inject feature), the CLI fails to discover previews located in Android library modules and flavored application modules.

### Details

1. **Library Modules Ignored by Auto-Inject**:
   The built-in auto-inject mechanism only applies `ee.schimke.composeai.preview` to projects that apply the `com.android.application` plugin. In modularized Android projects, `@Preview` composables typically live in library modules (e.g., `:core:ui`, `:core:designsystem`, `:feature:*` using `com.android.library`). Since the plugin is not applied to library modules, their previews are never scanned, resulting in `No previews found`.

2. **Flavored Application Modules Ignored**:
   If an application module does not have a default `debug` variant because it uses product flavors (e.g., `:app` has `demoDebug` and `prodDebug`), it is ignored during the CLI's module discovery phase, even if `--variant demoDebug` is passed. Only flavorless applications (like `:app-nia-catalog` with a default `debug` variant) are detected.

3. **Unable to override via global `init.d`**:
   Writing a custom Gradle init script under `~/.gradle/init.d/` to force-apply the plugin dynamically to library modules doesn't work because the CLI drives Gradle via the Tooling API, which bypasses `~/.gradle/init.d/` scripts.

### Steps to Reproduce
1. Clone the `nowinandroid` repository.
2. Run `compose-preview list` (or `compose-preview list --variant demoDebug`).
3. Observe that it only identifies `app-nia-catalog` as a preview module, and scans 0 previews.

### Suggested Fixes
* Allow the CLI's auto-inject init script to target both `com.android.application` and `com.android.library` modules.
* Fix the module discovery query to handle projects with product flavors/variants correctly (ensuring the `--variant` parameter is respected during the model configuration query).

PR #5170, opened 2026-09-05
ci: retire the integration push lane and file bugs for nightly failures

yschimke:
## Why

The external-repo integration matrix guards against **upstream** drift — a consumer bumping AGP, Kotlin or the Compose BOM — which correlates with time, not with our commits. The push-to-`main` lane never fit that shape.

An earlier change already trimmed the merge lane from 8 external cells to the two `pr_blocking` ones. What was left still ran `build-plugin` + 2 cells + `agp8-min` on **every** merge, under `cancel-in-progress: true`, with `main` merges landing a median 4 min apart — re-deriving signal the PR run had already produced minutes earlier on the same tree. (Measured previously: 72% of push runs were cancelled by the next merge before finishing.)

## What changed

- **`push` is gone from `on:`.** PRs keep their one cheap cell (`wear-os-samples (ComposeStarter)`) plus the change-scope gate; the nightly cron keeps the full matrix. `agp8-min` and `publish-previews` were already gated `!= 'pull_request'`, so they follow to cron/dispatch automatically.
- **`build-plugin`'s BuildFetch `rw-token` re-gated `push` → `schedule`.** It was keyed on an event this workflow no longer receives, so leaving it would have quietly retired a remote-cache writer. The invariant it encodes is unchanged, and is why this is not simply `!= 'pull_request'`: PR-head code must never see a write token, and `workflow_dispatch` can be aimed at a PR ref.
- **New `report-nightly-failures` job.** With no push lane, a nightly failure has no reader. It files **one issue per failing job**, comments on the open issue when the same job fails again, and closes the issue when that job goes green — so the open `integration-failure` set is exactly the set of currently-broken cells. Cron only: a PR failure is already a red check on the author's own PR, and a dispatch is someone watching the run they started. `cancelled` and `skipped` are not failures and are ignored.

Docs updated where they asserted the old cadence: `docs/VERSIONING.md` (§ 9 table and the § 6 line) and the remote-cache writers bullet in `docs/AGENT_GUIDE.md`.

## Effect on required checks

None. The `Protect Main` ruleset's required `wear-os-samples` legs are evaluated on the PR head, and the `pull_request` trigger is untouched — `release-please-gate` and `integration-scope-gate` still re-emit them for release PRs and out-of-scope diffs.

## Test plan

- `yaml.safe_load` on the workflow: parses; triggers are exactly `pull_request`, `schedule`, `workflow_dispatch`; the new job's `if` / `needs` / `permissions` resolve as intended.
- `bash -n` on both new `run` blocks.
- **Reconcile logic exercised against a stubbed `gh`** with a synthetic job list covering all five states. Observed, and correct on each:
  - new failure (`wear-os-samples (ComposeStarter)`) → `gh issue create`
  - repeat failure with an issue already open (`androidify`, #11) → `gh issue comment`, no duplicate issue
  - recovered job with a stale open issue (`nowinandroid`, #22) → `gh issue close --reason completed`
  - `cancelled` and `skipped` jobs → ignored
  - the reporter job's own in-progress entry (`conclusion: null`) → ignored
- Title matching verified against a cell name containing `#1546`, since one matrix cell embeds an issue reference.

Not exercised end-to-end: the real cron run, which is the first time `gh` is called for real. `report-nightly-failures` cannot fail the matrix — it runs after it — so a mistake there costs a missing bug report, not a red nightly.

No UI surface is touched, so there is no visual evidence to embed.

https://claude.ai/code/session_01Q1TLkhcQFjNqeqdkDcKVkS

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q1TLkhcQFjNqeqdkDcKVkS)_


yschimke:
`A11y + notification previews` is red on this PR, and it is not this PR's failure — it is broken on `main`.

**What is failing**

`Install CLI from source` → `./gradlew :cli:installDist`:

```
> Could not resolve all dependencies for configuration ':cli:runtimeClasspath'.
   > Could not find ee.schimke.composeai:compose-preview-render-host:3.0.0.
```

**Why it is not this PR's**

This PR changes `.github/workflows/integration.yml` and two docs. It touches no Gradle wiring and no version catalog.

The base commit is `2ee93ec` — "fix(deps): update dependency ee.schimke.composeai:compose-preview-serve to v3" (#5164), a Renovate automerge. CI run [12353](https://github.com/yschimke/compose-ai-tools/actions/runs/33958786288) failed on `main` at that same commit, before this branch existed.

**Root cause**

`gradle/libs.versions.toml` deliberately shares one version ref between the two artifacts:

```toml
composeai-preview-serve = "3.0.0"
composeai-preview-render-host = { module = "ee.schimke.composeai:compose-preview-render-host", version.ref = "composeai-preview-serve" }
```

The catalog comment justifies that with "a skew between the server and the render host it sits on top of is not a state that repository can produce". `compose-preview-server` [v3.0.0](https://github.com/yschimke/compose-preview-server/releases/tag/v3.0.0) produced exactly that state — it published `compose-preview-serve:3.0.0` but not `compose-preview-render-host:3.0.0`:

| artifact | 3.0.0 pom on Central | latest published |
| --- | --- | --- |
| `compose-preview-serve` | HTTP 200 | 3.0.0 |
| `compose-preview-render-host` | **HTTP 404** | 2.23.0 |

Not CDN propagation: the release published at 08:42Z and the 404 is direct on the pom, over an hour later.

**No fix exists to port.** There is no open PR in this repository addressing it, so there is nothing to bring into this branch. I am not widening a CI-cadence PR into a dependency fix.

**Proposed patch** (for a separate PR): revert #5164, returning `composeai-preview-serve` to `2.23.0`. Renovate will re-propose the bump once `compose-preview-render-host:3.0.0` is on Central. Splitting the version ref would work mechanically but contradicts the deliberate single-ref design, so the real fix is upstream — publish the render host at 3.0.0 from `compose-preview-server`, or yank/repair the v3.0.0 release.

**No re-run.** The one permitted re-run exists to confirm a failure is not this PR's; a direct 404 on the pom plus the same failure on `main` already establishes that more firmly than a re-run would, and a re-run cannot conjure a missing artifact.

Keeping this PR watched until it is green.

---

Issue #2824, opened 2026-07-27
BuildFetch remote cache serves truncated entries (2nd occurrence); composeai.cacheSalt doesn't cover non-Kotlin tasks


yschimke:
Two entries in the BuildFetch remote build cache are **stored truncated** and are served to every consumer under HTTP 200. Any build whose task keys resolve to them fails hard and cannot self-heal.

This is the **second** occurrence in this cache (the first is recorded in the `composeai.cacheSalt` history in `gradle.properties`), and the first one our existing escape hatch does **not** cover.

## Symptom

```
* What went wrong:
Execution failed for task ':cli:compileTestKotlin'.
> Could not resolve all files for configuration ':cli:detachedConfiguration2'.
   > Failed to transform gradle-api-9.6.1.jar to match attributes {artifactType=classpath-entry-snapshot, …}.
      > Execution failed for ClasspathEntrySnapshotTransform: …/generated-gradle-jars/gradle-api-9.6.1.jar.
         > Failed to load cache entry 12fcc978271d75e6470ed675c023a05d for ClasspathEntrySnapshotTransform:
           Could not load from remote cache: Unexpected end of ZLIB input stream
```

```
* What went wrong:
Execution failed for task ':samples:android-screenshot-test:mergeDebugResources'.
> Failed to load cache entry 877aca9ad8277aac3a9adf9a18efc1ec for task ':samples:android-screenshot-test:mergeDebugResources':
  Could not load from remote cache: Unexpected end of ZLIB input stream
```

Both hit on PR #2822 (run [30293634994](https://github.com/yschimke/compose-ai-tools/actions/runs/30293634994), jobs `Build CLI` + `Build Samples`), within one minute of each other, on a two-file diff in `renderers/android` that touches neither `:cli` nor `:samples`.

## Evidence that the stored objects are corrupt

Fetched directly from the cache with the read-only token, from a container on an unrelated network, hours after the CI failure:

| key | task / transform | HTTP | `content-length` | bytes received | inflates to | gzip EOF reached |
|---|---|---|---|---|---|---|
| `12fcc978271d75e6470ed675c023a05d` | `ClasspathEntrySnapshotTransform` (`gradle-api-9.6.1.jar`) | 200 | 868352 | 868352 | 12,170,543 | **no** |
[...]

Issue #4832, opened 2026-08-29
Refactor the render and history types out of cli/serve, so serve can move


yschimke:
## Why

The preview server moved to [yschimke/compose-preview-server](https://github.com/yschimke/compose-preview-server) and ships as `compose-preview-serve` 2.0.0. [#4831](https://github.com/yschimke/compose-ai-tools/pull/4831) retired the split instruments and the deployment lane here, but deliberately left `cli/serve` in place, because removing it takes **three more commands** with it:

| command | coupling |
| --- | --- |
| `browse` | a 62-line wrapper whose body is `ServeCommand(serveArgs(args), browseProject = true).run()` — it dies outright |
| `bundle render` | 18 usages of `ServeRenderHost` / `ServeBundleDaemon` / `RenderOutcome` / `SvgOutcome` |
| `history manifest` | 17 usages of `PreviewHistory` / `PreviewHistoryManifest` |

Those are not really *server* types. They are shared render and history types that happen to live in the `serve` package because that is where they were written.

## What the types actually are

Classified while scoping #4831:

**`ServeRenderHost.kt` is 2,001 lines and mixes both kinds.** Shape: `ServeMotion`, `ServePreview`, `ServeDeviceFrame`, `CatalogRenderFailure`, `RenderFailureFrame`, `ServeTheme`, and the sealed hierarchies `RenderOutcome`, `SvgOutcome`, `SlotsOutcome`, `AnnotationsOutcome`, `A11yOutcome`. Behaviour: `class ServeRenderHost` itself, plus `detectedFeaturesOf`, `declaredThemesFromPreviews`, `inferredThemeMode`. Splitting that file is most of the work.

**`PreviewHistoryManifest.kt`** (269 lines) is `@Serializable` DTOs — shape.

**`PreviewHistory.kt`** (364 lines) shells out to **git** (`core.quotePath=false`, `logArgs`, `java.io.File`) — behaviour, so it cannot go to `compose-preview-contracts` under that repository's "shape, never behaviour" rule.

**`ServeBundleDaemon.kt`** (960 lines) — behaviour.

## The destination question

Two candidates, and the split is not clean:

- **`compose-preview-contracts`** takes shape only. The outcome DTOs and `PreviewHistoryManifest` qualify. But every module there is a published contract with `explicitApi()` and a recorded ABI, so adding them is a real commitment and reaching them requires a contracts release.
[...]

PR #5175, opened 2026-09-05
docs: move settings.gradle.kts narrative into docs/build-scripts/

### LINKED ISSUE #5169: docs: move build-script narrative out of settings.gradle.kts into docs/
Found during a tech-debt sweep of the repository.

## The numbers

- Comments are **27% of non-blank lines** across the 406k lines of Kotlin (210k main / 183k test).
- Comments are **41% of non-blank lines** in the Gradle build scripts.
- `settings.gradle.kts` alone is **851 lines**, the majority of it prose.

## Why the build scripts specifically

`settings.gradle.kts` has become the changelog for the build. It carries multi-paragraph explanations of decisions, including passages that are explicitly historical:

> Historical note on why the snapshot lane exists at all: `androidx.compose.remote.foundation` (which `wear.compose.remote:remote-material3` needs) was, for a long time, published only on androidx-main, so the release line could not resolve the Wear widget layer. It now ships on Google Maven, which is what let the default flip back to `release`. […]

and

> The per-project conventions that used to live in the root build's `allprojects {}` block […] now live in `ComposeAiBaseConventionsPlugin` (build-logic) […]

This is good writing about real decisions — it is simply in a file that is read by the build on every configuration, rather than by a person looking for the rationale. `docs/` already has 193 markdown files and 35k lines and is where this material belongs; the build script needs the *constraint* ("both Remote Compose lines must move together"), not the archaeology.

## Related surface

The same weight shows up in CI: **39 workflows / 13k lines of YAML**, and 38 scripts across `scripts/` and `.github/scripts/`, nine of which are `test-*.sh` companions testing the other scripts. That is effectively a second, untyped codebase with its own test suite — worth a separate look, but noting it here because it is the same phenomenon.

## Proposed shape

For `settings.gradle.kts`, and then the other large build scripts (`cli/build.gradle.kts` at 925 lines, `samples/desktop-daemon-bench/build.gradle.kts` at 970):

- keep the comment that states the live constraint a reader must not break;
- move the decision history to a `docs/build/` page and link to it from a one-line comment.

## Done looks like

`settings.gradle.kts` is materially shorter, the rationale is in `docs/` under a stable anchor, and the remaining comments describe constraints rather than history.

yschimke:
Closes #5169.

`settings.gradle.kts` had become the changelog for the build — multi-paragraph decision history, superseded designs and module archaeology, re-read by the build on every configuration rather than by a person looking for the rationale.

**851 → 677 lines.** The prose moves to a new `docs/build-scripts/SETTINGS.md` under stable anchors; the script keeps the *constraint* a reader must not break, plus a one-line link.

### What moved

| Anchor | What it now holds |
| --- | --- |
| `#base-conventions` | why conventions are applied per-module rather than from `allprojects {}` (Isolated Projects, `beforeProject` can't apply an included-build plugin, keeping `googleStyle()` typed) |
| `#remote-compose-lane` | the release/snapshot table, and the historical note on why the snapshot lane exists at all (`compose.remote.foundation` was androidx-main-only) |
| `#catalog-override` | why `create("libs")` and not `named` |
| `#robolectric-snapshots` | the Sonatype 2024–2025 migration and why both endpoints are declared |
| `#repositories-mode` | the wasmJs Node.js Ivy repository |
| `#build-cache` | the four-step token resolution order and the push policy |
| `#local-cache-always-on` | the full `isEnabled = !remotePushEnabled` post-mortem, including the 2026-08-09 measurement |
| `#remote-cache-kill-switch` | the #2824 truncated-entry analysis and why `composeai.cacheSalt` can't reach it |
| `#flat-data-paths`, `#serve-wasm-fork`, `#jdk21-samples`, `#ktfmt-project-paths` | the module-layout rationale |
| `#extractions`, `#render-host` | a table of the modules that left this build (preview server, session-viewer clients, `:cli-scripting`, rc-players, the Remote Compose catalog) and the one that arrived |

### What stayed in the script

Only live constraints, e.g.:

```kotlin
// CONSTRAINT: the whole trio moves together. They only work when built against the same
// `remote-creation*`, so the mode flips all three keys at once and one group must never straddle
// the two lines. docs/build-scripts/SETTINGS.md#remote-compose-lane
```

### Two comment blocks were fused, and are un-fused here

Not a deliberate part of the sweep, but visible once the file was readable:

- the `:preview-data-api` "published wire-format DTOs" paragraph sat above `include(":build-host-protocol")`;
- the render-session-library paragraph sat above `include(":render-host")`.

Each is now next to the module it actually describes.

### On the directory name

The issue proposed `docs/build/`. `.gitignore`'s Gradle `build/` rule swallows that path — the page was silently not committed on the first attempt. `docs/build-scripts/` avoids the collision without carving an exception into a rule that exists for build output. Say the word if you'd rather have `!docs/build/` in `.gitignore` instead.

### Scope

The issue also names `cli/build.gradle.kts` (925 lines) and `samples/desktop-daemon-bench/build.gradle.kts` (970) as "and then". This PR does `settings.gradle.kts` only — that is what the issue's *Done looks like* is scoped to, and it keeps the diff reviewable. Happy to follow up on the other two.

### Test plan

- `./gradlew help` — exit 0, so the settings script still parses and configures. (Comment-only Kotlin change; no behaviour touched.)
- `scripts/check-agent-entrypoints.sh --report` — green; AGENTS.md grows by one map row, to 9,619 bytes / 29% of Codex's 32 KiB budget.
- Every `docs/build-scripts/SETTINGS.md#…` link in the script resolves to an `<a id>` in the page (checked mechanically).
- Non-visual change, so no render evidence.

https://claude.ai/code/session_01ULG2c6Rw8ceLMmFTshQXrW

---
_Generated by [Claude Code](https://claude.ai/code/session_01ULG2c6Rw8ceLMmFTshQXrW)_


yschimke:
`Apply compose-preview` is failing on this head, and it is not this PR's failure — this branch changes comments and markdown only. **`main` is broken.**

```
Could not resolve all dependencies for configuration ':cli:runtimeClasspath'.
  > Could not find ee.schimke.composeai:compose-preview-render-host:3.0.0.
```

### Root cause

Checked against Maven Central directly:

| Artifact | 2.21.0 | 3.0.0 |
| --- | --- | --- |
| `compose-preview-serve` | 200 | **200** |
| `compose-preview-render-host` | 200 | **404** |

`compose-preview-render-host` is not published at 3.0.0, because it came home to this repository as `:render-host` (#4832). `render-host/build.gradle.kts` says so explicitly, and correctly predicts this state:

> Coordinate change, deliberate: it published as `compose-preview-render-host` from the server and publishes as `render-host` from here. […] the old one stays resolvable at its final 2.x for anyone pinned to it.

But `gradle/libs.versions.toml` still points the old coordinate at the *server's* version ref:

```toml
composeai-preview-render-host = { module = "ee.schimke.composeai:compose-preview-render-host", version.ref = "composeai-preview-serve" }
```

with the shared-ref comment resting on an assumption that stopped being true when the module moved:

> ONE version ref for both, deliberately: they are cut from one build and one tag over there, and a skew between the server and the render host it sits on top of is not a state that repository can produce.

That repository can now produce exactly that state — it releases the server without the render host. So #5164 (Renovate, `compose-preview-serve` → v3) dragged the old render-host coordinate to a version that does not exist, and merged.

### Two candidate fixes, neither of which belongs in this PR

**A — unblock, mechanical.** Give the old coordinate its own version ref pinned at its final 2.x, so a server bump can't drag it again:

```toml
composeai-preview-serve = "3.0.0"
composeai-preview-render-host = "2.21.0"
```
```toml
composeai-preview-render-host = { module = "…:compose-preview-render-host", version.ref = "composeai-preview-render-host" }
```

**B — finish the move.** `:cli` still reaches for the *published* artifact (`cli/build.gradle.kts:276` `api(libs.composeai.preview.render.host)`, and `:441` `testFixtures(...)`) even though `:render-host` is in this build. Point those at `project(":render-host")` and drop the stale coordinate. This looks like what #4832 intended, and it also retires the shared-ref hazard rather than pinning around it.

B is the real fix but it is an architecture call, not mine to make on a docs PR — so I am not widening this one. Happy to open a separate PR for whichever you prefer.

### On re-running

Not re-running the job. The failure is a deterministic 404 on a coordinate that does not exist, verified against Maven Central above rather than by repetition — a re-run would only burn a runner. I'll keep this PR watched; it should go green on its own once `main` is fixed and this branch picks the fix up.

---

Issue #5176, opened 2026-09-05
Decide where the MCP server belongs under the layer rule


yschimke:
## What prompted this

Closing the dependency cycle ([compose-preview-server#180](https://github.com/yschimke/compose-preview-server/issues/180)) meant making `serve` a launcher and dropping `compose-preview-serve` from `:cli`. The expectation was that the Ktor server floor would leave the CLI distribution with it.

**It did not**, and measuring rather than assuming is what surfaced this. What actually left: the `compose-preview-serve` jar, `jmdns`, `ktor-server-compression`, `ktor-server-auto-head-response`. Seventy-two jars to sixty-nine.

What stayed, and why:

```
io.ktor:ktor-server-core / -cio / -content-negotiation / -sse / -websockets
  ├── project ':mcp'                                    (ktor-server-cio directly)
  └── io.modelcontextprotocol:kotlin-sdk-server-jvm      (the rest)
```

`compose-preview mcp` runs a server of its own. Five `ktor-server-*` artifacts are on the CLI's runtime classpath for it, and they were never `serve`'s to take.

## The question

`docs/design/REPOSITORY_LAYERS.md` places a module by one test: *does it need an HTTP server, a browser, or the UI builder to do its job?* If yes, it is layer 2 — compose-preview-server.

`:mcp` is an HTTP server. By the letter of the rule it is layer 2, and the fact that it sits in layer 1 is the same kind of mis-filing `:render-host` was — a module that ended up where it was standing rather than where the rule puts it.

But the rule was written with `serve` in mind, and MCP is not obviously the same shape:

- **It is an agent entry point, not a preview surface.** `compose-preview mcp serve | install | doctor` exists so an agent can drive *this CLI*. Moving it to the preview server puts a CLI front door in the repository that owns the web UI.
- **It depends on layer-1 internals.** `:daemon:core`, `:render-session-api`, `:daemon-client` as `api`, plus `data-layoutinspector-core`. Those are exactly the things the CLI has and the server consumes across a boundary — so a move trades a Ktor floor here for a wider published surface there.
- **17 files, ~8.4k lines**, so it is not a mechanical lift.

## Options, roughly

1. **Move `:mcp` to compose-preview-server.** Honours the rule literally; removes the last `ktor-server-*` from the CLI distribution. Cost: the agent entry point lives in the server repo, and whatever it needs from layer 1 becomes published surface.
[...]

PR #5177, opened 2026-09-05
feat(serve): launch the published server instead of linking it

yschimke:
The last step of [compose-preview-server#180](https://github.com/yschimke/compose-preview-server/issues/180). **The forward edge is gone and `CheckLayerBoundary`'s allowlist is empty** — which is the mechanical proof, not an assertion.

## Why

`serve` and `browse` were the last things holding the preview server on this module's classpath. `ServeCommand` implemented the server's `ServeBuildHost` and ran `ServeRunner` in-process, so an offline CLI linked a web server in order that four commands which never open a socket could reach types filed in the same package as one.

They are launchers now. `ServeCommand` execs the published `compose-preview-server` and forwards argv untouched; the Gradle work travels the other way, with the server spawning `compose-preview build-host --stdio` ([#5147](https://github.com/yschimke/compose-ai-tools/pull/5147), [compose-preview-server#294](https://github.com/yschimke/compose-preview-server/pull/294)). Neither side links the other.

`ServerBinaryDiscovery` mirrors the server's build-host discovery deliberately — flag, environment, `PATH`, most explicit first — so learning one teaches the other. Unlike the server's, a miss here is a failure: `serve` has nothing to fall back to, so it prints how to install the binary.

`:cli` also takes `:render-host` as a project now rather than the server's old published coordinate.

## The one edge that survives, and why

`compose-preview-serve` remains a **`testImplementation`**. Two tests drive the CLI's own HTTP clients against a real `ServeHttpServer`, and their stated purpose is to catch the two repositories' independently-declared wire types drifting apart:

> …so the two halves of the wire protocol are checked against each other **rather than each against its own idea of the other**.

I tried both alternatives and neither survives contact:

- **Move them to the server repo** — they need `AgentAccessClient`, `AgentAccessStore`, `ServeImageUploader`, all `internal` CLI classes reached same-package.
- **Stub the server** — a stub *is* "each against its own idea of the other". Seven tests would stay green while testing nothing they exist for, which is worse than deleting them because it looks like coverage. Five of them also approve/deny grants by reaching into the server's store, which only works in-process.

So the edge stays where it earns its keep and nowhere else. The gate reads `runtimeClasspath` and so does not see it — correctly: the claim it enforces is about what ships. The exclusion on that dependency is not optional, and the build file says why.

## Measured, not assumed

I expected the Ktor floor to leave with `serve`. **It didn't**, and checking the built distribution is what caught it:

| left | stayed |
| --- | --- |
| `compose-preview-serve` jar, `jmdns`, `ktor-server-compression`, `ktor-server-auto-head-response` | `ktor-server-core`, `-cio`, `-content-negotiation`, `-sse`, `-websockets` |

72 jars → 69. The five that stay arrive via `:mcp` and the MCP Kotlin SDK, because `compose-preview mcp` runs a server of its own — never `serve`'s to take. An earlier draft of the build comment claimed otherwise; it's corrected, and where MCP belongs is now [#5176](https://github.com/yschimke/compose-ai-tools/issues/5176).

## Tests replaced, not dropped

`ServeCommandTest` and `ServeOptionsDelegationTest` pinned wiring between `ServeCommand` and `ServeRunner`/`ServeCommandOptions`. There is no such wiring now, so `ServeLauncherTest` pins what replaced it — the argv, which is the launcher's whole job.

One case is deliberately not reproduced: `ServeOptionsDelegationTest`'s init-script check guarded `ServeCommand` overriding `autoInjectInitScriptArgs` with the same name as the function it delegated to, a pair that can silently recurse. `BuildHostCommand` declares no such override, so the hazard is gone with the override.

## Behaviour change worth flagging

Preview selection in `serve` now uses the server's `previewIdMatchesStandaloneRequest` rather than the CLI's shared rule, because the CLI no longer constructs `ServeCommandOptions`. The server's matcher was written to mirror it, but they are separate declarations.

## Test plan

- `./gradlew :cli:test` — **844 tests, 0 failures**, including the two wire-drift tests still running against a real server.
- `./gradlew :cli:checkLayerBoundary` — green **with an empty allowlist**.
- `./gradlew :cli:installDist` — distribution inspected for the table above.
- `./gradlew :cli:ktfmtCheck*`, `scripts/check-agent-entrypoints.sh` — green.

## Breaking

`compose-preview serve` / `browse` now require the `compose-preview-server` binary, which ships separately: on `PATH`, via `COMPOSE_PREVIEW_SERVER`, or `--server-binary`. Every other command is unaffected.

**This needs the installer in [`yschimke/skills`](https://github.com/yschimke/skills) to fetch the server binary** — that repo is outside this session's scope, and until it does, `serve` prints the installation hint instead of running.

https://claude.ai/code/session_014NtzoUCr4pou3y1qRDxcpZ

---
_Generated by [Claude Code](https://claude.ai/code/session_014NtzoUCr4pou3y1qRDxcpZ)_


yschimke:
fix merge conflict

PR #5173, opened 2026-09-05
chore: remove dead scaffolding — placeholder tests and the history/diff gate

yschimke:
Closes #5168 (items 1 and 2; item 3 deliberately left alone).

Two independent commits — the second can be dropped without affecting the first.

## 1. `test(daemon)`: drop six empty placeholder `@Ignore` harness tests

The S9/S10/S11 real-mode classes in `:daemon:harness` asserted nothing and never had — each body was a single comment, and the `@Ignore` reason pointed at a KDoc explaining why the test did not exist yet:

```kotlin
@Ignore("S11 real-mode desktop placeholder — see KDoc")
class S11HistoryPruneDesktopRealModeTest {
  @Test
  fun s11_history_prune_real_mode_desktop_placeholder() {
    // Placeholder body — the @Ignore annotation skips this test in JUnit.
  }
}
```

The only references to them were between the six themselves — verified before deleting, and no dangling references remain.

## 2. `feat(daemon)`: retire the `history/diff` experimental gate

`composeai.experimental.historyDiff` held `history/diff` at method-not-found by default. Three `TODO(1.1)` markers stated the precondition for removing it — *"land H5 + the remaining roadmap items"*. That precondition is met:

| Item | Landed in |
| --- | --- |
| pixel mode (H5) | #1892 |
| data mode — a11y / semantics / theme | #1873, #1949 |
| semantics mode | #1785, #1862 |
| per-call `ref` param for list/read/diff | #1909 |

`docs/daemon/PROTOCOL.md` already described the precondition as satisfied — *"`-32012` `HistoryPixelNotImplemented` is retired now that H5 has landed"* — while the dispatcher still gated on the flag. The repository is at 1.79.0, well past the 1.0 the gate was written for.

`history/diff` now behaves like its three siblings — served whenever `HistoryFeature.ENABLED` is on, returning the same `HISTORY_DISABLED_MESSAGE` when it is not:

```kotlin
"history/diff" ->
  if (HistoryFeature.ENABLED) handleHistoryDiff(req)
  else sendErrorResponse(req.id, ERR_METHOD_NOT_FOUND, HISTORY_DISABLED_MESSAGE)
```

The `historyDiffExperimental` constructor parameter and the `HISTORY_DIFF_EXPERIMENTAL_PROP` constant are gone, along with the two test call sites that passed it, the `HistoryFeature` KDoc sentence claiming diff carries an extra gate, and the experimental admonition in `PROTOCOL.md`.

[...]

PR #5179, opened 2026-09-05
refactor(daemon): dispatch RPC through a handler registry

### LINKED ISSUE #5166: refactor(daemon): break up JsonRpcServer and DaemonMcpServer via a handler registry
Found during a tech-debt sweep of the repository.

## The problem

Four files carry a disproportionate share of the codebase, each as essentially one class:

| File | Lines | Classes | Functions |
| --- | --- | --- | --- |
| `mcp/src/main/kotlin/ee/schimke/composeai/mcp/DaemonMcpServer.kt` | 5102 | 1 | 108 |
| `daemon/core/src/main/kotlin/ee/schimke/composeai/daemon/JsonRpcServer.kt` | 4675 | 3 | 95 |
| `daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/RenderEngine.kt` | 3537 | — | 51 |
| `daemon/desktop/src/main/kotlin/ee/schimke/composeai/daemon/RenderEngine.kt` | 3223 | — | 76 |

`JsonRpcServer` dispatches **38 RPC methods** from a single `when (req.method)` at line 725. Every feature that adds a method grows this one file, and every feature's handler lives far from the feature it serves.

## Why the seam is already there

`daemon/core` has the right idea elsewhere — `ExtensionRegistry`, `RecordingScriptHandlerRegistry`, `DataProductRegistry`, `FrameStreamRegistry` all exist and all do exactly this job for their own domain. The `JsonRpcServer` constructor KDoc even records the migration:

> Replaces the per-feature lists this constructor used to take (`dataProducts`, `dataExtensions`, `previewExtensions`).

RPC dispatch never got the same treatment.

## Proposed shape

An `RpcMethodHandler` registry mirroring `ExtensionRegistry`, letting each method's handler live beside the feature it serves rather than in the central `when`.

The `history/*` cluster is the natural first slice — `history/list`, `history/read`, `history/diff`, `history/prune` and their handlers would lift roughly 1000 lines out of `JsonRpcServer` into `daemon/core/.../history/`, where `HistoryManager` and friends already live.

## Done looks like

`JsonRpcServer` dispatches through a registry instead of a 38-arm `when`, and the history handlers live in the `history` package.

## Note

This is a large refactor with real regression risk on the protocol surface. Worth doing incrementally, one method cluster at a time, with the daemon harness suite green between each.

yschimke:
Closes #5166.

## The change

`JsonRpcServer.handleRequest` dispatched 24 request methods from a single `when (req.method)`, so every feature that added a method grew this one file and every feature's handler lived far from the feature it serves. `daemon/core` already solves that shape elsewhere — `ExtensionRegistry`, `RecordingScriptHandlerRegistry`, `DataProductRegistry`, `FrameStreamRegistry` — but RPC dispatch never got the same treatment.

**`rpc/RpcMethodRegistry.kt`** (new) — a method-name → handler map built once at construction, plus a narrow `RpcPeer` port carrying the four things a handler needs from the connection: decode params, encode a result, send a response, send an error. `handleRequest` is now a lookup and a call. Duplicate registration throws rather than silently shadowing; an unregistered method is what `method not found` means.

`RpcPeer` is implemented as an adapter object rather than `JsonRpcServer : RpcPeer`, so the reply helpers stay private to that file and the server's **public API is unchanged** apart from the three new `rpc` types (added to `core.api`).

**`history/HistoryRpcHandlers.kt`** (new) — the first slice the issue names. `history/list`, `history/read`, `history/diff` and `history/prune` move next to the `HistoryManager` they dispatch into, lifting ~470 lines out of `JsonRpcServer` (4675 → 4215 lines). Handler bodies are unchanged apart from reaching the connection through `RpcPeer`; `encodeHistoryEntry` becomes an internal top-level helper in the `history` package, still used by the `historyAdded` notification path.

The `HistoryFeature.ENABLED` gate moves into a handler wrapper that replies with the gate's error, so a gated method is still *registered* and "no handler" keeps its unambiguous meaning. Same error codes, same messages, same degradation when no manager is wired.

The remaining 20 methods register their existing `JsonRpcServer` handlers as-is; moving each cluster to its own file is the incremental follow-up the issue asks for.

## Merged with main

Rebased onto `main` twice by merge, resolving conflicts with #5173 (which retired the `history/diff` experimental sysprop — that gate is gone from the extracted handler too, matching main's shape) and #5182 (the typed `DaemonProperties` registry — import-only conflict). `main` is fully merged into the head; `:daemon:core:test` and `:daemon:harness:test` were re-run green after each merge.

## Test plan

- `./gradlew :daemon:core:test` — green, including `JsonRpcServerHistoryIntegrationTest` and `HistoryDiffTest`, which exercise the whole `history/…` wire surface end to end and are the regression net for the extraction.
- `./gradlew :daemon:harness:test` — green (42 classes, 0 failures), including the `S9`/`S10`/`S11` history scenarios.
- New `RpcMethodRegistryTest` pins the invariants the `when` used to hold by construction: dispatch, an unregistered method having no handler, duplicate registration failing loudly, and `methods` reporting the registered set.
- `./gradlew :daemon:core:ktfmtFormatMain :daemon:core:ktfmtFormatTest` run and staged.

Non-visual change: dispatch plumbing only, no rendered surface touched, so no before/after images.

Docs: added the `RpcMethodRegistry` seam to `docs/AGENT_GUIDE.md` (so the next method registers a handler instead of growing a central table) and the two new files to the `docs/daemon/DESIGN.md` module map.

https://claude.ai/code/session_01PFNgD8XmKnf31V8ZXAVwDA


yschimke:
fix merge conflicts

PR #5187, opened 2026-09-05
fix(scroll): verify every LONG stride and seam, and stop black overlaps winning the stitch

yschimke:
## Summary

Wear `@ScrollingPreview(LONG)` stitches of a `ScalingLazyColumn` under a pinned `TimeText` came out with `10:10` stamped into the middle of the list and a section header repeated at the seam. On horologist's `sample` module, 4 of the 6 `SectionedList*` LONG renders were broken this way (`SectionedListExpandableScreenPreview` on both round devices, `SectionedListStateful`/`Stateless` on Small Round).

**Root cause.** The stitcher's overlap matcher weighted rows by horizontal variance but nothing stopped a candidate whose whole overlap was low-signal from winning. With the time drawn *over* the head of every slice, the true 307 px shift (on a 384 px viewport) scored 8.5/px, while a 372 px shift over a 12-row all-black overlap scored 0.3/px and won — the next slice was painted from its top row, `TimeText` and all, sixty rows too high. Two more latent problems surfaced on the way: the reported-hint window `[hint/3, hint×3]` could not see an end-of-list landing (82 px against a 363 px hint), and the driver credited a full stride whenever a lazy list's `ScrollAxisRange.value` (which is `index × 500 + offset`, not pixels) said so.

**What changes.**

- **Stitcher (`data/scroll/core`)** — a candidate shift is only eligible when its overlap is at least a tenth of the viewport *and* its informative rows (stddev ≥ 10) sum to enough signal to be a feature rather than a stray edge; candidates within a small margin of the best score are a tie and the one nearest the hint wins (identical chip bodies one pitch apart); a reported hint bounds the search above only. Every seam now gets a `ScrollSeam` verdict — `verified`, `measured` (no pixel signal, placed by the driver's measured stride), `low_signal`, `mismatch` — with its residual and signal.
- **Driver (`data/scroll/android`)** — `driveScrollByViewport` snapshots the `positionInRoot` of the scrollable's descendant semantics nodes before each stride, advances the paused clock frame by frame until they stop moving, measures the median displacement, and dispatches up to three corrective `ScrollBy`s when the landing misses the plan by more than a pixel while the scroller still has room. The measured travel is the hint the stitcher gets, flagged so its search window is ±2 % of the viewport instead of a third-to-triple. (`boundsInRoot` is clipped to the viewport, so an item sliding off the top reports `top = 0` and its displacement saturates — that cost one round of debugging and is written up on `ContentAnchors.snapshot`.) Every stride reports a `ScrollStep`.
- **Detection** — anything unverified is written to `<png>.warnings.json` as `unlandedScrollSteps` / `unverifiedScrollSeams` (additive to the schema, empty arrays on a clean render), on both the Robolectric and daemon render paths. Seams whose rows the EdgeButton anchor path discards are not reported. `COMPOSEAI_KEEP_SCROLL_SLICES=1` keeps the slices, the final frame, the reported offsets and the driver's trace next to the output; `ScrollSliceStitcherHarnessTest` re-stitches such a directory offline.

## Before / after

The four broken horologist renders (`SectionedListExpandableScreenPreview` on Small Round — the file in the report — and Large Round, `Stateful` and `Stateless` on Small Round), same plugin build except for this change:

| before | after |
| --- | --- |
| before, expandable small round | after, expandable small round |
| before, expandable large round | after, expandable large round |
| before, stateful small round | after, stateful small round |
| before, stateless small round | after, stateless small round |

With the fix, all six horologist previews stride exactly (every step lands, every seam verified; the only remaining warning is the pre-existing `still_changing` on one placeholder-shimmer slice). `renders/wear-long-scroll-seams/README.md` has the provenance.

## Test plan

- [x] `:data-scroll-core:test` — existing stitcher fixtures (honest / over-reported / under-reported offsets, blank-background alignment, end-of-scroll no-move) still byte-exact; `ScrollSliceStitcherHarnessTest` re-stitches the six kept horologist slice sets with every seam verified
- [x] `:renderer-android:testDebugUnitTest --tests '*Scroll*' --tests '*Sidecar*' --tests '*VisualSettle*'` — new `WearPinnedTimeTextLongScrollTest` (Wear Material `ScalingLazyColumn` + pinned `TimeText` through `handleLongCapture`: every stride lands, every seam verifies, the `TimeText` band never recurs) and `ScrollDriveDiagnosticsTest` (recording, draining, sidecar shape); `AndroidScrollDriverStepTest`, `LongScrollGhostOverlayTest`, `WearTlcScrollSemanticsProbeTest` unchanged
- [x] `:samples:wear:testDebugUnitTest` — `LongScrollPreviewPixelTest` on the `TransformingLazyColumn` + `EdgeButton` fixtures, plus a new assertion that their sidecars carry no unlanded stride and no unverified seam
- [x] horologist `:sample:composePreviewRender -PcomposePreview.filter='SectionedList*'` against the mavenLocal snapshot of this branch — the four broken renders above are clean, the two that were fine are unchanged
- [x] `ktfmtFormat` on the touched modules

https://claude.ai/code/session_01WY2JFjCSGsFeq4AQuNE3m2

---
_Generated by [Claude Code](https://claude.ai/code/session_01WY2JFjCSGsFeq4AQuNE3m2)_


yschimke:
**`Actions Script Tests` is red here, and it is not this PR's.**

The failing step is *Run daemon-launch schema gate tests* (`scripts/test_check_daemon_launch_schema.py::test_repo_is_green`): the JVM contracts reader reports that `SANDBOX_COUNT_PROP = DaemonProperties.Names.SANDBOX_COUNT` in `daemon/android/.../DaemonMain.kt` and `.../pool/SandboxProcessPool.kt` no longer matches the `"composeai.daemon.sandboxCount"` literal in `DaemonLaunchDescriptor.kt`, because the checker resolves literals, not the new typed constant.

- Same step fails on `main` at ac87b28 ([run 33960679144](https://github.com/yschimke/compose-ai-tools/actions/runs/33960679144/job/101291976952)), where `Build CLI` is also red. It came in with the #5182 sysprops-registry refactor, which this branch picked up by rebasing onto `main`.
- This diff touches none of those files (its only `daemon/` change is three lines in `RenderEngine.kt` draining scroll diagnostics into the warnings sidecar).
- No fix PR exists yet, so there is nothing to port; a re-run would not change a deterministic gate. I'll keep this PR watched and rebase once `main` carries the fix.

---

PR #5192, opened 2026-09-05
fix(cli): make optional captures visible in show output

### LINKED ISSUE #5174: `optional` captures are invisible in `show` output: `[no PNG]` is untagged and `counts` don't sum to `total`
## Summary

`Capture.optional` is honoured by the render gate and by `counts.missing`, but **not** by the two channels a consumer actually reads per preview. The result is that a best-effort miss that is expected by design is indistinguishable from a genuine render failure, and the JSON counts silently stop adding up.

Concretely, on a run with 4 activity previews that produced no PNG — 1 required, 3 `optional`:

```console
$ compose-preview show --module app
…
WidgetPreviewActivity (activity__WidgetPreviewActivity) [no PNG]
MainActivity (activity__MainActivity) [no PNG]
RedirectUriReceiverActivity (activity__RedirectUriReceiverActivity) [no PNG]
AuthorizationManagementActivity (activity__AuthorizationManagementActivity) [no PNG]
Render task completed but produced no PNG for 1 of 37 preview(s):
  - activity__MainActivity (app) — no PNG for: default
```

Four previews are tagged as having no PNG; the summary immediately below says one. Both numbers are "right" under their own policy, but nothing in the output explains the discrepancy, and the reader has no way to tell *which* three of the four are fine.

## The two policies

The summary path is optional-aware. `cli/src/main/kotlin/ee/schimke/composeai/cli/Commands.kt:1575`:

```kotlin
internal fun previewsMissingPng(results: List<PreviewResult>): List<PreviewResult> =
  results.filter { r ->
    r.params.kind !in NON_PNG_PREVIEW_KINDS && r.captures.any { it.pngPath == null && !it.optional }
  }
```

The per-preview status tag is not. `Commands.kt:1711` (and identically for the multi-capture branch at `:1723`):

```kotlin
val statusTag =
  when {
    r.pngPath == null -> " [no PNG]"
    r.anyChanged() -> " [changed]"
    else -> ""
  }
```

No `optional` check. `ResourceCommands.kt:475` / `:487` have the same shape.

This is working as designed for the activity previews — `AppTourDiscovery.kt:224` is explicit that only the launcher activity is required:

> The launcher activity's capture is required (the hero must render); every other activity is `optional` — real screens often need intent extras or session state discovery can't guess, and a best-effort `.error.json` beside a missing PNG beats failing the whole render for it.

That is a good policy. It just isn't visible in the output.

## The counts bug

`countsOf` (`Commands.kt:958`) puts an optional-missing preview into **no bucket at all**. It isn't `changed`; it isn't `unchanged` (that requires some capture with a non-null `pngPath`); and it is excluded from `missing` by the `captures.any { !c.optional }` guard. So:

```json
"counts": { "total": 37, "changed": 0, "unchanged": 33, "missing": 1 }
```

`0 + 33 + 1 = 34`, not 37. The three missing entries are exactly the three `optional` activity previews. Any consumer that computes a residual from these counts, or asserts they partition `total`, is quietly wrong — and the error scales with how many optional captures a project has.

Verified against the raw JSON for the same run:

```
activity__WidgetPreviewActivity          pngPath=None  captures=[(has_png=False, optional=True)]
activity__MainActivity                   pngPath=None  captures=[(has_png=False, optional=False)]
activity__RedirectUriReceiverActivity    pngPath=None  captures=[(has_png=False, optional=True)]
activity__AuthorizationManagementActivity pngPath=None captures=[(has_png=False, optional=True)]
```

## Why it matters

The skill documentation points agents at `show --json` and tells them to read the PNGs whose state changed, using the counts to reason about the run. An agent seeing four `[no PNG]` tags against a summary of one has to either open the JSON and reimplement the `optional` policy itself, or guess. The most likely guess — "the render is broken, four previews failed" — is wrong three times out of four, and on this project it points the investigation at OAuth/redirect activities that were never expected to render.

It also makes the one *genuine* failure harder to see, which is the opposite of what the tagging is for.

## Suggested fix

1. **Make the status tag optional-aware.** Distinguish the expected miss from the real one, e.g. `[no PNG, optional]` / `[skipped]` / `[best-effort]` versus a bare `[no PNG]`. Both the preview-level tag (`Commands.kt:1711`) and the capture-level tag (`:1723`) need it, plus the two in `ResourceCommands.kt`. Reusing the `previewsMissingPng` predicate for the tag would keep the two channels honest by construction.
2. **Make the counts partition `total`.** Either add an `optionalMissing` (or `skipped`) bucket, or fold optional-missing into `unchanged`. Adding a bucket is the clearer of the two, and a unit test asserting `changed + unchanged + missing + skipped == total` would pin it.
3. **Consider surfacing the `.error.json` reason for optional misses too** — `AppTourDiscovery` already writes a best-effort sidecar; a one-line "why" beside an optional skip would answer the reader's next question without another command.

(1) alone removes the misleading part; (2) is what makes the JSON safe to compute on.

## Environment

- compose-preview 1.79.0, Gradle 9.7.1, JDK 17 (Temurin 17.0.19)
- Reproduced on `yschimke/homeassistant-remotecompose`, `:app` module (4 manifest activities, 1 launcher)

yschimke:
Fixes #5174.

`Capture.optional` was honoured by the render gate and by `counts.missing`, but not by the two channels a consumer actually reads per preview. A best-effort miss that is expected by design looked exactly like a render failure, and the JSON counts silently stopped adding up.

## What changed

**1. The status tag is optional-aware.** A bare `[no PNG]` is now reserved for the misses the missing-render gate flags — exactly the rows the "produced no PNG for N of M preview(s)" summary below the listing enumerates. Everything else says why it is empty:

```
WidgetPreviewActivity (activity__WidgetPreviewActivity) [no PNG, optional]
MainActivity (activity__MainActivity) [no PNG]
RedirectUriReceiverActivity (activity__RedirectUriReceiverActivity) [no PNG, optional]
AuthorizationManagementActivity (activity__AuthorizationManagementActivity) [no PNG, optional]
Render task completed but produced no PNG for 1 of 37 preview(s):
  - activity__MainActivity (app) — no PNG for: default
```

An `XR_SUBSPACE` preview, whose empty `pngPath` was never a failure either, reads `[no PNG, by design]`. Both the preview-level tag and the per-capture tag are covered, in `Commands.kt` and `ResourceCommands.kt` alike.

The gate predicate is extracted as `previewMissesRequiredPng` and now drives the tag, the diagnostic and the counts bucket, so the three channels a consumer reads cannot disagree about the same preview — which is what went wrong here. A test asserts the agreement over a mixed result set.

**2. The counts partition `total`.** `PreviewCounts` gains a `skipped` bucket, and `countsOf` classifies each preview exactly once (`previewCountBucket`) instead of evaluating four independent predicates. An optional-missing preview used to land in no bucket at all:

```json
"counts": { "total": 37, "changed": 0, "unchanged": 33, "missing": 1 }
```

Now `changed + unchanged + missing + skipped == total` holds by construction, and a unit test pins it across every result shape (changed, unchanged, required miss, optional miss, non-PNG kind, no captures, partial fan-out). The addition is backwards compatible: `skipped` defaults to `0`, the other three buckets keep their existing meanings and values, and no consumer in the repo reads a residual.

**3. `show-resources` too.** Resource captures have no `optional` flag; their equivalent is the renderer's `skipped` error status (a known degradation, not a failure), which now tags `[no PNG, skipped]`. Resource counts already partitioned, so they are unchanged.

## Not done

Suggestion (3) in the issue — surfacing the `.error.json` reason beside an optional skip — is left out. That reason comes from `diagnoseMissingRenders`, whose output is a carefully invariant-checked diagnostic paragraph (`MissingRenderMessage.kt`), and extending it to expected misses is a larger change than the visibility bug needs. The tag now answers "is this one a problem?"; `--json` still carries `optional` per capture for the follow-up question.

## Test plan

- `./gradlew :cli:test` — green. New `OptionalCaptureVisibilityTest` covers the tags, the tag/gate agreement and the counts partition; `ResourceCommandsTest` gains the resource-side tag case.
- `./gradlew :cli:ktfmtFormatMain :cli:ktfmtFormatTest` — clean.

Text-only change (CLI stdout tags and a JSON field), so there is no rendered surface to show before/after; the console excerpt above is the affected output.

https://claude.ai/code/session_01FZxi7UdPeKoss1yyGmWkwB

---
_Generated by [Claude Code](https://claude.ai/code/session_01FZxi7UdPeKoss1yyGmWkwB)_

### DIFF
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/Commands.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/Commands.kt
@@ -64,8 +64,32 @@ data class PreviewListResponse(
   val counts: PreviewCounts? = null,
 )
 
+/**
+ * The `counts` block of a `show` / `a11y` JSON envelope.
+ *
+ * The four buckets **partition** [total] — every preview lands in exactly one, and `changed +
+ * unchanged + missing + skipped == total` for any run (asserted in `PreviewCountsPartitionTest`).
+ * Consumers can therefore compute a residual from them, which they could not before [skipped]
+ * existed: a preview whose only captures were `optional` and produced no PNG fell into no bucket at
+ * all, so on a project with best-effort captures the counts silently stopped adding up
+ * (issue #5174).
+ *
+ * @property changed at least one capture's sha256 differs from the previous run.
+ * @property unchanged nothing changed, and at least one capture has a PNG.
+ * @property missing no PNG at all, and the miss is a render failure — the set `--missing-renders`
+ *   gates on (see [previewsMissingPng]).
+ * @property skipped no PNG at all, and the miss is **expected**: every absent capture is
+ *   `optional`, or the preview is one of [NON_PNG_PREVIEW_KINDS]. Text output tags these rows `[no
+ *   PNG, optional]` / `[no PNG, by design]` rather than a bare `[no PNG]`.
+ */
 @Serializable
-data class PreviewCounts(val total: Int, val changed: Int, val unchanged: Int, val missing: Int)
+data class PreviewCounts(
+  val total: Int,
+  val changed: Int,
+  val unchanged: Int,
+  val missing: Int,
+  val skipped: Int = 0,
+)
 
 /**
  * Compact response shape emitted under `--brief`. Drops everything an agent already had from a
@@ -898,8 +922,7 @@ abstract class Command(
   }
 
   /** True if the preview has at least one capture with `changed = true`. */
-  protected fun PreviewResult.anyChanged(): Boolean =
-    captures.any { it.changed == true } || changed == true
+  protected fun PreviewResult.anyChanged(): Boolean = hasChangedCapture()
 
   /** Filters by `--id` / `--filter` and (optionally) `--changed-only`. */
   protected fun applyFilters(all: List<PreviewResult>): List<PreviewResult> =
@@ -959,22 +982,7 @@ abstract class Command(
     )
   }
 
-  private fun countsOf(results: List<PreviewResult>) =
-    PreviewCounts(
-      total = results.size,
-      changed = results.count { it.anyChanged() },
-      unchanged = results.count { !it.anyChanged() && it.captures.any { c -> c.pngPath != null } },
-      // Exclude kinds that never emit a PNG (see [NON_PNG_PREVIEW_KINDS]) and `optional` captures
-      // so `counts.missing` matches what `--missing-renders` actually gates on — an
-      // `@XrSubspacePreview` with no composite or a desktop `@ColorCatalog` sheet still isn't a
-      // render failure.
-      missing =
-        results.count { r ->
-          r.params.kind !in NON_PNG_PREVIEW_KINDS &&
-            r.captures.all { c -> c.pngPath == null } &&
-            r.captures.any { c -> !c.optional }
-        },
-    )
+  private fun countsOf(results: List<PreviewResult>): PreviewCounts = previewCountsOf(results)
 
   protected fun matchesRequest(preview: PreviewInfo): Boolean =
     previewIdMatchesRequest(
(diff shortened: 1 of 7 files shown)


yschimke:
`Actions Script Tests` was red on `5c7efc0` — `test_check_daemon_launch_schema.py::RealTree::test_repo_is_green`, on the two android mirrors of `SANDBOX_COUNT_PROP` reading `DaemonProperties.Names.SANDBOX_COUNT` where the gate expected a string literal.

Not this PR's: the diff touches `cli/` only. It reproduces identically on `origin/main` (28fb93b) once the contracts checkout the CI job supplies is present — locally the mirrored-constant comparison is skipped without it, which is why it looks green in a bare clone:

```
$ COMPOSE_PREVIEW_CONTRACTS_ROOT=… python3 scripts/check-daemon-launch-schema.py   # on origin/main
check-daemon-launch-schema: FAILED
  daemon/android/…/DaemonMain.kt: `SANDBOX_COUNT_PROP` = DaemonProperties.Names.SANDBOX_COUNT, but …DaemonLaunchDescriptor.kt says "composeai.daemon.sandboxCount".
  daemon/android/…/pool/SandboxProcessPool.kt: …
```

The cause is #5182, which gave the `composeai.daemon.*` sysprops a typed registry.

A fix exists — https://github.com/yschimke/compose-ai-tools/pull/5191, which teaches the gate to follow a dotted constant reference — so I've cherry-picked it into this branch (95ada85) rather than wait for it to merge; it no-ops once `main` carries it. Verified here with the contracts checkout: `check-daemon-launch-schema: OK`, and `scripts/test_check_daemon_launch_schema.py` runs 51 tests green.

---

PR #5193, opened 2026-09-05
test: pin the two preview-selector rules against a shared table

### LINKED ISSUE #5185: Nothing pins the two preview-selector rules against each other
Small, and filed because it is the kind of divergence that goes unnoticed until someone's `--preview` selects the wrong thing.

## What changed

Before #5177, `compose-preview serve` built its options with the CLI's shared selector rule bound in:

```kotlin
previewMatcher = { id, exactId, filter, previewRef, className, functionName ->
  previewIdMatchesRequest(id, exactId = exactId, filter = filter, …)
}
```

`serve` is a launcher now, so the CLI no longer constructs `ServeCommandOptions` at all. Selection is done by the server's own `previewIdMatchesStandaloneRequest`.

## The problem

They are **separate declarations in separate repositories**, and nothing checks that they agree. The server's was written to mirror the CLI's, and as far as I can tell it does — but "as far as I can tell" is the whole issue. `--id`, `--filter` and `--preview` are user-facing selectors, and the failure mode is quiet: a preview that used to match stops matching, or the reverse, with no error anywhere.

This is the same shape as the wire-type drift that `AgentAccessClientIntegrationTest` exists to catch, minus the test.

## What would settle it

Either:

- **Pin it.** A shared table of (id, selectors) → expected outcome, exercised on both sides. The rule is small and pure, so a fixture file both repositories read is cheap — the same pattern as `docs/daemon/protocol-fixtures/` in compose-preview-contracts.
- **Or make one of them go away.** If the selector rule is genuinely a wire-level concern now, it belongs in one place, and the launcher change is the moment that became true.

Worth checking the two implementations line up before doing either — this issue asserts that nothing *pins* them, not that they currently differ.

## Context

Fell out of [compose-preview-server#180](https://github.com/yschimke/compose-preview-server/issues/180) step 4 (#5177), which flagged it as a behaviour change worth watching rather than fixing in that PR.

yschimke:
Closes #5185.

## First, the question the issue asked

The issue asserts that nothing *pins* the two rules, not that they differ. They do not differ. Read side by side:

| | `previewIdMatchesRequest` (`cli/.../Commands.kt`) | `previewIdMatchesStandaloneRequest` (server) |
| --- | --- | --- |
| `--id` | `id != exactId` → drop | same |
| `--filter` | `!id.contains(filter, ignoreCase = true)` → drop | same |
| `--preview` | delegates to `previewMatchesReference`: exact id, `Class.function`, bare function name, case-insensitive substring, OR'ed | the same four forms, inlined |
| combination | intersection | intersection |

The server's copy is a faithful inline of this repository's, down to the `className != null && functionName != null` guard on the qualified form. So this PR pins them rather than fixing a divergence.

## Which of the two options

The issue offered *pin it* or *make one go away*. Making one go away means the CLI calling the server's rule across a repository boundary — the CLI depends on the published `compose-preview-serve`, so it is mechanically possible, but it would promote an internal predicate to published API, make every non-`serve` command's selection semantics move on the server's release line, and put the CLI's behaviour behind a version bump. The rule is small and pure and sits on opposite sides of the layer split (`docs/design/REPOSITORY_LAYERS.md`), which is exactly the case a table settles more cheaply than a dependency.

## What this adds

- **`docs/serve/preview-selector-fixtures.json`** — 24 cases: `(id, className?, functionName?)` × the selectors present → expected outcome, each with the reason it is in the table. Covers the id-only call sites (no metadata, so the `--preview` metadata forms cannot fire), the deliberate looseness of `--preview Foo` also selecting `FooBar`, the empty-string selectors, and that a precise selector never promotes a preview past a loose one that misses.
- **`PreviewSelectorFixturesTest`** — runs every case through `previewIdMatchesRequest`, plus a coverage guard: each of the eight selector combinations must appear, and each non-empty one in **both** outcomes. Without that second test a rule change could add a branch and land green against a table that never reaches it, and a rule that accepted everything would satisfy a `true`-only table.
- **A `Preview selector fixtures` surface in `.github/ci/consumer-contract-notice.py`**, naming `compose-preview-server` and what it owes once this ships.
- KDoc on the rule saying it is stated twice and where the other statement lives.

Same shape as `docs/daemon/protocol-fixtures/`, and the same obligation: change the rule, change the table, in the same PR.

## The other half

[yschimke/compose-preview-server#303](https://github.com/yschimke/compose-preview-server/pull/303) vendors the same file, runs it through `previewIdMatchesStandaloneRequest`, and adds `scripts/sync-preview-selector-fixtures.sh` plus a `selector-fixtures` CI job that diffs the vendored copy against the compose-ai-tools release its `composeai-tools` pin names. It goes red on a pin bump that changed the table — the moment the obligation is real — and no-ops while the pinned release predates the file, so neither side deadlocks on the other.

## Test plan

- `./gradlew :cli:test --tests '*PreviewSelectorFixturesTest*'` — green.
- `./gradlew :cli:ktfmtCheckMain` — green.
- `python3 .github/ci/consumer-contract-notice.py docs/serve/preview-selector-fixtures.json` — reports the new surface; unrelated paths still return `NO_CONTRACT_SURFACES`.
- Server side: `./gradlew :server:test --tests '*PreviewSelectorFixturesTest*'` green on the linked PR, which is the actual agreement check — the same 24 cases, the other implementation.

Not UI-affecting: no rendered surface changes.

https://claude.ai/code/session_01M6te8X7U6KobwV82izPhxe

### DIFF
--- a/.github/ci/consumer-contract-notice.py
+++ b/.github/ci/consumer-contract-notice.py
@@ -7,6 +7,7 @@ After the split (#4732) two consumers live elsewhere:
 
   * yschimke/compose-preview-vscode    — the VS Code extension
   * yschimke/compose-preview-contracts — the published wire contracts
+  * yschimke/compose-preview-server    — the preview server behind `serve`
 
 Both pin a *released* version of this repository and vendor copies of what they
 need. Their own gates catch drift — but only against the release they pin, which
@@ -39,6 +40,7 @@ SENTINEL = "NO_CONTRACT_SURFACES"
 
 EXT = "yschimke/compose-preview-vscode"
 CONTRACTS = "yschimke/compose-preview-contracts"
+SERVER = "yschimke/compose-preview-server"
 
 # Each entry: the surface, the paths that constitute it, and per consumer what
 # that consumer has to do once this change is released. Keep the follow-ups
@@ -115,6 +117,27 @@ SURFACES: list[dict] = [
                 "**no gate covers it** — `--check` here compares only the Kotlin and C++ "
                 "mirrors, because this repository cannot write into that one. Regenerate it "
                 "there with `gen-spatial-scene.mjs --emit-typescript`."
+            ),
+        },
+    },
+    {
+        "name": "Preview selector fixtures",
+        "patterns": ["docs/serve/preview-selector-fixtures.json"],
+        "why": (
+            "The golden table for `--id` / `--filter` / `--preview`. Two implementations "
+            "answer that question — `previewIdMatchesRequest` here and "
+            "`previewIdMatchesStandaloneRequest` in the server, which builds "
+            "`ServeCommandOptions` itself now that `serve` is a launcher (#5177) — and this "
+            "table is the only thing pinning them together (#5185). The failure mode is "
+            "silent: a preview that stops matching produces no error, it is just absent."
+        ),
+        "consumers": {
+            SERVER: (
+                "vendors this at `docs/serve/preview-selector-fixtures.json` and runs it through "
+                "its own rule in `PreviewSelectorFixturesTest`. Its `selector-fixtures` CI job "
+                "diffs that copy against the compose-ai-tools release its `composeai-tools` "
+                "version pin names, so it goes red on the pin bump unless the same commit runs "
+                "`scripts/sync-preview-selector-fixtures.sh`."
             ),
         },
     },
--- a/cli/src/main/kotlin/ee/schimke/composeai/cli/Commands.kt
+++ b/cli/src/main/kotlin/ee/schimke/composeai/cli/Commands.kt
@@ -1329,6 +1329,14 @@ internal val NON_PNG_PREVIEW_KINDS = setOf("XR_SUBSPACE")
  * a manifest or a [PreviewResult] where the metadata is available, or from an already-resolved id
  * list where it isn't). They only affect `--preview`: without them the `Class.function` and bare
  * function-name forms can't be recognised and the ref falls back to its id-only forms.
+ *
+ * **This rule is stated twice.** `compose-preview serve` is a launcher (#5177), so the server
+ * builds its own `ServeCommandOptions` and answers the same question with
+ * `previewIdMatchesStandaloneRequest` in yschimke/compose-preview-server. The two live in separate
+ * repositories and are pinned against each other by a shared golden table,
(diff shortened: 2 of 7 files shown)


yschimke:
`Actions Script Tests` failed on d6df172 — not this PR's failure, and now fixed by a ported commit rather than left waiting.

**What failed:** `test_repo_is_green` in `scripts/test_check_daemon_launch_schema.py`, on two findings from `check-daemon-launch-schema`:

```
daemon/.../DaemonMain.kt: `SANDBOX_COUNT_PROP` = DaemonProperties.Names.SANDBOX_COUNT,
but daemon/protocol/.../DaemonLaunchDescriptor.kt says "composeai.daemon.sandboxCount".
```

**Why it is not this PR's:** the diff here is a JSON fixture table, a new `:cli` test, a KDoc paragraph and a `SURFACES` entry in `consumer-contract-notice.py` — it touches none of those daemon files, that checker, or its allowlist. The cause is #5182, which gave the `composeai.daemon.*` sysprops a typed registry, so the two mirrors became `DaemonProperties.Names.SANDBOX_COUNT` — a *safer* declaration that cannot drift, but one the scanner read as reference text and compared against the descriptor's string literal. Reproduced on `origin/main` with nothing of mine applied, with a contracts checkout in `COMPOSE_PREVIEW_CONTRACTS_ROOT` so the mirrored-constant check is not skipped:

| tree | `check-daemon-launch-schema.py` |
| --- | --- |
| `origin/main` (28fb93b), untouched | exit 1, same 2 findings |
| this branch at d6df172 | exit 1, same 2 findings |

**What I did:** cherry-picked 6bd60a5 from #5191 (*fix(ci): follow a sysprop mirror that reads the shared constant*) onto this branch. It teaches the resolver to follow a `const val` whose value is another constant through a new `constantRegistries` map, keeping an unregistered qualifier a failure so the gate never passes on a value it could not read. It no-ops here once #5191 lands on main — porting it beat waiting for that merge.

Verified after the port, same contracts checkout:

- `scripts/check-daemon-launch-schema.py` → `OK — 12 writer field(s), 1 reader(s) agree (jvm), 4 schema-version site(s) at v2`, exit 0
- `scripts/test_check_daemon_launch_schema.py` → 51 tests, OK (was 43 with 1 failure)

No re-run spent — the failure is real and reproducible, not a flake.

---

PR #5195, opened 2026-09-05
feat(cli): add `ui-builder`, the launcher for the server's `ui` command

yschimke:
The CLI half of yschimke/compose-preview-server#301 — the server side is yschimke/compose-preview-server#306, which gives `compose-preview-server` a `ui` command that builds a local project's previews and opens the Compose UI builder against them.

`compose-preview ui-builder` is the third launcher beside `serve` and `browse`.

It adds nothing of its own — no defaults, no flag rewriting — because unlike `browse` there is no project-mode choice to make on this side. Everything `ui` implies (discovery, the packaged builder distribution, the module's component record, the page to open) is decided by the server, which is also the half that knows where those things are. `--help` therefore reaches the server too, which is where the answer lives.

`ServeCommand` gained a `serverCommand` parameter so one launcher covers both commands rather than two copies of the discovery, argv-forwarding and exit-code handling. The Gradle work still travels the other way: the server spawns `compose-preview build-host --stdio` when it needs a build, so this process is a launcher and not a host.

Wiring, all in step because `CliRouterTest` pins it: the dispatch table, the `share` group (`serve · ui-builder · build-host · share-preview` — the same binary, its Gradle half, and now its builder command), `help --all`, and a `CliFlagValidation` allowlist. `--no-open` is listed there because it belongs to the lane rather than the server.

## Test plan

- `ServeLauncherTest` gains a case pinning that `ui-builder` launches the server's `ui` command with the caller's argv and adds nothing.
- `CliRouterTest` already pins that the dispatch table, `KNOWN_FLAT` and `CliFlagValidation.BY_COMMAND` agree — it fails if any of the three is missed.
- `./gradlew :cli:test --tests '*CliRouterTest*' --tests '*ServeLauncherTest*' --tests '*CliFlagsRegistryTest*'` green; `:cli` ktfmt clean.

Non-visual change: a command word and its routing.

https://claude.ai/code/session_015yYDifT4DZ9fmuMbraHZLL

---
_Generated by [Claude Code](https://claude.ai/code/session_015yYDifT4DZ9fmuMbraHZLL)_

### DIFF
--- a/.github/ci/consumer-contract-notice.py
+++ b/.github/ci/consumer-contract-notice.py
@@ -3,14 +3,14 @@
 
 `compose-ai-tools` used to contain every consumer of its own contracts, so a
 protocol change and the code reading it moved in one commit and CI checked both.
-After the split (#4732) two consumers live elsewhere:
+After the split (#4732) three consumers live elsewhere:
 
   * yschimke/compose-preview-vscode    — the VS Code extension
   * yschimke/compose-preview-contracts — the published wire contracts
   * yschimke/compose-preview-server    — the preview server behind `serve`
 
-Both pin a *released* version of this repository and vendor copies of what they
-need. Their own gates catch drift — but only against the release they pin, which
+Each pins a *released* version of this repository and vendors copies of what it
+needs. Their own gates catch drift — but only against the release they pin, which
 means a contract change here is invisible to them until somebody bumps that pin.
 That is the gap #4732 records as "a `daemon-protocol` bump has to fail the
 consumer's CI, not just the producer's".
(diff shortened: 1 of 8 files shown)


yschimke:
`Actions Script Tests` failed on `f693e5b`, and it is not this PR's: this branch changes nothing under `.github/` (`git diff origin/main -- .github/` is empty), and the failure reproduces at the base commit.

```
FAIL: test_consumer_repos_are_the_known_two (SurfaceMapIsCurrent)
AssertionError: {'yschimke/compose-preview-server'} not less than or equal to
  {'yschimke/compose-preview-vscode', 'yschimke/compose-preview-contracts'} : Preview selector fixtures
```

#5193 added the server to the surface map in `.github/ci/consumer-contract-notice.py` without widening the `known` set in the test that guards it, so `main` has been red since it landed.

Fixed in #5196, and ported here as `3e13c3f` so this PR does not wait on that merge — it no-ops once `main` carries it. `python3 .github/ci/test_consumer_contract_notice.py` (the same command the job runs) is now 13/13 on this head.

---

Issue #5183, opened 2026-09-05
`compose-preview serve` needs a server binary the installer does not fetch


yschimke:
**User-visible regression from #5177.** Filing it against this repository because this is where the broken command lives, but the fix is in [`yschimke/skills`](https://github.com/yschimke/skills), which owns the installer.

## What happens now

`serve` and `browse` stopped linking the preview server and became launchers — that is what closed the dependency cycle ([compose-preview-server#180](https://github.com/yschimke/compose-preview-server/issues/180)). They exec `compose-preview-server`, found by `--server-binary`, then `COMPOSE_PREVIEW_SERVER`, then `PATH`.

**Nothing installs that binary.** The one-line installer in `yschimke/skills` fetches the CLI and the skill bundle. So for anyone who installs the documented way, `compose-preview serve` now prints an installation hint and exits 1 instead of serving.

Every other command is unaffected — `render`, `show`, `bundle`, `history`, `a11y`, `mcp` need no server.

## What needs to happen

In the installer:

- Fetch the `compose-preview-server` distribution alongside the CLI, at a version that matches. The server publishes as `ee.schimke.composeai:compose-preview-serve` (3.0.0 at time of writing) and its distribution provides the binary.
- Put it where the CLI's discovery finds it — on `PATH`, or set `COMPOSE_PREVIEW_SERVER`.
- Decide whether it is unconditional or opt-in. It is a bigger download than the CLI and only `serve` / `browse` / a future `ui` need it, so "fetch on first use" is worth considering over "fetch always".

Possibly here instead, and worth deciding rather than assuming: the CLI already resolves Maven coordinates to local jars for `bundle daemon` and `bundle render` (`:bundle-coordinates`). `serve` could resolve and cache the server itself on first use, which would need no installer change at all. That is more code here but removes a cross-repo coupling from the install story.

## Acceptance

- A fresh install of the documented one-liner can run `compose-preview serve` without further steps.
- The version relationship between CLI and server is stated somewhere — they release on separate cadences, and a launcher execing an arbitrarily old server is a new way to skew.
- `compose-preview doctor` reports whether the server binary is present and which one it found, the way it reports other environment facts.

## Until then

The hint `serve` prints names the binary, the environment variable and the flag, so someone who has the server can point at it. That is a workaround, not a fix.

PR #5200, opened 2026-09-05
feat(mcp): launch the MCP server instead of bundling it

### LINKED ISSUE #5189: Move :mcp to compose-preview-server and make `compose-preview mcp` a launcher
Steps 2 and 3 of the decision on #5176: `:mcp` is layer 2 by the letter of the rule in `docs/design/REPOSITORY_LAYERS.md`, and moves.

Step 1 landed in https://github.com/yschimke/compose-ai-tools/pull/5188 — `MatrixAxes`, `MatrixCell` and `ContactSheet` are `:render-matrix` now, so the CLI's offline `render-matrix` command no longer compiles against an MCP server.

## What is left

2. **Stand the module up in compose-preview-server**, against the published layer-1 coordinates it currently reaches as projects: `daemon-core`, `render-session-api`, `daemon-client` (all `api` on `:mcp`'s surface) and `data-layoutinspector-core`. That is the cost the decision accepted — those become published surface across the boundary rather than project dependencies.
3. **Turn `compose-preview mcp` into a launcher and delete `:mcp` here**, the way `serve` and `browse` became launchers over the published server binary (compose-preview-server#180). `install` and `doctor` write local config and stay CLI-side; `serve` execs the published MCP binary.

## Done looks like

`CheckHttpServerFloor.httpServerProjects` is empty. It holds `:mcp` and the `:cli` that bundles it today, and emptying it is what removes the last five `ktor-server-*` artifacts from the CLI distribution — the ones the `serve` change was expected to take and did not.

Also worth checking on the way: `:mcp` carries layer-1 behaviour that only it uses today (`DaemonSupervisor`, `HistoryStore`, `DaemonClientRenderSession`, `WatchPropagator`, `FreshnessMetrics`). Anything a layer-1 caller acquires before the move has to be lifted first, exactly as `:daemon-client` (#3824) and `:render-matrix` were.

yschimke:
Closes #5189. Step 3 of the move #5176 decided, and the end of it: `:mcp` is gone from this repository, `compose-preview mcp serve` execs the `compose-preview-mcp` binary published from compose-preview-server ([#308 there](https://github.com/yschimke/compose-preview-server/pull/308)), and `CheckHttpServerFloor`'s allowlist is **empty**.

## The launcher

`serve` is the precedent and now the shared implementation. `ReleasedDistribution` names the two things this CLI launches but does not contain, and `ServerBinaryDiscovery` / `ServerDistributionProvision` take one instead of growing a second copy of the flag → env → PATH → cache ordering and the fetch-on-first-use. `--mcp-binary` / `COMPOSE_PREVIEW_MCP` mirror the server's pair; the cache roots are separate so an interrupted fetch of one can never satisfy the other's completeness check.

Two ways it differs from `serve`, both pinned by tests: no subcommand word is added (`compose-preview-mcp` *is* the server), and stdio is inherited rather than piped — the agent host's streams reach the MCP server directly, so this process cannot reframe or buffer a JSON-RPC message.

**`mcp install` and `mcp doctor` do not move.** They read Gradle descriptors, run discovery and write agent-host config — offline behaviour that opens no socket — and the config they write still names *this* CLI as the agent's command, so the flow an agent follows is unchanged.

## One pin, both archives

Both distributions ride the same compose-preview-server release, so a single reviewed pin governs the pair, and `check_preview_server_pin.py` now fails a pin whose release carries one and not the other. A `serve` that works beside an `mcp serve` that 404s is exactly the silent half-break that gate exists to catch — it is the #5183 shape, one command over.

## What actually left the distribution

Measured by diffing the built `lib/` against main's, not claimed: **sixteen jars, seventy to fifty-four, nothing added.**

`compose-preview-mcp`, `kotlin-sdk-core`, `kotlin-sdk-server`, the five `ktor-server-*` the issue named (`-core`, `-cio`, `-content-negotiation`, `-sse`, `-websockets`), the two `ktor-serialization-*` behind them, `ui-builder-protocol` — and the ones nobody predicted: `kotlin-reflect`, `config`, `kotlin-logging`, `kotlinx-collections-immutable`, `kotlinx-serialization-json-io`. An offline CLI was carrying a reflection library because an MCP SDK wanted one.

## Test plan

```
./gradlew :cli:test checkHttpServerFloor      # the floor across every project, empty allowlist
python3 .github/ci/test_check_preview_server_pin.py   # 26 tests, incl. the new half
.github/scripts/test-sweep-stranded-releases.sh
./gradlew :cli:installDist                    # lib/ diffed for the numbers above
```

All green.

## The one blocker

**The pin stays at 3.0.0 and the new gate half fails until compose-preview-server cuts a release carrying `compose-preview-mcp-<version>.tar.gz`.** v3.0.0 predates #308, so no published release has that asset yet. The gate is red *correctly* — that is the state it was written to catch — and the fix is one line here once that release exists. Everything else in the diff is green.

Non-visual change: a launcher, build wiring, CI gates and docs.

https://claude.ai/code/session_016Uxgbygt7e2LEBDruCGeKY

---
_Generated by [Claude Code](https://claude.ai/code/session_016Uxgbygt7e2LEBDruCGeKY)_


yschimke:
`XR Composite Pin` is red on `check_preview_server_pin.py`, and it is this PR's own new gate half doing exactly what it was written to do:

```
yschimke/compose-preview-server v3.0.0 does not carry compose-preview-mcp-3.0.0.tar.gz.
  Attached to that Release:
    compose-preview-server-3.0.0.tar.gz
```

v3.0.0 predates [compose-preview-server#308](https://github.com/yschimke/compose-preview-server/pull/308), so no published release has the MCP archive yet. That is not a defect in the diff — it is the gate refusing to let a `compose-preview mcp serve` ship that would 404 on first use, which is the #5183 shape one command over.

**No fix to push, and this should not merge until there is one.** What unblocks it, in order:

1. compose-preview-server cuts a release (its `main` carries `:mcp` and the tarball upload since #308).
2. `composeai-preview-serve` moves from `3.0.0` to that version in `gradle/libs.versions.toml` — one line, and the gate goes green because the release then carries both archives.

Everything else on this head is green: `:cli:test`, `checkHttpServerFloor` across every project on an empty allowlist, the 26 pin-gate unit tests (the offline half passes; only the online half fails), and the release-sweeper script tests.

I'm watching this PR and will push the pin bump as soon as that release exists.

---

yschimke:
Bump to 3.1.0?

yschimke:
Yes — bumping to 3.1.0. v3.1.0 published a few minutes ago carrying both archives, so the pin gate's online half now passes locally:

```
ok: yschimke/compose-preview-server v3.1.0 exists and carries
    compose-preview-server-3.1.0.tar.gz, compose-preview-mcp-3.1.0.tar.gz
```

Holding the push for one thing: `compose-preview-serve:3.1.0` is not on Maven Central yet (the GitHub release assets are up, the Central sync lags), and `:cli` resolves that coordinate as a `testImplementation` for the two wire-drift tests. Pushing now would turn CI red on dependency resolution rather than on anything in the diff. I'm polling Central and will push the bump as soon as `:cli:test` passes against it.

---

yschimke:
Central caught up, and the bump surfaced a second problem — this one not in this PR.

**`compose-preview-serve:3.1.0` is unresolvable.** Its POM names

```xml
<groupId>compose-preview-server</groupId>
<artifactId>ui-builder-export-jvm</artifactId>
<version>unspecified</version>
```

so `:cli:testRuntimeClasspath` fails at resolution, not at anything in this diff. `:server` gained `implementation(project(":ui-builder-export"))` over there while that module had no publishing configuration, and a project dependency reaches the POM as a coordinate. 3.0.0's POM has no such entry.

Fix is up as [compose-preview-server#316](https://github.com/yschimke/compose-preview-server/pull/316): publish the module in lockstep, plus `checkPublishedPomCoordinates` reading the generated POM so the next unpublished project dependency fails the build instead of the download.

So this PR now waits on a **3.1.1** release rather than on 3.1.0. The pin bump is committed here locally and I'll push it once there is a release that both carries `compose-preview-mcp-<v>.tar.gz` and resolves — pushing 3.1.0 would only trade one red check for another.

---

yschimke:
`Module Unit Tests` was red on `:daemon:core:checkKotlinAbi`, and that one is **not this PR's**: it fails identically on `origin/main` at 734e906 in an untouched checkout. The committed dump claims a synthetic constructor `RpcMethodRegistry` no longer emits.

Fix opened as [#5201](https://github.com/yschimke/compose-ai-tools/pull/5201) and **ported here** (c680968), so this PR does not wait on it — it no-ops once main carries it.

Also **dropped the 3.1.0 pin bump** from this branch rather than pushing it. 3.1.0 carries the MCP tarball but its `compose-preview-serve` POM is unresolvable ([compose-preview-server#316](https://github.com/yschimke/compose-preview-server/pull/316) fixes that), so pinning it would trade the pin-gate red for a dependency-resolution red. The pin stays at 3.0.0 until 3.1.1 exists, and that remains this PR's one expected failure.

State on this head: `:cli:test`, `:daemon:core:checkKotlinAbi` and `checkHttpServerFloor` all green locally; `check_preview_server_pin.py`'s online half is the only red, for the documented reason.

---

PR #5203, opened 2026-09-05
feat(screen): let a generated screen carry an opt-in @Preview

yschimke:
## Summary

A UI-builder design exported as Compose comes back as a bare `@Composable` — correct Kotlin, but pasting it into Android Studio shows nothing until you hand-write a `@Preview` wrapper and its import. This adds one, as an **opt-in** parameter:

```kotlin
ScreenGenerator.generate(document, components, preview = ScreenGenerator.Preview(
  widthDp = 411, heightDp = 914, fontScale = 1.3, locale = "en-US", darkMode = true,
))
```

emits, after the screen:

```kotlin
import androidx.compose.ui.tooling.preview.Preview
…
@Preview(
    widthDp = 411,
    heightDp = 914,
    fontScale = 1.3f,
    locale = "en-US",
    showBackground = true,
    uiMode = android.content.res.Configuration.UI_MODE_NIGHT_YES,
)
@Composable
private fun HomeScreenPreview() {
    HomeScreen()
}
```

**Opt-in rather than always.** `@Preview` lives in `androidx.compose.ui.tooling.preview`, an Android tooling dependency a consumer of this generator need not have on its compile classpath — the Gradle plugin generates screens into builds that do not. Emitting it unconditionally would trade "the file compiles" for "the file previews", which is backwards for a caller that only wanted source. Default is `null`; nothing changes for existing callers.

**A wrapper, not an annotation on the screen.** The screen is the artifact you paste and then *call* from your own code; annotating it directly would put a preview on every call site, and `@Preview` expects a zero-argument function — a constraint belonging to the preview, not the screen.

**The design's environment, not `@Preview`'s defaults.** A design authored at 411×914 in dark at 1.3 font scale, previewed at Android Studio's defaults, is a different picture from the one its author approved — which defeats the point of pasting it in. A light design gets no `uiMode` at all rather than `UI_MODE_NIGHT_NO`: unspecified and "explicitly not night" are not the same thing to a device configuration.

## Two hazards this closes rather than leaves open

**`locale` is wire data reaching a string literal.** It arrives from a design document over HTTP. A value carrying a quote would close the literal and continue in code — in a file this generator signs. It is validated against a language-tag shape and refused otherwise, rather than escaped: a locale is not typed prose, and a wrong one no Android runtime resolves is worth naming, not smuggling through.

**Two names the file now spends can shadow a catalog component.** `Preview` itself, and the wrapper `<Screen>Preview`. The second is the sharper one — the wrapper is a top-level declaration, so it *beats* an import of the same simple name, and a component called `HomeScreenPreview` placed in a `HomeScreen` would silently become a call to the wrapper, which calls the screen, which renders it: a stack overflow standing in for the component somebody placed. Both are added to the existing simple-import exclusion, and **only when a preview is emitted**, so no other screen pays for it.

Non-finite/non-positive sizes and font scales are refused too, all problems at once rather than one per run. `!(it > 0)` for the scale so `NaN` — which loses every comparison — is caught here instead of reaching the file as `fontScale = NaNf`.

## Test plan

Six new tests in `ScreenGeneratorTest`:

- no preview unless asked for (no annotation, no tooling import)
- the full environment maps through, `fontScale` as a `Float` literal, wrapper emitted after the screen
- a light design gets no `uiMode`
- a locale that is not a language tag is refused, not escaped in
- zero/negative size and zero font scale are refused, and all three problems are named
- a component named `Preview`, and one named `HomeScreenPreview`, are called fully qualified when a preview is emitted — and still simply imported when it is not

Verified locally:

- `:gradle-plugin:preview-discovery:test` — green
- `:screen-model:compileKotlinWasmJs` — green; this generator is shared with the browser editor via the `screen/generator` source directory, so the wasm target has to keep compiling
- `:gradle-plugin:preview-discovery:ktfmtCheckMain` / `ktfmtCheckTest` — clean

No ABI dump covers these modules, so none needed updating.

## Follow-up

`compose-preview-server`'s `ScreenGeneratorComposeExportExecutor` is the caller that should pass this, built from `DesignEnvironmentV1`. That needs a release of `screen-model` first, so it lands separately.

https://claude.ai/code/session_01UA6pgDdzR5xgKh6F7dG6eJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01UA6pgDdzR5xgKh6F7dG6eJ)_

### DIFF
--- a/gradle-plugin/preview-discovery/src/test/kotlin/ee/schimke/composeai/discovery/ScreenGeneratorTest.kt
+++ b/gradle-plugin/preview-discovery/src/test/kotlin/ee/schimke/composeai/discovery/ScreenGeneratorTest.kt
@@ -64,6 +64,39 @@ class ScreenGeneratorTest {
   private fun refusal(document: ScreenDocument, catalog: ComponentRecordFile) =
     (ScreenGenerator.generate(document, catalog) as ScreenGenerator.Result.Refused).reasons
 
+  private fun generated(
+    document: ScreenDocument,
+    catalog: ComponentRecordFile,
+    preview: ScreenGenerator.Preview,
+  ) =
+    ScreenGenerator.generate(document, catalog, preview = preview) as ScreenGenerator.Result.Emitted
+
+  private fun previewRefusal(
+    document: ScreenDocument,
+    catalog: ComponentRecordFile,
+    preview: ScreenGenerator.Preview,
+  ) =
+    (ScreenGenerator.generate(document, catalog, preview = preview)
+        as ScreenGenerator.Result.Refused)
+      .reasons
+
+  /** The same nested card-and-text screen several preview tests generate from. */
+  private fun screen() =
+    ScreenDocument(
+      name = "HomeScreen",
+      root =
+        ScreenNode(
+          componentId = card.canonicalId,
+          slots =
+            mapOf(
+              "content" to
+                listOf(
+                  ScreenNode(text.canonicalId, arguments = mapOf("text" to ScreenValue.Text("Hi")))
+                )
+            ),
+        ),
+    )
+
   @Test
   fun `a screen binds the document's values and nests children into slots`() {
     val screen =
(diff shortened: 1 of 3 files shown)


yschimke:
`XR Composite Pin` was red here, and it is not this PR's failure — this diff is a code generator and cannot affect which assets a GitHub Release carries.

```
yschimke/compose-preview-server v3.0.0 does not carry compose-preview-mcp-3.0.0.tar.gz.
  Attached to that Release:
    compose-preview-server-3.0.0.tar.gz
```

`check_preview_server_pin.py` resolves the live release named by `composeai-preview-serve`, still `3.0.0` on `main`. #5200 landed the `compose-preview mcp serve` launcher against a release predating the MCP module's move, so the archive it fetches was never attached. The gate was `skipped` on `main`'s own run, which is why `main` looks green while every PR that scopes it in goes red.

#5202 is the fix and I have ported its commit here verbatim, so this branch is green on the gate now rather than waiting on that merge; it no-ops once `main` carries it.

Worth recording why the pin skips 3.1.0, since I verified it independently before finding #5202: that release carries the MCP archive but its published `compose-preview-serve` POM names `compose-preview-server:ui-builder-export-jvm:unspecified` — an unpublished project dependency, with the Gradle root project name as its group. `:cli:test` cannot resolve it at all. 3.0.0's POM has 33 dependencies and none broken; 3.1.0 has 35 and one unresolvable. compose-preview-server#316 fixed it and 3.2.0's POM is clean (35 dependencies, zero `unspecified`).

Verified on this branch after the port:

```
python3 .github/ci/check_preview_server_pin.py
ok: yschimke/compose-preview-server v3.2.0 exists and carries
    compose-preview-server-3.2.0.tar.gz, compose-preview-mcp-3.2.0.tar.gz

./gradlew :cli:test                                  # green — resolves 3.2.0, which 3.1.0 could not
./gradlew :gradle-plugin:preview-discovery:test      # green
```

---

PR #5252, opened 2026-09-06
fix: merge the duplicate env block that broke the v2.1.0 release

yschimke:
**Releases are broken on `main` right now.** My #5251 left `release.yml`'s Maven Central publish step with two `env:` keys — the existing credentials block, and the `CORE_NEEDED` / `DATA_NEEDED` pair added for per-train gating.

```
Invalid workflow file: .github/workflows/release-please.yml#L187
error parsing called workflow
".github/workflows/release-please.yml"
-> "./.github/workflows/release.yml" (source branch with sha:57ae39ea9)
: (Line: 310, Col: 9): 'env' is already defined
```

Because `release.yml` is a *called* workflow, the rejection failed the whole v2.1.0 `release-please` run with **zero jobs, no logs, and nothing pointing at the cause**.

## Why it got through

Every "does this file parse?" check I ran used PyYAML's `safe_load`, which **silently keeps the last of a duplicated key** instead of erroring. Actions does the opposite and refuses the file.

The silent-discard behaviour also made the bug worse than a plain parse error: the block that survived was the *second* one, so the Maven Central credentials had been dropped from the step. Had Actions been as permissive as PyYAML, the publish would have run without them.

## The fix

One merged `env:` block carrying all four variables. The per-train gating logic is untouched — this is purely the YAML defect.

Plus `check-workflow-yaml.sh`, which models what Actions actually does: a `SafeLoader` subclass that refuses duplicate mapping keys, run over every workflow file. Wired into CI beside the other script tests.

## Test plan

- **Reproduced the failure and proved the checker catches it.** Run against the broken commit `57ae39ea9`:
  ```
  ::error file=...::duplicate key — line 310, column 9: 'env' is already defined
  1 workflow file(s) rejected.
  ```
  That is GitHub's own message — same line, same column, same wording — derived independently.
- Against the fixed tree: `40 workflow file(s) OK.`
- Confirmed all four variables survive the merge: `ORG_GRADLE_PROJECT_mavenCentralUsername`, `ORG_GRADLE_PROJECT_mavenCentralPassword`, `CORE_NEEDED`, `DATA_NEEDED`.

## One thing I checked and did *not* change

My first hypothesis was the folded `if:` I added, on the theory that a more-indented continuation line preserves a newline inside `${{ }}`. It does preserve the newline — but that is not the bug: four conditions already on `main` (`ci.yml`, `claude.yml`, `pr-body-syntax.yml`, `pr-commands.yml`) wrap the same way and run fine. I reverted that edit rather than ship a change justified by a wrong diagnosis.

Non-visual change; no render evidence applicable.

## Recovery for v2.1.0: nothing to undo

The run was rejected at parse time, so **no job ran at all** — including release-please's own. Verified: there is no `v2.1.0` tag (latest is `v2.0.0`), no draft release, and nothing reached Maven Central.

So there is no repair path to follow. Merging this restores `release.yml`, and the merge itself re-triggers `release-please`, which re-proposes the 2.1.0 release PR from the same commits.

https://claude.ai/code/session_01VDdkQ4Ch9VwYPJav4dbED2

### DIFF
--- a/.github/scripts/check-workflow-yaml.sh
+++ b/.github/scripts/check-workflow-yaml.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+# Reject duplicate keys in workflow YAML.
+#
+# PyYAML's `safe_load` — and every "does this file parse?" check built on it — silently keeps the
+# LAST of a duplicated key and discards the rest. GitHub Actions does the opposite: it refuses the
+# whole file. So a workflow can pass local validation and still be rejected at parse time, and the
+# rejection is about as unhelpful as CI failures get: the calling run fails with ZERO jobs, no logs
+# and no annotation on the job that caused it.
+#
+# That is not hypothetical. A second `env:` on one step of `release.yml` shipped exactly that way:
+#
+#     Invalid workflow file: .github/workflows/release-please.yml#L187
+#     error parsing called workflow ... (Line: 310, Col: 9): 'env' is already defined
+#
+# It failed the v2.1.0 release. And the silent-discard behaviour made it worse than a parse error
+# would have been: the surviving `env:` was the second one, so the Maven Central credentials had
+# been dropped from the step. Had Actions been as permissive as PyYAML, the publish would have run
+# without them.
+#
+# Usage: check-workflow-yaml.sh [<file>...]     # defaults to .github/workflows/*.yml
+set -euo pipefail
+
+cd "$(dirname "$0")/../.."
+
+if [ "$#" -gt 0 ]; then
+  files=("$@")
+else
+  mapfile -t files < <(find .github/workflows -maxdepth 1 -name '*.yml' -o -maxdepth 1 -name '*.yaml' | sort)
+fi
+
+python3 - "${files[@]}" <<'PY'
+import sys
+import yaml
+
+
+class DuplicateKeyError(Exception):
+    pass
+
+
+class StrictLoader(yaml.SafeLoader):
+    """A SafeLoader that refuses duplicate mapping keys, the way Actions does."""
+
+
+def _no_duplicates(loader, node, deep=False):
+    seen = {}
+    for key_node, value_node in node.value:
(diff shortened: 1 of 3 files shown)


yschimke:
**The two failing checks are not this PR's diff, and they cannot go green until this PR merges.** Flagging the ordering because it is a deadlock, not a normal red build.

`Apply compose-preview` and `A11y + notification previews` both die before running anything:

```
##[error]Unable to resolve action `yschimke/compose-ai-tools@v2.1.0`, unable to find version `v2.1.0`
```

Six composite actions self-reference the released tag, marked for release-please to bump:

```
.github/actions/apply/action.yml:715          …/install-cli@v2.1.0  # x-release-please-version
.github/actions/apply/action.yml:725          …/install@v2.1.0      # x-release-please-version
.github/actions/a11y-report/action.yml:33     …/apply@v2.1.0        # x-release-please-version
.github/actions/notification-previews/…:37    …/apply@v2.1.0        # x-release-please-version
.github/actions/preview-baselines/…:47        …/apply@v2.1.0        # x-release-please-version
.github/actions/preview-comment/…:50          …/apply@v2.1.0        # x-release-please-version
```

Merging the release PR bumped all six to `v2.1.0` and set the manifest to `2.1.0` — but the release run that would have *created* that tag was rejected at parse time by the duplicate `env:` this PR fixes. So `main` now points every preview job at a tag that does not exist, and **every PR in the repo is red for the same reason**, not just this one.

`git tag -l v2.1.0` is empty; latest is `v2.0.0`.

**The deadlock:** these checks need `v2.1.0` to exist → the tag needs release-please to complete → release-please needs a parseable `release.yml` → that is this PR. It cannot make itself green.

**Recovery:** merge this PR (the two red checks need an admin override if they are required). The merge push re-runs `release-please.yml`, which finds the still-pending release and cuts `v2.1.0`; the tag then resolves and CI recovers repo-wide on the next run.

Nothing reached Maven Central, so there is nothing irreversible to unwind.

---

Issue #525, opened 2026-05-02
 Parse `shape=round` in `spec:` device strings


yschimke:
**<sub><sub>![P2 Badge](https://img.shields.io/badge/P2-yellow?style=flat)</sub></sub>  Parse `shape=round` in `spec:` device strings**

The new `spec:` parsing path only sets `isRound` from `isRound=true`, so inputs like `spec:shape=round,width=...,height=...,dpi=...` still resolve with `isRound=false`. That drops round-display metadata for callers using the standard `shape` form, even though the new helper already treats `shape=round` as a round marker. Please include `shape=round` (and case-insensitive variants) in the `spec:` branch so round specs are preserved consistently (this same logic exists in both `gradle-plugin` and `daemon/core` copies).

Useful? React with 👍 / 👎.

_Originally posted by @chatgpt-codex-connector in https://github.com/yschimke/compose-ai-tools/pull/522#discussion_r3176948597_

PR #5258, opened 2026-09-07
fix(samples): regenerate the Wear widget fixture from the current generator

yschimke:
## Why

`samples/wear-widget/src/main/kotlin/com/example/wearwidget/ActivitySummaryWidget.kt` is generated by `WearWidgetCodeExporter` in [`yschimke/compose-preview-server`](https://github.com/yschimke/compose-preview-server) and checked in here, because compiling and rendering it is the half of "the generator works" that no test on that side of the split can answer.

Two changes landed there, and the checked-in copy predated both — so the fixture had quietly stopped being what the generator writes:

| | |
| --- | --- |
| [#525](https://github.com/yschimke/compose-preview-server/pull/525) | A row's own `verticalAlignment` reaches the generated `RemoteRow`. The canvas has always read it and defaults a row to `CenterVertically`; the emitter read only the children's modifiers, so a generated widget drew top-aligned. |
| [#524](https://github.com/yschimke/compose-preview-server/pull/524) | The generated `@Preview` is one preview at the container's widest footprint rather than `@PreviewParameter`'s fan-out. |

## What it looks like

| Before (stale fixture) | After (regenerated) |
| --- | --- |
| steps sitting at the top of its row | steps centred against the number |

`steps` sat at the top of its row, against the top of `8,412`; it is now centred against it. Both are this fixture compiled into `:samples:wear-widget` and rendered by its own preview lane at 216×124dp.

The preview change is not visible in a single frame — it shows up as **one** render file instead of two: `…Squircle_Preview-a0071b12.png` replacing `…_PARAM_0.png` and `…_PARAM_1.png`.

## The diff

Six lines, and nothing hand-edited — the file is the generator's output through `ktfmtFormat`:

```diff
-import androidx.compose.ui.tooling.preview.PreviewParameter
         RemoteRow(
           modifier = RemoteModifier.fillMaxWidth(),
           horizontalArrangement = RemoteArrangement.spacedBy(6.rdp),
+          verticalAlignment = RemoteAlignment.CenterVertically,
         ) {
-fun ActivitySummaryWidgetSquirclePreview(
-  @PreviewParameter(SquircleLargeWidgetPreviewParams::class) params: WearWidgetParams
-) = WearWidgetPreview(ActivitySummaryWidget(), params)
+fun ActivitySummaryWidgetSquirclePreview() =
+  WearWidgetPreview(
+    ActivitySummaryWidget(),
+    SquircleLargeWidgetPreviewParams().values.maxBy { it.widthDp },
+  )
```

## Test plan

- `./gradlew :samples:wear-widget:composePreviewRenderAll` — renders, producing the "after" image above.
- `./gradlew :samples:wear-widget:testDebugUnitTest` — green. Worth stating explicitly: `WearWidgetDocCaptureTest` reads `.rc` and PNGs out of `build/compose-previews/renders/`, and dropping the fan-out **changes those filenames** from a `_PARAM_n` pair to a single file. I ran it for that reason rather than as a formality, and it passes.
- `./gradlew :samples:wear-widget:ktfmtCheckMain` — clean.

---

https://claude.ai/code/session_01FjLofgNZiAcViCWYxucmHq

---
_Generated by [Claude Code](https://claude.ai/code/session_01FjLofgNZiAcViCWYxucmHq)_


yschimke:
On the accessibility report above: I checked it rather than assuming, and it is **not this PR's** — filed as #5259.

Five errors, all identical, all `SpeakableTextPresentCheck` on `androidx.compose.remote.player.view.RemoteComposePlayer`. Two things rule this diff out:

- **Four of the five flagged previews are files this PR does not touch** — `ImageWidgetDeviceSpecPreview`, `ImageWidgetFixedPreview`, `ImageWidgetSquircleLargePreview`, `ImageWidgetSquirclePreview` all live in `WearWidgetPreviews.kt` / `RemoteImageWidget.kt`, neither of which is in the diff.
- **The diff cannot affect labels.** It is six lines: a `verticalAlignment` argument on a `RemoteRow`, and a preview signature losing its `@PreviewParameter`. Nothing about semantics, content descriptions, or the view tree.

The cause is structural: a Remote Compose widget is played rather than composed, so the whole document renders inside one `RemoteComposePlayer` view, and the scanner sees a single unlabelled view. It is a fixed floor of one error per widget preview, independent of what the widget contains — which is why five completely different widgets produce byte-identical findings.

Fixing it here would turn a mechanical regeneration into an accessibility change, so I have left it to #5259, which sets out the three possible shapes.

---

PR #5267, opened 2026-09-07
fix(deps): update rc-players to 1.59.3

yschimke:
`rcplayers` 1.57.0 → 1.59.3, four releases on — and **`RC_PLAYERS_PIN` moves with it**, which is the half a version-catalog bump alone would miss.

## The pin is the point

`design-artifacts-reusable.yml` checks `RC_PLAYERS_PIN` against the `rcplayers` the export driver consumes, and skips the **vendored-Android** and **cmp-jvm** columns on any mismatch. The pin was already at `v1.59.2` — *ahead* of the 1.57.0 catalog, which that check calls the ordinary self-healing direction. Moving `rcplayers` to 1.59.3 while leaving the pin would have flipped it to *behind*, which the same check calls an editing mistake, and two columns would have quietly vanished from the wall. Both literals now read 1.59.3.

## What the four releases carry

| release | change |
| --- | --- |
| 1.58.0 | a lenient CMP playback mode, text on a circle (op 57), the vendored Horologist Lottie → Remote Compose compiler |
| 1.59.0 | font variations get the player's own type — **breaking**, see below |
| 1.59.2 | a derived colour reaching the text that names it, a disabled label drawn at all, letter spacing read as an em multiple |
| 1.59.3 | a NaN-boxed paint float resolved rather than read as a literal, a bitmap declared in a component's canvas stream registered, the `google:Inter` faces vendored |

## The 1.59.0 breaking change does not reach this repo — checked, not assumed

`RcTypefaceLoader.typeface` and `RcFontFaces.family` take `RcFontVariations?` instead of `FontVariation.Settings?`. Both live in **`rc-player-compose`**. This repo declares that coordinate in the catalog but **no build script consumes it** (`rcplayer.compose`, `.runtime`, `.protocol`, `.trace` are each referenced zero times outside the catalog), and nothing here implements `RcTypefaceLoader`. The only `FontVariation.Settings` in the tree is `androidx.compose.ui.text.font.FontVariation`, reached reflectively in `ComposeSemanticsDataProduct.kt` — Compose's own type, unrelated.

The five coordinates actually consumed are the two `dist` zips (`rcplayer-wasm-dist`, `rcplayer-js-dist`) and the embedded players (`rcplayer-embedded-android`, `rcplayer-embedded-jvm`) — none of which carry that API.

## Test plan

Build and the design-artifacts lanes, once Maven Central carries the artifacts.

**CI will be red until then.** v1.59.3 was tagged and released at 09:41 UTC; all eight coordinates were still 404 on Central when this was pushed. It clears with a re-run once the sync completes.

https://claude.ai/code/session_01DNvhkeMBYsxwduigaVvXqy

---
_Generated by [Claude Code](https://claude.ai/code/session_01DNvhkeMBYsxwduigaVvXqy)_


yschimke:
Two corrections to the PR body, and a real fix pushed as `a47ca9c`.

## There is a THIRD pin, and it is the one that mattered

Once Central synced, `Build plugin + CLI` failed again — a real failure, not the 404. Reproduced locally with the same command the job runs:

```
Could not resolve ee.schimke.composeai:third-party-rc-embedded-player:1.59.3.
   > Cannot find a version that satisfies the version constraints:
       Dependency path:  … --> third-party-rc-embedded-player:1.59.3
       Constraint path:  … --> third-party-rc-embedded-player:{strictly 1.57.0}
       because of the following reason: Dependency version enforced by Dependency Locking
```

**Gradle dependency locking** pins this coordinate independently of the version catalog, and `data/remotecompose/connector/gradle.lockfile` still held `1.57.0`. So the two pins disagreed and resolution failed — nothing to do with the artifact being absent.

Regenerated with the repo's own tooling rather than by hand, matching what `.github/workflows/dependency-locks.yml` runs:

```
./gradlew resolveAndLockAll --write-locks --no-configuration-cache
```

One line moves. It is the only lockfile of the **94** in the tree that names an rc-players coordinate. Verified by reproducing the failure on this branch and re-running `./gradlew publishToMavenLocal` clean afterwards — **BUILD SUCCESSFUL**, 1285 tasks.

## `RC_PLAYERS_PIN` was not mine to catch

The body presents the pin bump as the half a catalog-only change would miss. That is true of the *invariant*, but I overstated my part in it: Renovate keeps that literal in step automatically through its `# renovate: datasource=github-tags depName=yschimke/rc-players` annotation, and did exactly this bump independently in #5268 while I was working. Both sides set it to `v1.59.3`, so the merge is clean — I test-merged `origin/main` to confirm.

The pin genuinely does have to move with the catalog, and the workflow's own check enforces it. It just is not something a human or an agent needs to remember here.

---

Issue #5279, opened 2026-09-07
figma-svg: three published catalogs export every text run as missing-glyph boxes


yschimke:
`preview.coo.ee/remote-m3/compare` serves every SVG in the comparison wall as missing-glyph boxes, beside a Rendered PNG that draws real type. Reproduce right now:

```
$ curl -s https://preview.coo.ee/remote-m3/render/appcard__ideal__content-image__compact.svg \
    | grep -c 'ComposeAI Missing Font'
5
```

Two other catalogs are in the same state — `jetcaster` and `jetcaster-wear` — in a shape that may not share a cause; see the last section.

## The boxes are the guard, working as designed

`TofuFont` builds a real TrueType face whose one glyph is a hollow rectangle and maps every emitted code point to it. `ComposeFigmaSvgDataProducer` switches the whole document to it when `FigmaSvgRenderedFonts.unnamedIn(named)` is non-empty — the render drew a face the export cannot name, so it draws boxes rather than silently substituting the Material default. That decision is right and is not what this issue asks to change.

What is wrong is that the guard is firing on a document the export **can** reproduce. The served SVG has:

- the correct faces embedded — two real Roboto woff2, at weights **550** and **450**
- **no `font-family` on any `<text>`**; every run inherits the document default
- `google:Roboto Flex` in the `.rc` the same page serves

So the render drew Roboto Flex, the recorder published `Roboto Flex`, and the export — seeing no captured family — named the default `Roboto` via `FigmaLayeredSvg.embedFamily(null, …)`. `unnamedIn` compares `Roboto Flex` against `Roboto`, correctly concludes a face was lost, and boxes the document.

## Root cause (measured)

`Font(File, …)` resolves to `AndroidFileFont` on Android. Its entire declared surface:

```
public final java.io.File getFile();
public java.lang.String getCacheKey();
```

No `getIdentity()` — a desktop concept — and no `getResId()`, the Android *resource* one. Every branch of `ComposeSemanticsDataProducer.fontIdentity` misses it, so `fontFamilyLabel` returns null and the text node reaches the export with no family.

The embedded player reaches `Font(file = …)` through `GoogleFontFamilies.composeFontFamily`'s static branch whenever a font cache is configured — which the daemon always sets — and these documents name `google:Roboto Flex` with no variation axes, which is exactly that branch.

Fixed in #5273. **Whether that clears the wall is not yet established** — see below.

## Ruled out, so a future hunt does not redo them

[...]

Issue #3024, opened 2026-07-30
figma-svg embedded Android resource font has wider browser metrics and overflows text


yschimke:
## Summary

After #2886, figma-svg correctly resolves and embeds Android resource-backed Montserrat, but the browser-rendered SVG still uses materially wider text metrics than the Android PNG. Explicit line breaks captured from Compose no longer fit the same width, so text overflows instead of matching the render.

## Reproduction

Fresh compose-samples artifacts from compose-ai-tools 0.19.9:

- previews commit: ff8fa52964c58f559c8468fcea993390d7a32f7a
- workflow: https://github.com/yschimke/compose-samples/actions/runs/30577222458
- compare page: https://htmlpreview.github.io/?https://github.com/yschimke/compose-samples/blob/design-artifacts/jetnews/compare.html
- direct SVG: https://preview.coo.ee/jetnews/render/screens-article__ideal__default__compact.svg?mode=web

JetNews Screens/Article remains at 72.2% structural match. The PNG keeps the Montserrat article title inside the compact card. The SVG renders the same captured tspans visibly wider; the final title line extends beyond the right edge.

The current SVG is no longer exhibiting the original #2886 failure. It contains an embedded @font-face for Montserrat and emits text with font-family="Montserrat". This is a residual metrics/shaping mismatch after resource resolution, not a missing or numeric family.

Reader/Article body also remains at 89.1%, consistent with smaller instances of the same typography drift.

## Expected

An embedded Android resource font should reproduce the Android paragraph glyph advances closely enough that captured line breaks fit the same bounds. If browser SVG text cannot preserve those advances, export should apply measured text scaling/positioning or use the text raster fallback for the affected paragraph.

Please add an end-to-end Android resource-font fixture that compares a multi-line Montserrat paragraph near its width boundary, not only successful face resolution/embedding.

yschimke:
## Investigation: this is a font *size* bug, not a metrics/shaping one

Measured the published artifacts for `jetnews/screens-article__ideal__default__compact` (SVG + PNG from the run linked above).

### The embedded face is fine

Chromium loads and uses it — `document.fonts` reports `Montserrat/400/loaded` and `Montserrat Medium/500/loaded`, and `getComputedTextLength("Language to Kotlin")` is 1211px with the embedded family vs 979.8px with a deliberately-missing family. #2886's resolution + embedding really is working.

### The SVG declares a font-size 1.5× larger than the render drew

[...]

PR #5283, opened 2026-09-07
fix: name a family-less text run the family the render recorded

yschimke:
Closes part of #5279.

## The export manufactured its own mismatch

For a text run that states no family, `resolveFonts` filled the gap with a hardcoded constant:

```kotlin
val name = FigmaLayeredSvg.embedFamily(captured, DEFAULT_EMBED_FAMILY) ?: return   // "Roboto"
```

…and then, a few lines earlier in the same function, compared that guess against what the render had recorded:

```kotlin
val unnamed = FigmaSvgRenderedFonts.unnamedIn(named)   // ["Roboto Flex"]
```

`Roboto Flex` vs `Roboto`, disagreement, boxes. The answer was in hand three lines before the guess was made.

## What this does

When the render drew exactly **one** family, that is by construction what the family-less runs drew. Use it — for the `<text>`, for the `@font-face`, and as the key the bytes are looked up under.

The two ways it deliberately declines to help are as important as the case it fixes, and both have tests:

- **Two or more recorded families** → `singleRenderedFamily()` returns null and nothing changes. Some run drew one and some another; a family-less node cannot say which, and picking either would be the original defect wearing a new default. The guard is right to fire.
- **A face the resolver cannot supply** → still boxes. Naming a family the SVG carries no `@font-face` for is exactly the wrong-and-plausible failure the boxes exist to prevent, so the default only applies when the face can actually be embedded.

The recorder half publishes the resolved file under that family, so the export embeds the bytes the render drew rather than re-fetching a woff2 by name — the contract `recoverDownloadableFont` already has for downloadable faces, extended to the file-backed shape.

## Why the recorder's name and not the document's

The `.rc` is captured and published alongside the SVG, and it names the family outright — so reading it is the obvious alternative. Two reasons this doesn't:

1. **The namespace.** The document says `google:Roboto Flex`. `google:` is a *player resolution* namespace (`device:` is its sibling), not a CSS family. Emitting it would produce a `font-family` meaningless to CSS and Figma, and would send the `?mode=web` rewrite after a Google Fonts family that does not exist — the shape #3024 fixed for `Montserrat+Medium`. What the recorder holds is read out of the face's own bytes by `FigmaResourceFonts.familyName`, so it is already the clean `Roboto Flex` and no namespace can leak. Measured: `RobotoFlex.ttf` has no nameID 16 and nameID 1 `Roboto Flex`.
2. **No new seam.** The capture walks Compose text nodes, not the document, so routing the document's name to it means a new contract between `rc-embedded-player` and this repo. This change is entirely inside the export.

The document stays the right escalation for the case this declines: with several families recorded, it is the only thing that can say which run drew which. That is worth doing when something needs it, and it would need the `google:`/`device:` strip that `GoogleFontFamilies.googleFamilyName()` already implements.

## Test plan

Three cases added to `FigmaSvgTofuFallbackTest` (9 tests, was 6):

| test | asserts |
| --- | --- |
| `a family-less run is named the one family the render recorded` | the `remote-m3` shape — names `Roboto Flex`, no boxes, no warning sidecar |
| `two recorded families leave a family-less run to the guard` | still boxes, still warns |
| `a recorded face the resolver cannot supply still boxes` | the default never outruns what can be embedded |

**Reproduced first.** With `recordedFamily` forced to null (the old behaviour):

```
FigmaSvgTofuFallbackTest > a family-less run is named the one family the render recorded FAILED
tests=9 failures=1
```

…and with the change, `tests=9 failures=0`. The two guard-preserving tests pass either way, which is the point — they pin behaviour this must not change.

Full run, all green in 12m56s:

```
./gradlew :data-layoutinspector-connector:test :daemon:android:testDebugUnitTest ktfmtCheckAll
BUILD SUCCESSFUL
```

`:daemon:android:testDebugUnitTest` covers `FigmaSvgRemoteComposeFontEmbedTest` and `RenderEngineTest`, so the downloadable-font path and the existing tofu behaviour are unchanged.

Non-visual: no rendered surface changes in this diff. Whether it clears the `remote-m3` wall is still unmeasured for the reasons #5279 records — that needs a published release and a catalog re-render.

https://claude.ai/code/session_01JHXPXkZtZ2w6KpLtjjDvd5

---
_Generated by [Claude Code](https://claude.ai/code/session_01JHXPXkZtZ2w6KpLtjjDvd5)_

### DIFF
--- a/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/FontResolverRecorder.kt
+++ b/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/FontResolverRecorder.kt
@@ -49,6 +49,16 @@ class FontResolverRecorder(private val context: Context? = null) {
     // exporter uses, so the rendered-font audit compares `Roboto` with `Roboto` rather than a cache
     // filename with `Roboto` and falsely switches the whole document to tofu (issue #4935).
     val fileFamily = matched?.let(::recoverFileFontFamily)
+    // Publish the file under that family as well. The export reaches a face through
+    // [FigmaResourceFonts] by NAME, and a text run that states no family is looked up under the
+    // family this recorder published — so without this the export knows which face was drawn but
+    // not where its bytes are, and re-fetches a woff2 by name instead of embedding the ones the
+    // render used. Same contract as [recoverDownloadableFont] below, for the file-backed shape.
+    if (fileFamily != null && matched != null) {
+      fileFor(matched)?.let {
+        FigmaResourceFonts.register(fileFamily, weight, style == "italic", it.absolutePath)
+      }
+    }
     val displayName = matched?.let { displayFamilyName(fontLabel(it)) }
     // A downloadable `Font(GoogleFont("Lato"), …)` resolved to a real TTF in the renderer's font
     // cache for the PNG. Publish that file so the figma-svg export embeds the same bytes rather
--- a/data/layoutinspector/connector/src/main/kotlin/ee/schimke/composeai/daemon/ComposeFigmaSvgDataProduct.kt
+++ b/data/layoutinspector/connector/src/main/kotlin/ee/schimke/composeai/daemon/ComposeFigmaSvgDataProduct.kt
@@ -158,7 +158,22 @@ object ComposeFigmaSvgDataProducer {
         deviceBackground = background,
         backgroundMode = mode,
       )
-    val fonts = fontResolver?.let { resolveFonts(model, it, fileSystem) }
+    // The face the render actually drew for text that states no family of its own. The export used
+    // to hardcode [DEFAULT_EMBED_FAMILY] here and then compare that guess against what the render
+    // recorded — manufacturing its own mismatch and boxing the whole document whenever the two
+    // differed. When the render drew exactly ONE family, that is by construction what the
+    // family-less runs drew, so it is the answer rather than something to guess at.
+    //
+    // A singleton is the whole of the claim. Two or more recorded families means some run drew one
+    // and some another, and nothing here can say which is which — the guard is then right to fire.
+    // The recorded name is already the clean CSS family (`Roboto Flex`): it is read out of the
+    // face's own bytes by `FigmaResourceFonts.familyName`, never from the document, so a player's
+    // resolution namespace (`google:` / `device:`) cannot leak into a `font-family`.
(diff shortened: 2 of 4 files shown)


yschimke:
`wear-os-samples (ComposeStarter)` is failing on this head, and it is not this PR's.

```
.../_temp/00c3bf49-….sh: line 16: /home/runner/work/compose-ai-tools/compose-ai-tools/.github/scripts/retry-gradle-repository.sh: No such file or directory
##[error]Process completed with exit code 127.
```

The wrapper added by #5276 invokes `"$GITHUB_WORKSPACE/.github/scripts/retry-gradle-repository.sh"`, but the `integration` matrix job never checks this repository out. Its first step is `Checkout ${{ matrix.repo }}` with `path: external` (`integration.yml:778`), and the only other checkout is the optional secondary — so `$GITHUB_WORKSPACE` holds the *consumer's* tree plus downloaded artifacts, and `.github/scripts/` is not in it. The script itself is fine: committed at mode `100755`, and it exists on this branch.

Not this PR's, on two counts. The diff here touches four Kotlin files and no CI. And the same leg failed identically on **#5282** (`chore(main): release 2.2.2`, head `2b289be`, [run 34157289737](https://github.com/yschimke/compose-ai-tools/actions/runs/34157289737)) — a release-please PR whose diff is version bumps and a changelog, which cannot make a script disappear.

**No re-run.** The error is a deterministic missing path, not a transient repository 403 — the very condition the wrapper exists to retry. Re-running would spend the one re-run to reproduce a certainty, so I am leaving it unspent.

I have not fixed it here, because it is a shared-workflow change and widening this PR to carry it is the wrong shape. The patch, for whoever picks it up:

```yaml
- name: Checkout compose-ai-tools (for .github/scripts)
  if: ${{ env.CELL_ACTIVE == '1' }}
  uses: actions/checkout@…
  with:
    path: tools
    sparse-checkout: .github/scripts
    persist-credentials: false
```

…and point the wrapper at `"$GITHUB_WORKSPACE/tools/.github/scripts/retry-gradle-repository.sh"`. Inlining the retry loop into the step, or reverting #5276's wrapper for this job, would work equally well — the choice is about where that logic should live, not about the failure.

Worth noting the blast radius: this leg is a **required** status check on `pull_request`, and the job has no push-to-`main` lane, so nothing caught it at merge time and every PR opened since #5276 hits it.

---

PR #5292, opened 2026-09-08
fix: carry the figma-svg font warning and refuse to publish boxes

yschimke:
## The fault

A `compose/figma-svg` export that cannot name a font family the render drew falls back to missing-glyph boxes, and records exactly what it lost in `compose-figma-fonts.warnings.json` beside the SVG.

**Nothing has ever read that file.** It is written into the render's data dir, and from there: not collected by `DaemonSemanticsFetcher`, not injected into the bundle, not published to `design-artifacts/<system>`, not served. So a sheet of boxes shipped to a live catalog and was found by eye a day later — with the one artefact naming the lost face still sitting on the build machine's disk.

The tofu itself was fixed by #5283. This is about the part that let it ship unnoticed, which is still true for the next one.

## Follow the signal to the end

**Collect and carry it.** The sidecar now travels the same route as every other per-preview sidecar — collected by `DaemonSemanticsFetcher`, injected as `previews/<id>.figma-fonts.warnings.json`, and split out alongside the sticker it explains, so it reaches the delivery branch beside the SVG it is about. Only a degraded preview has one, so this is a no-op on a healthy sheet, and an entry is itself the signal.

**Refuse to publish it.** `bundle pack --with-semantics` exits non-zero when any preview degraded, naming each one and pointing at the sidecar:

```
bundle pack: 2 preview(s) exported their text as missing-glyph boxes because the
figma-svg export could not name a font family the render drew:
  - AppCard_ideal
  - Chip_default
Each carries previews/<id>.figma-fonts.warnings.json in out/wear-m3-bundle.png
naming the face that was lost. Fix the export, or pass --allow-lost-font-families
to publish the boxes deliberately.
```

`design-artifacts.yml` runs exactly that command under `set -euo pipefail`, so this fails the job before the publish step.

**Hand it to the reader.** `ServePreview.lostFontFamilies` is where the server lands the sidecar's contents so the viewer can badge the sticker. The server reads the file and draws the badge — that half lives in `compose-preview-server` and is blocked on this releasing, since it compiles against a published `render-host`.

## Why the gate is at the pack, not the export

Throwing from `ComposeFigmaSvgDataProducer.writeSvg` looks like the obvious place and does nothing. Post-capture extensions are dispatched inside a `catch (t: Throwable)` in `RenderEngine` that logs and continues, deliberately — an extension failure must not strand the PNG the user came for. A gate there would print one more stderr line nobody reads, and the tofu SVG (already written) would publish regardless. The publish is where a gate actually stops something.

`--allow-lost-font-families` is the deliberate override, matching the render's own `-Dcomposeai.fonts.failOnFallback`, which is default-on and fails a render that drew the wrong typeface for exactly this reason.

## Note on the default

This is default-on by design, so **a catalog that currently degrades will now fail its publish instead of shipping boxes.** That is the intent; the escape hatch is one flag if a sheet needs to ship while the export is fixed.

## Also

The sidecar's filename moves to `:data-fonts-core` — the one module both the producer (`:data-layoutinspector-connector`) and the reader (`:cli`) already depend on — so it is spelled once rather than mirrored. `bundle-format` and `render-host` ABI dumps are regenerated for the new suffix, injector and field.

## Test plan

- `LostFontFamilyRefusalTest` — healthy pack not refused; degraded pack names every preview, the sidecar and the override; the override publishes deliberately.
- `BundleSemanticsInjectTest` — the sidecar lands beside the `.figma.svg` it explains without disturbing it; a healthy sheet carries none at all.
- `BundleSplitTest` — the sidecar follows its sticker through a split (the delivery branch serves split bundles, so dropping it there is what left the boxes unexplained downstream).
- `./gradlew :cli:test :bundle-format:test :data-fonts-core:test :data-layoutinspector-connector:test :render-host:test :bundle-format:checkKotlinAbi :render-host:checkKotlinAbi` — green.

Text-only change (CLI output, bundle entries, one DTO field); no rendered surface moves.

https://claude.ai/code/session_01YAWxMuzo5hFagcDzxY4rUD

### DIFF
--- a/.github/scripts/sweep-stranded-releases.sh
+++ b/.github/scripts/sweep-stranded-releases.sh
@@ -62,9 +62,16 @@ CENTRAL_BASE_URL="${CENTRAL_BASE_URL:-https://repo1.maven.org/maven2}"
 now="$(date -u +%s)"
 exit_code=0
 
-minutes_since() { # <ISO-8601 timestamp> → whole minutes, or empty if unreadable
+minutes_since() { # <ISO-8601 timestamp> → whole minutes, or empty if absent/unreadable
   local epoch
-  epoch="$(date -u -d "${1:-}" +%s 2>/dev/null)" || return 0
+  # An EMPTY timestamp has to answer "unknown", not a number. GNU `date -u -d ""` does not fail —
+  # it reads the empty string as midnight today and exits 0 — so without this guard "no recovery
+  # run has ever been dispatched" came back as "dispatched <minutes since UTC midnight> ago". For
+  # the first RECOVERY_COOLDOWN_MINUTES of every UTC day that is inside the cooldown, so the
+  # sweeper declined to dispatch the rebuild it exists to dispatch, and its self-test failed for
+  # the same six hours daily.
+  [[ -n "${1:-}" ]] || return 0
+  epoch="$(date -u -d "$1" +%s 2>/dev/null)" || return 0
   [[ -n "${epoch}" ]] && echo $(( (now - epoch) / 60 ))
 }
 
--- a/.github/scripts/test-sweep-stranded-releases.sh
+++ b/.github/scripts/test-sweep-stranded-releases.sh
@@ -139,6 +139,12 @@ CURL_OK=false run central
 logged central "PATCH" && fail "a release whose plugin is not on Central must stay a draft"
 
 # 5. Assets missing means the build never finished — rebuild it for the tag rather than publish.
+#    No ${DISPATCHES} fixture is set, so this is also the "no recovery run has ever been
+#    dispatched" case, and it is the one that caught `date -u -d ""` reading an empty timestamp as
+#    midnight today rather than failing: the sweeper saw a phantom recovery <minutes since UTC
+#    midnight> old and sat inside its own cooldown. That made both assertions below — and the real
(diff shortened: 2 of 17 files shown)


yschimke:
`Actions Script Tests` was red on `def422c`. It is not this PR's — this diff touches no release scripts — and it is red on `main` too:

```
```

Root cause: `sweep-stranded-releases.sh`'s `minutes_since` passes its argument straight to `date -u -d`, and GNU `date -u -d ""` does not fail — it reads the empty string as midnight today and exits 0. So an absent `last_recovery` ("no recovery run has ever been dispatched") came back as *"dispatched (minutes since UTC midnight) ago"*, which is inside the 360-minute cooldown for the first six hours of every UTC day. Those two assertions are case 5, which deliberately sets no dispatches fixture, so they have been failing on every PR raised between 00:00 and 06:00 UTC and passing the other eighteen hours.

It is a real bug in the sweeper too, not only its test: for those six hours it declines to dispatch the rebuild it exists to dispatch.

Fixed in [#5294](https://github.com/yschimke/compose-ai-tools/pull/5294) and the same one-line guard is cherry-picked here as `ec5ff6c`, so this PR goes green without waiting on that one to merge; it no-ops once `main` carries it. Verified locally at 00:27 UTC — `sweep-stranded-releases.sh: all tests passed.`

---

PR #5293, opened 2026-09-08
fix: clear every Kotlin compile warning in the build

yschimke:
## What

`./gradlew build -x test` printed 18 Kotlin warnings across `:daemon:android`, `:samples:cmp-wasm-catalog`, `:samples:wear` and `:samples:remotecompose`. It now prints none. Every one is fixed at the source; one is suppressed, with the reason written beside it.

| Warning | Where | Fix |
| --- | --- | --- |
| Unnecessary safe call on a non-null `SemanticsInputTarget` ×4 | `AndroidInteractiveSession:285-286` | `hasTarget` already implies `rawTarget != null` |
| `Condition is always 'true'` | `FontResolverRecorder:57` | `fileFamily` is `matched?.let { … }`, so `matched != null` was the same fact twice |
| Overrides a deprecated member ×3 | `PlaceholderFallbackResources:83,118,134` | the class already suppresses `DEPRECATION` for the `Resources` accessors it wraps; `OVERRIDE_DEPRECATION` joins it |
| Unnecessary `!!` ×3 | `RenderEngine:616`, `RobolectricHost:744,3211` | the receiver is already smart-cast by the preceding check — the very next line was reading it without one |
| `No cast needed` | `RenderEngine:2822` | `rootView` is already a `View` |
| Unnecessary safe call on a non-null `JsAny` ×3 | `cmp-wasm-catalog` `CompileClient:161`, `Main:351,372` | `Promise.catch` hands over a non-null value, so the `?: "fetch failed"` fallback was unreachable |
| `Icons.Filled.VolumeUp` is deprecated | `samples/wear` `Previews:373` | the auto-mirrored one, as the deprecation says |
| Calling a UI Composable where a RemoteCompose Composable was expected | `samples/remotecompose` `WearWidgetPreviews:68` | `@Suppress("COMPOSE_APPLIER_CALL_MISMATCH")` — see below |

## The suppression

`SquircleRemoteWidgetWrapper.Wrap` frames `super.Wrap(content)` in a `Box(Modifier.clip(…))`. The applier check reads the override as RemoteCompose-targeted and `Box` as a UI composable, so it flags the call — but that framing is exactly the host/preview concern this wrapper exists to add *around* the captured document, as the class's own KDoc explains at length. Emitting a `RemoteBox` instead would put the shape into the widget payload, which is the thing the file is written to avoid. Suppressed at the one function, with the reasoning beside it rather than in a commit message.

## Test plan

- `./gradlew build -x test -x kotlinWasmToolingSetup --continue` — no `w:` lines anywhere in the log (was 18)
- `./gradlew :samples:remotecompose:compileDebugKotlin :samples:wear:compileDebugKotlin` — clean
- `./gradlew ktfmtFormat` — clean

Not green in this sandbox, and unrelated to this diff — all three reproduce on `main` here:

- `:kotlinWasmToolingSetup`, `:cli:serve-wasm:wasmJsBrowserTest` / `wasmJsBrowserProductionWebpack` — the environment's proxy 403s `codeload.github.com/Kotlin/karma`, so the Wasm test tooling never installs. Excluded; the Wasm *compile* tasks ran and are clean.
- `:renderer-android:lintDebug` (`NewApi` on `java.time.Duration#ofMillis`), `:samples:android:lintDebug` (`DisallowLookaheadAnimationVisualDebug`), `:samples:design-catalog-wear-m3:lintDebug` (`WearStandaloneAppFlag`) — all in files this PR does not touch.

Text-only PR: no rendered surface changes. The one pixel-adjacent edit is `Icons.Filled.VolumeUp` → `Icons.AutoMirrored.Filled.VolumeUp` in a Wear sample's settings list, which is the same glyph, mirrored only under RTL.

https://claude.ai/code/session_018TiJw44mgFLFjaf19Y7X1T

---
_Generated by [Claude Code](https://claude.ai/code/session_018TiJw44mgFLFjaf19Y7X1T)_

### DIFF
--- a/.github/scripts/sweep-stranded-releases.sh
+++ b/.github/scripts/sweep-stranded-releases.sh
@@ -62,9 +62,14 @@ CENTRAL_BASE_URL="${CENTRAL_BASE_URL:-https://repo1.maven.org/maven2}"
 now="$(date -u +%s)"
 exit_code=0
 
-minutes_since() { # <ISO-8601 timestamp> → whole minutes, or empty if unreadable
+minutes_since() { # <ISO-8601 timestamp> → whole minutes, or empty if absent or unreadable
   local epoch
-  epoch="$(date -u -d "${1:-}" +%s 2>/dev/null)" || return 0
+  # An empty argument is "no such timestamp", and it has to be rejected before `date` sees it:
+  # GNU `date -d ""` does not fail, it answers *midnight today*. So an absent `last_recovery`
+  # read as a recovery dispatched however many minutes we are past 00:00 UTC, and every sweep
+  # between midnight and the 6h cooldown refused to dispatch a rebuild it should have.
+  [[ -n "${1:-}" ]] || return 0
+  epoch="$(date -u -d "$1" +%s 2>/dev/null)" || return 0
   [[ -n "${epoch}" ]] && echo $(( (now - epoch) / 60 ))
 }
 
--- a/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidInteractiveSession.kt
+++ b/daemon/android/src/main/kotlin/ee/schimke/composeai/daemon/AndroidInteractiveSession.kt
@@ -282,8 +282,8 @@ internal constructor(
     if (hasTarget && replyMatched?.get() == false) {
       val message =
         "AndroidInteractiveSession.dispatch: target did not resolve to a node " +
-          "(ref=${rawTarget?.ref}, testTag=${rawTarget?.testTag}, role=${rawTarget?.role}, " +
-          "text=${rawTarget?.text})"
+          "(ref=${rawTarget.ref}, testTag=${rawTarget.testTag}, role=${rawTarget.role}, " +
+          "text=${rawTarget.text})"
       val reason =
         replyUnresolvedReasonJson?.get()?.let { json ->
           runCatching {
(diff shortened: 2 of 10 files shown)


yschimke:
`Actions Script Tests` went red on `7920659`, and it is not this PR's — this branch touches no `.github` script.

The failure is `test-sweep-stranded-releases.sh` cases 5 and 6:

```
```

`minutes_since ""` in `sweep-stranded-releases.sh` hands the empty string to `date -u -d`, which does not fail on it — GNU `date` answers midnight today. With no prior recovery run, `last_recovery` is empty and the sweeper reads it as a recovery dispatched however many minutes we are past 00:00 UTC, which inside the 6h cooldown is a refusal. The job started at 00:27 UTC.

It is a real bug, not just a test artifact: every scheduled sweep between 00:00 and 06:00 UTC declines to dispatch the rebuild a stranded draft needs, warning about a run that never happened.

Reproduced on `origin/main` at the same hour with nothing else changed, so it predates this branch. Fixed in #5295, and the same one-line commit is cherry-picked here (`9d3947f`) so this PR can go green rather than wait on that merge — it no-ops once `main` carries it.

---

yschimke:
Correction to my comment above: the canonical fix for that `Actions Script Tests` failure is **#5294**, not #5295.

#5294 was opened four minutes before mine, by a session driving #5292, with the same root cause and the same guard — we collided on it. I have closed #5295 as the duplicate.

Nothing changes here: `9d3947f` carries the same one-line guard, so this PR is green now, and the commit no-ops once `main` has #5294.

---

PR #5296, opened 2026-09-08
feat: let a Remote Compose capture defer density to the player

yschimke:
`?fontScale=` against a `remote-m3` sticker has never moved a pixel — e.g. [`appcard__ideal__default__compact`](https://preview.coo.ee/remote-m3/p/appcard__ideal__default__compact?fontScale=1), where `fontScale=1` and `fontScale=2` return byte-identical PNGs (`md5 855640…`, both `25538` bytes) under `x-compose-preview-generation: catalog-cache` — a cached **live** render, never a baked fallback.

**This PR does not fix that.** It fixes the capture half, which is a prerequisite, and it documents precisely what is still missing. Please read the evidence section before merging on the strength of the title.

## The cause is two defects in the capture, not one

`RemoteOverridablePreview` built its `RemoteDensity` with `RemoteDensity.from(displayInfo)`. Disassembling `remote-creation-compose:1.0.0-alpha18`:

```java
// RemoteDensity.Companion.from(info)
new RemoteDensity(rf(info.density.density), rf(info.density.fontScale))   // literal constants

// RemoteDensity.Companion.getHost()
density   = RemoteFloat(RemoteContext.FLOAT_DENSITY)
fontScale = RemoteFloat(Rc.System.FONT_SIZE) / 14f / density              // host variables
```

1. `from(...)` folds density **and** font scale into literal `RemoteFloat` constants, so every sp→px is resolved before a byte is written and no player can move it afterwards.
2. Independently, the call passed no font scale into `RemoteCreationDisplayInfo`, whose own parameter defaults to `1f` (`fconst_1` in the `$default` bridge). So the constant baked in was `1` however the render spec was configured.

Defect 2 is fixed unconditionally: read `LocalDensity.current.fontScale`, which the Android lane already sets per render spec.

Defect 1 becomes a build-wide setting rather than a straight flip. `RemoteDensity.Host` — Kotlin-cased `Host`, a companion `val`, not an enum constant — defers both to the player's variables. But it defers **density** as well as font scale, turning the replay density into a correctness input where it used to be ignored. That is a change to what a captured `.rc` means about geometry, not only about text, so `composePreview.rcDensity` defaults to `fixed` and a catalog opts in when it is ready to re-bake and read the pixels. `serve` resolves replay density from each preview's declared device (`renderDensityFor`), the same source the bake used, so the two agree.

## Evidence

`:samples:remotecompose` with the existing `accessibility` permutation, which draws every preview a second time at `fontScale = 2`. Each pair is one preview against **its own** `_fontscale-2x` variant.

Baseline — the defect, reproduced locally:

baseline 1x
baseline 2x

`rcDensity=host`:

host 1x
host 2x

Control — an ordinary Compose preview, same module, same run, same lane, which **does** scale:

control 1x
control 2x

| Pair | `.rc` document | pixels |
| --- | --- | --- |
| baseline | identical, 1943 B | **identical** ← the bug |
| `rcDensity=host` | identical, 2407 B | **identical** |
| control (non-remote) | no document | **differ** |

**Two of these pairs are identical images, and that is the finding.** The capture change does what it claims — the document grew by the expression graphs that replaced the folded constants, and is now byte-identical across font scales because the scale is no longer stored in it. The control proves the harness applies `fontScale=2` to the composition. But the stickers still rasterise identically, so a link between the document's variable reference and the drawn glyphs is missing and this PR does not close it.

For the record, the player is not obviously the culprit: `third-party-rc-embedded-player` (Android, the `capturePlayer` these stickers record) does write the variable, from the composition's own `Density`:

```java
ctx.loadFloat(33, 14.0f * density.getFontScale() * density.getDensity());  // ID_FONT_SIZE
ctx.loadFloat(27, density.getDensity());                                   // ID_DENSITY
```

Next step is to dump the ops from the 2407-byte document and check whether the text op references variable 33 at all.

## Also here

`RcJvmRenderSpec` gains a `fontScale`, forwarded to the cmp-jvm sidecar as `--fontScale`. That axis is inert until `lib-rcjvm` grows the parameter — `renderRemoteDocumentToPng(bytes, w, h, density, seeds, theme, …)` takes no font scale and builds its scene with `Density(density)`, whose `fontScale` defaults to `1f` — so it needs a change in `yschimke/rc-players`. The flag is safe to send meanwhile: `RcJvmRenderMain.parseArgs` collects `--flag value` pairs into a map and only ever does `Map.get` on keys it knows, so an older sidecar ignores it. The pooled lane, whose frame is pinned at `PROTOCOL_VERSION` 2 and cannot spell the field, is skipped for scaled requests rather than answering warm with pixels that ignored the axis.

## Test plan

- `:data-remotecompose-connector:testDebugUnitTest --tests '*RemoteDensitySelectionTest*'` — new, green
- `:render-host:checkKotlinAbi` — green (`render-host/api/render-host.api` regenerated for the new field)
- `:gradle-plugin:ktfmtCheckMain`, `:render-host:ktfmtCheckMain` — green
- `:samples:remotecompose:composePreviewRenderAll` on all three configurations above — green

## Related

- `yschimke/wear-m3-catalog` — the `rcDensity=host` opt-in, which should **not** merge until the chain above completes.
- `yschimke/compose-preview-server` — corrects the stale explanation of why `fontScale` looked inert on `remote-m3`. The matching cmp-jvm threading is blocked there until this repo releases `render-host`, since that repo compiles against the published coordinate.

https://claude.ai/code/session_01Mu5a7gLKePVTyjQ95zs3dX

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mu5a7gLKePVTyjQ95zs3dX)_


yschimke:
**`Actions Script Tests` is red, and it is not this PR's.** Standing down on it rather than widening the PR.

Failing on `d6b6c36`:

```
2 sweep-stranded-releases.sh test(s) failed.
```

Both are case 5 of `.github/scripts/test-sweep-stranded-releases.sh` — release automation. This PR's diff is Kotlin capture/render code, a Gradle plugin, an ABI dump and PNGs; it touches nothing under `.github/`.

Rather than re-run the job to test for a flake, I reproduced it deterministically on a clean worktree of `origin/main` with none of this branch's commits:

```
$ git worktree add /tmp/mainwt origin/main
$ ./.github/scripts/test-sweep-stranded-releases.sh
2 sweep-stranded-releases.sh test(s) failed.
```

That is stronger evidence than a re-run would be — it is red on the base branch, every time, so I have not spent a CI re-run confirming what the base already shows. Happy to trigger one if you'd rather see it.

**No fix exists to port**, so there is nothing for me to carry into this branch. What I can offer toward one, from a short look:

`required-release-assets.sh 1.78.0` lists three assets:

```
compose-preview-1.78.0.tar.gz
compose-preview-1.78.0.zip
compose-preview-android-daemon-1.78.0.zip
```

Case 5 fixtures only `compose-preview-1.78.0.zip`, so two are missing and `sweep-stranded-releases.sh` should take its dispatch branch — but no `POST …/release.yml/dispatches` is logged. So either the required-asset list has grown past what the fixture models, or the dispatch is being suppressed before it is reached (the cooldown guard that case 6 exercises is the obvious suspect). I stopped there: root-causing the release sweeper is a separate change with its own blast radius, and it is not mine to fold into a font-scale PR.

The rest of this PR's checks are unaffected by it. I'll keep watching and will re-check on my next pass.

---

PR #5324, opened 2026-09-09
perf(serve): stop re-hashing and re-formatting a theme cache file name

### LINKED ISSUE #5322: perf(serve): ThemeCacheStore.Generation.fileName hashes and hex-formats per membership test, costing ~1.5M String.format calls per /status
`ThemeCacheStore.Generation.fileName` is called once per `contains()`, and each call does a fresh
SHA-256 plus a `String.format("%02x")` **per digest byte**. `CatalogThemeCache.snapshot()` calls
`contains` once per themed render, so the cost scales with the size of the thing the cache exists to
grow.

On `preview.coo.ee` (render-host 2.3.0) that is **~48,400 themed renders across 16 catalogs**, so one
full snapshot pass costs roughly:

- **1,549,000** `String.format` calls (48,400 x 32 digest bytes), and
- **48,400** `MessageDigest.getInstance("SHA-256")` provider lookups.

## Evidence

From `javap -c` on `render-host-2.3.0.jar`, `ThemeCacheStore$Generation.fileName(String)`:

```
0:  ldc_w  "SHA-256"
3:  invokestatic  MessageDigest.getInstance      <- fresh instance per call
23: invokevirtual MessageDigest.digest
51: invokestatic  ArraysKt.joinToString$default  <- over the 32 digest bytes
```

and the `joinToString` transform, `fileName$lambda$0`:

```
0:  ldc_w  "%02x"
25: invokestatic  String.format:(String;[Object;)String
```

`contains`, `isDirtyName` and `get` each route through it:

```
13: invokespecial fileName:(String;)String
16: invokeinterface Set.contains
```

Thread dumps of the deployed server under concurrent load, aggregated over 4 samples, land squarely
on that lambda:

```
40  java.util.Formatter.format
21  kotlin.collections.ArraysKt___ArraysKt.joinToString
21  ThemeCacheStore$Generation.fileName$lambda$0
21  ThemeCacheStore$Generation.fileName
21  ThemeCacheStore$Generation.contains
21  CatalogThemeCache.contains
21  CatalogThemeCache.snapshot
20  java.lang.String.format
18  java.util.Formatter.parse
```

`Formatter.parse` re-parses the two-character format string on every one of those calls.

## Suggested fix

Two independent changes, either of which helps and both of which are local to `fileName`:

1. **Drop `String.format` for hex.** `java.util.HexFormat` (Java 17+, and this is on the 17 floor)
   or a manual nibble lookup removes the formatter entirely. This is the big one — it is 32 of the
   32 formatter parses per call.
2. **Stop re-deriving the name.** The digest is a pure function of the preview name and the
   generation, so it can be memoised per `Generation` (or the `present` set can be keyed on the raw
   name instead of the hashed file name, which removes the hash from the membership path
   altogether).

(1) alone should take the bulk of it; (2) makes `contains` a plain set lookup.

## Scope

This was found while profiling the `compose-preview-server` front door, where
`rememberCatalogMeta` called `themeOptimizationSnapshot()` for every listed catalog on every home
index request — about 88% of that page's server time. That caller is fixed downstream in
yschimke/compose-preview-server (the home index does not render progress counters, so it no longer
asks for them).

`/status` is not fixable that way: it genuinely needs the snapshot, and still pays the full cost on
every load. On the deployed box `/status` measures ~1.7s server-side against ~750ms for `/`, which
is consistent with this being the dominant term.

yschimke:
Fixes #5322.

`ThemeCacheStore.Generation.fileName` ran a fresh `MessageDigest.getInstance("SHA-256")` and then `String.format("%02x")` **per digest byte**, on every call — and `contains`, `isDirtyName`, `get`, `wasAdopted` and `put` all route through it. `CatalogThemeCache.snapshot()` calls `contains` once per themed render, so the work scaled with the size of the thing the cache exists to grow.

Both halves the issue suggests, since they are independent and the second makes the path a plain lookup:

- **`java.util.HexFormat` replaces the per-byte `String.format`.** This is the big one: `Formatter.parse` re-parsed the same two-character format string 32 times per name, which is where the thread dumps landed.
- **The name is memoised per generation.** The digest is a pure function of the cache key, so `contains` becomes a map lookup after the first ask.

## Measured

48,400 distinct keys — the `preview.coo.ee` figure from the issue — one snapshot pass, on this JVM:

| | | |
| --- | ---: | ---: |
| old (`getInstance` + `%02x`×32) | 616.4 ms | |
| `HexFormat` only | 21.5 ms | 28.7× |
| `HexFormat` + memo, cold | 30.8 ms | 20.0× |
| `HexFormat` + memo, warm | **7.3 ms** | **84.5×** |

Cold is the first pass after a generation opens; warm is every `/status` after that. Consistent with the issue's ~1.7 s server-side measurement being dominated by this term.

## The part that had to be proved rather than assumed

The names must be **byte-identical**. `present` is built by listing the generation's directory, so a changed name does not fail — it silently orphans every PNG a previous build wrote and the catalog re-renders from scratch, which is precisely the failure this store exists to prevent.

`java.util.Formatter` renders a negative `Byte` under `%x` as the value plus 2⁸, which is exactly the unsigned hex `HexFormat.formatHex` writes. Verified over all 256 byte values and several keys including non-ASCII, and pinned by the test below.

## Bounding

The memo is bounded the way `present`, `dirtyNames` and `adopted` already are: one entry per cache key this generation is asked about, i.e. one per themed render in one catalog. Every caller walks a catalog's own renders, so nothing reaches it with unbounded keys. Said out loud in the KDoc, since "add a map" deserves an answer to "how big does it get".

## Test plan

`ThemeCacheStoreFileNameTest` is new — there was no test over this store at all, which is why nothing was holding the on-disk contract.

- It computes the expected name using the **old** `"%02x".format` path and asserts the file appears on disk under it. That makes it an independent equivalence check rather than a restatement of the new code, and it will fail if either the digest or the encoding ever drifts.
- Repeated membership through the memo, and a miss.
- A reopened generation finds what the previous one wrote — the memo is per generation, so the second one recomputes and has to agree.

Plus `:render-host:test --tests '*ThemeCache*'` and ktfmt.

Text-only evidence: nothing renders.

https://claude.ai/code/session_01YKaGa8Rtm7fRwkrwwiWJ3h

---
_Generated by [Claude Code](https://claude.ai/code/session_01YKaGa8Rtm7fRwkrwwiWJ3h)_

### DIFF
--- a/.github/workflows/design-artifacts-reusable.yml
+++ b/.github/workflows/design-artifacts-reusable.yml
@@ -1912,23 +1912,40 @@ jobs:
           # was the fix; checking it unconditionally was the same fix left half-applied — the Gradle
           # side learned ownership and this pre-flight did not.
           #
-          # With no named module every discovered module is rendered, and one of them may have
-          # neither file, so the root stays reachable and is checked.
+          # The question is reachability, and only the set of modules being RENDERED can answer it.
           #
-          # `ALL_MODULES` is the same question asked properly. It is true for `module: ''` AND for
-          # `modules: all`, and the second of those can carry a named module — so reading emptiness
-          # alone, a caller passing both got ownership decided by the one module it named while the
-          # job went on to render every discovered one. Any sibling with neither authored file falls
-          # back to the root, so a malformed root policy skipped this pre-flight entirely and was
-          # dropped after the render: the catalog published with no builder data, which is the exact
-          # silence this step exists to end.
-          module_owns_itself=false
-          if [ "${ALL_MODULES:-false}" = "false" ] && [ -n "$MODULE" ] &&
+          # A single-module render asks it of that module. An all-module render cannot: `ALL_MODULES`
+          # is true for `module: ''` AND for `modules: all`, the second of which may still carry a
+          # named module, so deciding from `$MODULE` there let one module's ownership speak for a job
+          # that renders every discovered one — and a sibling with neither authored file falls back
+          # to a root that then skipped this check. Deciding instead that every all-module run
+          # reaches the root is the same mistake mirrored: where every discovered module owns its
+          # own files, nothing falls back, and validating the root turns another catalog's policy
+          # into a failure of a render that would never have read it.
+          #
+          # So an all-module run walks the discovery list and asks whether ANY module would fall
+          # back. Absent list, absent answer: the root stays checked, because being unable to prove
+          # nothing reads it is not proof that nothing does.
+          root_reachable=true
+          if [ "${ALL_MODULES:-false}" = "true" ]; then
+            if [ -f "$GITHUB_WORKSPACE/preview-module-sources.txt" ]; then
+              root_reachable=false
+              while IFS= read -r discovered_source; do
+                [ -n "$discovered_source" ] || continue
+                discovered_dir="${discovered_source%/}"
+                if [ ! -f "$discovered_dir/ui-builder.policy.json" ] &&
+                  [ ! -f "$discovered_dir/catalog.spec.json" ]; then
+                  root_reachable=true
+                  break
+                fi
+              done < "$GITHUB_WORKSPACE/preview-module-sources.txt"
+            fi
+          elif [ -n "$MODULE" ] &&
             { [ -f "$(to_dir "$MODULE")/ui-builder.policy.json" ] ||
               [ -f "$(to_dir "$MODULE")/catalog.spec.json" ]; }; then
-            module_owns_itself=true
+            root_reachable=false
           fi
-          if [ "$module_owns_itself" = "false" ] && [ -f "${RENDER_DIR:-.}/ui-builder.policy.json" ]; then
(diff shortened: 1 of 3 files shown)


yschimke:
The P1 on this PR — "Bound memoization to configured cache keys" — was right, and it landed unaddressed. Followed up in [#5327](https://github.com/yschimke/compose-ai-tools/pull/5327).

I verified the claim rather than taking it: `CatalogThemeCache.get` forwards whatever key a request produces to `Generation.wasAdopted` and `Generation.get`, and `CatalogThemeCache.put` declines to persist exactly those ad-hoc override keys, for the reason its own comment gives — "on a public box a visitor could mint distinct keys indefinitely". So the memo retained every unpersistable disk miss for the generation's lifetime, outside both the render LRU and the disk budget. My "nothing reaches this with unbounded keys" in the KDoc and the PR body was simply wrong about the read path.

#5327 gates the insert on membership (`name in present || name in adopted`) instead of memoising on the way in, which bounds the map by the same `--theme-cache-max-bytes` that bounds the tier it describes, with no cap to pick. The `HexFormat` win here is untouched.

---

PR #5383, opened 2026-09-10
feat(screen-model): generate stateful loops and reusable composables

yschimke:
Screen documents can now express typed state selection, authored row loops and reusable composables. The generator validates their scopes and types, then emits ordinary Compose source: `when`, typed `forEach` loops and private `@Composable` functions. This supplies the shared generator primitives used by the existing UI-builder Code pane and MCP export.

- `ScreenSelection` preserves literal case values and an optional fallback. Branches remain in the existing slots map; enclosing layout receivers and modifiers keep their usual behavior.
- `ScreenValue.ActionLambda` carries ordered, checked state actions in nested expressions such as `Modifier.clickable(onClick = …)`.
- `ScreenRepetition` declares concrete row types and authored data. Templates read their innermost row; nested initializers can read the enclosing row. Callbacks capture the correct row instance.
- `ScreenFunction` declares explicit concrete values and zero-argument Unit callbacks. Calls validate every argument. Functions cannot implicitly capture screen state, caller rows or layout receivers; recursion and invalid scopes refuse generation.

Document DTOs now belong to [compose-preview-contracts#65](https://github.com/yschimke/compose-preview-contracts/pull/65), published as `screen-document`. Both generator builds API-export that artifact, retaining the existing Kotlin package and serialized forms. Validation, source generation, editing and discovery stay here. Naming regression tests prevent authored functions from colliding with generated preview declarations, callback parameters from shadowing `kotlin`, and the chosen package from capturing the reusable-function `OptIn` annotation.

Both generator builds now consume published contracts **2.18.0** from Maven Central. Local publication staging remains available for future contract changes.

The server contracts pin is updated in [compose-preview-server#758](https://github.com/yschimke/compose-preview-server/pull/758); its dependency allowlist update landed in #756.

The consumer has landed in [compose-preview-server#708](https://github.com/yschimke/compose-preview-server/pull/708) behind its default-off `uiBuilderRemoteCompose` build flag. These additive generator primitives have no editor/server dependency. Local dependency staging allows the consumer to verify them before a release.

### Validation

Verified locally against released contracts 2.18.0, without local dependency overrides:

- 63 screen-model JVM tests and 575 discovery tests pass.
- WASM compilation and the full Kotlin formatting gate pass. Neither generator JAR defines `ScreenDocument`; both resolve it from contracts.
- The existing server/UI-builder consumer passes 90 export tests, 971 editor tests (4 skipped), 201 runtime tests, WASM compilation, the runtime ABI check and both dependency boundaries with Remote Compose enabled and all three artifacts staged locally.
- All six `ScreenGeneratorCompileFunctionalTest` cases pass. They discover real Material 3 components and compile the exact generated source. Selection switches between branches and fallback; row/function tests physically click Alpha, Beta, Gamma and Alpha again at densities 1 and 2. The function test also verifies per-instance 0/8/16 dp modifier insets.
- Committed repetition source and captures have been refreshed from this run. Selection and function source match their committed examples exactly.

```sh
./gradlew :screen-model:jvmTest :screen-model:compileKotlinWasmJs \
  :gradle-plugin:preview-discovery:test :gradle-plugin:functionalTest \
  --tests '*ScreenGeneratorCompileFunctionalTest'
./gradlew ktfmtCheckAll
```

### Compiled UI evidence

Generated state selection, before and after clicking:

Initial branch
Selected second branch

Reusable composable instances, before and after the second callback:

Initial reusable instances
Second instance selects Beta

[Selection semantics and source](https://github.com/yschimke/compose-ai-tools/blob/eeab218432848db3c551a3d5b63444b094b80c5f/docs/design/SCREEN_SELECTION.md) · [row semantics and source](https://github.com/yschimke/compose-ai-tools/blob/eeab218432848db3c551a3d5b63444b094b80c5f/docs/design/SCREEN_REPETITION.md) · [function scopes and source](https://github.com/yschimke/compose-ai-tools/blob/eeab218432848db3c551a3d5b63444b094b80c5f/docs/design/SCREEN_FUNCTIONS.md).

### Limits

Rows are authored and eager. Runtime lists, composable slots, event-value callback parameters, generic/nullable function parameters and per-instance retained state remain outside this change. Selection follows ordinary Compose composition behavior; transitions and retention of inactive branches are not added.

Nested expression factories and modifier calls do not carry discovered parameter signatures; their state actions are validated here, while Kotlin checks the enclosing overload. Declaration-only experimental type opt-ins are not represented, so function parameter and empty-row field types must be usable without additional type-level markers.


yschimke:
is this meant to land?

PR #5393, opened 2026-09-11
fix(usage-source-psi): register an extensions storage before parsing

yschimke:
`KotlinCoreEnvironment.createForProduction` reads `configuration.extensionsStorage` while wiring compiler-plugin extension points. A bare `CompilerConfiguration()` carries none, and on Kotlin 2.4.20 that is fatal:

```
java.lang.IllegalStateException: Extensions storage is not registered
  at ExtensionPointUtilsKt.getCompilerExtensions(ExtensionPointUtils.kt:14)
  at KotlinCoreEnvironment$Companion.configureProjectEnvironment(KotlinCoreEnvironment.kt:654)
  at KotlinCoreEnvironment$Companion.createForProduction(KotlinCoreEnvironment.kt:435)
```

Every `UsageSourceAnalyzer` parse dies in the constructor, so the analyzer yields nothing and each consumer's assertions fail.

**Why this repository is green and still has the bug.** Not the Kotlin version — `main` is already on 2.4.20 (#5385). It is coverage: `usage-source-psi` has `src/main` and nothing else, no test source set, and no other module here references `UsageSourceAnalyzer`. The class is built and published from here and exercised nowhere in this repository.

Where it *is* exercised is compose-preview-server, which vendors this file under a hash pin (`usage-source-psi/upstream.json`) and has **21 tests** failing on it: `UsageSourceParserTest`, `PreviewUsageIndexTest`, `PlaygroundSourceCleanerTest`, and the PSI spike. The pin is why the fix belongs here — the consumer cannot edit the file without failing its vendor gate.

Parsing needs no compiler plugins, so an empty `ExtensionStorage` is the entire fix. `extensionsStorage` is present in 2.4.10 as well as 2.4.20, so this is not a version-gated branch.

## Test plan

- `./gradlew :usage-source-psi:build` — BUILD SUCCESSFUL. Worth being honest about what that does and does not prove: with no test source set it establishes compilation, not behaviour.
- `./gradlew :usage-source-psi:ktfmtFormatMain` clean.
- The behavioural evidence is downstream: applying this exact change to the vendored copy in a compose-preview-server checkout takes `:server:test` from 21 failures to none across all four affected test classes.

That asymmetry — a module whose only coverage lives in the repository that vendors it — is worth a follow-up on its own, and is how a fatal bug in it shipped green. Out of scope here.

https://claude.ai/code/session_014dZ7SNn7muC8ahNUuXJ5tz

### DIFF
--- a/usage-source-psi/src/main/kotlin/ee/schimke/composeai/usagepsi/UsageSourceAnalyzer.kt
+++ b/usage-source-psi/src/main/kotlin/ee/schimke/composeai/usagepsi/UsageSourceAnalyzer.kt
@@ -2,12 +2,15 @@ package ee.schimke.composeai.usagepsi
 
 import org.jetbrains.kotlin.CoreEnvironmentDeprecation
 import org.jetbrains.kotlin.K1Deprecation
+import org.jetbrains.kotlin.cli.extensionsStorage
 import org.jetbrains.kotlin.cli.jvm.compiler.EnvironmentConfigFiles
 import org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreEnvironment
 import org.jetbrains.kotlin.com.intellij.openapi.Disposable
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer
 import org.jetbrains.kotlin.com.intellij.psi.PsiFileFactory
 import org.jetbrains.kotlin.com.intellij.psi.util.PsiTreeUtil
+import org.jetbrains.kotlin.compiler.plugin.CompilerPluginRegistrar
+import org.jetbrains.kotlin.compiler.plugin.ExperimentalCompilerApi
 import org.jetbrains.kotlin.config.CompilerConfiguration
 import org.jetbrains.kotlin.idea.KotlinFileType
 import org.jetbrains.kotlin.psi.KtCallExpression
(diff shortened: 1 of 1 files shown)


yschimke:
`Analyze (java-kotlin)` (CodeQL) failed on `99db53c`, and it is **not this PR's failure**.

```
Found 8 raw diagnostic messages.
There are multiple results visible on the status page for the diagnostic 'Kotlin version too new'.
```

`Kotlin version too new` is the CodeQL Kotlin extractor refusing the toolchain, which is set by `gradle/libs.versions.toml`. This PR touches exactly one file, `UsageSourceAnalyzer.kt`, and no version catalog — nothing in the diff can change which Kotlin the extractor meets.

The cause is #5385, which moved `main` to Kotlin 2.4.20 at 05:44 today; CodeQL's extractor does not support it yet. `main`'s last CodeQL run predates that bump (2026-09-09, green), so this PR's run is the first to meet 2.4.20 rather than the first to break.

No fix exists to port: it clears when the CodeQL bundle adds 2.4.20 support, or by pinning the extractor / excluding `java-kotlin` from the matrix — both of which are decisions about the security scanning setup, not about this diff, so I am not making either here.

I have not spent a re-run on it. A configuration error of this kind reproduces deterministically and a re-run would only confirm what the diagnostic already states.

Also correcting something in this PR's original description, since it was load-bearing: I first wrote that this repository is green because it still builds on 2.4.10. That was wrong — `main` is on 2.4.20, as this very failure shows. The real reason is that `usage-source-psi` has no test source set and no other consumer here, so nothing in this repository exercises the class. The description now says so.

---

PR #5406, opened 2026-09-11
feat(remote-json): preserve independent mutable string state

yschimke:
AndroidX’s stock string declaration interns equal initial values, so independently mutable variables can share one text ID. Add the explicit `compose-preview-state-v1` compiler profile: `mutableString` allocates each declaration with AndroidX’s named-string writer API and emits ordinary NamedVariable/TextData operations. The profile includes existing integer expressions; the stock and integer-only profiles keep their behavior.

Names and declaration fields are validated, and the additive profile constant is included in the API dump. The real CMP player proof starts with two variables and a literal all containing Ready. A click changes only the first variable; a named host update changes only the second. Six exact consumer-export fixtures additionally verify repeated ordered assignments, Unicode, empty strings, reference-like literals and generated-name collisions. String equality/selection is outside this profile.

Validation: all 31 `:remotecompose-json:check` tests pass on current main, with formatting, ABI and module boundaries. The existing ui-builder consumes a locally staged Maven publication: 52 shared-export tests and 20 server tests pass, and its real WASM text-state edits download JSON/RC identical to hosted MCP without saving. Consumer changes and reproduction are in yschimke/compose-preview-server#708; no release is required to prototype.

Before interaction, using the new profile:

Independent text before interaction

After the click and host update:

Independent text after interaction

[Source and evidence](https://github.com/yschimke/compose-ai-tools/blob/9db1139da3986edc120a25595db228b1d10b0480/docs/evidence/mutable-string-state/README.md).


yschimke:
Is this a bug? If so create a tracking ticket to report upstream

PR #5477, opened 2026-09-16
fix: regenerate the render-host ABI dump for the design-pages accessors

yschimke:
`:render-host:checkKotlinAbi` fails on `main` itself, so every PR opened since #5474 inherits a red **Module Unit Tests** that belongs to none of them.

#5474 made three accessors public on `ServeDesignPageStore` (`render-host/src/main/kotlin/ee/schimke/composeai/cli/serve/ServeDesignPages.kt:89-96`) without updating the dump the gate compares against:

```
+	public final fun asset (Ljava/lang/String;)Lee/schimke/composeai/designpages/PageAsset;
+	public final fun assets ()Ljava/util/Collection;
+	public final fun background (Lee/schimke/composeai/designpages/DesignPage;)Ljava/util/List;
```

Regenerated with `./gradlew :render-host:updateKotlinAbi` — what the gate's own failure message says to run — rather than hand-edited.

## Test plan

Reproduced first, on a clean checkout of `origin/main` (`28e9c68`) with no other change in the tree:

```
$ ./gradlew :render-host:checkKotlinAbi
  <<<ABI has changed>>>
  ...
  +	public final fun asset (Ljava/lang/String;)Lee/schimke/composeai/designpages/PageAsset;
  +	public final fun assets ()Ljava/util/Collection;
  +	public final fun background (Lee/schimke/composeai/designpages/DesignPage;)Ljava/util/List;
BUILD FAILED                                                    # exit 1
```

Then, after `:render-host:updateKotlinAbi`:

```
$ git diff --stat
 render-host/api/render-host.api | 3 +++

$ ./gradlew :render-host:checkKotlinAbi                         # exit 0
```

The diff is those three declarations and nothing else, which is the check worth making on a generated file: a dump that had moved further than the change did would mean it was stale for some other reason too.

Found while driving #5476, which is unrelated to `render-host` and was red on this.

https://claude.ai/code/session_01Ep7XZbBTFzCZcQ9dY8nVjS

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ep7XZbBTFzCZcQ9dY8nVjS)_


yschimke:
**`Analyze (java-kotlin)` is red and cannot be this PR's — this PR changes three lines of a `.api` text file and no Kotlin at all.**

The root cause, from the job log rather than the status summary:

```
> Task :gradle-plugin:daemon-launch-builder:compileKotlin FAILED
```

CodeQL 2.27.0's Java/Kotlin extractor refuses Kotlin 2.4.20, which is what `gradle/libs.versions.toml` pins. The scan runs `build-mode: manual` (this workflow's own header explains why: Default Setup's autobuild injects `allprojects { … }` init scripts that this repo's `configuration-cache=true` + `problems=fail` + `isolated-projects=true` config rejects), so the extractor has to compile — and the compile is what the version gate stops. No source is analysed, and the job ends `CodeQL job status was configuration error`.

It is repository-wide, not per-PR: identical on `83183ac` and `2dcb285` on #5476, and now on `3694408` here, whose entire diff is

```
 render-host/api/render-host.api | 3 +++
```

Not re-run — a version gate reproduces rather than flakes, and re-running spends a cycle to learn nothing.

I have not attempted a fix, because both routes are decisions rather than repairs and neither belongs in a PR about an ABI dump: `build-mode: none` is ruled out by the workflow's own note (*"Kotlin isn't supported in `none` mode"*), which leaves waiting for a CodeQL release that accepts 2.4.20, or holding Kotlin below it. Worth its own issue.

Everything else on this PR's head is green, and `:render-host:checkKotlinAbi` — the check this PR exists to fix — passes.

---

PR #5480, opened 2026-09-16
build: take daemon versions from the published BOM

yschimke:
## Why

`compose-preview-daemon` publishes **only the modules a release changes** since its 3.6.0 ([compose-preview-daemon#123](https://github.com/yschimke/compose-preview-daemon/pull/123)), so its coordinates no longer all sit at one version.

This catalog pinned all nineteen of them to a single shared version ref. That resolves right up until the first reduced release, and then fails on whichever module did not publish at the new version — **a build break arriving from a repository we did not change, on a version bump that looks routine.**

3.6.0 published everything (it changed shared build inputs, which forces a full publish), so nothing is broken today. The exposure starts at the next daemon release, which is the first likely to be reduced. This closes it first.

## What

`compose-preview-daemon-bom` is the published record of which versions belong together, so it is now the only coordinate here that names one. The nineteen entries lose their `version.ref`; the BOM keeps it, so **Renovate still has exactly one thing to bump** and the whole stack follows.

`ComposeAiBaseConventionsPlugin` puts the platform on `api` and `implementation` rather than each module declaring it. I measured the surface before choosing: **47 `implementation`, 20 `api`, 2 `testImplementation`** (extends `implementation`, as do the Android variant configurations) — so those two buckets carry 67 of the 69 declarations. The other two are the `daemonBench` configurations in the sample benchmarks, which are resolvable configurations of their own and inherit nothing, so they add the platform themselves.

The twelve `composeai-contracts` coordinates are **deliberately untouched**: they come from `compose-preview-contracts`, which publishes every module on one line and has no BOM.

## Test plan

Verified by **actually resolving**, not by building the catalog — `./gradlew :bundle-format:dependencies --configuration compileClasspath`:

```
+--- ee.schimke.composeai:daemon-client -> 3.6.0
|    +--- ee.schimke.composeai:daemon-core:3.6.0
```

`daemon-client` carries no version in the catalog and resolves to 3.6.0 from the platform. Zero `FAILED` nodes; the BOM appears on both buckets.

**That check earned its place.** The first attempt looked fine and was completely inert — it used a one-shot `configurations.findByName` at plugin-apply time, before the Java/Kotlin/Android plugin creates `api` and `implementation`, so it silently added nothing. The same command reported:

```
+--- ee.schimke.composeai:preview-data-api FAILED
+--- ee.schimke.composeai:daemon-client FAILED
BUILD SUCCESSFUL in 3m 17s
```

Note the `BUILD SUCCESSFUL` — the `dependencies` task exits 0 while printing `FAILED` per unresolvable node, so a naive exit-code check would have called that a pass. It now uses `configurations.all`, which fires for configurations created later and is therefore independent of plugin ordering.

Upstream state this depends on: daemon 3.6.0 is live, all 69 coordinates plus the BOM resolve on Central, and the BOM POM carries 69 constraints all at 3.6.0.

---


yschimke:
## `Analyze (java-kotlin)` is not this PR's failure

Standing down on this one check, with evidence.

**What it says:**

```
BUILD FAILED in 1m 25s
```

**Why it is not this PR's:**

- `kotlin = "2.4.20"` is on `origin/main`. This PR's diff is 4 files — `ComposeAiBaseConventionsPlugin.kt`, `gradle/libs.versions.toml`, and two sample build scripts — and touches no Kotlin version line. Confirmed with `git diff origin/main...HEAD -- gradle/libs.versions.toml | grep '^[+-]kotlin'`, which returns nothing.
- The same job fails identically on an unrelated PR: the Renovate run [35086399350](https://github.com/yschimke/compose-ai-tools/actions/runs/35086399350) has `Analyze (java-kotlin) = failure` with everything else green. An older branch that predates main's Kotlin bump ([35067771409](https://github.com/yschimke/compose-ai-tools/actions/runs/35067771409)) passes it.
- It is a CodeQL extractor version ceiling, not a code defect. No fix exists to port: it clears when CodeQL ships support for Kotlin 2.4.20, or when the repo's Kotlin version moves below it.

**No re-run spent.** The cause is deterministic and stated by the tool itself, and the unrelated-PR comparison is stronger evidence than a repeat would be.

**Also worth noting:** the CodeQL *workflow run* concludes `success` on this head — the `java-kotlin` job failure is already tolerated at the workflow level. So this does not gate the PR, and the check-run notification is the only place it surfaces as red.

Everything else on this head is mine and is being driven to green as normal.

---

Issue #4772, opened 2026-08-29
Apply for maven-central-publishing-limits


yschimke:
https://central.sonatype.org/publish/maven-central-publishing-limits/

yschimke:
## Where we actually stand against the limits

The org is over on all three axes. Dashboard, September 2026, namespace `ee.schimke`:

| Axis | Used | Limit | p95 |
| --- | ---: | ---: | ---: |
| Release Size | 2.26 GB | 80 MB | 247 MB |
| File Count | 129,620 | 1,000 | 2,805 |
| Release Count | 206 | 7 | 13 |

I measured what's behind those numbers directly from `repo1.maven.org` rather than inferring it from the build files. Method: enumerate every coordinate under `ee/schimke/composeai/`, count version directories dated `2026-09`, and read the per-version file listings. Size is estimated as *current artifact size × September publish count*, so it runs slightly high for artifacts that grew during the month — it lands at 2.71 GB against the dashboard's 2.26 GB, and file count at 141,780 against 129,620 (~9% over). Directionally reliable, not exact.

### The file count is a fixed multiple, which makes the limit much tighter than it looks

Every coordinate publish is **exactly 20 files**: `jar`, `-sources.jar`, `-javadoc.jar` (a 261-byte stub), `.pom`, `.module` — each with `.asc`, `.md5`, `.sha1`. Central requires all of them, so there is nothing to trim per-publish.

That converts the limits into something more actionable:

- File Count 1,000 = **50 coordinate publishes per month**
- p95 2,805 = **140 coordinate publishes per month**
- September actual = **7,089**

There are 159 coordinates in the namespace. At p95 we could publish the entire surface *less than once a month*.

### Answers to the three questions

**Are we frequently publishing large artifacts?** Frequently, yes — extremely. Large, mostly no. The median coordinate is ~1 MB. Size is concentrated in three payloads:

| Coordinate | Per release | September | Status |
| --- | ---: | ---: | --- |
| `compose-preview-ui-builder-web` | 13.73 MB | 686 MB | stopped 2026-09-11 |
| `compose-preview-serve` | 8.53 MB | 427 MB | stopped 2026-09-11 |
| `rc-player-wasm-dist` | 8.17 MB | 188 MB | still publishing |

[...]

PR #5485, opened 2026-09-16
feat(build): publish only the modules a release changes

yschimke:
This repository releases **~123 times in 30 days** across 26 coordinates. At 20 files per coordinate-publish that's **~66,000 files a month** against an org-wide Maven Central file-count limit — the largest single contributor of the four repositories that publish (#4772).

Measured over `v2.10.0..v2.17.0`, **10 real intervals, 260 module-publications: 42% is a version-bumped rebuild of identical code.**

| Interval | Publishes / 26 |
| --- | ---: |
| v2.11.0 → v2.11.1 | 4 |
| v2.13.0 → v2.13.1 | 0 |
| v2.13.1 → v2.14.0 | 4 |
| v2.14.1 → v2.15.0 | 0 |
| (5 others) | 26 — a shared build input moved |

## How it decides

A module publishes when a file under it changed **since the tag it last published at**, when something it depends on is publishing, or when a shared build input changed. Comparing against its own baseline rather than the last release is what makes a gap safe: a module skipped for three releases is still diffed against the version it actually shipped. Every uncertainty resolves to "publish" — Central refuses a second upload, so an unnecessary publish costs quota while a wrongly-skipped one is unrepairable.

Skipped modules keep their **recorded** version, which is the whole mechanism: a POM names its project dependencies at their `project.version`, so a module built at the tag names its skipped dependencies at versions that are actually on Central. Stamping the tag onto a module that didn't publish is the break this repository shipped in v2.2.1 (yschimke/wear-m3-catalog#350).

Verified end to end — under `-Pcomposeai.publishSet=render-host`, `:bom`'s POM gives `render-host` the build version and `screen-model` its recorded `2.17.0`.

## Three things this repo needs that compose-preview-daemon#123 didn't

- **The `gradle-plugin` included build always publishes.** `:cli` bakes a plugin coordinate into `cli-version.properties` and falls back to the CLI's own version, which `release.yml` deliberately leaves unset. A CLI shipped at 2.18.0 telling Gradle to resolve `compose-preview-plugin:2.18.0` when the plugin was skipped is a user-facing break on the first command they run. Pinning the four costs ~3 points (45% → 42%) and removes the class.
- **`record-published` is its own `continue-on-error` job that opens a PR.** In the daemon this began as a step pushing to `main`; the branch ruleset rejected it and failed the release *after* the artifacts were on Central — unrepairable (its #127).
- **`release-please.yml` grants `pull-requests: write`.** A called workflow cannot elevate what its caller granted, and asking for more is a **startup failure of the entire run**, not a job failure — the daemon's #131, where release-please silently stopped for two pushes. Both halves land together here rather than as a follow-up repair.

The manifest is seeded from **Maven Central, not from the tag**: all 26 coordinates were confirmed to resolve at 2.17.0 before it was committed.

## Test plan

- Plan script exercised over 10 real intervals → 42%.
- A no-change release publishes **4** (the pinned plugin set); a shared-input release publishes all **26** — including the four plugin coordinates, which an earlier draft silently dropped because the early exit ran before they were registered. Caught by running it, not by reading it.
- `printPublishTasks` verified for absent / empty / partial sets (24 / BOM-only / BOM+named).
- `record-published.py` round-trips, moving only the named modules.
- `build-logic` tests and `ktfmtFormat` green; both workflows parse.

Non-visual: build and release configuration only.

---


yschimke:
**`Analyze (java-kotlin)` is red on `1fd4101`, and it is not this PR's.** Standing down on it; no re-run spent.

```
Execution failed for task ':gradle-plugin:preview-discovery:compileKotlin'
...
```

Three reasons:

1. **It fails in a module this diff does not touch.** `:gradle-plugin:preview-discovery:compileKotlin` — the change is `publishing-manifest.json`, two `.github/scripts/`, `release.yml`, `release-please.yml`, `build.gradle.kts`'s `printPublishTasks` filter, and `publishedVersion()` in build-logic. None of it alters how that module compiles.
2. **The cause is a toolchain ceiling, not the code.** The CodeQL java-kotlin extractor supports Kotlin *below* 2.4.20; this repository is on 2.4.20, from `main`. CodeQL says so itself and classifies the run as `configuration error` rather than a finding.
3. **Same failure, same cause, on an unrelated PR earlier today** (#5480), where it was established against `origin/main`'s Kotlin version and an unrelated Renovate PR showing the identical error.

No re-run: this is deterministic — a version comparison, not a flake — so a re-run would spend the one allowed and return the same line. The fix is an upstream extractor release or a Kotlin downgrade, neither of which belongs in this PR.

---

PR #5488, opened 2026-09-16
fix(build-logic): resolve the included build's real artifact ids

yschimke:
The v2.18.0 release job ([run 35155606835](https://github.com/yschimke/compose-ai-tools/actions/runs/35155606835)) failed in `publish-gradle-plugin` while configuring the `gradle-plugin` included build:

```
* Where:
Build file '.../gradle-plugin/build.gradle.kts' line: 3

* What went wrong:
An exception occurred applying plugin request [id: 'composeai.maven-publishing']
> Failed to apply plugin 'composeai.maven-publishing'.
   >  is not in the publish set and has no entry in publishing-manifest.json, so there is no version it can safely carry. Add it to the manifest, or publish it.
```

Leading space and all — the artifact id was the empty string. The tag `v2.18.0` exists and `release` succeeded, but `record-published`, `finalize-release` and `verify-maven-readiness` all skipped, so nothing reached Central and `releases/latest` is still v2.17.0.

## Cause

`publishedArtifactId()` derived the id by flattening the project path, and two of the included build's four published projects do not follow from their path at all:

| project path | declares | flattens to |
| --- | --- | --- |
| `:` | `compose-preview-plugin` | *(empty string)* |
| `:gradle-plugin-config` | `compose-preview-config` | `gradle-plugin-config` |
| `:preview-discovery` | `preview-discovery` | `preview-discovery` ✓ |
| `:daemon-launch-builder` | `daemon-launch-builder` | `daemon-launch-builder` ✓ |

Neither of the first two is ever in a publish set, so **every** reduced-publish release fails there — the failing run's set named all 26 real coordinates and it still died. Nothing noticed before because the id was never asked for until the publish plan landed (#5483); without a plan, `PublishedVersions.resolve` returns the tag for everything.

`PublishedArtifactIdTest` did not catch it: one case pins the *main* build's path-flattening assumption, the other pins *which* four coordinates the included build carries. Neither checks that the included build's coordinates follow from its paths.

## Change

- The derivation moves into `PublishedArtifactIds`, a pure function over (root project name, path) carrying the two exceptions. Keyed by root project name because an included build's paths are relative to its own root: `:` in the main build is a different project from `:` in `gradle-plugin`, and only the latter publishes.
- `PublishedArtifactIdTest` gains a case that walks the included build and asserts every published project resolves to the coordinate it declares — the assumption that was missing rather than wrong.
- `PublishedVersions.resolve`'s error now quotes the artifact id, so an empty one is legible rather than a leading space.

No behaviour change for the main build or the BOM: `forProject` is the old flattening for every root project that is not `gradle-plugin`.

## Test plan

- `./gradlew --project-dir build-logic test` — 3 tests, 0 failures (the new case included).
- Reproduced the release configuration locally, which previously failed to configure:
  ```
  PLUGIN_VERSION=2.18.0 ./gradlew --project-dir gradle-plugin \
    -Pcomposeai.publishSet=compose-preview-config,compose-preview-plugin,daemon-launch-builder,preview-discovery \
    properties :gradle-plugin-config:properties :preview-discovery:properties :daemon-launch-builder:properties
  ```
  All four now resolve `version: 2.18.0`.

Once this lands, the v2.18.0 release needs a `workflow_dispatch` re-run to actually upload — the tag is already cut.

Text-only change to build logic; no visual surface.

https://claude.ai/code/session_01VuDi28WkdKpC6TcHWYzeCp

---
_Generated by [Claude Code](https://claude.ai/code/session_01VuDi28WkdKpC6TcHWYzeCp)_


yschimke:
Two things from a parallel session that this PR should know about — the diagnosis here is right, and matches what I found independently.

**1. `main` already carries a fix for this.** [#5487](https://github.com/yschimke/compose-ai-tools/pull/5487) merged ~20 minutes ago (`1ce35f7`, which is this PR's base). It takes the other branch of the same fork: rather than resolving the included build's real ids, it returns the tag outright for any project in an included build, on the grounds that `maven-publish-plan.sh` does `dirty.update(INCLUDED_BUILD_IDS)` unconditionally — so those four always publish and the tag is always their correct version.

The approach in this PR is the more general one: it yields the *correct id*, which the bypass does not, and that matters if anything else ever needs to address those projects by coordinate. Worth keeping on those grounds, but it is a refactor of working code now, not a release fix — and the two will collide in `publishedVersion()`.

**2. The re-run plan does not work as written**, and I had the same error in my own write-up:

> Once this lands, the v2.18.0 release needs a `workflow_dispatch` re-run to actually upload — the tag is already cut.

`publish-gradle-plugin` checks out `ref: ${{ env.TAG_NAME }}`, so a dispatch for `v2.18.0` builds **the tree at that tag** — which is `39ad1f7`, predating both fixes:

```
$ git merge-base --is-ancestor 1ce35f7 v2.18.0   # the fix in main
NO — the tag predates the fix
$ git show v2.18.0:build-logic/.../ComposeAiMavenPublishingPlugin.kt | grep -c gradle.parent
0
```

So a re-run fails identically, wherever the fix lands. v2.18.0 can only ship by moving that tag onto a fixed commit, or by cutting a new version that contains one. Nothing is stranded either way — `render-host`, `compose-preview-plugin` and the BOM are all 404 on Central at 2.18.0, and the GitHub Release is still an untagged draft.

---

yschimke:
Superseded by #5487, which landed on `main` while this was open. Closing rather than resolving the conflict.

Same root cause, different answer, and #5487's is the better one to keep: it returns the tag outright for any included build (`gradle.parent != null`) instead of teaching the derivation the two exceptions. That is correct for the reason its comment gives — `maven-publish-plan.sh` marks all four included-build ids dirty unconditionally, so they publish at every release and never need the lookup. This PR's `PublishedArtifactIds` table would only start earning its keep if the included build ever became skippable, and #5487's test pins the fact that the path convention cannot name those coordinates, which is the part that was actually missing.

The conflict is real, not spurious: both diffs edit the same lines of `publishedVersion()` and add a case to `PublishedArtifactIdTest`. Nothing here is worth carrying forward on top — the one remaining difference is that `PublishedVersions.resolve`'s error quotes the artifact id, and with the bypass in place an empty id can no longer reach it.

`v2.18.0` still needs a `workflow_dispatch` re-run of the release to upload: the tag is cut and `release` succeeded, but `publish-gradle-plugin` failed, so `record-published`, `finalize-release` and `verify-maven-readiness` all skipped and `releases/latest` is still v2.17.0.

---

PR #5490, opened 2026-09-16
fix(build): read the publish baseline from Central, and fix the recovery path

yschimke:
Two fixes to the release plumbing, both found by running it rather than reading it. **This should land before 2.18.1 is cut.**

## 1. A `workflow_dispatch` recovery run published only the BOM

The step chose its task list on whether the publish set was non-empty, then passed `-Pcomposeai.publishSet=` *unconditionally*:

```bash
if [ -n "$PUBLISH_SET" ]; then TASKS=…-Pcomposeai.publishSet="$PUBLISH_SET"
else                           TASKS=…-Pcomposeai.publishSet=          # ← empty, not absent
fi
./gradlew $TASKS -Pcomposeai.publishSet="$PUBLISH_SET"                 # ← empty again
```

Absent and empty are different properties and the whole design turns on it: absent means *"no plan ran, publish everything"*, empty means *"the plan ran and found nothing"* — the BOM alone. So the no-plan path sent the wrong one. Measured on this branch:

| `printPublishTasks` with… | tasks |
| --- | --- |
| `-Pcomposeai.publishSet=` (before) | **1** — `:bom` and nothing else |
| no property (after) | **24** |

That is the one entry point that exists to republish after a failure, and it is exactly the path a v2.18.0 recovery would have taken. My bug, from #5485.

## 2. The baseline comes from Maven Central, not from git

The plan needs one fact per module: the version it is currently published at. That was a committed `publishing-manifest.json`, written back after each release by a job that opened a PR — which needed a human to merge, and was wrong until they did. In compose-preview-daemon the equivalent PR took **2h12m** to land, and a release cut in that window would have diffed against a stale baseline and republished everything. Here it never ran at all: v2.18.0 died before it, so the committed file still reads `2.17.0` for all 26 coordinates.

`maven-metadata.xml` answers the same question authoritatively, so the plan asks Central (26 coordinates in **1.9s**) and `--write-manifest` drops the resolved baseline into the workspace for Gradle to read. Nothing is committed. A coordinate Central has never seen, or cannot be reached about, resolves to **publish** — the same fail-safe as before, and the safe direction.

Gone with it: `record-published.py`, its job, and the `pull-requests: write` on the caller that existed only for it. No job in `release.yml` needs that scope any more, and the comment now says to keep it that way — a job asking for more than the caller grants is a startup failure of the whole run (compose-preview-daemon#131).

Companion PR in the daemon: [yschimke/compose-preview-daemon#144](https://github.com/yschimke/compose-preview-daemon/pull/144), where the write-back also dragged an empty release behind every real one.

## Test plan

- Central-derived baseline vs the committed file: **26/26, both 2.17.0** — identical, so no behaviour change on the next release.
- `maven-publish-plan.sh --head v2.18.0` → `baseline: Maven Central (26 of 26 coordinates)`, then every module dirty via the shared-input rule, as before.
- End-to-end handoff: plan writes the manifest, then `PLUGIN_VERSION=2.19.0 -Pcomposeai.publishSet=render-host` gives `:render-host` → **2.19.0** and `:bundle-format` → **2.17.0**.
- `printPublishTasks` with **no** property → 24 rows, and needs no manifest file at all.
- `ktfmtCheckAll` green; `./gradlew -p build-logic test` green (including the Build Logic Tests job added in #5487).

Text-only evidence: nothing here changes what anything draws.

---


yschimke:
`Analyze (java-kotlin)` is red here and is **not this PR's**. From [its log](https://github.com/yschimke/compose-ai-tools/actions/runs/35160316679/job/105009374295):

```
Execution failed for task ':gradle-plugin:preview-discovery:compileKotlin'
…
```

The extractor cannot build this repository at its Kotlin version, so the check fails on every PR here regardless of content — this diff is a shell script, two workflows, a `.gitignore` and comments, none of which CodeQL reaches. No fix exists on our side short of holding Kotlin back or waiting for extractor support, and I am not spending a re-run on it: the failure is deterministic and the log names the cause rather than suggesting a transient one.

Everything else on this head is green.

---

PR #5500, opened 2026-09-17
feat(actions): a reusable workflow that renders UI-builder designs on a PR

yschimke:
A catalog's review surface is pixels, and `apply` provides it. A **UI-builder design** cannot use it: a design is a committed JSON document of nodes, properties and modifiers — no Gradle module to discover, nothing for `compose-preview show --json` to enumerate. So a change to one of the nineteen designs `compose-ui-builder` commits could only be reviewed as a text diff, which says nothing about whether the screen still looks right.

This adds `ui-builder-designs-reusable.yml`, which renders them and comments.

```yaml
jobs:
  designs:
    permissions:
      contents: write
      pull-requests: write
    uses: yschimke/compose-ai-tools/.github/workflows/ui-builder-designs-reusable.yml@main
    with:
      catalog-bundle: https://raw.githubusercontent.com/yschimke/m3-catalog/design-artifacts/m3-catalog/bundle/bundle.png
      components: |
        m3-catalog=docs/design/fixtures/ui-builder/m3-catalog-components-v1.json
```

The pipeline is `compose-preview-server design render --local` — the design's generated Kotlin, compiled and rasterised in-process — over documents the caller's own `design-sync.mjs` replays out of its fixtures. Changed designs only on a PR, a sticky before/after comment against a baseline branch, baseline refreshed on push to the development branch. Same shape `apply` gives a catalog; deliberately not the same pipeline, which is also why `design-renders.py` sits beside `compare-previews.py` rather than inside it.

## Two decisions worth reading

**It renders inside `ghcr.io/<owner>/compose-preview-host`.** A local render needs four sidecars beside the install — `lib-bta/` (the Kotlin compiler), `lib-daemon-desktop/`, `lib-renderer/`, and the Skiko **native** for the host — and compose-preview-server's released tarball carries neither the first nor the last. Without the native, every frame dies inside `Library`'s static initialiser and reaches the lane as a picture of nothing. That image already assembles all four with the binary on PATH, so using it is what makes this a workflow rather than a week of sidecar archaeology. Hence the container job, and `host-image` to pin a version.

**Three jobs, split on write access.** `stage` runs the caller's `design-sync.mjs`; `render` runs the caller's designs through a compiler. Both hold `contents: read` and hand on artifacts. Only `publish`, which runs none of the caller's code, pushes the render branch and posts the comment — the same argument `design-artifacts-reusable.yml` makes for its publish split. Workflow-level `permissions: {}`, inputs consumed through `env:` so expression expansion cannot alter a shell, per this repo's conventions.

## A refusal is a row, not a failure

`design render` can refuse a design whose values the projection has no Kotlin for, and that is the lane's most useful output: a design that cannot be exported cannot be reviewed as pixels either. The comment names it with the compiler's own reason. And because a frame the daemon never drew lands as a **zero-byte file** rather than an error, an empty render is recorded as a refusal too — without that the lane publishes blank pictures and calls it a pass.

## Checks

- `actionlint` clean on the new workflow and on `ci.yml`
- `python3 .github/actions/lib/test_design_renders.py` — 15 tests, pure stdlib, wired into **Actions Script Tests** (no perceptual-diff install needed: a design render is compared by the sha256 its manifest carries, so nothing here opens an image)
- The **staging half is proven locally** against a real committed fixture: `ui-builder-mobile` replays to an 80-node `DesignDocumentV1`, and `index.json` — which `publish` writes beside the documents and which an earlier draft of the render loop would have tried to render as a design — is skipped by name.

The **render half needs the host image and a catalog bundle**, so CI is the first place the assembled lane runs. The command itself is not new or unproven: it is the one behind the committed evidence renders in `compose-preview-server/docs/evidence/ui-builder-design-export-render/`.

The first consumer is yschimke/compose-ui-builder#6.

https://claude.ai/code/session_01Q4HmbbqexFPp9UTPEucD47

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q4HmbbqexFPp9UTPEucD47)_


yschimke:
**Closing: this belongs in compose-preview-server, not here.** Re-landed as yschimke/compose-preview-server#947.

The lane invokes `compose-preview-server design render --local` inside `ghcr.io/yschimke/compose-preview-host`. **Neither artefact is this repository's** — that repo releases both. The only thing of compose-ai-tools' it used was `design-renders.py`, which came with the PR, so the coupling was entirely self-inflicted. What it left behind was a workflow in one repository and the two artefacts it drives in another, with nothing policing the version skew between them — the class of problem `apply`'s own `skew-check` input exists to catch on the plugin/CLI seam.

Putting it next to `apply` looked right because this is where the reusable workflows and the render-comment precedent live, and because the catalog repos already `uses:` from here. That is a discoverability argument, and it lost to the skew one.

Nothing of this repository's was changed except the one line wiring the tests into **Actions Script Tests**, which goes away with the PR. `design-renders.py` deliberately did not merge into `compare-previews.py`, and that reasoning survives the move: the comparator is coupled to `compose-preview show --json`, a preview id and a Gradle module, and a design document has none of the three.

The split question this raised — whether `design` should be its own project, published as a tarball, rather than shipping inside a server distribution — is measured and sequenced in `docs/design/COMPOSE_PREVIEW_DESIGN_SPLIT.md` in that PR. Short version: `--local` borrows `PlaygroundCompileService` and friends, 6,093 lines with ten other consumers in `:server`, so it is an engine extraction rather than a move, and it queues behind compose-ui-builder's publishing lane.

---

PR #5508, opened 2026-09-19
fix(cli): launcher allowlists cover the server's documented flags

yschimke:
## Summary

`compose-preview ui-builder --no-project --agent-grants --catalog-mcp --agent-grant-capabilities …` is the documented local MCP command line. Every one of those flags was reported as unrecognised by the launcher — `--catalog-mcp` and `--agent-grants` were missing from the `serve` allowlist too.

To be precise about what was and was not broken: **the flags were forwarded and took effect** (the launcher passes argv through; verified against server 3.40.0 — grants, the capability ceiling, the `/mcp` endpoint and the builder all came up). The damage was the warning: on a documented command line it read as though the command were wrong. #5501 already made the note say *forwarded* instead of *ignored*; this removes the note for every flag the server documents.

- The `serve` flag set moves to one `serveFlags` val.
- `ui-builder` takes it wholesale — the server's `ui` command accepts every `serve` flag, as its own help says — plus the builder-only options.
- Eight flags the 3.40.0 server documents that neither entry listed are added: `--admin-read-token`, `--catalog-mcp`, `--onboard-cache`, `--spare-sandboxes`, `--ui-builder-comment-webhook`, `--ui-builder-comment-webhook-format`, `--ui-builder-packs`, `--wasm-ui-dir`.
- `design` gains its local replay lane (`--local`, `--document`, `--catalog`, `--assets`, `--components`), documented by `design --help` and previously warned about.

Verified live: the exact command line above now produces **zero** notes, with grants and `/mcp` up.

The other half of the reported friction — `--module :remote-catalog` not scoping the build — is a server bug fixed in yschimke/compose-preview-server#956.

## Test plan

- [x] `ui-builder accepts every flag serve accepts` pins the shared set structurally.
- [x] The reported command line and the `design --local` lane are asserted unknown-flag-free.
- [x] `:cli:test` and `ktfmtCheckAll` green.
- [x] Live run against 3.40.0 as above.
- [ ] CI.


yschimke:
The two red `Build` jobs are not this change — main is broken by Renovate's #5505: it bumped the shared `composeai-contracts` ref to `3.1.1`, but only `data-layoutinspector-core` is published at 3.1.1; `screen-document`, `daemon-protocol`, `daemon-devices` and `data-preview-overrides-core` are not (404 on Central), so `:samples:android` fails to configure:

```
> Could not find ee.schimke.composeai:screen-document:3.1.1.
```

main's own `Samples-SDK21` run is red at the same commit for the same reason. `:cli:build :bundle-viewer:build` passes locally on this branch. The fix belongs on main (split the layoutinspector version from `composeai-contracts`, or revert the bump until the contracts line publishes 3.1.1).

